PR #6, opened 2016-12-26
making it testable with recent D2 compiler

MartinNowak:
Sorry for the giant diff, don't know how to deal with the D1/D2 conversion.
Just putting my D2 fixes up here b/c it might be useful for you as well, they are grouped by compiler change.
My goal was only to find dlang regressions, though thus far I'm only at compiling, not linking/running.
Also see https://trello.com/c/6rNVdDEr/100-fix-and-test-ocean-library


leandro-lucarella-sociomantic:
Thanks Martin, this is AWESOME. For the time being we'll take this PR as some sort of example, because we still couldn't move our internal Ocean development to the public repo, but we are finally really close.

@mihails-strasuns-sociomantic surely can tell you how to better deal with D1/D2 changes. I wonder if we should opensource some of the internal docs with recommendations on how to develop/test in both D1/D2 for potential contributors to ocean.

src/ocean/util/log/PeriodicTrace.d
```diff
@@ -192,7 +192,7 @@ struct PeriodicTracer
     {
         va_list ap;
 
-        va_start(ap, __va_argsave);
+        va_start(ap, fmt);
```
mihails-strasuns-sociomantic, line 195: Have cherry picked all changes in this PR apart from this line because it clashes with D1 runtime which has `va_start` defines like this: https://github.com/sociomantic-tsunami/tangort/blob/v1.x.x/src/core/stdc/stdarg.d#L103-L106

Any ideas how to hack this other than porting compiler intrinsic (https://github.com/dlang/druntime/blob/master/src/core/stdc/stdarg.d#L194) to D1 branch? This is a a valid option too, but least desired one for obvious reason.

src/ocean/util/cipher/gcrypt/core/Gcrypt.d
```diff
@@ -352,8 +353,6 @@ private class GcryptBase ( Algorithm algorithm, Mode mode )
 
     version ( UnitTest )
     {
-        import ocean.core.Test;
```
mihails-strasuns-sociomantic, line 355: Doesn't seem to be directly related to the comment description. Is the change in wrong commit or I am missing the context?

src/ocean/util/cipher/gcrypt/HMAC.d
```diff
@@ -32,12 +32,12 @@ module ocean.util.cipher.gcrypt.HMAC;
 
 import ocean.util.cipher.gcrypt.core.MessageDigestCore;
 import ocean.transition;
+import ocean.util.cipher.gcrypt.c.md;
 
 /******************************************************************************/
 
 public class HMAC: MessageDigestCore
 {
-    import ocean.util.cipher.gcrypt.c.md;
```
mihails-strasuns-sociomantic, line 40: This kind of changes is one of main reasons why I was reluctant to hurry with 2.071 update. Problem is that it doesn't only affect ocean - any downstream library that used to call `HMAC.some_imported_md_symbol` will break. I am going to ignore this commit for now and re-add it later after testing how it affects downstream projects.

src/ocean/io/device/File.d
```diff
@@ -185,8 +185,13 @@ class File : Device, Device.Seek, Device.Truncate
 
                 if (msg.length == 0)
                 {
+                    static if (__VERSION__ < 2070)
```
mihails-strasuns-sociomantic, line 188: Just checked, `static if` is not necessary here, can be simply changed to `import core.stdc.string` unconditionally. Has to be at top level though, because D1 does not support imports inside function scope.

mihails-strasuns-sociomantic (commented):
Have just created PR for primary internal repo with all changes but 1a5f575212f8e0741f82ec22c2e527e2afcfda36 (see review comment).

Should be in public repository upon next sync merge.

mihails-strasuns:
@MartinNowak 

> Sorry for the giant diff, don't know how to deal with the D1/D2 conversion

Internally we always do development on D1 version, running `make d2conv` each time on top to test D2 build. For public repository we could provide automatically generated branch with already converted sources but I am not sure how much advantage does it offer because all changes have to be submitted against D1 base anyway.

Note that we are going to make public Sociomantic docker repository soon which will contain base image with necessary tools and libraries. Hopefully that will make external testing/experimenting much easier.

mathias-lang-sociomantic:
Isn't this outdated now ?

mihails-strasuns-sociomantic:
It sort of was since the very start :) I kept the issue open as a way to report situation back to Martin but it is probably not relevant anymore. `v3.x.x` already has basic 2.071 compatibility and more will follow when we have travis testing matrix in power.

PR #12, opened 2017-02-08
travis: Fix docker image naming

leandro-lucarella-sociomantic:
The name for the docker image for PRs was wrongly set as the branch
image, and images for push builds where plain wrong (:ci-false).

This commit fixes that and makes pulling of images smarter for PRs. If
a PR image doesn't exist already, the base branch image will be fetched
instead (and tagged as the PR image). Since docker image building
scripts are very unlikely to change, the base branch image will probably
be the same as the PR image anyway, so it will be used as a cache.

### DIFF
--- a/ci/travis-image-name.sh
+++ b/ci/travis-image-name.sh
@@ -1,10 +1,2 @@
 #!/bin/sh
-
-# Build docker image name
-img="sociomantictsunami/$(basename $TRAVIS_REPO_SLUG):ci"
-if test "$TRAVIS_PULL_REQUEST" != "false"
-then
-    echo "$img-$TRAVIS_BRANCH"
-else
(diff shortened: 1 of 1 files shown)


nemanja-boric-sociomantic (approved):

leandro-lucarella-sociomantic:
Ah, damn! I just remembered why I didn't wanted to use PR images at all! PRs can't push to docker hub (it would be unsafe to let anyone sending a PR to get the docker hub password or even just push. I will change to always use the base branch, it should make everything simpler (although rebuilding PRs that change the docker image will take ages, at least until we have a base docker image).

leandro-lucarella-sociomantic:
Green, now the change is trivial, so I'll be bold and merge, I want to see if it fixes the push build too.

PR #14, opened 2017-02-23
travis: Don't push docker images when building tags

leandro-lucarella-sociomantic:
There is no easy way to identify which branch a tag belongs to (and it
might not belong to any), and saving images of tags as a cache makes no
sense at all, so just avoid pushing images of tag builds.

### DIFF
--- a/ci/travis-install.sh
+++ b/ci/travis-install.sh
@@ -11,7 +11,7 @@ docker build --pull -t "$img" .
 
 # And pushing will also be very fast (done only if there are valid credentials)
 set +x # No peeking of password!
-if test -n "$DOCKER_PASSWORD"
+if test -n "$DOCKER_PASSWORD" -a -z "$TRAVIS_TAG" # Don't push tags
 then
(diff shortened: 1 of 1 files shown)


mihails-strasuns-sociomantic:
Remind me please, why does pushing is needed at all here (considering cache is not working)?

leandro-lucarella-sociomantic:
> Remind me please, why does pushing is needed at all here (considering cache is not working)?

Just because of faith in getting it to work. I already contacted Travis. We could disable it for now though, yeah.

mihails-strasuns-sociomantic:
Yeah I think it is better to comment it out completely for now with mention of upstream bug

leandro-lucarella-sociomantic:
There is no upstream bug, just private e-mail conversations and some experimental SSH access to the VM to investigate it myself...

leandro-lucarella-sociomantic:
Anyway, that's a different issue, please create one if you want it, but this is a fix that is needed anyway :)

mihails-strasuns-sociomantic (approved):

PR #18, opened 2017-03-09
travis: Don't push docker images when building tags

leandro-lucarella-sociomantic:
There is no easy way to identify which branch a tag belongs to (and it
might not belong to any), and saving images of tags as a cache makes no
sense at all, so just avoid pushing images of tag builds.


leandro-lucarella-sociomantic:
Actually v2.4.x is going out of support like in 5 minutes...

PR #23, opened 2017-03-09
README: Update build badge for v2.6.x

leandro-lucarella-sociomantic:
This might be annoying to remember, I don't really know what to do about
it. We can try to keep it up to date for now.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,4 +1,4 @@
-.. image:: https://travis-ci.org/sociomantic-tsunami/ocean.svg?branch=v2.x.x
+.. image:: https://travis-ci.org/sociomantic-tsunami/ocean.svg?branch=v2.6.x
   :alt: Build status
   :target: https://travis-ci.org/sociomantic-tsunami/ocean
 


mathias-lang-sociomantic:
Could we just use the default branch ? It doesn't bring too much to do it every time, does it ?

mihails-strasuns-sociomantic:
Yeah I agree with @mathias-lang-sociomantic , the main purpose of the badge is to show current build status in the rendered readme - keeping it tied to current branch is just an extra maintenance effort.

mihails-strasuns-sociomantic:
Updated to use default branch instead

leandro-lucarella-sociomantic:
> Could we just use the default branch ? It doesn't bring too much to do it every time, does it ?

That's misleading, what sense does it make to see that some other branch is green when I'm looking at another particular branch. I would rather completely remove the badge instead if we don't care to keep it up to date.

I don't think is that much of a trouble to update the badge (it will just be easy to forget). We already have a lot of bookeeping to do when we make a release, updating the badge should be a non-brainer when we have scripts for this.

leandro-lucarella-sociomantic:
Sorry, force pushed again with my original commit, I prefer to leave the changes in this PR consistent with what the PR was created for, if you want another solution just create another PR.

mihails-strasuns-sociomantic:
I'd like to close this PR as wontfix in that case, as I strongly object the practice you are trying to introduce here.

mathias-lang-sociomantic:
> That's misleading, what sense does it make to see that some other branch is green when I'm looking at another particular branch.

The root issue here is that there is no way to use the current branch as a parameter to anything in markdown. It's a well-known Github limitation, and I don't find it particularly misleading (but would really like for it to be solved as well).

mihails-strasuns-sociomantic:
Competing PR to remove the badge: https://github.com/sociomantic-tsunami/ocean/pull/29

leandro-lucarella-sociomantic:
Again I don't think is a huge burden to keep it updated and it can be automatized, but I'm also fine with removing the badge completely. When I see an alternative PR to solve this in another consistent way, I'll happily close as won't fix.

PR #26, opened 2017-03-10
Smart union: add active_name() and callWithActive()

gavin-norman-sociomantic:


nemanja-boric-sociomantic:
You seems to have D2 failures: https://travis-ci.org/sociomantic-tsunami/ocean/builds/209681331

```
./src/ocean/core/SmartUnion.d(252): Error: function ocean.core.SmartUnion.callWithActive!(print, SmartUnion!(TestUnion)).callWithActive cannot access frame of function ocean.core.SmartUnion.__unittestL259_270.printActiveUnionField
./src/ocean/core/SmartUnion.d(252): Error: function ocean.core.SmartUnion.callWithActive!(print, SmartUnion!(TestUnion)).callWithActive cannot access frame of function ocean.core.SmartUnion.__unittestL259_270.printActiveUnionField
./src/ocean/core/SmartUnion.d(283): Error: template instance ocean.core.SmartUnion.callWithActive!(print, SmartUnion!(TestUnion)) error instantiating
make: *** [/home/travis/build/sociomantic-tsunami/ocean/build-d2/production/tmp/allunittests] Error 1
```

gavin-norman-sociomantic:
Updated.

src/ocean/core/SmartUnion.d
```diff
@@ -254,6 +289,33 @@ private template MemberList ( uint i, size_t len, U )
...
+    static if ( i == U.tupleof.length )
+    {
+        const MemberStringList = "";
+    }
+    else
+    {
+        const MemberStringList = `,"` ~ FieldName!(i, U) ~ `"` ~
+            MemberStringList!(i + 1, U);
+    }
+}
```
david-eckardt-sociomantic: Suggest generating the actual array of strings here like
```d
// Template code would probably be cumbersome so use CTFE
private static istring[] memberStringList ( U ) ( )
{
    istring[] member_strings = ["none"[]];
    foreach (i, Field; typeof(U.tupleof))
        member_strings ~= FieldName!(i, U);
    return member_strings;
}

// inside SmartUnionIntern
istring[] active_names = memberStringList!(U)();
```
Btw. why did you put `MemberStringList` outside the `SmartUnion` and `SmartUnionIntern` struct?

src/ocean/core/SmartUnion.d
```diff
@@ -212,6 +212,86 @@ version (UnitTest)
...
+        smart_union = smart-union instance whose active field should be passed
+            to Callable
+
+*******************************************************************************/
+
+public void callWithActive ( alias Callable, SU ) ( SU smart_union )
+{
+    alias TemplateInstanceArgs!(SmartUnion, SU) SUArgs;
+    static assert(SUArgs.length == 1);
+    alias SUArgs[0] U;
```
david-eckardt-sociomantic: Hm, why didn't you put this method in `SmartUnion` so you don't have to derive the union type? Btw. I would probably use a simple `alias typeof(smart_union._) U`.

david-eckardt-sociomantic: > Btw. I would probably use a simple `alias typeof(smart_union._) U`.

Never mind, this may not work.

david-eckardt-sociomantic: > Btw. I would probably use a simple `alias typeof(smart_union._) U`.

`alias typeof(smart_union._.u) U`, of course.

src/ocean/core/SmartUnion.d
```diff
@@ -212,6 +212,86 @@ version (UnitTest)
...
+
+    if ( smart_union._.active == smart_union.Active.none )
+        return;
+
+    assert(smart_union._.active >= 1);
```
david-eckardt-sociomantic: `smart_union.Active.none` is intentionally defined to be zero to allow for
```d
if ( !smart_union._.active ) return;
```
in situations like this, which makes the assertion unnecessary (at least for me ;) ).

src/ocean/core/SmartUnion.d
david-eckardt-sociomantic: > Btw. why did you put `MemberStringList` outside the `SmartUnion` and `SmartUnionIntern` struct?

I see why: `MemberList` is outside as well. I didn't read this code for a long time ;)
Still you *could* put `MemberStringList` inside `SmartUnion/Intern`.

src/ocean/core/SmartUnion.d
gavin-norman-sociomantic: I see, yeah that's a bit cleaner.

src/ocean/core/SmartUnion.d
gavin-norman-sociomantic: Regarding both points (the location of the template and using a template vs CTFE), I simply wrote code in the same style as the existing code in this module. I agree that the CTFE solution looks a lot nicer, though, so will change that.

src/ocean/core/SmartUnion.d
gavin-norman-sociomantic: >Hm, why didn't you put this method in SmartUnion so you don't have to derive the union type?

```
    Note: declared at module-scope (rather than nested inside the SmartUnion
    template) to work around limitations of template alias parameters. (Doing it
    like this allows it to be called with a local name.)
```

gavin-norman-sociomantic:
Updated:
* CTFE names list.
* `alias typeof(smart_union._.u) U`

src/ocean/core/SmartUnion.d
```diff
@@ -220,6 +330,34 @@ private struct SmartUnionIntern ( U )
...
 
     Active active;
+
+    /***************************************************************************
+
+        List of active state names
+
+    ***************************************************************************/
+
+    istring[] active_names = member_string_list();
```
mathias-lang-sociomantic: `const` ?

src/ocean/core/SmartUnion.d
```diff
@@ -220,6 +330,34 @@ private struct SmartUnionIntern ( U )
...
+
+        CTFE function to generate the list of active state names for union U.
+
+        Returns:
+            a list containing the names of each of the active states of the
+            smart-union (i.e. the names of the fields of U)
+
+    ***************************************************************************/
+
+    static private istring[] member_string_list ( )
```
mathias-lang-sociomantic, line 352: `private static Immut!(istring[])` so you can assign it at CT ?

src/ocean/core/SmartUnion.d
```diff
@@ -179,6 +212,83 @@ version (UnitTest)
...
+            the provided smart-union
+        SU = type of smart-union to operate on
+        smart_union = smart-union instance whose active field should be passed
+            to Callable
+
+*******************************************************************************/
+
+public void callWithActive ( alias Callable, SU ) ( SU smart_union )
+{
+    static assert(is(TemplateInstanceArgs!(SmartUnion, SU)));
```
david-eckardt-sociomantic, line 236: You like to go on number secure :)

src/ocean/core/SmartUnion.d
david-eckardt-sociomantic: :+1: but I think it is actually assigned at CT already.

mathias-lang-sociomantic: No it's not, because it isn't assigned to a manifest constant (` istring[] active_names = member_string_list();`).

src/ocean/core/SmartUnion.d
gavin-norman-sociomantic: I don't want people passing any old crap to my function!

src/ocean/core/SmartUnion.d
mihails-strasuns-sociomantic: All initializers in D are required to be CTFE-evaluated, not matter if LHS is constant or not. Also it is allowed to implicitly convert any kind of mutability qualifier from RHS to the one required by LHS upon initialization for that exact reason.

src/ocean/core/SmartUnion.d
david-eckardt-sociomantic: Funktionstürsteher

src/ocean/core/SmartUnion.d
gavin-norman-sociomantic: Updated.

src/ocean/core/SmartUnion.d
```diff
@@ -103,6 +95,31 @@ struct SmartUnion ( U )
 ///
 unittest
 {
+    union MyUnion
+    {
+        int    x;
```
nemanja-boric-sociomantic: nitpick: indentation here is screwed.

src/ocean/core/SmartUnion.d
```diff
@@ -226,6 +253,34 @@ private struct SmartUnionIntern ( U )
...
+
+        Returns:
+            a list containing the names of each of the active states of the
+            smart-union (i.e. the names of the fields of U)
+
+    ***************************************************************************/
+
+    static private istring[] member_string_list ( )
+    {
+        istring[] names = ["none"[]];
```
nemanja-boric-sociomantic, line 354: I wonder if the empty string would be a better choice here (if the user is allowed to have the `none` as a member)?

src/ocean/core/SmartUnion.d
```diff
@@ -212,6 +212,83 @@ version (UnitTest)
...
+    }
+}
+
+///
+version (D_Version2) unittest
+{
+    // This example is D2 only because it requires a function template,
+    // `print`, and D1 doesn't allow defining a function template at the scope
+    // of a function, including a `unittest`. In D1 this example works if
+    // `print`, is defined outside of function scope.
```
nemanja-boric-sociomantic, line 262: "outside of `unittest` scope"?

nemanja-boric-sociomantic (approved):
LGTM

src/ocean/core/SmartUnion.d
gavin-norman-sociomantic: The smart-union enum member is actually called `none`, so this is entirely accurate. https://github.com/gavin-norman-sociomantic/ocean/blob/52961089d40181042c1680ad7af43166c4d34861/src/ocean/core/SmartUnion.d#L247

src/ocean/core/SmartUnion.d
gavin-norman-sociomantic: Hm, it's already mentioned that function scope = unittest scope.

src/ocean/core/SmartUnion.d
nemanja-boric-sociomantic: Sorry, I've missed this!

src/ocean/core/SmartUnion.d
nemanja-boric-sociomantic: Yeah, true, it's just noise from me.

src/ocean/core/SmartUnion.d
gavin-norman-sociomantic: Ha oh yeah. Updated.

nemanja-boric-sociomantic:
Maybe @david-eckardt-sociomantic wants to give it a final look?

david-eckardt-sociomantic (approved):

PR #30, opened 2017-03-10
Fix flaky signalfd tests

### LINKED ISSUE #25: Fix and re-enable disabled integration test
#24 disables the `signalfd` test due to a level of flakiness. This should be fixed / rewritten and re-enabled.

nemanja-boric-sociomantic:
Because child may handle only some of the signals that parent has sent
(when handle returns in the middle of the parent sending signals), we
need to make sure that we will handle signals in child only after parent
sent all signals to the child.

Fixes #25

### DIFF
--- a/Build.mak
+++ b/Build.mak
@@ -112,11 +112,6 @@ TEST_FILTER_OUT += \
 # This is an integration test that depends on Collectd -- Don't run it
 TEST_FILTER_OUT += $C/test/collectd/main.d
 
-# integration test which is temporarily disabled due to flakiness
-# to be fixed and re-enabled
-TEST_FILTER_OUT += \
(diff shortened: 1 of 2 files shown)


nemanja-boric-sociomantic:
I will need to change `Build.mak` after merging `2.5.x` into `v2.x.x`

nemanja-boric-sociomantic:
Updated with reenabling the test in `Build.mak`

mihails-strasuns-sociomantic (approved):

nemanja-boric-sociomantic:
Ping

mihails-strasuns-sociomantic:
I was hoping someone more familiar with signals will chime in but that seems like naive expectation ;)

PR #32, opened 2017-03-10
StopWatch: Use the monotonic clock, fix minor bugs

david-eckardt-sociomantic:
- Use the monotonic timer rather than the wall clock time to avoid incorrect results if the clock is adjusted.
- Avoid huge time differences resulting from integer overflow if the clock is inconsistent.
- Remove `version(Posix)`, tidy up imports, correct documentation.

### DIFF
--- a/src/ocean/time/StopWatch.d
+++ b/src/ocean/time/StopWatch.d
@@ -43,13 +43,23 @@ extern (C) private
         // ...
 
         double i = elapsed.stop;
+        ulong us = elapsed.microsec;
         ---
 
-        The measured interval is in units of seconds, using floating-
-        point to represent fractions. This approach is more flexible
-        than integer arithmetic since it migrates trivially to more
-        capable timer hardware (there no implicit granularity to the
-        measurable intervals, except the limits of fp representation)
+        The measured interval is either an integer value in units of
+        microseconds -- returned by `microsec` -- or a floating-point value in
+        units of seconds with fractions -- returned by `stop`.
+        The integer value has always a precision of 1µs and is recommended if
+        you want to compare it with reference values such as checking if it's
+        below or above 1ms (e.g. `elapsed.microsec <= 1000`).
+        Although floating-point representation seems often more convenient, bear
+        in mind that
+          - The precision is relative to the value (i.e. the greater the values
+            the lower the absolute precision).
+          - Comparing fractions can yield unexpected results due to rounding and
+            because decimal fractions have no exact binary floating-point
+            representation. To avoid surprises like this using the integer
+            represenation of time spans is in general recommended.
 
         There is some minor overhead in using StopWatch, so take that into
         account
--- a/src/ocean/time/StopWatch.d
+++ b/src/ocean/time/StopWatch.d
@@ -95,7 +95,7 @@ public struct StopWatch
 
         double stop ()
         {
-                return multiplier * (timer - started);
+                return multiplier * this.microsec;
         }
 
         /***********************************************************************
@@ -106,7 +106,7 @@ public struct StopWatch
 
         ulong microsec ()
         {
-                 return (timer - started);
+                 return (timer >= started)? (timer - started) : 0;
         }
 
         /***********************************************************************
--- a/src/ocean/time/StopWatch.d
+++ b/src/ocean/time/StopWatch.d
@@ -17,6 +17,19 @@
 
 module ocean.time.StopWatch;
 
+// CLOCK_MONOTONIC and clock_gettime will be added to core.sys.posix.time in
+// tangort v1.7.0, see tangort issue #6.
+import core.sys.posix.time; // clockid_t, timespec
+extern (C) private
+{
+    enum: clockid_t
+    {
+        CLOCK_MONOTONIC = 1
+    }
+
+    int clock_gettime(clockid_t clk_id, timespec* t);
+}
+
 /*******************************************************************************
 
         Timer for measuring small intervals, such as the duration of a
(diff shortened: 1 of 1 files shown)


src/ocean/time/StopWatch.d
```diff
@@ -50,6 +49,11 @@ module ocean.time.StopWatch;
 public struct StopWatch
 {
         import ocean.stdc.posix.sys.time;
+        // TODO: Import clock_gettime and CLOCK_MONOTONIC on the next tangort
```
nemanja-boric-sociomantic: Perhaps be more clear here with `on tangort v1.7.0` release.

src/ocean/time/StopWatch.d
```diff
@@ -17,6 +17,9 @@
 
 module ocean.time.StopWatch;
 
+// TODO: Remove this on the next tangort release.
```
nemanja-boric-sociomantic: Perhaps be more clear here with on tangort v1.7.0 release.

src/ocean/time/StopWatch.d
```diff
@@ -99,7 +99,7 @@ public struct StopWatch
 
         ulong microsec ()
         {
-                 return (timer - started);
+                 return (timer >= started)? (timer - started) : 0;
```
nemanja-boric-sociomantic: I guess this would be more an underflow than overflow :-)

nemanja-boric-sociomantic (approved):

nemanja-boric-sociomantic:
I wonder if we need release notes for this one. I don't feel like it's needed, because code now does what it was expected.

src/ocean/time/StopWatch.d
```diff
@@ -17,6 +17,9 @@
```
gavin-norman-sociomantic, line 17: What's the relevance of the comment "On Linux clock_getres(CLOCK_MONOTONIC) reports a resolution of 1ns."? Maybe explain this a little bit more.

src/ocean/time/StopWatch.d
david-eckardt-sociomantic: Righ, but AFAIK _overflow_ is used for both directions.

nemanja-boric-sociomantic: Yes, I think you're right. I can't find any authoritative source on _integer underflow_ (as oposed to the floating point one) :+1:

src/ocean/time/StopWatch.d
```diff
@@ -60,8 +58,15 @@ version (Posix)
 
 public struct StopWatch
 {
+        import ocean.stdc.posix.sys.time;
+        // TODO: Import clock_gettime and CLOCK_MONOTONIC on tangort v1.7.0
```
mihails-strasuns-sociomantic: How about "TODO" now? :)

david-eckardt-sociomantic: That was quick.

nemanja-boric-sociomantic: We haven't done it yet :-)

src/ocean/time/StopWatch.d
david-eckardt-sociomantic: It is the reason why I removed the paragraph in the documentation claiming on Linux the time resolution would be 1µs at best.

src/ocean/time/StopWatch.d
david-eckardt-sociomantic: Too late :P

src/ocean/time/StopWatch.d
gavin-norman-sociomantic: Yeah I guessed so. It's not clear from the commit message, though. Could you add a little more explanation?

gavin-norman-sociomantic:
`./src/ocean/time/StopWatch.d(107): Error: undefined identifier clock_gettime`

nemanja-boric-sociomantic:
I will release runtime tomorrow.

david-eckardt-sociomantic:
Updated, adding the missing C bindings to the `StopWatch` module and rebasing.

david-eckardt-sociomantic:
ping

nemanja-boric-sociomantic:
@gavin-norman-sociomantic had some comments, are they resolved?

david-eckardt-sociomantic:
Ah yes, one still needs to be addressed.

david-eckardt-sociomantic:
Updated, addressing the last comment.

src/ocean/time/StopWatch.d
```diff
@@ -96,11 +107,11 @@ public struct StopWatch
...
         {
-                timeval tv;
-                if (gettimeofday (&tv, null))
-                    throw new PlatformException ("Timer :: linux timer is not available");
+                timespec t;
+                if (clock_gettime(CLOCK_MONOTONIC, &t))
+                    throw new PlatformException ("Timer :: CLOCK_MONOTONIC is not available");
 
-                return (cast(ulong) tv.tv_sec * 1_000_000) + tv.tv_usec;
+                return t.tv_sec * 1_000_000_000UL + t.tv_nsec / 1_000UL;
```
nemanja-boric-sociomantic: Should this be `tv_sec * 1_000_000 + t.tv_nsec / 1_000` ? Looks like you're adding nanoseconds (from the first member) and microseconds (from the second member).

david-eckardt-sociomantic:
Updated, fixing the `tv_sec` factor (to 1E6, was 1E9)

nemanja-boric-sociomantic:
Changed base to v2.6.x.

PR #34, opened 2017-03-11
Provide the ability to specify custom layouts from the config

mathias-lang-sociomantic:
Issue reported by @daniel-zullo-sociomantic :

> Some configureLoggers() were deprecated in v2.6.0 and the suggested method to use instead configureOldLoggers() does not deal with the template parameters FileLayout and ConsoleLayout.

This provides a more generic approach, less tied to the internals of this function (and that can be done at runtime).

CC @daniel-zullo-sociomantic @hatem-oraby-sociomantic for testing / sanity.

### DIFF
--- a/relnotes/makelayout.feature.md
+++ b/relnotes/makelayout.feature.md
@@ -1,0 +1,37 @@
+* `ocean.util.log.Config`
+
+    An extra overload of the `configureLogger` and `configureOldLoggers` functions
+    were introduced, with an additional `makeLayout` parameter.
+    This parameter is used to instantiate a `Layout` from a name, allowing
+    users to have custom layouts usable from their config files.
+    For example, a config entry would be:
+    ```
+    [LOG.myapp.supermodule]
+    level = info
+    file  = log/supermodule.log
+    file_layout = aquatic
+    console_layout = submarine
+    ```
+    And calling `configureOldLoggers` should be done as such:
+    ```
+    void myConfigureLoggers (
+        ClassIterator!(Config, ConfigParser) config,
+        MetaConfig m_config,
+        Appender delegate ( istring file, Layout layout ) file_appender,
+        bool use_insert_appender = false)
+    {
+        Layout makeLayout (cstring name)
+        {
+            if (name == "aquatic")
+                return new AquaticLayout;
+            if (name == "submarine")
+                return new SubmarineLayout;
+            return ocean.util.log.Config.newLayout(name);
+        }
+        ocean.util.log.Config.configureOldLoggers(config, m_config,
+            file_appender, &makeLayout, use_insert_appender);
+    }
+    ```
+    Note that the previous example is kept short for simplicity. `makeLayout` gets the raw name,
+    and as such may make the comparison as loose as wanted
+    (by lowercasing, removing `"layout"` prefix/suffix, etc...).
--- a/relnotes/makelayout.feature.md
+++ b/relnotes/makelayout.feature.md
@@ -35,3 +35,12 @@
     Note that the previous example is kept short for simplicity. `makeLayout` gets the raw name,
     and as such may make the comparison as loose as wanted
     (by lowercasing, removing `"layout"` prefix/suffix, etc...).
+
+* `ocean.util.app.LogExt : LogExt`
+
+    This class has a new constructor, which accepts a delegate with the same signature as `makeLayout`
+    previously described, and will pass it to `configureOldLoggers` accordingly.
+
+* `ocean.util.app.DaemonApp : DaemonApp.OptionalSettings`
+
+    This struct gained a new member, `make_layout`, which is passed to `LogExt` and ultimately `configureOldLoggers`.
--- a/src/ocean/util/app/DaemonApp.d
+++ b/src/ocean/util/app/DaemonApp.d
@@ -218,6 +218,7 @@ public abstract class DaemonApp : Application,
 
     public static struct OptionalSettings
     {
+        import ocean.util.log.Appender;
         import core.sys.posix.signal : SIGHUP;
 
         /***********************************************************************
@@ -287,6 +288,9 @@ public abstract class DaemonApp : Application,
         ***********************************************************************/
 
         int reopen_signal = SIGHUP;
+
+        /// Delegate for LogExt that instantiates a `Appender.Layout` from a name
+        Appender.Layout delegate (cstring name) make_layout;
     }
 
     /***************************************************************************
@@ -363,7 +367,8 @@ public abstract class DaemonApp : Application,
         this.args_ext.registerExtension(this.config_ext);
 
         // Create and register log extension
-        this.log_ext = new LogExt(settings.use_insert_appender);
+        this.log_ext = new LogExt(settings.make_layout,
+                                  settings.use_insert_appender);
         this.config_ext.registerExtension(this.log_ext);
 
         // Create and register version extension
--- a/src/ocean/util/app/ext/LogExt.d
+++ b/src/ocean/util/app/ext/LogExt.d
@@ -79,6 +79,10 @@ class LogExt : IConfigExtExtension
     public bool use_insert_appender;
 
 
+    /// If set, will be used by `makeLayout` to map a name to a `Layout`
+    private Appender.Layout delegate (cstring name) layout_maker;
+
+
     /***************************************************************************
 
         Constructor.
@@ -89,11 +93,33 @@ class LogExt : IConfigExtExtension
 
     ***************************************************************************/
 
-    this ( bool use_insert_appender = false )
+    this ( bool use_insert_appender )
     {
-        this.use_insert_appender = use_insert_appender;
+        this(null, use_insert_appender);
     }
 
+
+    /***************************************************************************
+
+        Constructor.
+
+        Params:
+            make_layout         = A delegate that instantiates an
+                                  `Appender.Layout` from a name. If null,
+                                  defaults to `ocean.util.Config: newLayout`.
+            use_insert_appender = true if the InsertConsole appender should be
+                                  used (needed when using the AppStatus module)
+
+    ***************************************************************************/
+
+    public this ( Appender.Layout delegate (cstring name) make_layout = null,
+                  bool use_insert_appender = false )
+    {
+        this.layout_maker = make_layout is null ? &this.makeLayoutDefault
+            : make_layout;
+
+        this.use_insert_appender = use_insert_appender;
+    }
 
     /***************************************************************************
 
@@ -148,7 +174,7 @@ class LogExt : IConfigExtExtension
         enable_loose_parsing(conf_ext.loose_config_parsing);
 
         LogUtil.configureOldLoggers(log_config, log_meta_config, &appender,
-            this.use_insert_appender);
+            this.layout_maker, this.use_insert_appender);
 
         foreach (ext; this.extensions)
         {
@@ -198,5 +224,27 @@ class LogExt : IConfigExtExtension
     {
         // Unused
         return files;
+    }
+
+
+    /***************************************************************************
+
+        Default for layout_maker. DMD complains because `LogUtil.newLayout`
+        is a function, not a delegate.
+
+        Params:
+            name = name of the Layout to instantiate.
+
+        Throws:
+            `Exception` if it cannot match the name
+
+        Returns:
+            A new Layout instance matching name
+
+    ***************************************************************************/
+
+    private Appender.Layout makeLayoutDefault ( cstring name )
+    {
+        return LogUtil.newLayout(name);
     }
 }
--- a/src/ocean/util/log/Config.d
+++ b/src/ocean/util/log/Config.d
@@ -319,7 +319,7 @@ public Layout newLayout ( cstring layout_str )
             // yields `cstring` instead of `istring`.
             istring msg = "Invalid layout requested : ";
             msg ~= layout_str;
-            throw new Exception(msg);
+            throw new Exception(msg, __FILE__, __LINE__);
     }
 
     return layout;
--- a/src/ocean/util/log/Config.d
+++ b/src/ocean/util/log/Config.d
@@ -277,7 +277,7 @@ alias Appender.Layout Layout;
 
 *******************************************************************************/
 
-public Layout newLayout ( istring layout_str )
+public Layout newLayout ( cstring layout_str )
 {
     Layout layout;
 
@@ -315,7 +315,11 @@ public Layout newLayout ( istring layout_str )
             break;
 
         default:
-            throw new Exception("Invalid layout requested : " ~ layout_str);
+            // Has to be 2 statements because `istring ~ cstring`
+            // yields `cstring` instead of `istring`.
+            istring msg = "Invalid layout requested : ";
+            msg ~= layout_str;
+            throw new Exception(msg);
     }
 
     return layout;
@@ -427,11 +431,15 @@ public void configureLoggers ( Source = ConfigParser, FileLayout = LayoutDate,
 
     enable_loose_parsing(loose);
 
+    scope Layout delegate(cstring) makeLayout
+        = (cstring v) { return newLayout(v); };
+
     configureLoggers!(Logger, Source, FileLayout, ConsoleLayout)
         (config, m_config,
          (cstring n) { return !n.length ? Log.root : Log.lookup(n); },
          file_appender,
-         (Layout l) { return console_appender_fn(use_insert_appender, l); });
+         (Layout l) { return console_appender_fn(use_insert_appender, l); },
+         makeLayout);
 }
 
 
@@ -457,6 +465,39 @@ public void configureOldLoggers (
     Appender delegate ( istring file, Layout layout ) file_appender,
     bool use_insert_appender = false)
 {
+    configureOldLoggers(config, m_config, file_appender,
+        (cstring v) { return newLayout(v); }, use_insert_appender);
+}
+
+
+/*******************************************************************************
+
+    Sets up logging configuration for `ocean.util.log.Log`
+
+    Calls the provided `file_appender` delegate once per log being configured
+    and passes the returned `Appender` to the `Logger.add` method.
+
+    This is an extra overload because using a delegate literal as a parameter's
+    default argument causes linker error in D1.
+
+    Params:
+        config   = an instance of an class iterator for Config
+        m_config = an instance of the MetaConfig class
+        file_appender = delegate which returns appender instances to write to
+                        a file
+        makeLayout = A delegate that returns a `Layout` instance from
+                     a name, or throws on error.
+                     By default, wraps `ocean.util.log.Config.newLayout`
+        use_insert_appender = true if the InsertConsole appender should be used
+                              (needed when using the AppStatus module)
+
+*******************************************************************************/
+
+public void configureOldLoggers (
+    ClassIterator!(Config, ConfigParser) config, MetaConfig m_config,
+    Appender delegate ( istring file, Layout layout ) file_appender,
+    Layout delegate (cstring) makeLayout, bool use_insert_appender = false)
+{
     // DMD1 cannot infer the common type between both return, we have to work
     // around it...
     static Appender console_appender_fn (bool insert_appender, Layout l)
@@ -476,9 +517,35 @@ public void configureOldLoggers (
                        { return console_appender_fn(use_insert_appender, l); };
 
     configureLoggers!(Logger, ConfigParser, LayoutDate, LayoutSimple)
-        (config, m_config, lookup, file_appender, appender_dg);
+        (config, m_config, lookup, file_appender, appender_dg, makeLayout);
 }
 
+
+///
+unittest
+{
+    // In a real app those would be full-fledged implementation
+    alias LayoutSimple AquaticLayout;
+    alias LayoutSimple SubmarineLayout;
+
+    void myConfigureLoggers (
+        ClassIterator!(Config, ConfigParser) config,
+        MetaConfig m_config,
+        Appender delegate ( istring file, Layout layout ) file_appender,
+        bool use_insert_appender = false)
+    {
+        Layout makeLayout (cstring name)
+        {
+            if (name == "aquatic")
+                return new AquaticLayout;
+            if (name == "submarine")
+                return new SubmarineLayout;
+            return ocean.util.log.Config.newLayout(name);
+        }
+        ocean.util.log.Config.configureOldLoggers(config, m_config,
+            file_appender, &makeLayout, use_insert_appender);
+    }
+}
 
 /*******************************************************************************
 
@@ -502,6 +569,8 @@ public void configureOldLoggers (
         console_appender = Delegate which returns an Appender suitable to use
                            as console appender. Might not be called if console
                            writing is disabled.
+        makeLayout      = A delegate that returns a `Layout` instance from
+                          a name, or throws on error.
 
 *******************************************************************************/
 
@@ -511,7 +580,8 @@ private void configureLoggers
     (ClassIterator!(Config, Source) config, MetaConfig m_config,
      LoggerT delegate (cstring name) lookup,
      Appender delegate (istring file, Layout layout) file_appender,
-     Appender delegate (Layout) console_appender)
+     Appender delegate (Layout) console_appender,
+     Layout delegate (cstring) makeLayout)
 {
     // It is important to ensure that parent loggers are configured before child
     // loggers. This is because parent loggers will override the settings of
@@ -568,7 +638,8 @@ private void configureLoggers
         configureLogger!(FileLayout, ConsoleLayout, LoggerT)
             (lookup(name), settings, name,
              file_appender, console_appender,
-             console_enabled, syslog_enabled, m_config.buffer_size);
+             console_enabled, syslog_enabled, m_config.buffer_size,
+             makeLayout);
     }
 }
 
@@ -596,6 +667,9 @@ private void configureLoggers
         console_enabled = `true` if a console appender should be added (by
                           calling `console_enabled`).
         syslog_enabled  = `true` if a syslog appender should be added.
+        makeLayout      = A delegate that returns a `Layout` instance from
+                          a name, or throws on error.
+                          By default, wraps `ocean.util.log.Config.newLayout`.
 
 *******************************************************************************/
 
@@ -605,7 +679,8 @@ public void configureLogger
     (LoggerT log, Config settings, istring name,
      Appender delegate ( istring file, Layout layout ) file_appender,
      Appender delegate (Layout) console_appender,
-     bool console_enabled, bool syslog_enabled, size_t buffer_size)
+     bool console_enabled, bool syslog_enabled, size_t buffer_size,
+     Layout delegate (cstring) makeLayout = (cstring v) { return newLayout(v); })
 {
     if (settings.buffer_size)
         buffer_size = settings.buffer_size;
@@ -623,7 +698,7 @@ public void configureLogger
     if (settings.file.set)
     {
         Layout file_log_layout = (settings.file_layout.length)
-            ? newLayout(settings.file_layout)
+            ? makeLayout(settings.file_layout)
             : new FileLayout;
         log.add(file_appender(settings.file(), file_log_layout));
     }
@@ -634,7 +709,7 @@ public void configureLogger
     if (console_enabled)
     {
         Layout console_log_layout = (settings.console_layout.length)
-            ? newLayout(settings.console_layout)
+            ? makeLayout(settings.console_layout)
             : new ConsoleLayout;
         log.add(console_appender(console_log_layout));
     }


mathias-lang-sociomantic:
Note: A similar fix will need to be applied to v3.x.x (and the merge will conflict since it's touching deprecated code).

relnotes/makelayout.migration.md
```diff
@@ -0,0 +1,43 @@
...
+        }
+        ocean.util.log.Config.configureOldLoggers(config, m_config,
+            file_appender, use_insert_appender, makeLayout);
+    }
+    ```
+    Note that the previous example is kept short for simplicity. `makeLayout` gets the raw name,
+    and as such should make sure to make the comparison as loose as wanted (by lowercasing,
+    removing `"layout"` prefix/suffix, etc...).
+
+    In addition, the members `console_layout` and `file_layout` in the `Config` class
```
gavin-norman-sociomantic: I wonder, would it be possible to leave the default values as an empty string and add an extra check that converts `""` -> `"simple"` or `"date"` in the point where the fields are referred to, in `configureLogger`? Would that remove the need for this migration instruction and make this a pure new feature?

mathias-lang-sociomantic: I think I got mixed up, because this is v3.0.0 material. Removed it.

mathias-lang-sociomantic:
Updated:
```diff
diff --git a/relnotes/makelayout.feature.md b/relnotes/makelayout.feature.md
index f92a097..dabb75b 100644
--- a/relnotes/makelayout.feature.md
+++ b/relnotes/makelayout.feature.md
@@ -1,6 +1,6 @@
 * `ocean.util.log.Config`
 
-    The `configureLogger`, `configureLoggers` overloads and `configureOldLoggers`
+    The `configureLogger` and `configureOldLoggers` functions
     have gained an additional `makeLayout` parameter.
     This parameter is used to instantiate a `Layout` from a name, allowing
     users to have custom layout usable from their config files.
@@ -35,9 +35,3 @@
     Note that the previous example is kept short for simplicity. `makeLayout` gets the raw name,
     and as such should make sure to make the comparison as loose as wanted (by lowercasing,
     removing `"layout"` prefix/suffix, etc...).
-
-    In addition, the members `console_layout` and `file_layout` in the `Config` class
-    (used to configure a logger instance) now default to `"simple"` and `"date"`, respectively.
-    This means that if any code relying on them being empty to instantiate the default values
-    (previously provided as `FileLayout` and `ConsoleLayout` template parameters) should now
-    rely on providing a cusom `makeLayout` parameter.
diff --git a/src/ocean/util/log/Config.d b/src/ocean/util/log/Config.d
index 18e9e58..ff5ec23 100644
--- a/src/ocean/util/log/Config.d
+++ b/src/ocean/util/log/Config.d
@@ -151,7 +151,7 @@ class Config
 
     ***************************************************************************/
 
-    public istring console_layout = "simple";
+    public istring console_layout;
 
     /***************************************************************************
 
@@ -167,7 +167,7 @@ class Config
 
     ***************************************************************************/
 
-    public istring file_layout = "date";
+    public istring file_layout;
 
     /***************************************************************************
 ```

First sentence of relnotes changed because it was wrong, `configureLoggers` wasn't changed (at least not the public ones).

File: src/ocean/util/log/Config.d [hunk 2]
```diff
@@ -447,8 +453,11 @@ public void configureLoggers ( Source = ConfigParser, FileLayout = LayoutDate,
         m_config = an instance of the MetaConfig class
         file_appender = delegate which returns appender instances to write to
                         a file
+        makeLayout      = A delegate that returns a `Layout` instance from
+                          a name, or throw on error.
+                          By default, wraps `ocean.util.log.Config.newLayout`
         use_insert_appender = true if the InsertConsole appender should be used
```
daniel-zullo: the `makeLayout` parameter only exists for the function that overloads `ConfigureOldLoggers()`

src/ocean/util/log/Config.d
```diff
@@ -457,6 +466,18 @@ public void configureOldLoggers (
...
+    // link failures in D1
+    configureOldLoggers(config, m_config, file_appender,
+        (cstring v) { return newLayout(v); }, use_insert_appender);
+}
+
+/// Ditto
+public void configureOldLoggers (
+    ClassIterator!(Config, ConfigParser) config, MetaConfig m_config,
+    Appender delegate ( istring file, Layout layout ) file_appender,
+    Layout delegate (cstring) makeLayout, bool use_insert_appender = false)
```
daniel-zullo: The order of the last two parameters does not seem to be correct for the overloading function. I think it should be the other way around (`use_insert_appender` should go first)

daniel-zullo (commented):
I'm about to test this patch in one application. Anyway I've done a couple of suggestions that you might want to address.

relnotes/makelayout.feature.md
```diff
@@ -0,0 +1,37 @@
...
+    ```
+    And calling `configureOldLoggers` should be done as such:
+    ```
+    void myConfigureLoggers (
+        ClassIterator!(Config, ConfigParser) config,
+        MetaConfig m_config,
+        Appender delegate ( istring file, Layout layout ) file_appender,
+        bool use_insert_appender = false)
+    {
+        scope makeLayout = (cstring name)
```
daniel-zullo: I don't think this is correct. What would be the type of `makeLayout`? It might work for D2 though

relnotes/makelayout.feature.md
```diff
@@ -0,0 +1,37 @@
...
+            if (name == "aquatic")
+                return new AquaticLayout;
+            if (name == "submarine")
+                return new SubmarineLayout;
+            return ocean.util.log.Config.newLayout(name);
+        }
+        ocean.util.log.Config.configureOldLoggers(config, m_config,
+            file_appender, use_insert_appender, makeLayout);
+    }
+    ```
```
daniel-zullo: Wouldn't it be more convenient to add the example in a unit test rather than in the release notes? Or maybe in both places

src/ocean/util/log/Config.d
mathias-lang-sociomantic: It cannot go first because it has a default value

src/ocean/util/log/Config.d
mathias-lang-sociomantic: This comment is on L450 / 459, and I'm not sure what the problem is ? This is the documentation for `configureOldLoggers`.

relnotes/makelayout.feature.md
mathias-lang-sociomantic: `scope Layout delegate (cstring)`

D1 is capable of inferring it (there are some bugs, but usually it works).
Maybe it's surprising because we usually use it with `class`es, but it works just the same with delegates.
Would writing the complete type make it more clear ?

src/ocean/util/log/Config.d
daniel-zullo: I just wanted to point out that `makeLayout` only exists in one of the `configureOldLoggers()` and it might not be right to have the same documentation for both methods when the signatures are different.

mathias-lang-sociomantic: Ah I see. I don't know if we have any guidelines about it. It's the style used in Phobos AFAIR, and DDOC just seem to handle it correctly.

relnotes/makelayout.feature.md
daniel-zullo: > D1 is capable of inferring it (there are some bugs, but usually it works).

Then I am afraid it will be more reliable to define the complete type (and clear).
The complete type might look a bit long though
```D
scope Layout delegate(cstring) makeLayout = (cstring name)
...
```

src/ocean/util/log/Config.d
daniel-zullo: Right, i wonder if it makes sense and it's doable to provide a default value for `makeLayout`
(I am not a big fan of switching the order of the parameters when overloading a function)

mathias-lang-sociomantic:
So we tried it with @daniel-zullo-sociomantic and hit a little snag: In order to work reliably, it also needs to change `LogExt` to use the user defined `makeLayout`.

I was wondering how to better handle this: By introducting something specific to `makeLayout` (e.g. an extra parameter to`DaemonApp`'s ctor), something a bit more generic towards `LogExt` (e.g. a ctor that accepts a `LogExt` / `protected void setupLogExtension` function), or something else ?
CC @gavin-norman-sociomantic

src/ocean/util/log/Config.d
gavin-norman-sociomantic: This is one reason I don't like `/// ditto` generally -- it's usually an over-simplification. (There must be _some_ difference between the two things that's worth explaining, otherwise why have them both?)

gavin-norman-sociomantic:
A new field in [`DaemonApp.OptionalSettings`](https://github.com/sociomantic-tsunami/ocean/blob/v2.x.x/src/ocean/util/app/DaemonApp.d#L219) perhaps?

src/ocean/util/log/Config.d
mathias-lang-sociomantic: > otherwise why have them both

Originally I went with a single overload with 2 default parameter, but then the linker said no (IOW, I hit a DMD bug and it's a workaround).

src/ocean/util/log/Config.d
mathias-lang-sociomantic: See my comment [here](https://github.com/sociomantic-tsunami/ocean/pull/34#discussion_r106218478). I tried the default argument approach but hit a bug.
It's mentioned at L469 - 470:
```diff
+    // Using a dg literal as default argument in parameter list causes
+    // link failures in D1
```

mathias-lang-sociomantic:
> A new field in DaemonApp.OptionalSettings perhaps?

Done. Will test with @daniel-zullo-sociomantic tomorrow.

daniel-zullo:
The current patch is running smoothly in one application. You could check the reference to the workaround done in the application side to make it work (it's a hacky version). 
So I'd say now we can proceed with the review.

src/ocean/util/log/Config.d
```diff
@@ -596,6 +621,9 @@ private void configureLoggers
         console_enabled = `true` if a console appender should be added (by
                           calling `console_enabled`).
         syslog_enabled  = `true` if a syslog appender should be added.
+        makeLayout      = A delegate that returns a `Layout` instance from
+                          a name, or throw on error.
```
daniel-zullo: throw**s**?

src/ocean/util/log/Config.d
```diff
@@ -633,8 +662,8 @@ public void configureLogger
 
     if (console_enabled)
     {
-        Layout console_log_layout = (settings.console_layout.length)
-            ? newLayout(settings.console_layout)
+       Layout console_log_layout = (settings.console_layout.length)
```
daniel-zullo: The indentation seems to be wrong in this line

src/ocean/util/log/Config.d
```diff
@@ -502,6 +523,8 @@ public void configureOldLoggers (
         console_appender = Delegate which returns an Appender suitable to use
                            as console appender. Might not be called if console
                            writing is disabled.
+        makeLayout      = A delegate that returns a `Layout` instance from
+                          a name, or throw on error.
```
daniel-zullo: ditto if my similar comment is taken

src/ocean/util/log/Config.d
```diff
@@ -447,8 +453,11 @@ public void configureLoggers ( Source = ConfigParser, FileLayout = LayoutDate,
         m_config = an instance of the MetaConfig class
         file_appender = delegate which returns appender instances to write to
                         a file
+        makeLayout      = A delegate that returns a `Layout` instance from
+                          a name, or throw on error.
+                          By default, wraps `ocean.util.log.Config.newLayout`
         use_insert_appender = true if the InsertConsole appender should be used
-                              (needed when using the AppStatus module)
+                              (needed when using the AppStatus module).
```
daniel-zullo: the change in this line might not be related.

src/ocean/util/app/ext/LogExt.d
```diff
@@ -199,4 +215,26 @@ class LogExt : IConfigExtExtension
         // Unused
         return files;
     }
+
+
+    /***************************************************************************
+
+        Default for layout_maker. DMD complains. because `LogUtil.newLayout`
```
daniel-zullo: You might want to delete the dot `.` appended to `complains`

src/ocean/util/app/ext/LogExt.d
```diff
@@ -79,22 +79,38 @@ class LogExt : IConfigExtExtension
...
+    this ( bool use_insert_appender )
     {
-        this.use_insert_appender = use_insert_appender;
+        this(null, use_insert_appender);
     }
 
 
+    /// Ditto
+    public this ( Appender.Layout delegate (cstring name) make_layout = null,
+                 bool use_insert_appender = false)
```
daniel-zullo: The code style of the ctor signature doesn't seem to be consistent. Either you keep space for open/closed parenthesis or you remove for both of them

src/ocean/util/app/ext/LogExt.d
```diff
@@ -79,22 +79,38 @@ class LogExt : IConfigExtExtension
...
+
+
     /***************************************************************************
 
         Constructor.
 
         Params:
+            make_layout         = A delegate that instantiate an
+                                  `Appender.Layout` from a name. If null,
+                                   defaults to `ocean.util.Config: newLayout`.
```
daniel-zullo: Please check indentation. Also this documentation should be added to the constructor below and not here. Again the `ditto` situation...

src/ocean/util/app/DaemonApp.d
```diff
@@ -287,6 +288,9 @@ public abstract class DaemonApp : Application,
         ***********************************************************************/
 
         int reopen_signal = SIGHUP;
+
+        /// Delegate for LogExt that instantiate a `Appender.Layout` from a name
```
daniel-zullo: instantiate**s**?

relnotes/makelayout.feature.md
```diff
@@ -0,0 +1,46 @@
+* `ocean.util.log.Config`
+
+    The `configureLogger` and `configureOldLoggers` functions
+    have gained an additional `makeLayout` parameter.
+    This parameter is used to instantiate a `Layout` from a name, allowing
+    users to have custom layout usable from their config files.
```
daniel-zullo: either `a custom layout` or `custom layouts`

relnotes/makelayout.feature.md
```diff
@@ -0,0 +1,46 @@
...
+            if (name == "submarine")
+                return new SubmarineLayout;
+            return ocean.util.log.Config.newLayout(name);
+        }
+        ocean.util.log.Config.configureOldLoggers(config, m_config,
+            file_appender, use_insert_appender, makeLayout);
+    }
+    ```
+    Note that the previous example is kept short for simplicity. `makeLayout` gets the raw name,
+    and as such should make sure to make the comparison as loose as wanted (by lowercasing,
```
daniel-zullo: Maybe replace `make sure` to `ensure`.

src/ocean/util/app/ext/LogExt.d
```diff
@@ -79,22 +79,38 @@ class LogExt : IConfigExtExtension
...
             use_insert_appender = true if the InsertConsole appender should be
                                   used (needed when using the AppStatus module)
 
     ***************************************************************************/
 
-    this ( bool use_insert_appender = false )
+    this ( bool use_insert_appender )
```
daniel-zullo, line 96: Why the default value is removed here?

src/ocean/util/app/ext/LogExt.d
```diff
@@ -79,22 +79,38 @@ class LogExt : IConfigExtExtension
...
+        this(null, use_insert_appender);
     }
 
 
+    /// Ditto
+    public this ( Appender.Layout delegate (cstring name) make_layout = null,
+                 bool use_insert_appender = false)
+    {
+        this.layout_maker = make_layout !is null ? make_layout
+            : &this.makeLayoutDefault;
```
daniel-zullo: What about doing it the other way around:
```D
    this.layout_maker = make_layout is null ? &this.makeLayoutDefault
        : make_layout;
```

src/ocean/util/log/Config.d
```diff
@@ -447,8 +453,11 @@ public void configureLoggers ( Source = ConfigParser, FileLayout = LayoutDate,
         m_config = an instance of the MetaConfig class
         file_appender = delegate which returns appender instances to write to
                         a file
+        makeLayout      = A delegate that returns a `Layout` instance from
+                          a name, or throw on error.
```
daniel-zullo: throw**s**?

src/ocean/util/app/ext/LogExt.d
```diff
@@ -199,4 +215,26 @@ class LogExt : IConfigExtExtension
...
     }
+
+
+    /***************************************************************************
+
+        Default for layout_maker. DMD complains. because `LogUtil.newLayout`
+        is a function, now a delegate.
+
+        Params:
+            name = Name of the Layout to instantiate.
```
daniel-zullo: `Name` is Capitalized

src/ocean/util/app/ext/LogExt.d
```diff
@@ -199,4 +215,26 @@ class LogExt : IConfigExtExtension
         // Unused
         return files;
     }
+
+
+    /***************************************************************************
+
+        Default for layout_maker. DMD complains. because `LogUtil.newLayout`
+        is a function, now a delegate.
```
daniel-zullo: `now a delegate`? what do you mean here?

daniel-zullo (commented):
I've finished my review round. I just have some minor comments and a couple of questions. We are also still discussing a couple of issues related to the design of this feature.

src/ocean/util/app/ext/LogExt.d
mathias-lang-sociomantic: Because there are 2 ways to call this ctor:
`this()` and `this(bool)`.
I don't want a breaking change, so `this(bool)` needs to still work, but `this()` now calls the other ctor (which is compatible).

src/ocean/util/app/ext/LogExt.d
mathias-lang-sociomantic: Typo, should be `not`.

src/ocean/util/app/ext/LogExt.d
daniel-zullo: Got it, thanks for the clarification.

src/ocean/util/log/Config.d
mathias-lang-sociomantic: It was wrong before actually.

relnotes/makelayout.feature.md
mathias-lang-sociomantic: Is the current wording incorrect ? Ensure sounds more like an absolute requirement, which this is a pretty loose area.

mathias-lang-sociomantic:
Updated accorded to @daniel-zullo-sociomantic 's (except the ones I answered to).

src/ocean/util/app/ext/LogExt.d
```diff
@@ -79,22 +79,39 @@ class LogExt : IConfigExtExtension
...
+        this(null, use_insert_appender);
     }
 
 
+    /// Ditto
+    public this ( Appender.Layout delegate (cstring name) make_layout = null,
+                 bool use_insert_appender = false )
+    {
+        this.layout_maker = make_layout is null ? &this.makeLayoutDefault
+            : make_layout
```
daniel-zullo: semicolon is missing at the end of this line

src/ocean/util/log/Config.d
daniel-zullo: mmm...I don't think so. You might need to add one space before `Layout`

mathias-lang-sociomantic:
Fixed and green

src/ocean/util/app/ext/LogExt.d
mathias-lang-sociomantic: Forgot to mention, this was addressed.

src/ocean/util/log/Config.d
daniel-zullo: it seems this comment was not addressed, was it?

src/ocean/util/log/Config.d
daniel-zullo: also here you might want to append `s` to `throw`

src/ocean/util/log/Config.d
daniel-zullo: and this one?

src/ocean/util/log/Config.d
```diff
@@ -427,11 +429,15 @@ public void configureLoggers ( Source = ConfigParser, FileLayout = LayoutDate,
 
     enable_loose_parsing(loose);
 
+    scope Layout delegate(cstring) makeLayout
+                     = (cstring v) { return newLayout(v); };
```
daniel-zullo: it seems the indentation is not right

src/ocean/util/app/ext/LogExt.d
```diff
@@ -79,22 +79,39 @@ class LogExt : IConfigExtExtension
...
+    /// If set, will be used by `makeLayout` to map a name to a `Layout`
+    private Appender.Layout delegate (cstring name) layout_maker;
+
+
     /***************************************************************************
 
         Constructor.
 
         Params:
+            make_layout         = A delegate that instantiate an
```
daniel-zullo: instantiate**s**

daniel-zullo (commented):
I've made just a couple of cosmetic comments and LGTM. Also I've tested with one application and it works fine. Anyway it might be worth to have a full review from someone with more experience with the `Logger`

mathias-lang-sociomantic:
Addressed comments by @daniel-zullo-sociomantic

src/ocean/util/log/Config.d
mathias-lang-sociomantic: > mmm...I don't think so. You might need to add one space before `Layout`

One level of indent = 4 spaces. There were 5 before. I do not do this manually, my editor caught it.

mathias-lang-sociomantic:
Pingitty ping

nemanja-boric-sociomantic:
I'll look into it, but @gavin-norman-sociomantic has more experience with Config parsers.

src/ocean/util/log/Config.d
```diff
@@ -476,10 +497,35 @@ public void configureOldLoggers (
                        { return console_appender_fn(use_insert_appender, l); };
 
     configureLoggers!(Logger, ConfigParser, LayoutDate, LayoutSimple)
-        (config, m_config, lookup, file_appender, appender_dg);
+        (config, m_config, lookup, file_appender, appender_dg, makeLayout);
 }
 
 
+unittest
```
mathias-baumann-sociomantic: Should this maybe be a documented unittest, serving as example as well?

src/ocean/util/app/ext/LogExt.d
```diff
@@ -79,22 +79,39 @@ class LogExt : IConfigExtExtension
...
+    this ( bool use_insert_appender )
     {
-        this.use_insert_appender = use_insert_appender;
+        this(null, use_insert_appender);
     }
 
 
+    /// Ditto
+    public this ( Appender.Layout delegate (cstring name) make_layout = null,
+                 bool use_insert_appender = false )
```
mathias-baumann-sociomantic: I think you need one whitespace more here

mathias-baumann-sociomantic (changes requested):

mathias-lang-sociomantic:
Addressed @mathias-baumann-sociomantic 's comments

mathias-baumann-sociomantic (approved):

relnotes/makelayout.feature.md
gavin-norman-sociomantic: "and as such may make the comparison as loose as wanted" is clearer.

src/ocean/util/log/Config.d
```diff
@@ -315,7 +315,9 @@ public Layout newLayout ( istring layout_str )
             break;
 
         default:
-            throw new Exception("Invalid layout requested : " ~ layout_str);
+            istring msg = "Invalid layout requested : ";
```
gavin-norman-sociomantic: Why this code formatting change? Adding file and line is unrelated to this commit, right?

src/ocean/util/log/Config.d
gavin-norman-sociomantic: I'd go with duplicating the doc (don't use `/// ditto`) and mention the reason for the two overloads in the doc for the second (i.e. move it out of the method body and into its "header" doc).

relnotes/makelayout.feature.md
```diff
@@ -0,0 +1,37 @@
+* `ocean.util.log.Config`
+
+    The `configureLogger` and `configureOldLoggers` functions
```
gavin-norman-sociomantic: The way this is described "have gained" makes this sound like a breaking change. It should be made clear that these are overloads, not replacements.

gavin-norman-sociomantic (commented):
Apart from a few minor points, LGTM. (Sorry for the delay.)

src/ocean/util/log/Config.d
mathias-lang-sociomantic: The change was "necessary" because `istring ~ cstring` yields `cstring` (stupid, I know). Thus `throw new Exception("..." ~ layout_str);` stopped to compile after I changed `layout_str` to `cstring`.
The options, were to add a `cast` (unnecessary and prone to refactoring errors), use `format` (adding a new dependency seemed overkill), or just break down the statement. And since I was changing those line, I figured I might just add `__FILE__` and `__LINE__`.

gavin-norman-sociomantic: Ha gotcha. Pretty please split the file & line change into another commit? Just a minor thing, but I think that'll make the other change here -- which is actually connected to this commit -- easier to spot.

mathias-lang-sociomantic:
Updated:

```diff
diff --git a/relnotes/makelayout.feature.md b/relnotes/makelayout.feature.md
index e50d128..903a01f 100644
--- a/relnotes/makelayout.feature.md
+++ b/relnotes/makelayout.feature.md
@@ -1,7 +1,7 @@
 * `ocean.util.log.Config`
 
-    The `configureLogger` and `configureOldLoggers` functions
-    have gained an additional `makeLayout` parameter.
+    An extra overload of the `configureLogger` and `configureOldLoggers` functions
+    were introduced, with an additional `makeLayout` parameter.
     This parameter is used to instantiate a `Layout` from a name, allowing
     users to have custom layouts usable from their config files.
     For example, a config entry would be:
@@ -33,8 +33,8 @@
     }
     ```
     Note that the previous example is kept short for simplicity. `makeLayout` gets the raw name,
-    and as such should make sure to make the comparison as loose as wanted (by lowercasing,
-    removing `"layout"` prefix/suffix, etc...).
+    and as such may make the comparison as loose as wanted
+    (by lowercasing, removing `"layout"` prefix/suffix, etc...).
 
 * `ocean.util.app.LogExt : LogExt`
```

```diff
diff --git a/src/ocean/util/log/Config.d b/src/ocean/util/log/Config.d
index 8a861e3..8c7659a 100644
--- a/src/ocean/util/log/Config.d
+++ b/src/ocean/util/log/Config.d
@@ -453,9 +453,6 @@ public void configureLoggers ( Source = ConfigParser, FileLayout = LayoutDate,
         m_config = an instance of the MetaConfig class
         file_appender = delegate which returns appender instances to write to
                         a file
-        makeLayout      = A delegate that returns a `Layout` instance from
-                          a name, or throws on error.
-                          By default, wraps `ocean.util.log.Config.newLayout`
         use_insert_appender = true if the InsertConsole appender should be used
                               (needed when using the AppStatus module)
 
@@ -466,13 +463,34 @@ public void configureOldLoggers (
     Appender delegate ( istring file, Layout layout ) file_appender,
     bool use_insert_appender = false)
 {
-    // Using a dg literal as default argument in parameter list causes
-    // link failures in D1
     configureOldLoggers(config, m_config, file_appender,
         (cstring v) { return newLayout(v); }, use_insert_appender);
 }
 
-/// Ditto
+
+/*******************************************************************************
+
+    Sets up logging configuration for `ocean.util.log.Log`
+
+    Calls the provided `file_appender` delegate once per log being configured
+    and passes the returned `Appender` to the `Logger.add` method.
+
+    This is an extra overload because using a delegate literal as a parameter's
+    default argument causes linker error in D1.
+
+    Params:
+        config   = an instance of an class iterator for Config
+        m_config = an instance of the MetaConfig class
+        file_appender = delegate which returns appender instances to write to
+                        a file
+        makeLayout      = A delegate that returns a `Layout` instance from
+                          a name, or throws on error.
+                          By default, wraps `ocean.util.log.Config.newLayout`
+        use_insert_appender = true if the InsertConsole appender should be used
+                              (needed when using the AppStatus module)
+
+*******************************************************************************/
+
 public void configureOldLoggers (
     ClassIterator!(Config, ConfigParser) config, MetaConfig m_config,
     Appender delegate ( istring file, Layout layout ) file_appender,
```

mathias-lang-sociomantic:
Updated:

```diff
diff --git a/src/ocean/util/log/Config.d b/src/ocean/util/log/Config.d
index 8c7659a..a816a1f 100644
--- a/src/ocean/util/log/Config.d
+++ b/src/ocean/util/log/Config.d
@@ -315,6 +315,8 @@ public Layout newLayout ( cstring layout_str )
             break;
 
         default:
+            // Has to be 2 statements because `istring ~ cstring`
+            // yields `cstring` instead of `istring`.
             istring msg = "Invalid layout requested : ";
             msg ~= layout_str;
             throw new Exception(msg, __FILE__, __LINE__);
```

+ split the `__FILE__` and `__LINE__` additions.

src/ocean/util/log/Config.d
mathias-lang-sociomantic: Sure, done, and added a comment.

gavin-norman-sociomantic (approved):
LGTM! @daniel-zullo-sociomantic as the person who filed the issue, does this meet your approval too?

src/ocean/util/app/ext/LogExt.d
```diff
@@ -79,22 +79,39 @@ class LogExt : IConfigExtExtension
...
 
-    this ( bool use_insert_appender = false )
+    this ( bool use_insert_appender )
     {
-        this.use_insert_appender = use_insert_appender;
+        this(null, use_insert_appender);
     }
 
 
+    /// Ditto
```
daniel-zullo: You might not use Ditto here (to be consistent with other similar situations in other functions). So new documentation needs to be added to this constructor and the parameter `make_layout`  should be removed from the constructor above

src/ocean/util/log/Config.d
```diff
@@ -457,6 +465,39 @@ public void configureOldLoggers (
...
+    Params:
+        config   = an instance of an class iterator for Config
+        m_config = an instance of the MetaConfig class
+        file_appender = delegate which returns appender instances to write to
+                        a file
+        makeLayout      = A delegate that returns a `Layout` instance from
+                          a name, or throws on error.
+                          By default, wraps `ocean.util.log.Config.newLayout`
+        use_insert_appender = true if the InsertConsole appender should be used
+                              (needed when using the AppStatus module)
```
daniel-zullo: The format of the documentation needs to be well formatted.

daniel-zullo (commented):
Apart from my 2 minor comments LGTM

src/ocean/util/log/Config.d
mathias-lang-sociomantic: It is literaly a copy/paste of what's before, just adding a parameter to the list. I don't see how it's badly formatted ?

mathias-lang-sociomantic:
Updated:

```diff
diff --git a/src/ocean/util/app/ext/LogExt.d b/src/ocean/util/app/ext/LogExt.d
index e8851af..ce27ffc 100644
--- a/src/ocean/util/app/ext/LogExt.d
+++ b/src/ocean/util/app/ext/LogExt.d
@@ -88,9 +88,6 @@ class LogExt : IConfigExtExtension
         Constructor.
 
         Params:
-            make_layout         = A delegate that instantiates an
-                                  `Appender.Layout` from a name. If null,
-                                  defaults to `ocean.util.Config: newLayout`.
             use_insert_appender = true if the InsertConsole appender should be
                                   used (needed when using the AppStatus module)
 
@@ -102,7 +99,19 @@ class LogExt : IConfigExtExtension
     }
 
 
-    /// Ditto
+    /***************************************************************************
+
+        Constructor.
+
+        Params:
+            make_layout         = A delegate that instantiates an
+                                  `Appender.Layout` from a name. If null,
+                                  defaults to `ocean.util.Config: newLayout`.
+            use_insert_appender = true if the InsertConsole appender should be
+                                  used (needed when using the AppStatus module)
+
+    ***************************************************************************/
+
     public this ( Appender.Layout delegate (cstring name) make_layout = null,
                   bool use_insert_appender = false )
     {
```

src/ocean/util/log/Config.d
daniel-zullo: I think `makeLayout` does not have the right format and that is a new parameter (the spaces between `makeLayout`  and `=` are not consistent to the other parameters)

mathias-lang-sociomantic:
Updated:

```diff
diff --git a/src/ocean/util/log/Config.d b/src/ocean/util/log/Config.d
index a816a1f..04cfa1a 100644
--- a/src/ocean/util/log/Config.d
+++ b/src/ocean/util/log/Config.d
@@ -485,9 +485,9 @@ public void configureOldLoggers (
         m_config = an instance of the MetaConfig class
         file_appender = delegate which returns appender instances to write to
                         a file
-        makeLayout      = A delegate that returns a `Layout` instance from
-                          a name, or throws on error.
-                          By default, wraps `ocean.util.log.Config.newLayout`
+        makeLayout = A delegate that returns a `Layout` instance from
+                     a name, or throws on error.
+                     By default, wraps `ocean.util.log.Config.newLayout`
         use_insert_appender = true if the InsertConsole appender should be used
                               (needed when using the AppStatus module)
 ```

daniel-zullo (approved):
LGTM, many thanks!

mathias-lang-sociomantic:
Thanks everyone for the review :)

PR #38, opened 2017-03-13
Add new function awaitRunningTasks()

gkotian:
This function suspends the current task until all tasks in the pool have
finished executing.

### DIFF
--- a/relnotes/taskpool.feature.md
+++ b/relnotes/taskpool.feature.md
@@ -1,0 +1,7 @@
+* `ocean.task.TaskPool`
+
+  A new function `awaitRunningTasks()` has been added. This function suspends
+  the current task until all running tasks in the pool have finished executing.
+
+  It is assumed that the current task (i.e. the one being suspended) is not
+  itself a task from the pool.
--- a/src/ocean/task/TaskPool.d
+++ b/src/ocean/task/TaskPool.d
@@ -33,12 +33,14 @@ import ocean.transition;
 import ocean.task.Task;
 import ocean.task.Scheduler;
 
+import ocean.core.Enforce;
 import ocean.core.Traits;
 import ocean.util.container.pool.ObjectPool;
 
 version (UnitTest)
 {
     import ocean.core.Test;
+    import ocean.task.util.Timer;
 }
 
 /*******************************************************************************
@@ -151,6 +153,45 @@ class TaskPool ( TaskT : Task ) : ObjectPool!(Task)
         }
 
         return true;
+    }
+
+    /***************************************************************************
+
+        Suspends the current task until all running tasks in the pool have
+        finished executing.
+
+        Note: care should be taken to ensure that the current task (i.e. the one
+        to be suspended) is not itself a task from the pool. If that is the
+        case, the current task will never be resumed, and this function will
+        never return.
+
+    ***************************************************************************/
+
+    public void awaitRunningTasks ()
+    {
+        if (!this.num_busy())
+            return;
+
+        auto current_task = Task.getThis();
+        enforce(current_task !is null,
+            "Current task is null in TaskPool.awaitRunningTasks");
+
+        scope tasks_iterator = this.new BusyItemsIterator;
+        int count;
+
+        foreach (task; tasks_iterator)
+        {
+            ++count;
+
+            task.terminationHook({
+                --count;
+                if (count == 0)
+                    current_task.resume();
+            });
+        }
+
+        if (count > 0)
+            current_task.suspend();
     }
 
     static if( hasMethod!(TaskT, "deserialize", void delegate(void[])) )
@@ -273,4 +314,50 @@ unittest
     theScheduler.eventLoop();
     test!("==")(RecycleTask.recycle_count, 2,
         "RecycleTask.recycle was not called the correct number of times");
+}
+
+unittest
+{
+    // Test for waiting until all running tasks of a task pool finish executing.
+
+    static class AwaitTask : Task
+    {
+        static int value;
+
+        public void copyArguments ( )
+        {
+        }
+
+        override public void run ( )
+        {
+            .wait(1); // so that the task gets suspended
+            value++;
+        }
+    }
+
+    class MainTask : Task
+    {
+        TaskPool!(AwaitTask) my_task_pool;
+
+        public this ( )
+        {
+            this.my_task_pool = new TaskPool!(AwaitTask);
+        }
+
+        override protected void run ( )
+        {
+            const NUM_START_CALLS = 6;
+
+            for (uint i; i < NUM_START_CALLS; i++)
+                this.my_task_pool.start();
+
+            this.my_task_pool.awaitRunningTasks();
+
+            test!("==")(AwaitTask.value, NUM_START_CALLS);
+        }
+    }
+
+    initScheduler(SchedulerConfiguration.init);
+    theScheduler.schedule(new MainTask);
+    theScheduler.eventLoop();
 }


src/ocean/task/TaskPool.d
```diff
@@ -153,6 +153,36 @@ class TaskPool ( TaskT : Task ) : ObjectPool!(Task)
...
+
+    ***************************************************************************/
+
+    public void awaitAllTasks ()
+    {
+        scope iter = this.new ReadOnlyBusyItemsIterator;
+        int count;
+
+        auto task = Task.getThis();
+        assert(task !is null, "Null task in TaskPool.await");
```
nemanja-boric-sociomantic: `awaitAllTasks`

relnotes/taskpool.feature.md
```diff
@@ -0,0 +1,4 @@
+* `ocean.task.TaskPool`
+
+  A new function `awaitAllTasks()` has been added. This function suspends the
+  current task until all tasks in the pool have finished executing.
```
gavin-norman-sociomantic: Not having looked at the code, I find myself wondering at this point "all tasks including the task that's being suspended?" (= impossible).

src/ocean/task/TaskPool.d
```diff
@@ -153,6 +153,36 @@ class TaskPool ( TaskT : Task ) : ObjectPool!(Task)
...
+
+        foreach (item; iter)
+        {
+            ++count;
+
+            item.terminationHook({
+                --count;
+                if (count == 0)
+                    task.resume();
```
nemanja-boric-sociomantic: Not sure if `if (count == 0 && task.suspended())` would be appropriate here.

mihails-strasuns-sociomantic: No, there is no `assign` behaviour here - main task is suspended unconditionally.

gkotian: I guess that would make sense.

src/ocean/task/TaskPool.d
```diff
@@ -153,6 +153,36 @@ class TaskPool ( TaskT : Task ) : ObjectPool!(Task)
...
     }
 
+    /***************************************************************************
+
+        Suspends the current task until all tasks in the pool have finished
+        executing.
+
+    ***************************************************************************/
+
+    public void awaitAllTasks ()
```
mihails-strasuns-sociomantic: I suggest name `awaitRunningTasks` to represent its meaning better.

src/ocean/task/TaskPool.d
```diff
@@ -153,6 +153,36 @@ class TaskPool ( TaskT : Task ) : ObjectPool!(Task)
...
+    {
+        scope iter = this.new ReadOnlyBusyItemsIterator;
```
mihails-strasuns-sociomantic: My memory is vague here - does `ReadOnly` here implies you shouldn't modify task items? (because you do)

src/ocean/task/TaskPool.d
mihails-strasuns-sociomantic: I'd also replace this with sanity `enforce` as there are no other safeguards here.

relnotes/taskpool.feature.md
gkotian: I see what you mean. In my particular case, the current task is not a task from the pool at all. It is simply a task which owns the pool. Any suggestion for how to make this better?

gkotian:
Fixed typo in assert message and added additional condition before attempting `resume()`.

mihails-strasuns-sociomantic (changes requested):

relnotes/taskpool.feature.md
gavin-norman-sociomantic: How about: "This function suspends the current task until all tasks in the pool have finished executing. It is assumed that the current task (the one being suspended) is _not_ a task in the pool." ?

Maybe add that to the function doc, as well. (I also wonder about asserting that this is the case, but I'm not sure if there's a way to do that without iterating over the whole pool, which seems a bit excessive.)

src/ocean/task/TaskPool.d
```diff
@@ -153,6 +153,36 @@ class TaskPool ( TaskT : Task ) : ObjectPool!(Task)
...
+        assert(task !is null, "Null task in TaskPool.awaitAllTasks");
+
+        foreach (item; iter)
+        {
+            ++count;
+
+            item.terminationHook({
+                --count;
+                if ((count == 0) && task.suspended())
+                    task.resume();
```
daniel-zullo: You might want to add `task.resume();` between curly brackets

daniel-zullo (commented):
Missing unit tests for the new functionality (at least one unit test should be added).

src/ocean/task/TaskPool.d
```diff
@@ -175,12 +180,16 @@ class TaskPool ( TaskT : Task ) : ObjectPool!(Task)
...
                 --count;
                 if ((count == 0) && task.suspended())
+                {
                     task.resume();
+                }
             });
         }
 
         if (count > 0)
+        {
```
gavin-norman-sociomantic: Our style guide doesn't require this btw.

src/ocean/task/TaskPool.d
gkotian: Yep you're right. I overlooked the fact that I'm modifying the pool item when I add the termination hook. I should be using `BusyItemsIterator` instead.

relnotes/taskpool.feature.md
gkotian: I'll update the release notes and function documentation as you suggested.

>I also wonder about asserting that this is the case, but I'm not sure if there's a way to do that without iterating over the whole pool, which seems a bit excessive.

We're iterating over all the busy tasks in the pool anyway, so we'd just need to extend that to iterate over all tasks in the pool. The additional safety that the assertion would provide would make it worthwhile imo.

The tricky part when iterating over all tasks is to identify whether a task is busy or not. For this, I'll need to add a new `isBusy()` function to the API, which would mean touching a couple of other modules.

So for now, I'd say we add this `awaitRunningTasks()` function with just the note in the documentation, and I'll submit another PR to add the assertion after this is merged. How does that sound?

src/ocean/task/TaskPool.d
gkotian: Yeah I did it based on a review comment. I don't really care either way.

relnotes/taskpool.feature.md
gavin-norman-sociomantic: Makes sense :+1:

gkotian:
Updated with:
- `s/awaitAllTasks/awaitRunningTasks`
- Add note to function doc & release notes that it is assumed that the current task (i.e. the one being suspended) is not itself a task from the pool.
- Check in the beginning of the function whether the pool has any busy items at all, and return early if it doesn't
- `s/ReadOnlyBusyItemsIterator/BusyItemsIterator`
- `s/assert/enforce`

Will add a unit test in a bit.

src/ocean/task/TaskPool.d
daniel-zullo: Well I made the comment for the `if` statement defined above and not for this one :)
And I still think it was the right thing to do, wasn't it?
```D
if ((count == 0) && task.suspended())
    task.resume();
```

mihails-strasuns-sociomantic: You really don't to check if it is suspended at that point, it can only be otherwise in case of programming error so better plain `resume` and let assert fire.

gkotian: >Well I made the comment for the if statement defined above and not for this one :)

Ah yeah, true. Sorry I mixed them up.

>so better plain resume and let assert fire.

OK, I'll remove the extra `task.suspended()` check.

gkotian:
Added a new unit test block to test for the new functionality.

src/ocean/task/TaskPool.d
```diff
@@ -274,3 +315,47 @@ unittest
...
+        TaskPool!(AwaitTask) my_task_pool;
+
+        public this ( )
+        {
+            this.my_task_pool = new TaskPool!(AwaitTask);
+        }
+
+        override protected void run ( )
+        {
+            this.my_task_pool.start();
```
mihails-strasuns-sociomantic: Running `start` 2-3 times here while keeping rest of test code the same will be a better check in my opinion (the one that ensures that _all_ tasks are waited for)

mihails-strasuns-sociomantic:
One minor test tweak and LGTM

src/ocean/task/TaskPool.d
gkotian: Good idea.

Wouldn't it be even better if I incremented `value` by one in each task? Then the final value should be equal to the number of times I called `this.my_task_pool.start()`. I'll push a fixup commit with this change, and you can tell me what you think about it.

mihails-strasuns-sociomantic: Yep, looks good!

gkotian:
Let me know if the fixup commit looks good and I'll squash it.

gkotian:
>Yep, looks good!

Fixup commit squashed.

mihails-strasuns-sociomantic (approved):

Issue #36, opened 2017-03-13
`TimerExt` should allow for GC-free allocations


nemanja-boric-sociomantic:
Currently, `TimerExt` of `DaemonApp` is based on the `TimerSet`, which with few more inheritance layer goes back to `TimeoutManager`. `TimeoutManager` by default uses `BucketElementGCAllocator` for allocating internals, which uses GC to allocate.

However, `TimeoutManager` allows for configuring this, so another allocation strategies may be used, such malloc/free via `BucketElementMallocAllocator`, for example.

For the applications which want to avoid GC activity, it should be allowed to use some strategy other than `BucketElementGCAllocator`.

nemanja-boric-sociomantic:
cc @vasileios-kyrillidis-sociomantic

gavin-norman-sociomantic:
Interesting. Is there an actual use-case for this? I mean, I agree that it's theoretically possible, but also feel that such implementation details should be avoided at the level of the app classes, wherever possible.

As far as I've seen, the `TimerExt` is normally just used at app startup to set up a few timers.

nemanja-boric-sociomantic:
There is an actual use-case for this. It all comes down to:

> As far as I've seen, the TimerExt is normally just used at app startup to set up a few timers.

apparently not being true. Application will actually allocate on every timer firing event (I will dig a little bit more to find out why), which generates a lot of opportunity for GC collections, even if there's no other GC activity in the application

nemanja-boric-sociomantic:
This is the map that actually allocates underneath: https://github.com/sociomantic-tsunami/ocean/blob/v2.x.x/src/ocean/time/timeout/TimeoutManager.d#L229

nemanja-boric-sociomantic:
Periodic stacktrace:


```
#0  gc_malloc (size=25, attrs=0, ptrmap=...) at /home/jenkins/docker/src/gc/cdgc/gc.d:2583
#1  0x00000000009ef89e in _d_newarrayiT (ti=0xdb0bc0 <TypeInfo_PS5ocean4util9container3map5model6Bucket346__T6BucketVm8TPS5ocean4util9container6ebtree8EBTree6413__T8EBTree64Z8EBTree64260__T4NodeTS5ocean4util9container6ebtree8EBTree649eb64_nodeTmS122_D5ocean4util9container6ebtree8EBTree6413__T8EBTr
[...]

PR #39, opened 2017-03-13
Use FreeList-based allocation for TimerExt internals

nemanja-boric-sociomantic:
Previously, TimerExt of DaemonApp is based on the TimerSet, which with
few more inheritance layer goes back to TimeoutManager. TimeoutManager
by default uses BucketElementGCAllocator for allocating internals, which
uses GC to allocate.

However, TimeoutManager allows for configuring this, so another
allocation strategies may be used, such malloc/free via
BucketElementMallocAllocator, for example.

In order to minimize GC activity, BucketElementFreeList allocator
is now used for TimerExt internal allocation.

Fixes sociomantic-tsunami/ocean#36

### DIFF
--- a/src/ocean/io/select/timeout/TimerEventTimeoutManager.d
+++ b/src/ocean/io/select/timeout/TimerEventTimeoutManager.d
@@ -75,6 +75,8 @@ static this ( )
 
 class TimerEventTimeoutManager : TimeoutManager
 {
+    import ocean.util.container.map.model.IAllocator;
+
     /***************************************************************************
 
         TimerEvent for absolute real-time that calls checkTimeouts() when fired.
@@ -138,10 +140,16 @@ class TimerEventTimeoutManager : TimeoutManager
 
         Constructor
 
+        Params:
+            allocator = use this bucket element allocator for the expiry
+                registration to ISelectClient map. If it is null the default map
+                allocator (BucketElementGCAllocator) is used.
+
     ***************************************************************************/
 
-    public this ( )
+    public this ( IAllocator allocator = null )
     {
+        super(allocator);
         this.event = this.new TimerEvent;
     }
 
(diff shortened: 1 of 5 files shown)


andrew-stevenson-sociomantic:
LGTM!

gavin-norman-sociomantic:
The change looks fine, but I would like to understand why the GC-based allocator allocates every time an event fires. That sounds wrong to me.

nemanja-boric-sociomantic:
Yes, I agree. See https://github.com/sociomantic-tsunami/ocean/issues/36#issuecomment-286375005

src/ocean/time/timeout/TimeoutManager.d
```diff
@@ -128,6 +128,20 @@ class TimeoutManager : TimeoutManagerBase
 
     /***************************************************************************
 
+        Constructor.
+
```
gavin-norman-sociomantic, line 141: Params:

src/ocean/io/select/timeout/TimerEventTimeoutManager.d
```diff
@@ -137,11 +139,15 @@ class TimerEventTimeoutManager : TimeoutManager
     /***************************************************************************
 
         Constructor
+            allocator = use this bucket element allocator for the expiry
```
gavin-norman-sociomantic: Params:

gavin-norman-sociomantic (commented):
Well done for navigating the horrors of the timeout manager!

src/ocean/time/timeout/TimeoutManager.d
nemanja-boric-sociomantic: Whoops!

nemanja-boric-sociomantic:
I've updated it with missing `Params:` and by extracting literal `1024` to `TimeoutManager.default_expected_no_elements` (let me know if you don't like it).

gavin-norman-sociomantic:
>extracting literal 1024 to TimeoutManager.default_expected_no_elements (let me know if you don't like it).

Ha, I almost commented on that before ;)

gavin-norman-sociomantic (approved):

nemanja-boric-sociomantic:
Green, merge at will.

PR #47, opened 2017-03-17
Pass inotify_handler to the delegate as a pointer

nemanja-boric-sociomantic:
inotify_handler is declared as:

```
struct
{
    int,
    uint,
    uint,
    uint len,
    char[0] name
}
```

where `name`'s length is `len` bytes (and is null-terminated in
addition). Dereferencing the pointer and passing it to the delegate
inside opApply doesn't copy the `name` field, so all the data behind it
is lost. Instead, passing by pointer is appropriate.

### DIFF
--- a/src/ocean/sys/Inotify.d
+++ b/src/ocean/sys/Inotify.d
@@ -105,7 +105,7 @@ public class Inotify : ISelectable
 
         ***********************************************************************/
 
-        public int opApply ( int delegate ( ref inotify_event ) dg )
+        public int opApply ( int delegate ( ref inotify_event* ) dg )
         {
             //255 is the default max filename length in linux)
             char[inotify_event.sizeof + 255 + 1] buffer_temp;
@@ -121,7 +121,7 @@ public class Inotify : ISelectable
                 for ( uint i; i < read_bytes; i += inotify_event.sizeof + i_event.len  )
                 {
                     i_event = cast(inotify_event*) &buffer[i];
-                    result = dg(*i_event);
+                    result = dg(i_event);
 
                     if (result)
                     {


gavin-norman-sociomantic:
`char[0]` ?

mihails-strasuns-sociomantic:
@gavin-norman-sociomantic it is C idiom for variadic length structs

mihails-strasuns-sociomantic:
@nemanja-boric-sociomantic maybe you can add `version(D_Version2) { @disable this(this); }` to ensure it is non-copyable by D2 compiler?

nemanja-boric-sociomantic:
> @nemanja-boric-sociomantic maybe you can add `version(D_Version2) { @disable this(this); }` to ensure it is non-copyable by D2 compiler?

Very nice suggestion, will update.

nemanja-boric-sociomantic:
Turns out, `inotify_event` is defined in runtime. I would try to submit a PR there.

nemanja-boric-sociomantic:
https://github.com/dlang/druntime/pull/1795

mihails-strasuns-sociomantic (approved):

schveiguy:
I don't know the context of this (I came over from the druntime PR, which I approve and agree with), but this wasn't exactly a bug, because inotify struct was still passed by reference to the delegate, no?

mihails-strasuns-sociomantic:
My understanding was that it still resulted in creation of temporary copy on the call site, but I haven't checked generated assembly.

nemanja-boric-sociomantic:
If I remember correctly, problem was in the line

```
    i_event = cast(inotify_event*)...;
    auto new_event = *i_event; // temporary copy here
    dg(new_event);
````

So, the actual dereferencing the pointer was creating temporary copy, but I need to confirm that now, you've put the bug in me ;-)

**Comment by nemanja-boric-sociomantic** [comment]
Yes, so the simple printout confirms that after you dereference the pointer to the `inotify_event` and pass it to delegate, delegate actually gets the different different chunk of memory (passed by reference, that's true), not including the data following the original pointer:

```
i_event = 7ffe9715dfb0 // pointer to our buffer containing the read data
&event = 7ffe9715de50 // inside delegate
```

**Comment by schveiguy** [comment]
Hm... don't see that in the diff. Diff shows this code:

```D
                    i_event = cast(inotify_event*) &buffer[i];
                    result = dg(*i_event);
```

Which should have passed just the pointer anyway.

nemanja-boric-sociomantic:
@schveiguy I've just got to work and took a look into this. You're most certainly right, the problem was not in that piece of the code - at the `opApply` definition site, the pointer is still passed.

However, the problem stands in the user's side. Instead of `foreach (el; inotify.readData());` user is forced to do `foreach (ref el; inotify.readData);` Former will create the copy of the structure.

I didn't know (and I still don't know) the language construct to enforce user to accept by reference in their foreach, so I passed the pointer instead. I'll make a new PR with better documentation, thank you very much!

Example code:

```d
struct S
{
    int x;
    char[0] y;

}

struct Fruct
{
    int opApply (int delegate (ref S) dg)
    {
        ubyte[64] arr = 1;
        S* s = cast(S*)&arr[0];

        dg(*s);
        return 0;
    }
}

void main()
{
    Fruct f;

    foreach (ref el; f)
    {
        // Passes
        assert (*cast(ubyte*)&el.y == 1);
    }

    foreach (el; f)
    {
        // Fails
        assert (*cast(ubyte*)&el.y == 1);
    }
}
```

schveiguy:
You are quite right! I had forgotten the compiler allows that implicitly. I think if you overload opapply you can direct what happens for ref vs non ref, but it would be nice if the compiler implicitly or forced the user to use ref.

PR #48, opened 2017-03-17
Enforce that the task calling `awaitRunningTasks()` doesn't belong to the pool

gkotian:
As promised in https://github.com/sociomantic-tsunami/ocean/pull/38#discussion_r106163458

### DIFF
--- a/relnotes/ipool.feature.md
+++ b/relnotes/ipool.feature.md
@@ -1,0 +1,4 @@
+* `ocean.util.container.pool.model.IPool`
+
+  A new function `isBusy()` has been added to the API of IPool. It can be used
+  to check whether a given pool item is currently busy or idle.
--- a/src/ocean/task/TaskPool.d
+++ b/src/ocean/task/TaskPool.d
@@ -160,10 +160,14 @@ class TaskPool ( TaskT : Task ) : ObjectPool!(Task)
         Suspends the current task until all running tasks in the pool have
         finished executing.
 
-        Note: care should be taken to ensure that the current task (i.e. the one
-        to be suspended) is not itself a task from the pool. If that is the
-        case, the current task will never be resumed, and this function will
-        never return.
+        Note: it is important to ensure that the current task (i.e. the one to
+        be suspended) is not itself a task from the pool. If that were allowed,
+        the current task would never get resumed, and this function would never
+        return.
+
+        Throws:
+            if the current task is null or if the current task belongs to the
+            task pool
 
     ***************************************************************************/
 
@@ -181,6 +185,9 @@ class TaskPool ( TaskT : Task ) : ObjectPool!(Task)
 
         foreach (task; tasks_iterator)
         {
+            enforce(!this.isSame(this.toItem(current_task), this.toItem(task)),
+                "Current task cannot be from the pool of tasks to wait upon");
+
             if (!this.isBusy(this.toItem(task)))
                 continue;
 
@@ -363,4 +370,37 @@ unittest
     initScheduler(SchedulerConfiguration.init);
     theScheduler.schedule(new MainTask);
     theScheduler.eventLoop();
+}
+
+unittest
+{
+    // Test that 'awaitRunningTasks()' cannot be called from a task that itself
+    // belongs to the pool.
+
+    static class AwaitTask : Task
+    {
+        void delegate () dg;
+
+        public void copyArguments ( void delegate () dg )
+        {
+            this.dg = dg;
+        }
+
+        override public void run ( )
+        {
+            testThrown!(Exception)(this.dg());
+        }
+    }
+
+    auto pool = new TaskPool!(AwaitTask);
+    initScheduler(SchedulerConfiguration.init);
+
+    pool.start(
+        {
+            // This delegate should throw as it will attempt to call the task
+            // pool's awaitRunningTasks() method from within a task that itself
+            // belongs to the pool.
+            pool.awaitRunningTasks();
+        }
+    );
 }
--- a/src/ocean/task/TaskPool.d
+++ b/src/ocean/task/TaskPool.d
@@ -176,11 +176,14 @@ class TaskPool ( TaskT : Task ) : ObjectPool!(Task)
         enforce(current_task !is null,
             "Current task is null in TaskPool.awaitRunningTasks");
 
-        scope tasks_iterator = this.new BusyItemsIterator;
+        scope tasks_iterator = this.new AllItemsIterator;
         int count;
 
         foreach (task; tasks_iterator)
         {
+            if (!this.isBusy(this.toItem(task)))
+                continue;
+
             ++count;
 
             task.terminationHook({
--- a/src/ocean/util/container/pool/model/IAggregatePool.d
+++ b/src/ocean/util/container/pool/model/IAggregatePool.d
@@ -204,32 +204,32 @@ public abstract class IAggregatePool ( T ) : IPool, IFreeList!(ItemType_!(T))
 
     public alias IPool.Item Item;
 
-    /**************************************************************************
+    /***************************************************************************
 
         List of items (objects) in pool for safe iteration. items is copied into
         this array on safe iterator instantiation.
 
-     **************************************************************************/
+    ***************************************************************************/
 
     protected Item[] iteration_items;
 
-    /**************************************************************************
+    /***************************************************************************
 
         true if a safe iterator instance exists currently, used for assertions
         to ensure that only a single safe iterator can exist at a time (as it
         uses the single buffer, iteration_items, above).
 
-     **************************************************************************/
+    ***************************************************************************/
 
     protected bool safe_iterator_open = false;
 
-    /**************************************************************************
+    /***************************************************************************
 
         Count of unsafe iterator instances which exist currently, used for
         assertions to ensure that while an unsafe iterator exists the object
         pool may not be modified.
 
-     **************************************************************************/
+    ***************************************************************************/
 
     protected size_t unsafe_iterators_open = 0;
 
@@ -319,7 +319,7 @@ public abstract class IAggregatePool ( T ) : IPool, IFreeList!(ItemType_!(T))
             return this.get(new T);
         }
 
-        /**********************************************************************
+        /***********************************************************************
 
             Ensures that the pool contains at least the specified number of
             items. Useful to pre-allocate a pool of a certain size.
@@ -330,7 +330,7 @@ public abstract class IAggregatePool ( T ) : IPool, IFreeList!(ItemType_!(T))
             Returns:
                 this
 
-         **********************************************************************/
+        ***********************************************************************/
 
         public typeof(this) fill ( size_t num )
         {
@@ -382,7 +382,7 @@ public abstract class IAggregatePool ( T ) : IPool, IFreeList!(ItemType_!(T))
         return this;
     }
 
-    /**************************************************************************
+    /***************************************************************************
 
         Recycles all items in the pool.
 
@@ -391,7 +391,7 @@ public abstract class IAggregatePool ( T ) : IPool, IFreeList!(ItemType_!(T))
         Returns:
             this instance
 
-     **************************************************************************/
+    ***************************************************************************/
 
     override public typeof(this) clear ( )
     {
@@ -401,7 +401,7 @@ public abstract class IAggregatePool ( T ) : IPool, IFreeList!(ItemType_!(T))
         return this;
     }
 
-    /**************************************************************************
+    /***************************************************************************
 
         Sets the limit of number of items in pool or disables limitation for
         limit = unlimited.
@@ -419,7 +419,7 @@ public abstract class IAggregatePool ( T ) : IPool, IFreeList!(ItemType_!(T))
             LimitExceededException if the number of busy pool items exceeds
             the desired limit.
 
-     **************************************************************************/
+    ***************************************************************************/
 
     override public size_t setLimit ( size_t limit )
     {
@@ -429,7 +429,7 @@ public abstract class IAggregatePool ( T ) : IPool, IFreeList!(ItemType_!(T))
         return super.setLimit(limit);
     }
 
-    /**************************************************************************
+    /***************************************************************************
 
         Obtains the n-th pool item. n must be less than the value returned by
         length().
@@ -444,7 +444,7 @@ public abstract class IAggregatePool ( T ) : IPool, IFreeList!(ItemType_!(T))
         Returns:
             n-th pool item
 
-    **************************************************************************/
+    ***************************************************************************/
 
     public ItemType opIndex ( size_t n )
     /+out (obj)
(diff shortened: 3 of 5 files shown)


relnotes/ipool.feature.md
```diff
@@ -0,0 +1,7 @@
+* `ocean.util.container.pool.model.IPool`, `ocean.util.container.pool.model.IAggregatePool`
+
+  A new abstract function `isBusy()` has been added to the API of IPool. Classes
+  wishing to inherit from `IPool` must provide an implementation for it.
```
nemanja-boric-sociomantic: `type-breaking`?

src/ocean/task/TaskPool.d
```diff
@@ -176,11 +176,14 @@ class TaskPool ( TaskT : Task ) : ObjectPool!(Task)
         enforce(current_task !is null,
             "Current task is null in TaskPool.awaitRunningTasks");
 
-        scope tasks_iterator = this.new BusyItemsIterator;
+        scope tasks_iterator = this.new AllItemsIterator;
         int count;
 
         foreach (task; tasks_iterator)
         {
+            if (!isBusy(this.toItem(task)))
```
nemanja-boric-sociomantic: `this.isBusy`?

src/ocean/task/TaskPool.d
```diff
@@ -181,6 +185,9 @@ class TaskPool ( TaskT : Task ) : ObjectPool!(Task)
 
         foreach (task; tasks_iterator)
         {
+            enforce(!isSame(this.toItem(current_task), this.toItem(task)),
```
nemanja-boric-sociomantic: `this.`?

nemanja-boric-sociomantic: I would go as far as asserting here?

nemanja-boric-sociomantic:
I wonder if, instead of checking at the iteration time, you could perform a search to find out if the current task is contained in a pool?

nemanja-boric-sociomantic:
i.e. if it would be better to provide `IAggregatePool.isOwner` or similar method.

relnotes/ipool.feature.md
gkotian: Yep, it'll be a breaking change for any class that inherits from `IPool`. Does that mean that this PR should then target `v3.x.x`?

nemanja-boric-sociomantic: Yes, it should target v3.x.x and the milestone would be v4.0.0. I wonder if this can be done differently (see my other comment), so we avoid breaking change.

src/ocean/task/TaskPool.d
gkotian: >I would go as far as asserting here?

Does it make a difference here? Using `enforce` allows me to unit test it using `testThrown()`, but I guess I won't be able to do that if I use `assert` instead.

relnotes/ipool.feature.md
mihails-strasuns-sociomantic: Why don't you move this method to `IPool` itself, possibly even marking as final? Both `num_busy` and `getItemIndex` are available in it.

mihails-strasuns-sociomantic (changes requested):

src/ocean/task/TaskPool.d
mihails-strasuns-sociomantic: `assert` is not applicable here by our guidelines as there are no other safety checks against the situation and it will be removed in `-release`.

nemanja-boric-sociomantic: Isn't it a programmer's error to end up in this situation? Is there any other way how would you end in this situation?

mihails-strasuns-sociomantic: > Isn't it a programmer's error to end up in this situation?

Yes and it doesn't make your life any better when deployed app silently hangs in production. You may need to re-read https://github.com/sociomantic/backend/blob/master/doc/coding/assert-vs-enforce.rst :P

relnotes/ipool.feature.md
gkotian: >Why don't you move this method to IPool itself, possibly even marking as final? Both num_busy and getItemIndex are available in it.

Ah yes indeed. I'm not sure why I didn't think of this simpler approach. Will try it out now.

src/ocean/util/container/pool/model/IAggregatePool.d
```diff
@@ -639,14 +639,31 @@ public abstract class IAggregatePool ( T ) : IPool, IFreeList!(ItemType_!(T))
...
+            item = item to check
+
+        Returs:
+            true if item is currently busy, false otherwise.
+
+    ***************************************************************************/
+
+    protected override bool isBusy ( Item item )
+    {
+        return this.getItemIndex(item) < this.num_busy_;
```
mihails-strasuns-sociomantic: Is it even correct? Are items moved in the pool when they are recycled?

src/ocean/task/TaskPool.d
nemanja-boric-sociomantic: Sure, I don't mind this, nor I will insist on `assert` here. I was just under impression that this check is very likely to fail the assertion during development time, where it can be safely removed from the production.

src/ocean/util/container/pool/model/IAggregatePool.d
nemanja-boric-sociomantic: looks like this might be correct, but it's implementation detail of `IPool` (one more vote for having it there: https://github.com/gkotian/ocean/blob/aab03da14b6012f85da7fc0cfe888a5878239f8e/src/ocean/util/container/pool/model/IPool.d#L413)

nemanja-boric-sociomantic: This is where the shuffling of the last busy element and recycled element is done: https://github.com/gkotian/ocean/blob/aab03da14b6012f85da7fc0cfe888a5878239f8e/src/ocean/util/container/pool/model/IPool.d#L480

gkotian:
Updated with `isBusy()` moved to `IPool`.

```diff
diff --git a/relnotes/ipool.feature.md b/relnotes/ipool.feature.md
index e1ea926..5470124 100644
--- a/relnotes/ipool.feature.md
+++ b/relnotes/ipool.feature.md
@@ -1,7 +1,4 @@
-* `ocean.util.container.pool.model.IPool`, `ocean.util.container.pool.model.IAggregatePool`
+* `ocean.util.container.pool.model.IPool`
 
-  A new abstract function `isBusy()` has been added to the API of IPool. Classes
-  wishing to inherit from `IPool` must provide an implementation for it.
-
-  The `IAggregatePool` class (which itself inherits from `IPool`) has been
-  updated to provide its implementation of the `isBusy()` method.
+  A new function `isBusy()` has been added to the API of IPool. It can be used
+  to check whether a given pool item is currently busy or idle.
diff --git a/src/ocean/util/container/pool/model/IAggregatePool.d b/src/ocean/util/container/pool/model/IAggregatePool.d
index 22bf0c1..f5605d3 100644
--- a/src/ocean/util/container/pool/model/IAggregatePool.d
+++ b/src/ocean/util/container/pool/model/IAggregatePool.d
@@ -648,29 +648,12 @@ public abstract class IAggregatePool ( T ) : IPool, IFreeList!(ItemType_!(T))
 
     /***************************************************************************
 
-        Checks if item is currently busy.
-
-        Params:
-            item = item to check
-
-        Returs:
-            true if item is currently busy, false otherwise.
-
-    ***************************************************************************/
-
-    protected override bool isBusy ( Item item )
-    {
-        return this.getItemIndex(item) < this.num_busy_;
-    }
-
-    /***************************************************************************
-
         Checks if item is null.
 
diff --git a/src/ocean/util/container/pool/model/IPool.d b/src/ocean/util/container/pool/model/IPool.d
index 696bced..6031190 100644
--- a/src/ocean/util/container/pool/model/IPool.d
+++ b/src/ocean/util/container/pool/model/IPool.d
@@ -324,6 +324,23 @@ public abstract class IPool : IPoolInfo, ILimitable
 
     /***************************************************************************
 
+        Checks if item is currently busy.
+
+        Params:
+            item = item to check
+
+        Returns:
+            true if item is currently busy, false otherwise.
+
+    ***************************************************************************/
+
+    public bool isBusy ( Item item )
+    {
+        return this.getItemIndex(item) < this.num_busy_;
+    }
+
+    /***************************************************************************
+
         Ensures that the pool contains at least the specified number of items.
         Useful to pre-allocate a pool of a certain size.
 
@@ -558,26 +575,12 @@ public abstract class IPool : IPoolInfo, ILimitable
 
     /***************************************************************************
 
-        Checks if item is currently busy.
-
-        Params:
-            item = item to check
-
-        Returs:
-            true if item is currently busy, false otherwise.
-
-    ***************************************************************************/
-
-    abstract protected bool isBusy ( Item item );
-
-    /***************************************************************************
-
         Checks if item is null.
 
```

gkotian:
With this, I suppose `type-breaking` can be removed.

Also, @nemanja-boric-sociomantic, is your comment about having a `IAggregatePool.isOwner` or similar method still applicable?

nemanja-boric-sociomantic:
> Also, @nemanja-boric-sociomantic, is your comment about having a IAggregatePool.isOwner or similar method still applicable?

No, now it's better, that was just one suggestion to avoid breaking change.

src/ocean/task/TaskPool.d
```diff
@@ -181,6 +185,9 @@ class TaskPool ( TaskT : Task ) : ObjectPool!(Task)
 
         foreach (task; tasks_iterator)
         {
+            enforce(!this.isSame(this.toItem(current_task), this.toItem(task)),
```
nemanja-boric-sociomantic, line 188: So, this works only with the `AllItemsIterator`, but not with `BusyItemsIterator`, right?

mihails-strasuns-sociomantic: Yeah, this PR introduces minor performance pessimization to verify the sanity condition. Should be fine for its use case but worth keeping in mind (maybe even put an explicit comment about it, @gkotian ?)

gkotian: Hah, no. In theory, I could've simply used `BusyItemsIterator` (and not even had to define the new `isBusy` method), but I just wanted to play it safe and deliberately iterate over all tasks in the pool, including the idle ones.

nemanja-boric-sociomantic: That's ok, but I don't understand why iteration over all tasks is necessary in this context. Because if the current task is part of the task pool, it must be non-idle element, or?

nemanja-boric-sociomantic: > Hah, no. In theory, I could've simply used BusyItemsIterator (and not even had to define the new isBusy method), but I just wanted to play it safe and deliberately iterate over all tasks in the pool, including the idle ones.

Yeah, I was trying to figure this out :-). I will defer to @mihails-strasuns-sociomantic here as I don't have Task profiling experience to be able to figure out how big the ratio of busy/all items is.

mihails-strasuns-sociomantic: Pedantic approach would be to iterate over `BusyItems` first and `enforce`, and after that call a delegate from within `assert` that iterates over `IdleItems` only. Latter indeed is not possible if rest of the scheduler is in sane state but can serve as an additional debugging helper (== ``assert`).

mihails-strasuns-sociomantic: (I don't mind either of "current state" vs "only busy items" vs "enforce busy, assert idle" options)

nemanja-boric-sociomantic: Me neither, I was just trying to understand the previous commit's message :-).

nemanja-boric-sociomantic (approved):

src/ocean/task/TaskPool.d
```diff
@@ -361,3 +371,44 @@ unittest
...
+        public void copyArguments ( void delegate () non_throwing_dg,
+            void delegate () throwing_dg )
+        {
+            this.non_throwing_dg = non_throwing_dg;
+            this.throwing_dg = throwing_dg;
+        }
+
+        override public void run ( )
+        {
+            this.non_throwing_dg();
```
mihails-strasuns-sociomantic: Why this no-op delegate call is needed at all? Looks confusing.

mihails-strasuns-sociomantic: Also can copy pool reference instead of delegate and call `pool.awaitRunningTasks` directly from `run`

mihails-strasuns-sociomantic (commented):
Few minor comments on test case

src/ocean/task/TaskPool.d
gkotian: >Why this no-op delegate call is needed at all? Looks confusing.

OK, I'll get rid of it.

>Also can copy pool reference instead of delegate and call `pool.awaitRunningTasks` directly from `run`

How can I copy it without running into the forward reference errors?
That is, if I do this:

```d
unittest
{
    // Test that 'awaitRunningTasks()' cannot be called from a task that itself
    // belongs to the pool.

    static class AwaitTask : Task
    { 
        private TaskPool!(AwaitTask) pool;

        public void copyArguments ( TaskPool!(AwaitTask) pool ) 
        { 
            this.pool = pool;
        } 

        override public void run ( ) 
        { 
            this.pool.awaitRunningTasks();
        } 
    } 

    auto pool = new TaskPool!(AwaitTask);
    initScheduler(SchedulerConfiguration.init);

    pool.start(pool);
}
```

it fails with:
```
Error: template instance ocean.core.Traits_tango.ParameterTupleOf!(copyArguments) forward reference of copyArguments
```

gkotian:
Removed the extra non-throwing delegate.

```diff
@@ -379,21 +379,16 @@ unittest
 
     static class AwaitTask : Task
     {
-        void delegate () non_throwing_dg;
-        void delegate () throwing_dg;
+        void delegate () dg;
 
-        public void copyArguments ( void delegate () non_throwing_dg,
-            void delegate () throwing_dg )
+        public void copyArguments ( void delegate () dg )
         {
-            this.non_throwing_dg = non_throwing_dg;
-            this.throwing_dg = throwing_dg;
+            this.dg = dg;
         }
 
         override public void run ( )
         {
-            this.non_throwing_dg();
-
-            testThrown!(Exception)(this.throwing_dg());
+            testThrown!(Exception)(this.dg());
         }
     }
 
@@ -402,12 +397,9 @@ unittest
 
     pool.start(
         {
-            // This is a non throwing delegate as it does nothing.
-        },
-        {
-            // On the other hand, this delegate should throw as it will attempt
-            // to call the task pool's awaitRunningTasks() method from within a
-            // task that itself belongs to the pool.
+            // This delegate should throw as it will attempt to call the task
+            // pool's awaitRunningTasks() method from within a task that itself
+            // belongs to the pool.
             pool.awaitRunningTasks();
         }
     );
```

src/ocean/task/TaskPool.d
mihails-strasuns-sociomantic: I see, didn't expect that to be an issue. Nevermind in that case, LGTM

mihails-strasuns-sociomantic (approved):

mihails-strasuns-sociomantic:
Thanks!

PR #49, opened 2017-03-17
Extend support for monitoring directories using FileSystemEvent

nemanja-boric-sociomantic:
Previously, although user was able to monitor the directory, there was no way to get file name of the files that triggered the event. This is now supported through the new handler which also accepts name of the
file.

### DIFF
--- a/relnotes/filesystemdir.feature.md
+++ b/relnotes/filesystemdir.feature.md
@@ -1,0 +1,7 @@
+* `ocean.io.select.client.FileSystemEvent`
+
+  `FileSystemEvent`'s constructor and `setHandler` now accepts
+  `FileSystemEvent.Notifier` which accepts `SmartUnion` of structs
+  `(path, events)` or `(path, name, events)` where the later member is
+  used when the changes are performed on the files inside monitored directory,
+  where the `name` will be the name of the file that triggered this event.
--- a/src/ocean/io/select/client/FileSystemEvent.d
+++ b/src/ocean/io/select/client/FileSystemEvent.d
@@ -27,12 +27,16 @@ module ocean.io.select.client.FileSystemEvent;
 *******************************************************************************/
 
 import ocean.sys.Inotify;
+import core.stdc.string;
 import core.sys.linux.sys.inotify;
 
 import core.sys.posix.unistd;
 
 import ocean.io.select.EpollSelectDispatcher;
 import ocean.io.select.client.model.ISelectClient: ISelectClient;
+import ocean.core.Buffer;
+import ocean.core.SmartUnion;
+import ocean.transition;
 
 
 /*******************************************************************************
@@ -76,13 +80,75 @@ enum FileEventsEnum : uint
 
 class FileSystemEvent : ISelectClient
 {
+    import ocean.core.Buffer;
+
     /***************************************************************************
 
         Alias for event handler delegate.
 
     ***************************************************************************/
 
+    // TODO: this is deprecated. Remove in the next major
     public alias void delegate ( char[] path, uint event ) Handler;
+
+
+    /***************************************************************************
+
+        Structure carrying only path and event.
+
+    ***************************************************************************/
+
+    public struct FileEvent
+    {
+        cstring path;
+        uint event;
+    }
+
+
+    /***************************************************************************
+
+        Structure carrying path, name and event, representing event that happened
+        on a file in a watched directory.
+
+    ***************************************************************************/
+
+    public struct DirectoryFileEvent
+    {
+        cstring path;
+        cstring name;
+        uint event;
+    }
+
+
+    /***************************************************************************
+
+        Union of possible events that could happen.
+
+    ***************************************************************************/
+
+    public union EventUnion
+    {
+        FileEvent file_event;
+        DirectoryFileEvent directory_file_event;
+    }
+
+
+    /***************************************************************************
+
+        SmartUnion alias for the possible events that could happen.
+
+    ***************************************************************************/
+
+    public alias SmartUnion!(EventUnion) RaisedEvent;
+
+
+    /***************************************************************************
+
+        Alias for the notifier delegate that receives the event as a smart union.
+
+    ***************************************************************************/
+
+    public alias void delegate ( RaisedEvent ) Notifier;
 
 
     /***************************************************************************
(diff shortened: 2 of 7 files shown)


nemanja-boric-sociomantic:
Blocked on #47

src/ocean/io/select/client/FileSystemEvent.d
```diff
@@ -153,6 +208,21 @@ class FileSystemEvent : ISelectClient
...
     /***********************************************************************
 
+        Replace the handle delegate
+
+        Params:
+            handler = event handler
+
+    ***********************************************************************/
+
+    public void setHandler ( DirHandler handler )
```
gavin-norman-sociomantic: It needs to be clearer that both a `Handler` _and_ a `DirHandler` can be set and that they will both be called.

Two questions on that topic, though:
1. Is there actually a use case for this or would it be better to have a union of handlers, instead?
2. Are the cases covered by `DirHandler` a superset of those covered by `Handler`? (i.e. now there's `DirHandler`, is there any reason to use `Handler`?) In this case, is it worth deprecating `Handler`?

gavin-norman-sociomantic (commented):
Initial questions before continuing reviewing.

src/ocean/io/select/client/FileSystemEvent.d
nemanja-boric-sociomantic: This is something I hoped I'll get the input on. Yes, all cases covered by `DirHandler` are covered by `Handler`. 

Initially I opted in keeping both, because  `DirHandler` contains one argument that's usable only in context of monitoring the directory. I felt user would get this dummy argument which is useless for most of the usage scenarios.

Ideally, I would like to rename `DirHandler` to `HandlerEx`, deprecate `Handler` and remove in the next major. In the meantime, based on the assumption, that this class is just a dummy wrapper around `inotify` and `ISelectClient`, expecting user to know internals of `inotify`, mechanism, I would create utility modules which covers the 99% of the usage - `FileCreateModifyEvent` and `FileModifyEvent` (for example) which would internally use `FileSystemEvent`.

gavin-norman-sociomantic: Ah I see, so the `name` argument of `DirHandler` would be unused, if called for a normal (i.e. not a directory) monitoring?

gavin-norman-sociomantic: Another approach would be for the delegate to have a single argument: a `SmartUnion` of different notification types.

nemanja-boric-sociomantic: > Ah I see, so the name argument of DirHandler would be unused, if called for a normal (i.e. not a directory) monitoring?
  
Correct.

> Another approach would be for the delegate to have a single argument: a SmartUnion of different notification types.

Ha, I wonder how I didn't think of this before.

nemanja-boric-sociomantic:
Updated to make handler accept `SmartUnion` and deprecate old ones.

relnotes/filesystemdir.feature.md
```diff
@@ -0,0 +1,6 @@
+* `ocean.io.select.client.FileSystemEvent`
+
+  `FileSystemEvent`'s constructor and `setHandler` now accepts
+  `FileSystemEvent.DirHandler` which accepts the name as well as the path.
```
gavin-norman-sociomantic: `HandlerEx`. Description generally outdated.

test/filesystemevent/main.d
```diff
@@ -174,8 +172,198 @@ class FileModificationTestTask: Task
...
+        this.inotifier  = new FileSystemEvent(&this.fileSystemHandler);
+        inotifier.watch(this.watched_path.dup,
+                       FileEventsEnum.IN_CREATE);
+
+        theScheduler.epoll.register(inotifier);
+
+        auto file_name = "test_file";
+        File.set(file_name, "".dup);
+
+        this.suspend();
```
gavin-norman-sociomantic, line 251: Is it possible that the event has already happened at this point? Then this suspend will hang forever.

test/filesystemevent/main.d
```diff
@@ -174,8 +172,198 @@ class FileModificationTestTask: Task
...
+
+    override public void run ( )
+    {
+        auto sandbox = DirectorySandbox.create();
+        scope (exit)
+            sandbox.exitSandbox();
+
+        this.watched_path = sandbox.path;
+
+        this.inotifier  = new FileSystemEvent(&this.fileSystemHandler);
```
gavin-norman-sociomantic: Maybe a comment at the top of each test describing what they do? (As far as I read this method, it contains two separate test cases.)

relnotes/filesystemdir.feature.md
nemanja-boric-sociomantic: Thanks, I'll update

test/filesystemevent/main.d
nemanja-boric-sociomantic: Yeah, I would factor this out into two separate methods. I wanted to use the same callback, so I could test `SmartUnion`, but there's no excuse for not extracting this bit.

test/filesystemevent/main.d
nemanja-boric-sociomantic: Hmm, given that the scheduler is not preemptive, I don't see how this could happen - the first time the event could happen is when we yield from the current task, or?

gavin-norman-sociomantic: Ah yes, you're right. I was thinking of the case of socket I/O, which can succeed (and thus call the "I'm finished" callback) without needing an epoll cycle. I presume this isn't the case here, though.

nemanja-boric-sociomantic: Yes, I don't think this is the case here - this requires epoll cycle to call handler.

nemanja-boric-sociomantic:
Updated and unblocked.

src/ocean/io/select/client/FileSystemEvent.d
```diff
@@ -101,11 +167,23 @@ class FileSystemEvent : ISelectClient
...
 
     /***************************************************************************
 
+        Event handler delegate, specified in the constructor and called whenever
+        a watched file system event fires, passing the SmartUnion describing the
+        event.
+
+    ***************************************************************************/
+
+    private HandlerEx handler_ex;
```
gavin-norman-sociomantic: Just a thought: to avoid `_ex` / `Ex`, you could call the new thing a "notifier" rather than a handler.

nemanja-boric-sociomantic:
Updated with renaming `HandlerEx` -> `Notifier`.

nemanja-boric-sociomantic:
Updated with a few more tidy ups in naming.

PR #50, opened 2017-03-19
BitArray: Fix padding bits in length()

daniel-zullo:
The function `length()` incorrectly changes the padding bits of the `BitArray` when decreasing the `length`.

Also add a unit test to check the correctness of the state of the bits when increasing or decreasing the length of the `BitArray`.

### DIFF
--- a/src/ocean/core/BitArray.d
+++ b/src/ocean/core/BitArray.d
@@ -1123,3 +1123,60 @@ struct BitArray
         assert( c == a );
     }
 }
+
+version (UnitTest)
+{
+    import ocean.core.Test : NamedTest;
+}
+
+unittest
+{
+    auto t = new NamedTest("Test increase and decrease length of the BitArray");
+
+    const size_of_uint = uint.sizeof * 8;
+    const num_bits = (size_of_uint * 5) + 15;
+
+    // Creates a BitArray and sets all the bits.
+    scope bool_array = new bool[num_bits];
+    bool_array[] = true;
+
+    BitArray bit_array;
+    bit_array = bool_array;
+
+    // Self-verification of the BitArray.
+    assert(bit_array.length == bool_array.length);
+    foreach (bit; bit_array)
+        t.test!("==")(bit, true);
+
+    // Increases the length of the BitArray and checks the former bits remain
+    // set and the new ones are not.
+    const greater_length = size_of_uint * 7;
+    static assert(greater_length > num_bits);
+    bit_array.length = greater_length;
+    foreach (pos, bit; bit_array)
+    {
+        if (pos < num_bits)
+            t.test!("==")(bit, true);
+        else
+            t.test!("==")(bit, false);
+    }
+
+    // Decreases the length of the BitArray to a shorter length than the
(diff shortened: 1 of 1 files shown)


src/ocean/core/BitArray.d
```diff
@@ -86,11 +86,13 @@ struct BitArray
                 uint[] buf = ptr[0 .. olddim];
 
                 buf.length = newdim; // realloc
+                enableStomping(buf);
+
                 ptr = buf.ptr;
-                if( newdim & 31 )
+                if( auto pad_bits = newlen & 31 )
```
nemanja-boric-sociomantic: I always need to read these types of expressions twice, especially given the differences in grouping between `&` and `&&`. Is there a way to make the assignment and bitwise AND more explicit? Meaning - it's subtle difference if the `pad_bits` contains new length or just the value of the 31st bit.

daniel-zullo: I am not sure if adding parenthesis would make it better
```D
if( auto pad_bits = (newlen & 31) )
```

Otherwise I could revert it to the original version and just replace `newdim` to `newlen` but then we will have to compute `newlen & 31` twice.

nemanja-boric-sociomantic: Well, I think that's the completely oposite of how I read this code :-). So - `pad_bits` is the value of the 31st bit? I think parenthesis makes this more clear IMHO.

daniel-zullo: ok, then i'll add parenthesis if it makes the code clear for the reader.

daniel-zullo:
Updated based on a suggestion from Nemanja.

Previous head was 640c0fb, new head at dd6b297. Diff between heads are:
```diff
diff --git a/src/ocean/core/BitArray.d b/src/ocean/core/BitArray.d
index 96a9c98..c8aa27c 100644
--- a/src/ocean/core/BitArray.d
+++ b/src/ocean/core/BitArray.d
@@ -89,7 +89,7 @@ struct BitArray
                 enableStomping(buf);
 
                 ptr = buf.ptr;
-                if( auto pad_bits = newlen & 31 )
+                if( auto pad_bits = (newlen & 31) )
                 {
                     // Set any pad bits to 0
                     ptr[newdim - 1] &= ~(~0 << pad_bits);
```

nemanja-boric-sociomantic:
In general, LGTM! How would you feel about squashing the unittest and the change commits together, or reordering them so that unittest is added after the change has been made. That would make bisecting of ocean easier as there wouldn't be commit that would fail on `make test` (even though rightfully?)

daniel-zullo:
I'd reorder them as suggested if it's fine with you.

nemanja-boric-sociomantic:
Fine by me, thanks!

daniel-zullo:
The commits were reordered, the commit with the unit test was added at the end.

nemanja-boric-sociomantic (approved):

nemanja-boric-sociomantic:
Thanks!

daniel-zullo:
Thanks to you Nemanja :+1:

PR #35, opened 2017-03-11
Format Time correctly with Formatter

mathias-lang-sociomantic:
CC @andrej-mitrovic-sociomantic : Do you need other format string than 'G' ?

Fixes #33

PR #51, opened 2017-03-20
Remove circular imports from ocean.time.Time

mathias-lang-sociomantic:
It currently fails with:

object.Exception@src/rt/minfo.d(167): Aborting: Cycle detected between modules with ctors/dtors:
ocean.text.convert.Layout_tango* ->
ocean.time.Time ->
ocean.core.Test ->
ocean.core.Enforce ->
ocean.text.convert.Format* ->
ocean.text.convert.Layout_tango

This is because Layout_tango needs to import Time to know about the struct type.
However, after ocean.core.Test is switched to the Formatter, the sink approach can be used which does not
require modifications on the Formatter's side.

### DIFF
--- a/src/ocean/time/Time.d
+++ b/src/ocean/time/Time.d
@@ -23,7 +23,10 @@ module ocean.time.Time;
 import ocean.transition;
 import ocean.text.convert.DateTime_tango;
 
-version (UnitTest)
+// This causes circular references because `Test` imports `Layout_tango`
+// which imports this. Enable again when `Test` switches to the Formatter.
+// Formatter imports `Test` through `Traits`, so we cannot use `assert` either
+version (none) version (UnitTest)
 {
     import ocean.core.Test;
     import ocean.text.convert.Formatter;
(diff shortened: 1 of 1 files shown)


mathias-lang-sociomantic:
P.S: This is a fixup for https://github.com/sociomantic-tsunami/ocean/pull/35

src/ocean/time/Time.d
```diff
@@ -23,7 +23,10 @@ module ocean.time.Time;
 import ocean.transition;
 import ocean.text.convert.DateTime_tango;
 
-version (UnitTest)
+// This causes circular references because `Test` imports `Layout_tango`
+// which imports this. Enable again when `Test` switches to the Formatter.
+// Formatter imports `Test` through `Traits`, so we cannot use `assert` either
```
mihails-strasuns-sociomantic, line 28: How about using `assert` in this block instead with same comment? Better than no testing at all.

mathias-lang-sociomantic: That's what I originally went with, but as the comment mention, `Formatter` imports `Test` too. So I cannot test `Formatter` from this module for this patch release.
However, starting from v2.6.x, `Test` uses `Formatter`. So you can just revert this on merging.

mihails-strasuns-sociomantic: Ah, I see, I didn't read original comment with sufficient attention.
GTG in that case.

mihails-strasuns-sociomantic (approved):

PR #53, opened 2017-03-21
Call enableStomping before & after appending \0 terminator in StringC

nemanja-boric-sociomantic:


src/ocean/text/util/StringC.d
```diff
@@ -89,7 +89,9 @@ class StringC
     {
         if (str.length && !!str[$ - 1])
         {
+            enableStomping(str);
             str ~= StringC.Term;
```
mihails-strasuns-sociomantic, line 93: This is wrong place to call `enableStomping` - it has to be done right after `str.length` is changed, not upon appending. The way you do it now simply hides missing `enableStomping` in some other place.

mihails-strasuns-sociomantic (changes requested):

src/ocean/text/util/StringC.d
nemanja-boric-sociomantic: Hmm, I thought there was a consensus that in library functions one should call `enableStomping` before and after changing the length of the user-provided buffer, just because this reason (because the library can't assert on the fact that user did call `enableStomping` where necessary). I have no problems to remove this call, if that's not the case.

mihails-strasuns-sociomantic: The reasoning you remember was mostly related to `ocean.core.array` and similar widely used array utilities where it is justified by amount of porting effort it saves. In general, calling `enableStomping` upon changing length is expected from downstream apps same as with libraries.

Not that appending (like `str -= StringC.Term`) is what I refer to as "changing length" when talking about stomping stuff. Only explicit assignment to `.length` property require `enableStomping` (and only if they may shrink length).

nemanja-boric-sociomantic: > The reasoning you remember was mostly related to ocean.core.array and similar widely used array utilities where it is justified by amount of porting effort it saves. In general, calling enableStomping upon changing length is expected from downstream apps same as with libraries.


Yeah, that's right, and it makes sense.

> Not that appending (like str -= StringC.Term) is what I refer to as "changing length" when talking about stomping stuff. Only explicit assignment to .length property require enableStomping (and only if they may shrink length).

Ha, yes, I missed the point of shrinking length. But, is then this PR even necessary? I don't think so?

mihails-strasuns-sociomantic: You can introduce Buffer overload of this function if that helps though :)

mihails-strasuns-sociomantic: > I missed the point of shrinking length. But, is then this PR even necessary? I don't think so?

Depends on where the buffer came from. If it was from your app, PR is not necessary. If your app called some other ocean utility which called this, that ocean utility may need to be adjusted instead.

nemanja-boric-sociomantic: Yes, correct. I did found what's wrong with my code, but I wanted to put this `enableStomping` _before_ the call, but that's unnecessary, so I'll just close this PR.

PR #56, opened 2017-03-21
BitArray: Optimize opSliceAssign()

daniel-zullo:
This patch uses D array functionality to copy all the content from another `BitArray` instead of applying multiple bit operations.

This patch copies all the bytes in the `BitArray` without resetting the padding bits as they will be reset anyway if the array is extended.

*description updated

### DIFF
--- a/src/ocean/core/BitArray.d
+++ b/src/ocean/core/BitArray.d
@@ -180,22 +180,22 @@ struct BitArray
      *  assert(ba2[0] == false);
      *  -------------------
      */
-     BitArray opSliceAssign(BitArray rhs)
-     in
-     {
-         assert(rhs.len == len);
-     }
-     body
-     {
-         size_t mDim=len/32;
-         ptr[0..mDim] = rhs.ptr[0..mDim];
-         int rest=cast(int)(len & cast(size_t)31u);
-         if (rest){
-             uint mask=(~0u)<<rest;
-             ptr[mDim]=(rhs.ptr[mDim] & (~mask))|(ptr[mDim] & mask);
-         }
-         return *this;
-     }
+    BitArray opSliceAssign(BitArray rhs)
+    in
+    {
+        assert(rhs.len == len);
+    }
+    body
+    {
+        size_t mDim=len/32;
+        ptr[0..mDim] = rhs.ptr[0..mDim];
+        int rest=cast(int)(len & cast(size_t)31u);
+        if (rest){
+            uint mask=(~0u)<<rest;
+            ptr[mDim]=(rhs.ptr[mDim] & (~mask))|(ptr[mDim] & mask);
+        }
+        return *this;
+    }
 
 
     /**
--- a/src/ocean/core/BitArray.d
+++ b/src/ocean/core/BitArray.d
@@ -165,8 +165,7 @@ struct BitArray
      * copy.
      *
      * Params:
-     *  rhs = A BitArray with at least the same number of bits as this bit
-     *  array.
+     *  rhs = A BitArray with the same number of bits as this bit array.
      *
      * Returns:
      *  A shallow copy of this array.
--- a/src/ocean/core/BitArray.d
+++ b/src/ocean/core/BitArray.d
@@ -186,13 +186,8 @@ struct BitArray
     }
     body
     {
-        size_t mDim=len/32;
-        ptr[0..mDim] = rhs.ptr[0..mDim];
-        int rest=cast(int)(len & cast(size_t)31u);
-        if (rest){
-            uint mask=(~0u)<<rest;
-            ptr[mDim]=(rhs.ptr[mDim] & (~mask))|(ptr[mDim] & mask);
-        }
+        auto dimension = this.dim();
+        this.ptr[0..dimension] = rhs.ptr[0..dimension];
         return *this;
     }
 
--- a/src/ocean/core/BitArray.d
+++ b/src/ocean/core/BitArray.d
@@ -1115,3 +1115,47 @@ struct BitArray
         assert( c == a );
     }
 }
+
+version (UnitTest)
+{
+    import ocean.core.Test : NamedTest;
+}
+
+unittest
+{
+    auto t = new NamedTest("Copy the bits from another array");
+
+    const num_bits = 80;
+
+    // Creates a BitArray and sets the bits only for even position.
+    BitArray src_bit_array;
+    src_bit_array.length = num_bits;
+
+    foreach (pos, ref bool_value; src_bit_array)
+    {
+        if (pos % 2 == 0)
+            bool_value = true;
+    }
+
+    BitArray dst_bit_array;
+    dst_bit_array.length = src_bit_array.length;
+
+    // test self-verification
+    t.test!("==")(dst_bit_array.length, num_bits);
+    foreach (pos, bit_value; dst_bit_array)
+        t.test!("==")(bit_value, false);
+
+    dst_bit_array[] = src_bit_array; // performs the copy
+
+    foreach (pos, bit_value; dst_bit_array)
+    {
+        if (pos % 2 == 0)
+            t.test!("==")(bit_value, true);
+        else
+            t.test!("==")(bit_value, false);
+    }
+
+    // Ensures both arrays do not share the same memory storage
+    src_bit_array[0] = false;
+    t.test!("==")(dst_bit_array[0], true);
+}


src/ocean/core/BitArray.d
```diff
@@ -1121,3 +1118,47 @@ struct BitArray
...
+    {
+        if (pos % 2 == 0)
+            assert(bit_value == true);
+        else
+            assert(bit_value == false);
+    }
+
+    // Ensures both arrays do not share the same memory storage
+    src_bit_array[0] = false;
+    assert(dst_bit_array[0] == true);
```
daniel-zullo: I'll replace all `assert()` to `t.test!()`

src/ocean/core/BitArray.d
```diff
@@ -189,13 +191,8 @@ struct BitArray
     }
     body
     {
-        size_t mDim=len/32;
-        ptr[0..mDim] = rhs.ptr[0..mDim];
```
nemanja-boric-sociomantic: Based on this line I'm likely to think we're missing something that we have removed with `if (rest)` branch. The reason for that is that this line `ptr[0..mDim] = rhs.ptr[0..mDim]` will not copy stuff bit by bit, but it will actually perform `memcpy` or similar thing inside runtime.

src/ocean/core/BitArray.d
```diff
@@ -189,13 +191,8 @@ struct BitArray
...
-        int rest=cast(int)(len & cast(size_t)31u);
```
nemanja-boric-sociomantic: Since I'm not familiar with `BitArray` I can't figure out from top of my head what this was doing, but it wasn't bit by bit copying as your commit message suggests.

nemanja-boric-sociomantic: I think it would perform assignment on the highest word of the bit array that's only partially set.

nemanja-boric-sociomantic: I think this is the point:

```
       *  rhs = A BitArray with at least the same number of bits as this bit
       *  array.
```

nemanja-boric-sociomantic: But then this `in` guard is broken here.

nemanja-boric-sociomantic:
Ok, this is the change where this in-contract was introduced:

http://www.dsource.org/projects/tango/changeset/4107

Here's one the ticket:

http://www.dsource.org/projects/tango/ticket/1371

But I can't find the ticket that explains why this particular change in the in-contract has been added. If we change it back to `assert (rhs.len > this.len)` the original code (but not your code) will work. I think either that or changing the documentation that requires *exact* length should be done (and then that `if` thing that copies only part of the bits should be removed). `memcpy` trick seems like it's not needed, because original implementation _does not iterate bit-by-bit_ over the source and destination bit array.

nemanja-boric-sociomantic (dismissed):
I can't find the ticket that explains why this particular change in the in-contract has been added. If we change it back to assert (rhs.len > this.len) the original code (but not your code) will work. I think either that or changing the documentation that requires exact length should be done (and then that if thing that copies only part of the bits should be removed). memcpy trick seems like it's not needed, because original implementation does not iterate bit-by-bit over the source and destination bit array.

src/ocean/core/BitArray.d
daniel-zullo: > Since I'm not familiar with BitArray I can't figure out from top of my head what this was doing, but it wasn't bit by bit copying as your commit message suggests.

True, the commit message needs to be changed to something like 
```TXT
This patch uses `memcpy()` to copy all the content from another BitArray instead of applying multiple bit operations.
```

> But then this in guard is broken here.

Why? `len` is the number of bits in the `BitArray`

nemanja-boric-sociomantic: Well because the documentation requires that the source array has *at least* the same number of bits, which contradicts this in contract which says the source array must have exact number of bits as the destination one.

daniel-zullo:
> Ok, this is the change where this in-contract was introduced:

Sorry, what do you mean here??

nemanja-boric-sociomantic:
I have pressed enter before I finished comment, sorry. I've edited the comment for the full length.

src/ocean/core/BitArray.d
nemanja-boric-sociomantic: > This patch uses `memcpy()` to copy all the content from another BitArray instead of applying multiple bit operations.

Yes, but then you could just remove this multiple bit operations and just write

```
    BitArray opSliceAssign(BitArray rhs)                                                                               
    in                                                                                                                 
    {                                                                                                                  
        assert(rhs.len == len);                                                                                        
    }                                                                                                                  
    body                                                                                                               
    {                                                                                                                  
        size_t mDim=len/32;                                                                                            
        ptr[0..mDim] = rhs.ptr[0..mDim];                                                                               
        return *this;
}
```

because druntime will actually perform `memcpy` on the array while copying.

daniel-zullo:
> I can't find the ticket that explains why this particular change in the in-contract has been added. If we change it back to assert (rhs.len > this.len) the original code (but not your code) will work. I think either that or changing the documentation that requires exact length should be done (and then that if thing that copies only part of the bits should be removed). memcpy trick seems like it's not needed, because original implementation does not iterate bit-by-bit over the source and destination bit array.

The patch is based on the current implementation, why do you want this patch to check for a former `in` contract? And it has been validated in the unit test that works (even if you extend or shrink the `BitArray`). It might not optimize a lot as it is claimed in the commit message but made the code much simpler (and a bit more efficient).

nemanja-boric-sociomantic:
> The patch is based on the current implementation, why do you want this patch to check for a former in contract?

Yes, I'm suggesting then the documentation should be fixed as well, because this patch contradicts the documentation, but not implementation.

> It might not optimize a lot as it is claimed in the commit message but made the code much simpler (and a bit more efficient).

I have doubts about this code:

```
    BitArray opSliceAssign(BitArray rhs)
    in
    {
        assert(rhs.len == len);
    }
    body
    {
        auto num_bytes = this.dim() * uint.sizeof;
        this.ptr[0..num_bytes] = rhs.ptr[0..num_bytes];
        return *this;
    }

```

being more complex than:

```
    BitArray opSliceAssign(BitArray rhs)
    in
    {
        assert(rhs.len == len);
    }
    body
    {
        auto num_bytes = this.dim() * uint.sizeof;
        memcpy(this.ptr, rhs.ptr, num_bytes);
        return *this;
    }
```

> and a bit more efficient

If you remove the extra `if` branch as suggested, I wonder if the performance improvement stays?

src/ocean/core/BitArray.d
daniel-zullo: ok, I see your point about using current array functionality rather than `memcpy`. Anyway your suggestion is incorrect in the way `mDim` is computed, it should be `(len+31)/32` or use `dim()` (because you are not copying the last dimension).

nemanja-boric-sociomantic: Yes, sorry, you're right, my other comment has this fixed: https://github.com/sociomantic-tsunami/ocean/pull/56#issuecomment-288678541

daniel-zullo:
> Yes, I'm suggesting then the documentation should be fixed as well, because this patch contradicts the documentation, but not implementation.

ok, I'll fix the documentation too (but it was wrong before applying this patch).

nemanja-boric-sociomantic:
> ok, I'll fix the documentation too (but it was wrong before applying this patch).

Yeah it was wrong. As you could see, the tango commit where the implantation was changed didn't carry a lot of effort to explain why it's done and to change documentation to follow the new implementation. Thanks!

daniel-zullo:
```D
    BitArray opSliceAssign(BitArray rhs)
    in
    {
        assert(rhs.len == len);
    }
    body
    {
        auto num_bytes = this.dim() * uint.sizeof;
        this.ptr[0..num_bytes] = rhs.ptr[0..num_bytes];
        return *this;
    }
```
Unfortunately using the suggestion above throws `unexpected exception object.Exception: overlapping array copy` and same thing happens if `ocean.core.array.copy()` is used.
I am trying to understand what is the reason.

nemanja-boric-sociomantic:
Looks like both `this.ptr` and `rhs.ptr` refer to the same block of memory.

nemanja-boric-sociomantic:
`memcpy` should fail as well in this scenario, `memmove` is safe to use for that purpose (but I think something else is wrong).

daniel-zullo:
`memcpy` does not fail but reading about copying D arrays they just do a `mencpy` internally as you said before.

daniel-zullo:
The test ensures both arrays do not share the same memory storage when using `memcpy()`

mihails-strasuns-sociomantic:
`memcpy` won't "fail" for overlapping copy but doing so is undefined behaviour (https://linux.die.net/man/3/memcpy)

nemanja-boric-sociomantic:
Ah, yes, of course.

daniel-zullo:
So should I go for `memmove()`? Or do you have any suggestion to use the array functionality instead?

nemanja-boric-sociomantic:
I think something is broken, as these two regions shouldn't overlap unless you assign bitarray to itself, right?

daniel-zullo:
Replacing `dst_bit_array[] = src_bit_array;` to `dst_bit_array = src_bit_array;` in the test don't throw the `overlapping array copy` but it ends up using the same memory storage for both arrays.

nemanja-boric-sociomantic:
```
$17 = (ocean.core.BitArray.BitArray *) 0x7fffffffd900
(gdb) p &src_bit_array
$18 = (ocean.core.BitArray.BitArray *) 0x7fffffffd8f0
(gdb) p dst_bit_array.ptr
$19 = (unsigned int *) 0x7ffff48a7b20
(gdb) p src_bit_array.ptr
$20 = (unsigned int *) 0x7ffff48a7b30
```

and in `opSliceAssign`:

```
(gdb) p num_bytes
$21 = 12
(gdb) p len
$22 = 80
```

I think `num_bytes` is calculated wrongly, because it should be 10, right?

nemanja-boric-sociomantic:
I think it should be just

```
this.ptr[0..this.dim()] = rhs.ptr[0..this.dim()]`
```

right?

daniel-zullo:
ha that's it :D

daniel-zullo:
We don't need to multiply for `uint.sizeof` anymore :s

nemanja-boric-sociomantic:
Yeah, because we will perform everything on the `uint` level, no need for `byte` level calculations (which lead us to this confusion as we instinctively thought `this.ptr()` returns byte array).

daniel-zullo:
Updated based on suggestions from Nemanja.

Previous head was 5b0c493, new head at 50c6439. Diff between heads are:
```diff
diff --git a/src/ocean/core/BitArray.d b/src/ocean/core/BitArray.d
index 15d3172..5daf065 100644
--- a/src/ocean/core/BitArray.d
+++ b/src/ocean/core/BitArray.d
@@ -20,8 +20,6 @@ import ocean.transition;
 
 import ocean.core.BitManip;
 
-import core.stdc.string; // memcpy()
-
 
 /**
  * This struct represents an array of boolean values, each of which occupy one
@@ -169,8 +167,7 @@ struct BitArray
      * copy.
      *
      * Params:
-     *  rhs = A BitArray with at least the same number of bits as this bit
-     *  array.
+     *  rhs = A BitArray with the same number of bits as this bit array.
      *
      * Returns:
      *  A shallow copy of this array.
@@ -191,8 +188,8 @@ struct BitArray
     }
     body
     {
-        auto num_bytes = this.dim() * uint.sizeof;
-        memcpy(this.ptr, rhs.ptr, num_bytes);
+        auto dimension = this.dim();
+        this.ptr[0..dimension] = rhs.ptr[0..dimension];
         return *this;
     }
 
```

nemanja-boric-sociomantic:
LGTM! My last minor comment would be the same as on #50 - can you squash or reorder the unittest change, so the ocean's commit can be built and the test could pass, since we're not using merge commit, it would allow for easier bisecting.

nemanja-boric-sociomantic:
One more question, can this go to v2.5.x, or does it have to go to v2.6.x?

daniel-zullo:
>  can you squash or reorder the unittest change, so the ocean's commit can be built and the test could pass, since we're not using merge commit, it would allow for easier bisecting.

sure.

> One more question, can this go to v2.5.x, or does it have to go to v2.6.x?

Of course it can go to v2.6.x, I wasn't sure about what was the right branch to send this PR as it isn't a bug fix.

Can we also block this PR until #50 is merged as they both share a commit for enable stomping?

nemanja-boric-sociomantic:
Yeah, if it isn't a bugfix, then rebase it to v2.x.x and we will release it in v2.7.0.

> Can we also block this PR until #50 is merged as they both share a commit for enable stomping?

Yeah, you can just remove this commit from here and it would appear in v2.x.x when v2.5.x. is merged into v2.x.x (unless it breaks tests, in which case we would block it).

nemanja-boric-sociomantic (approved):

daniel-zullo:
Updated base branch to v2.x.x and reordered commits

nemanja-boric-sociomantic:
Thanks a lot for patience!

daniel-zullo:
Thanks a lot!

PR #57, opened 2017-03-22
[docker/ci] Use tagged makd release for packaging dmds

nemanja-boric-sociomantic:
tags are for forever

### DIFF
--- a/docker/dmd-transitional.sh
+++ b/docker/dmd-transitional.sh
@@ -42,7 +42,7 @@ cd -
 #make -r -C phobos -f posix.mak MODEL=64 RELEASE=1
 
 # Make package using MakD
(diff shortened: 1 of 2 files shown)


leandro-lucarella-sociomantic (approved):
Yeah, that was a silly thing to do.

leandro-lucarella-sociomantic:
So this should be promptly merged to the other branches, right?

nemanja-boric-sociomantic:
Correct. But as a bonus point, I think we're at stage where we can release v2.5.6.

leandro-lucarella-sociomantic:
> Correct. But as a bonus point, I think we're at stage where we can release v2.5.6.

we don't really need to release to do a merge anyway. :)

PR #59, opened 2017-03-23
Don't require running scheduler state for await

### LINKED ISSUE #58: Check if scheduler state restriction is necesary in `await`
I can't remember the reason why https://github.com/sociomantic-tsunami/ocean/blob/v2.x.x/src/ocean/task/Scheduler.d#L455 was added, will need to investigate that again.

mihails-strasuns-sociomantic:
Further investigation has shown that this restriction is not necessary and is
either an artifact of earlier implementation or manifestation of defensive
programming.

Fixes #58


gavin-norman-sociomantic:
Ah that's good news!

PR #60, opened 2017-03-23
Implement hasFeaturesFrom helper

mihails-strasuns-sociomantic:
Follow-up to recent discussion about exposing library versions in D code for
`static if` shenanigans in downstream libraries.

### DIFF
--- a/src/ocean/LibFeatures.d
+++ b/src/ocean/LibFeatures.d
@@ -1,0 +1,26 @@
+/*******************************************************************************
+
+    Meta module defining constants that match ocean feature releases allowing
+    downstream libraries to conditionally implement functionality relying on
+    later ocean versions.
+
+    Copyright:
+        Copyright (c) 2009-2016 Sociomantic Labs GmbH.
+        All rights reserved.
+
+    License:
+        Boost Software License Version 1.0. See LICENSE_BOOST.txt for details.
+        Alternatively, this file may be distributed under the terms of the Tango
+        3-Clause BSD License (see LICENSE_BSD.txt for details).
+
+*******************************************************************************/
+
+module ocean.LibFeatures;
+
+const has_features_2_7 = true;
+const has_features_2_6 = true;
+const has_features_2_5 = true;
+const has_features_2_4 = true;
+const has_features_2_3 = true;
+const has_features_2_1 = true;
+const has_features_2_0 = true;
--- a/src/ocean/core/VersionCheck.d
+++ b/src/ocean/core/VersionCheck.d
@@ -1,0 +1,49 @@
+/*******************************************************************************
+
+    Helper utility for reflection on `LibFeatures` modules provided by
+    libraries. If a library provides a module named `libname.LibFeatures` which
+    defines constants of form `const has_features_x_y = true` where `x` is major
+    version number and `y` is minor version number, helper defined in this
+    module allows to query supported features with a standard API.
+
+    Copyright:
+        Copyright (c) 2009-2016 Sociomantic Labs GmbH.
+        All rights reserved.
+
+    License:
+        Boost Software License Version 1.0. See LICENSE_BOOST.txt for details.
+        Alternatively, this file may be distributed under the terms of the Tango
+        3-Clause BSD License (see LICENSE_BSD.txt for details).
+
+*******************************************************************************/
+
+module ocean.core.VersionCheck;
+
+import ocean.transition;
+import ocean.core.Traits;
+
+/// ditto
+public template hasFeaturesFrom ( istring libname, ulong major, ulong minor )
+{
+    const hasFeaturesFrom = hasFeaturesFromFunc!(libname, major, minor)();
+}
+
+///
+unittest
+{
+    static assert ( hasFeaturesFrom!("ocean", 2, 0));
+    static assert ( hasFeaturesFrom!("ocean", 2, 6));
+    static assert (!hasFeaturesFrom!("ocean", 100, 0)); // will have to be
+                                                        // updated one day
+}
+
+private bool hasFeaturesFromFunc ( istring libname, ulong major, ulong minor ) ()
+{
+    struct Library
+    {
+        mixin("import " ~ libname ~ ".LibFeatures;");
+    }
+
+    mixin("return is(typeof(Library.has_features_"
+        ~ ctfe_i2a(major) ~ "_" ~ ctfe_i2a(minor) ~ "));");
+}


src/ocean/LibVersion.d
```diff
@@ -0,0 +1,9 @@
+module ocean.LibVersion;
```
leandro-lucarella-sociomantic: Since this module is starting to only care about new features introduced in minor versions, I wonder if it shouldn't be named `libFeatures` instead.

leandro-lucarella-sociomantic (commented):
Besides the small bikeshedding, LGTM (not approving because it lacks release notes / docs, which I guess is intended being a WIP and all :P).

src/ocean/LibVersion.d
```diff
@@ -0,0 +1,9 @@
...
+
+const has_features_2_6 = true;
```
leandro-lucarella-sociomantic: Wait! Since previous versions won't have the `hasFeaturesFrom` (or even the features definition constants), does it make sense to define the features for old versions?

src/ocean/core/VersionCheck.d
```diff
@@ -0,0 +1,28 @@
...
+
+private bool hasFeaturesFromFunc ( istring libname, ulong major, ulong minor ) ()
+{
+    struct Library
+    {
+        mixin("import " ~ libname ~ ".LibVersion;");
+    }
+
+    mixin("return is(typeof(Library.has_features_"
+        ~ ctfe_i2a(major) ~ "_" ~ ctfe_i2a(minor) ~ "));");
```
mathias-lang-sociomantic: So in this case, how do we deprecate `has_features` symbols when a library is using `-de` ?

mihails-strasuns-sociomantic: Same as usual. Library usage of `-de` does not propagate to applications - deprecating symbol will ensure it is not directly used in libraries but will keep working if app upgrades to ocean version above supported to library. Which is exactly what we want, right?

src/ocean/LibVersion.d
mihails-strasuns-sociomantic: Yes, that is the plan. If design looks OK, I'll make separate pull requests per each supported branch to define consts separately.

src/ocean/LibVersion.d
mihails-strasuns-sociomantic: Sounds good.

src/ocean/LibVersion.d
mihails-strasuns-sociomantic: See https://github.com/sociomantic-tsunami/ocean/pull/64

src/ocean/core/VersionCheck.d
mathias-lang-sociomantic: Oh I understand better the original discussion then. I like it :+1:

src/ocean/LibVersion.d
leandro-lucarella-sociomantic: I still don't get it. Also, technically, this is a new feature (adding a module), so we shouldn't release that as a point release. We could make an exception, but I'd prefer not to unless there is rally a **very** strong case for it.

mihails-strasuns-sociomantic: AFAICS it is neither a feature nor a bug fix - it is purely infrastructural maintenance for a feature that will be added in v2.7 (this PR). To make `hasFeatureFrom` helper reliable we need to add that module to all supported branches in all core libraries - it has no effect on their own API/functionality.

leandro-lucarella-sociomantic: But again, you can't rely on this feature because then if you introduce this in v2.5.7, anyone still at v2.5.6 or lower will have the check failing. I know people is supposed to update to the latest patch release, but it won't always happen, and if you can't rely on this feature `const` being correctly defined, it means it's not really necessary to add the check at this point, so why add it?

For me is best and cleaner to do a clean cut and support this idiom only for v2.7+

mihails-strasuns-sociomantic: > anyone still at v2.5.6 or lower will have the check failing

We don't support non-latest point releases, such users are strictly obliged to upgrade.

mihails-strasuns-sociomantic: (same concern applies to v2.5.x and v2.6.x branches btw)

mihails-strasuns-sociomantic:
Update the PR

mihails-strasuns-sociomantic:
Green

src/ocean/core/VersionCheck.d
```diff
@@ -0,0 +1,49 @@
...
+    const hasFeaturesFrom = hasFeaturesFromFunc!(libname, major, minor)();
+}
+
+///
+unittest
+{
+    static assert ( hasFeaturesFrom!("ocean", 2, 0));
+    static assert ( hasFeaturesFrom!("ocean", 2, 6));
+    static assert ( hasFeaturesFrom!("ocean", 2, 7));
+    static assert (!hasFeaturesFrom!("ocean", 3, 0));
```
leandro-lucarella-sociomantic: This unit test will have to be adapted when moving to v3. Why not just using a feature that will never be set. Otherwise are we updating this test each time we branch a new major? :)

leandro-lucarella-sociomantic (approved):
Other than that (minor), LGTM.

src/ocean/core/VersionCheck.d
mihails-strasuns-sociomantic: Yeah makes sense

src/ocean/LibFeatures.d
```diff
@@ -0,0 +1,26 @@
+/*******************************************************************************
+
+    Meta module defining constants that match ocean feature releases allowing
+    downstream libraries to conditionally implement functionality relying on
+    later ocean versions.
+
+    Copyright:
+        Copyright (c) 2009-2016 Sociomantic Labs GmbH.
```
daniel-zullo, line 8: This is not right, we are in 2017.

src/ocean/core/VersionCheck.d
```diff
@@ -0,0 +1,49 @@
+/*******************************************************************************
+
+    Helper utility for reflection on `LibFeatures` modules provided by
+    libraries. If a library provides a module named `libname.LibFeatures` which
+    defines constants of form `const has_features_x_y = true` where `x` is major
+    version number and `y` is minor version number, helper defined in this
+    module allows to query supported features with a standard API.
+
+    Copyright:
+        Copyright (c) 2009-2016 Sociomantic Labs GmbH.
```
daniel-zullo, line 10: 2017

src/ocean/core/VersionCheck.d
```diff
@@ -0,0 +1,49 @@
...
+///
+unittest
+{
+    static assert ( hasFeaturesFrom!("ocean", 2, 0));
+    static assert ( hasFeaturesFrom!("ocean", 2, 6));
+    static assert (!hasFeaturesFrom!("ocean", 100, 0)); // will have to be
+                                                        // updated one day
+}
+
+private bool hasFeaturesFromFunc ( istring libname, ulong major, ulong minor ) ()
```
daniel-zullo, line 40: the methods in this module do not have documentation.

daniel-zullo (commented):
Please be more careful before merging PRs without the right documentation.

src/ocean/core/VersionCheck.d
mihails-strasuns-sociomantic: You are wrong on this part though :) The `ditto` bit will inherit public method documentation from the module header (and private one is not intended to be documented).

mihails-strasuns-sociomantic:
Copyright statements are contagious disease that was artifically create by evil legal world to drive developers to madness :grumpy:

src/ocean/LibFeatures.d
mihails-strasuns-sociomantic: https://github.com/sociomantic-tsunami/ocean/pull/66

src/ocean/core/VersionCheck.d
daniel-zullo: Sorry but I don't totally agree as the parameters documentation would be missing if we use `ditto` to inherit from the module header.

gavin-norman-sociomantic: I agree. This template and function ought to be documented in the usual style.

mathias-baumann-sociomantic: This is still pending

PR #62, opened 2017-03-24
Don't call exception_handler for EpollException

mihails-strasuns-sociomantic:
EpollException will be thrown for variety of socket/client errors thrown from
`checkKeyError` method. Trying out `exception_handler` in live projects has
shown that silent consumption of such exception is strictly required for some
of existing library code to work.

### DIFF
--- a/src/ocean/io/select/EpollSelectDispatcher.d
+++ b/src/ocean/io/select/EpollSelectDispatcher.d
@@ -137,7 +137,10 @@ public class EpollSelectDispatcher : IEpollSelectDispatcherInfo
 
     /***************************************************************************
 
-        Optional hook to be called on unhandled exceptions from events
+        Optional hook to be called on unhandled exceptions from event callbacks.
+
+        NB: it won't be called on actual event errors thrown as EpollException
+        as those are expected to be handled on select client level exclusively.
 
     ***************************************************************************/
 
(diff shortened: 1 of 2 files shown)


gavin-norman-sociomantic:
This is an interesting case. It changes the behaviour of a feature added in v2.6.0, but does not update the documentation (release notes) to describe the new behaviour. Any thoughts?

mihails-strasuns-sociomantic:
I decided to count is as a silent bug fix because it was not possible to use the feature without some bad hacks if the affected behaviour was triggerred. Basically existing documentation remains in full effect and we consider `EpollException` to be epoll internal implementation details which should not propagate to the user code.

gavin-norman-sociomantic:
Fair enough.

The doc for `EpollSelectDispatcher.unhandled_exception_hook` currently says "Optional hook to be called on unhandled exceptions from events". `SelectedKeysHandler.checkKeyError` is specifically checking the events which fired for the client, so it sounds like the doc needs to be updated to clarify this.

mihails-strasuns-sociomantic:
Check updated docs, something like this?

gavin-norman-sociomantic:
Yep, looks better. @david-eckardt-sociomantic sanity check?

david-eckardt-sociomantic (approved):

PR #64, opened 2017-03-24
Add ocean.LibFeatures module

mihails-strasuns-sociomantic:
Defines feature constants for this and previous v2.x releases

### DIFF
--- a/src/ocean/LibFeatures.d
+++ b/src/ocean/LibFeatures.d
@@ -1,0 +1,24 @@
+/*******************************************************************************
+
+    Meta module defining constants that match ocean feature releases allowing
+    downstream libraries to conditionally implement functionality relying on
+    later ocean versions.
+
+    Copyright:
+        Copyright (c) 2009-2016 Sociomantic Labs GmbH.
+        All rights reserved.
+
+    License:
+        Boost Software License Version 1.0. See LICENSE_BOOST.txt for details.
+        Alternatively, this file may be distributed under the terms of the Tango
+        3-Clause BSD License (see LICENSE_BSD.txt for details).
+
+*******************************************************************************/
+
+module ocean.LibFeatures;
+
+const has_features_2_5 = true;
+const has_features_2_4 = true;
+const has_features_2_3 = true;
+const has_features_2_1 = true;
+const has_features_2_0 = true;


src/ocean/LibFeatures.d
```diff
@@ -0,0 +1,24 @@
...
+        Boost Software License Version 1.0. See LICENSE_BOOST.txt for details.
+        Alternatively, this file may be distributed under the terms of the Tango
+        3-Clause BSD License (see LICENSE_BSD.txt for details).
+
+*******************************************************************************/
+
+module ocean.LibFeatures;
+
+const has_features_2_5 = true;
+const has_features_2_4 = true;
```
leandro-lucarella-sociomantic, line 21: I insist, why defining `has_features_2_4` (and lower) if those branches will never had and never will get the feature constants?

leandro-lucarella-sociomantic (commented):
Also, wrong commit author :innocent:

src/ocean/LibFeatures.d
mihails-strasuns-sociomantic: Remember that already released libraries often depend on unmaintained ocean versions (i.e. libblah v5.x.x depends on ocean 2.3.7) because of semver. So it may happen that I will need to check for features from ocean 2.4 when making new turtle feature release in that branch.

One may judge it as unlikely but adding those consts doesn't cost us anything so I don't see any reason to not do so.

mihails-strasuns-sociomantic:
Updated author

src/ocean/LibFeatures.d
leandro-lucarella-sociomantic: My point is, if somebody has ocean v2.4, then when you check for the constant it won't be there anyway, in that case you'll still be using features from v2.3, not v2.4, right? It will be when people upgrade to v2.7 (or v2.5 if we include it in that one first) that the feature check will work, so why not just checking for ocean v2.7 instead of v2.4 then? If the check will only pass for v2.7 and up?

mihails-strasuns-sociomantic: Remember that "somebody" is library when it comes to writing the condition, but it will be application which checks it. Applications are not supposed/allowed to use non-supported minor releases while libraries often have to.

Thus in your case library will be on version 2.3 but implementing a feature needing 2.4 under relevant `static if`. However application using that library will use 2.5 at least, which will have all flags for v2.x defined, including 2.5 one.

Alternative would be to check for version 2.5 in that library despite the fact needed feature actually came in 2.4 - I don't mind that too much, but such situation seems much more confusing to me when someone will read that library code.

leandro-lucarella-sociomantic: I still don't get this argument:

> However application using that library will use 2.5 at least

So why include feature constants for pre-v2.5 then? This is what I can't understand.



> Alternative would be to check for version 2.5 in that library despite the fact needed feature actually came in 2.4 - I don't mind that too much, but such situation seems much more confusing to me when someone will read that library code.

Yes, this is what I mean, for me this is less misleading, because in v2.4 there will be no feature constant so anyone using v2.4 will not get features from it used (simply because the feature constant will not be defined). So effectively features from v2.4 will only be used starting from v2.5 and on.

leandro-lucarella-sociomantic: BTW, this would be only like that in the transitional phase, of course afterwards one should use the constant from the version that really introduced the feature.

leandro-lucarella-sociomantic: > Yes, this is what I mean, for me this is less misleading, because in v2.4 there will be no feature constant so anyone using v2.4 will not get features from it used (simply because the feature constant will not be defined)

OK, here lies the point that I was missing. In v2.4 the new function is not defined so it simply won't compile. I still think is not ideal but I guess for the transitional period there is no ideal solution, so we can go as well with this one.

leandro-lucarella-sociomantic (approved):

mihails-strasuns-sociomantic:
After some more discussions via Slack we decided to go with 2.7.0 only path and release 2.7.0 asap instead, allowing new library major versions to start using it immediately.

PR #66, opened 2017-03-28
Adjust few wrong copyright statements

mihails-strasuns-sociomantic:
.


gavin-norman-sociomantic:
Included in v2.8.0, as it's not worth making a patch release for this change :D

PR #67, opened 2017-03-29
Add close-on-exec option for sockets, timers and events

david-eckardt-sociomantic:
Some of our applications spawn new processes, which inherit all file descriptors by default but neither use nor close them until they terminate. Consequently the parent process cannot close any device associated with a file descriptor because the child process keeps it open. One consequence, which caused a bug that was hard to reproduce and track down, is that file descriptors in epoll are not automatically unregistered when closing them.
The solution is to enable the close-on-exec option for every single file descriptor, which prevents the child process from inheriting it. Fortunately on Linux all functions that create a file descriptor accept an option to enable close-on-exec. Add this option to the ocean facilities that create file descriptors which are usually registered with epoll.
There are still two open points:
1. To avoid a breaking change the close-on-exec option is disabled by default so it is up to user code to enable it. I think it should really be enabled by default because it protects from sporadic errors that can be severe and hard to debug -- or do we have an application that spawns a process which needs to use the parent's socket, timer, event or epoll file descriptors?
2. `EpollProcess` is affected as well, in particular the `SIGCHLD` select event in its `GlobalProcessMonitor`. Since a global singleton instance of this class is used I am not sure how to add close-on-exec without a breaking change.

### DIFF
--- a/relnotes/cloexec.feature.md
+++ b/relnotes/cloexec.feature.md
@@ -1,0 +1,80 @@
+* `ocean.sys.CloseOnExec`
+
+    * `open_with_close_on_exec` has been added. It is a global configuration
+      variable to define whether the close-on-exec option should be set on
+      system calls that create a file descriptor in ocean modules.
+
+    * `setCloExec` is a helper function to set the close-on-exec bit in a bit
+      mask which specifies option flags for a system call that obtains a new
+      file descriptor, such as `open(2)`. On recent Linux all such
+      system/library functions support enabling the close-on-exec option; this
+      is a Linux extension to POSIX.
+
+    On a POSIX system, if a program calls `exec(2)` (or an equivalent function
+    from the `exec` family) to execute another program, then the devices
+    referred to by the file descriptors held by the original program stay open,
+    and the executed program inherits all of them.
+    This has the consequence that, if the executed program is unaware of the
+    original program's file descriptors so it doesn't close those it doesn't
+    need, then all devices will stay open until the process exits.
+
+    Most of the time an `exec` call is preceded by a `fork` call to execute a
+    program in a new process. As with `exec` the child process created by `fork`
+    inherits all file descriptors from its parent process that called `fork`.
+    However, each device referred to by one of these file descriptors stays open
+    until _both_ the parent and child process close them or exit. This means
+    that the parent process has no way of closing its devices any more unless
+    the child process cooperates and does so as well, which is in practice very
+    unlikely.
+
+    This can cause the following situation:
+     - A program opens a device (file, socket, timer or event fd).
+     - The program registers the device with `epoll`.
+     - The program uses a third-party library which starts a task in a separate
+       process using `fork` + `exec`.
+     - The original program (parent process) closes the device.
+     - The executed program (child process) is, as libraries are, unaware of
+       its parent's business so it doesn't close any of the inherited file
+       descriptors; it doesn't even know which it inherited. So the device stays
+       opened.
+     - Because the device is opened it stays registered with `epoll`. The parent
+       process cannot unregister it any more because it has closed the file
+       descriptor.
+     - Until the child process exits `epoll_wait(2)` keeps reporting events for
+       the device in the parent process.
+
+    This is a problem because it can cause sporadic erratic behaviour in a
+    program. The error may be hard to reproduce and track down.
+
+    To prevent this from happening POSIX provides a close-on-exec option for
+    each file descriptor. By default it is disabled. If enabled, `exec` will
+    close the file descriptor so that the executed program won't inherit it. But
+    it works only if the parent process explicitly enables this option for every
+    single file descriptor it obtains from the system.
+
+    The global `open_with_close_on_exec` variable in this module is read by all
+    functions and class constructors in ocean that obtain a file descriptor from
+    the system. If it is `true` then they set the close-on-exec option.
+
+    This does not affect `stdin`, `stdout` and `stderr`, which are opened before
+    the start of the program.
+
+    Currently the flag is `false` to keep the system's default behaviour. In
+    general it is recommended to set it to `true` to avoid the aforementioned
+    problem unless file descriptor inheriting is needed. Since there appears to
+    be no use case for it for us the default value of `open_with_close_on_exec`
+    will change to `true` in the next major ocean release.
+
+    IF YOU HAVE A USE CASE FOR INHERITING FILE DESCRIPTORS, PLEASE CONTACT THE
+    OCEAN MAINTAINERS.
+
+    The following ocean functions obtain a file descriptor and use this flag:
+
+    - `ocean.sys.Epoll`: `Epoll.create`
+    - `ocean.sys.EventFD`: `EventFD` constructor
+    - `ocean.sys.Inotify`: `Inotify` constructor
+    - `ocean.sys.SignalFD`: `SignalFD.register`
+    - `ocean.sys.TimerFD`: `TimerFD` constructor
+    - `ocean.sys.socket`: The `socket` and `accept` methods in all `*Socket`
+        classes
+    - `ocean.io.device.File`: `File.open`
--- a/src/ocean/sys/CloseOnExec.d
+++ b/src/ocean/sys/CloseOnExec.d
@@ -1,0 +1,125 @@
+/*******************************************************************************
+
+    Global configuration variable to define whether the close-on-exec option
+    should be set on system calls that create a file descriptor in ocean
+    modules.
+
+    On a POSIX system, if a program calls `exec(2)` (or an equivalent function
+    from the `exec` family) to execute another program, then the devices
+    referred to by the file descriptors held by the original program stay open,
+    and the executed program inherits all of them.
+    This has the consequence that, if the executed program is unaware of the
+    original program's file descriptors so it doesn't close those it doesn't
+    need, then all devices will stay open until the process exits.
+
+    Most of the time an `exec` call is preceded by a `fork` call to execute a
+    program in a new process. As with `exec` the child process created by `fork`
+    inherits all file descriptors from its parent process that called `fork`.
+    However, each device referred to by one of these file descriptors stays open
+    until _both_ the parent and child process close them or exit. This means
+    that the parent process has no way of closing its devices any more unless
+    the child process cooperates and does so as well, which is in practice very
+    unlikely.
+
+    This can cause the following situation:
+     - A program opens a device (file, socket, timer or event fd).
+     - The program registers the device with `epoll`.
+     - The program uses a third-party library which starts a task in a separate
+       process using `fork` + `exec`.
+     - The original program (parent process) closes the device.
+     - The executed program (child process) is, as libraries are, unaware of
+       its parent's business so it doesn't close any of the inherited file
+       descriptors; it doesn't even know which it inherited. So the device stays
+       opened.
+     - Because the device is opened it stays registered with `epoll`. The parent
+       process cannot unregister it any more because it has closed the file
+       descriptor.
+     - Until the child process exits `epoll_wait(2)` keeps reporting events for
+       the device in the parent process.
+
+    This is a problem because it can cause sporadic erratic behaviour in a
+    program. The error may be hard to reproduce and track down.
+
+    To prevent this from happening POSIX provides a close-on-exec option for
+    each file descriptor. By default it is disabled. If enabled, `exec` will
+    close the file descriptor so that the executed program won't inherit it. But
+    it works only if the parent process explicitly enables this option for every
+    single file descriptor it obtains from the system.
+
+    The global `open_with_close_on_exec` variable in this module is read by all
+    functions and class constructors in ocean that obtain a file descriptor from
+    the system. If it is `true` then they set the close-on-exec option.
+
+    This does not affect `stdin`, `stdout` and `stderr`, which are opened before
+    the start of the program.
+
+    Currently the flag is `false` to keep the system's default behaviour. In
+    general it is recommended to set it to `true` to avoid the aforementioned
+    problem unless file descriptor inheriting is needed. Since there appears to
+    be no use case for it for us the default value of `open_with_close_on_exec`
+    will change to `true` in the next major ocean release.
+
+    IF YOU HAVE A USE CASE FOR INHERITING FILE DESCRIPTORS, PLEASE CONTACT THE
+    OCEAN MAINTAINERS.
+
+    The following ocean functions obtain a file descriptor and use this flag:
+
+    - `ocean.sys.Epoll`: `Epoll.create`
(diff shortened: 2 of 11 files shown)


leandro-lucarella-sociomantic:
> To avoid a breaking change the close-on-exec option is disabled by default so it is up to user code to enable it. I think it should really be enabled by default because it protects from sporadic errors that can be severe and hard to debug -- or do we have an application that spawns a process which needs to use the parent's socket, timer, event or epoll file descriptors?

I'd suggest using this as the default starting from v4 if that makes sense.

leandro-lucarella-sociomantic:
> EpollProcess is affected as well, in particular the SIGCHLD select event in its GlobalProcessMonitor. Since a global singleton instance of this class is used I am not sure how to add close-on-exec without a breaking change.

Maybe some module global variable to set this up? Then also in v4 we can make it the default, but a mechanism should be still provided to get the old behaviour. I guess they might be legal use cases for it too.

src/ocean/sys/EventFD.d
```diff
@@ -118,7 +121,18 @@ import core.sys.posix.unistd: read, write, close;
 
 *******************************************************************************/
 
-private extern ( C ) int eventfd ( uint initval, int flags );
+extern ( C )
+{
+    private int eventfd ( uint initval, EFD flags );
```
leandro-lucarella-sociomantic: Wasn't this added already to tangort and druntime a long time ago?

src/ocean/sys/Inotify.d
```diff
@@ -165,18 +165,7 @@ public class Inotify : ISelectable
...
-
-        if ( fcntl(this.fd, F_SETFL, flags | O_NONBLOCK) < 0 )
-        {
-            scope (exit) errno = 0;
-
-            int errnum = errno;
-
-            throw this.e.set(errnum, identifier!(.fcntl));
-        }
+        this.fd = this.e.enforceRet!(inotify_init1)(&verify).call(IN_NONBLOCK);
```
gavin-norman-sociomantic: Could you explain why this change is made, in the commit message?

src/ocean/io/select/client/SelectEvent.d
```diff
@@ -87,11 +87,25 @@ public abstract class ISelectEvent : IAdvancedSelectClient
...
+          - `EFD.NONBLOCK`: Set the `O_NONBLOCK` file status flag on the new
+                open file description.
+          - `EFD.CLOEXEC`: Set the close-on-exec (`FD_CLOEXEC`) flag on the new
+                file descriptor. (See the description of the `O_CLOEXEC` flag
+                in `open(2)` for reasons why this may be useful.)
+          - `EFD.SEMAPHORE` (since Linux 2.6.30): Provide semaphore-like
+                semantics for reads from the new file descriptor.
+
+        Params:
+            initval = initial counter value
```
gavin-norman-sociomantic: I'd be inclined not to add the initial value setter. No one has ever needed this feature, so I wouldn't complicate the API for no reason.

src/ocean/io/select/client/SelectEvent.d
```diff
@@ -87,11 +87,25 @@ public abstract class ISelectEvent : IAdvancedSelectClient
         Constructor. Creates a custom event and hooks it up to the provided
         event handler.
 
+        Supported OR-combination of `flags`, starting with Linux 2.6.27:
+          - `EFD.NONBLOCK`: Set the `O_NONBLOCK` file status flag on the new
```
gavin-norman-sociomantic: Not sure about the value of exposing `NONBLOCK` and `SEMAPHORE` here. Again, it just makes the API more complicated for no real reason (i.e. there's no demand for those features). It's a grey area, I know... there _may_ be a future use case and just exposing the standard options is consistent with the system `eventfd` function. My feeling is slightly on the side of not copying the API of `eventfd` directly, though, as our higher level wrappers (like this class) have always presented a simplified API, just containing the functionality we need.

relnotes/cloexec-eventfd.feature.md
```diff
@@ -2,3 +2,24 @@
...
+  The `Inotify` and `TimerFD` constructor and `SignalFD.register` accept an
+  additional optional `close_on_exec` argument, which is `false` by default.
+  Setting this parameter to `true` enables the close-on-exec option on the
+  internal file descriptor of these facilites, which prevents inheriting this
+  file descriptor if the process calls `exec(2)`, which happens when spawning a
+  child process using `EpollProcess`, for example.
+  If the close-on-exec option is not set and a (parent) process spawns another
+  (child) process then the child process inherits the file descriptor. If the
+  parent process closes the file descriptor the associated device will stay open
+  until the child process terminstes or closes the same file descriptor as well.
```
gavin-norman-sociomantic: Typo "terminstes".

relnotes/cloexec-eventfd.feature.md
```diff
@@ -2,3 +2,24 @@
...
+  Most probably the child process will not do the latter because the program
+  that is executed does usually not know anything about file descriptors used by
+  the program that started it (perhaps with the exception of the standard I/O
+  devices).
+  A consequence of this is that file descriptore registered with epoll will not
```
gavin-norman-sociomantic: Typo "descriptore".

gavin-norman-sociomantic (changes requested):
Wow, this is a big change. All sociomantic programs that I can think of should set close-on-exec to true on all fds they create. Doing so would be very inconvenient, with the current form of this PR.

I wonder whether one of the following options might be better:
1. Just make this a breaking change in the next major version. v4.0.0 is (hopefully) a way off, though, so this isn't that great an option.
2. Add a global flag to toggle this option. It could default to false, but could then be switched on for all fds with a single assignment.

gavin-norman-sociomantic:
Re: the `EpollProcess.GlobalProcessMonitor` singleton...

>Maybe some module global variable to set this up? Then also in v4 we can make it the default, but a mechanism should be still provided to get the old behaviour. I guess they might be legal use cases for it too.

That would tie in with my proposal above (https://github.com/sociomantic-tsunami/ocean/pull/67#pullrequestreview-30002523).

david-eckardt-sociomantic:
Yes, this is a big change, and it's even bigger than this: All files opened by any program should be opened with close-on-exec, too. I didn't dare adding this to the Tango heritage.
When you look at the Linux manpages, `CLOEXEC` is everywhere where a file descriptor is created. I always thought this is an advanced feature for special purposes I'm not aware of, but it turns out the special purpose is actually _not_ using it.

src/ocean/sys/EventFD.d
david-eckardt-sociomantic: Oh, I found a couple of more places like this where functions are declared that are/should/will soon be in Tango. I would leave this cleanup to a separate PR.

src/ocean/io/select/client/SelectEvent.d
david-eckardt-sociomantic: It is true that this is a separate issue with no demand. Still I think it's bad design to not make these parameters available. But this should probably better be done in a separate PR.

relnotes/cloexec-eventfd.feature.md
david-eckardt-sociomantic: Latin. (Ok, ok, this is not ablative case.)

david-eckardt-sociomantic:
Updated with a completely different design: `ocean.sys.CloseOnExec` contains a global `open_with_close_on_exec` variable which is read by all functions in ocean that obtain a new file descriptor from the system. If it is true then they set the close-on-exec flag for the `open(2)`-like system call.
Right now `open_with_close_on_exec` is disabled by default so that this is not a breaking change, but it should be enabled in the next major ocean release unless someone has a use case for inheriting file descriptors.

relnotes/cloexec.feature.md
```diff
@@ -0,0 +1,77 @@
+* `ocean.sys.CloseOnExec`
+
+    * `open_with_close_on_exec` has been added. It is a global configuration
+      variable to define whether the close-on-exec option should be set on system
+      calls that create a file descriptor in ocean modules.
+
+    * `setCloExec` is a helper function o set the close-on-exec bit in a bit
```
gavin-norman-sociomantic: Typo "o set"

relnotes/cloexec.feature.md
```diff
@@ -0,0 +1,77 @@
...
+      mask which specifies option flags for a system call hat obtains a new file
```
gavin-norman-sociomantic: Typo "hat obtains"

relnotes/cloexec.feature.md
```diff
@@ -0,0 +1,77 @@
...
+     - A program opens a device (file, socket, timer or event fd).
+     - The program registers the device with `epoll`.
+     - The program uses a third-party library which starts a task in a separate
+       process using `fork` + `exec`.
+     - The original program (parent process) closes the device.
+     - The executed program (child process) is, as libraries are, unaware of
+       its parent's business so it doesn't close any of the inherited file
+       descriptors; it doesn't even know which it inherited. So the device stays
+       opened.
+     - Because the device is opened stays registered from `epoll`. The parent
```
gavin-norman-sociomantic: Missing "it"? "Because the device is open, it stays registered..."

gavin-norman-sociomantic: "Registered with" not "registered from"

relnotes/cloexec.feature.md
```diff
@@ -0,0 +1,77 @@
...
+       process cannot unregister it any more because it has closed the file
+       descriptor.
+     - Until the child process exits `epoll_wait(2)` keeps reporting events for
+       the device in the parent process.
+
+    This is a problem because can cause sporadic erratic behaviour in a program.
```
gavin-norman-sociomantic: "because can cause"?

src/ocean/sys/CloseOnExec.d
```diff
@@ -0,0 +1,122 @@
...
+    - `ocean.sys.EventFD`: `EventFD` constructor
+    - `ocean.sys.Inotify`: `Inotify` constructor
+    - `ocean.sys.SignalFD`: `SignalFD.register`
+    - `ocean.sys.TimerFD`: `TimerFD` constructor
+    - `ocean.sys.socket`: The `socket` and `accept` methods in all `*Socket`
+        classes
+    - `ocean.io.device.File`: `File.open`
+
+    Copyright:
+        Copyright (c) 2009-2017 Sociomantic Labs GmbH.
```
gavin-norman-sociomantic: Should only say 2017.

src/ocean/sys/CloseOnExec.d
```diff
@@ -0,0 +1,122 @@
...
+    descriptors.
+
+*******************************************************************************/
+
+mixin(global("bool open_with_close_on_exec = false"));
+
+/*******************************************************************************
+
+    Helper function to set the close-on-exec bit in a bit mask which specifies
+    option flags for a system call hat obtains a new file descriptor, such as
```
gavin-norman-sociomantic: "hat obtains"

src/ocean/sys/SignalFD.d
```diff
@@ -109,64 +109,17 @@ import ocean.core.Traits;
...
 
-import core.sys.posix.fcntl : O_NONBLOCK;
-
 import ocean.core.Array : contains;
 
 import ocean.transition;
 
 debug import ocean.io.Stdout;
 
-
```
gavin-norman-sociomantic, line 124: Is this change essentially unrelated?

gavin-norman-sociomantic (commented):
Much better.

src/ocean/sys/SignalFD.d
david-eckardt-sociomantic: Yes, it should be in a separate commit.

david-eckardt-sociomantic:
Update:
- Documentation
- Moved `SignalFD` cleanup to a separate module

david-eckardt-sociomantic:
An example what other people think:  https://mail.gnome.org/archives/gtk-devel-list/2015-March/msg00038.html

nemanja-boric-sociomantic:
Looks like Python opted in for CLOEXEC by default as well (if the `Final` status means it's accepted): https://www.python.org/dev/peps/pep-0446/

nemanja-boric-sociomantic:
Some other thoughts:

http://udrepper.livejournal.com/20407.html (from Ulrich Drepper)

> In glibc we do now in almost all interfaces open the file descriptor with the close-on-exec flag set. This means a lot of work but it has to be done.

gavin-norman-sociomantic:
So the alternative would be to modify [`Process`](https://github.com/sociomantic-tsunami/ocean/blob/v2.x.x/src/ocean/sys/Process.d) to `fork`, then close all fds, then `exec`?

nemanja-boric-sociomantic:
Yeah, but only to some extent (if the library itself calls `fork`/`exec`) and not via `ocean`'s `Process`, we're screwed again.

gavin-norman-sociomantic:
Ah, indeed. Do we use libraries that fork and exec? (Anyway, I think the close-on-exec solution is probably more robust.)

david-eckardt-sociomantic:
> So the alternative would be to modify Process to fork, then close all fds, then exec?

Yes, that seems to be the common pattern. However, _closing all file descriptors_ means brute force, i.e. call `close()` for each possible file descriptor number.
Right now I prefer flagging each file descriptor with `CLOEXEC`, though, to avoid hacking around with brute force (I'm not sure I'm never going to change my mind about it with more experience). A global variable should be sufficient for us. Maintainers of applications that make direct C library calls obtaining file descriptors and spawn new processes need to be aware of it, though.

david-eckardt-sociomantic:
> Do we use libraries that fork and exec?

@don-clugston-sociomantic, where is fork/exec called in your applications, in the 3rd-party libraries or in ocean?

nemanja-boric-sociomantic:
Not sure about if we're using libraries that do this. However, I think we all agree `O_CLOEXEC` solution is the most robust one and commonly used (according to all articles I could find).

The mailing list @david-eckardt-sociomantic linked explains later that the problem is the fact that GTK works on the platforms where thread-safe `O_CLOEXEC` approach is not reliable/possible, and that application developer needs to be careful. I don't think that applies to our usecase.

> Its a fact of life that O_CLOEXEC (and the threadsafe setting of it) is
not universally available on the systems we support glib on. This means
we *have* to close all open fds in e.g. g_spawn(), and no code can
safely rely on it having been used on all open fds.

> In general, setting O_CLOEXEC is a nice thing to do, but doing so does
not change the fundamental fact that you can't rely on it being set.

david-eckardt-sociomantic:
Updated, importing `ocean.sys.CloseOnExec` in every module it is used.

nemanja-boric-sociomantic (approved):

gavin-norman-sociomantic (approved):

gavin-norman-sociomantic:
Ok here goes nothing.

PR #68, opened 2017-04-07
Ameliorate DaemonApp and CliApp examples

daniel-zullo:
The variable `ver` is never used.

### DIFF
--- a/src/ocean/util/app/CliApp.d
+++ b/src/ocean/util/app/CliApp.d
@@ -233,7 +233,7 @@ unittest
             settings.help = "Actually, this program does nothing. Sorry!";
 
             // Call the super class' ctor.
-            super(name, desc, VersionInfo.init, settings);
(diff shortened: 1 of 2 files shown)


mihails-strasuns-sociomantic:
Change is OK but commit message is wrong `VersionInfo.init` does not initialize `VersionInfo`, it is a compile-time value of `VersionInfo` initializer ;)

daniel-zullo:
> Change is OK but commit message is wrong VersionInfo.init does not initialize VersionInfo, it is a compile-time value of VersionInfo initializer ;)

Right, I'll remove that part from the commit message and the PR description.

daniel-zullo:
Updated the commit message and issue description.

mihails-strasuns-sociomantic (approved):

daniel-zullo:
Thanks :+1:

PR #71, opened 2017-04-07
AddressIPSocket: Fix EFAULT on invalid IP address presentation string

david-eckardt-sociomantic:
Make AddressIPSocket methods fail with `EAFNOSUPPORT` as their documentation specifies, not `EFAULT`.

IP address presentation strings (that is, strings that are expected to contain the address in the "a.b.c.d" notation) accepted by `AddressIPSocket` methods are forwarded to `IPAddress.opCall()`, which calls `inet_pton(3)` to parse them, and returns a `sockaddr*` pointer on success or null on failure, i.e. `inet_pton` failed. On failure the returned null pointer was passed to `connect(2)`, which then fails with `EFAULT` _Bad address_ (this would be a segfault in user space).

Ironically _Bad address_ is actually a helpful error message to report an invalid IP address.

### DIFF
--- a/src/ocean/sys/socket/AddressIPSocket.d
+++ b/src/ocean/sys/socket/AddressIPSocket.d
@@ -173,8 +173,8 @@ class AddressIPSocket ( bool IPv6 = false ) : IPSocket!(IPv6), IAddressIPSocketI
             appropriately.
 
         Errors:
-            as above but also sets errno to EAFNOSUPPORT if the address does not
-            contain a valid IP address string.
+            as the overridden method but also sets errno to EAFNOSUPPORT if the
+            address does not contain a valid IP address string.
 
      **************************************************************************/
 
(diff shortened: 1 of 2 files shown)


nemanja-boric-sociomantic (approved):

david-eckardt-sociomantic:
Updated with improved tests, found a bug in `InetAddress.opCall()` and fixed it.

src/ocean/sys/socket/AddressIPSocket.d
```diff
@@ -393,3 +405,76 @@ class AddressIPSocket ( bool IPv6 = false ) : IPSocket!(IPv6), IAddressIPSocketI
         this.ip_address_len = this.in_address.inet_ntop(this.ip_address_).length;
     }
 }
+
+/// Test that methods that accept IP addresses in presentation notation
```
gavin-norman-sociomantic: I believe using `///` here will make the following unittest be rendered as a usage example, which isn't what we want.

gavin-norman-sociomantic (commented):
LGMT apart from one small point.

david-eckardt-sociomantic:
Updated, fixing unit test comment.

PR #72, opened 2017-04-07
Report an invalid IP address string via EINVAL, not EAFNOSUPPORT

david-eckardt-sociomantic:
Based on #71.

`EINVAL` is the appropriate code for this. `EAFNOSUPPORT` is wrong because it means a wrong address family was used, which is not the case here.
For functions that call `bind(2)`/`connect(2)` after `inet_ntop` this is in line with bind/connect error codes: Both return `EINVAL` for an invalid address argument as well.

### DIFF
--- a/relnotes/inetaddress-einval.migration.md
+++ b/relnotes/inetaddress-einval.migration.md
@@ -1,0 +1,9 @@
+* `ocean.sys.socket.InetAddress`, `ocean.sys.socket.AddressIPSocket`
+
+  Functions which accept an IP address string now report an invalid address by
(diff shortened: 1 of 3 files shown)


nemanja-boric-sociomantic:
This needs to go against `v4.x.x` branch. I guess it's time...

nemanja-boric-sociomantic:
v4.x.x is out, you can do the rebase on it @david-eckardt-sociomantic .

david-eckardt-sociomantic:
Ping.

mihails-strasuns-sociomantic (approved):

PR #61, opened 2017-03-23
Switch to using prebuilt docker image

mihails-strasuns-sociomantic:

PR #76, opened 2017-04-14
Allow contiguous copy of const structs

mihails-strasuns-sociomantic:
Serializer already supports it but `copy` definitions were not adjusted.

### DIFF
--- a/src/ocean/util/serialize/contiguous/Util.d
+++ b/src/ocean/util/serialize/contiguous/Util.d
@@ -21,10 +21,11 @@ module ocean.util.serialize.contiguous.Util;
 
 *******************************************************************************/
 
-import ocean.util.serialize.contiguous.Contiguous,
-       ocean.util.serialize.contiguous.Serializer,
-       ocean.util.serialize.contiguous.Deserializer;
+import ocean.util.serialize.contiguous.Contiguous;
+import ocean.util.serialize.contiguous.Serializer;
+import ocean.util.serialize.contiguous.Deserializer;
 
+import ocean.transition;
 import ocean.core.Test;
 
 /*******************************************************************************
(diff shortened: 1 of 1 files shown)


src/ocean/util/serialize/contiguous/Util.d
```diff
@@ -65,11 +66,25 @@ public Contiguous!(S) copy(S) ( Contiguous!(S) src, ref Contiguous!(S) dst )
 
 *******************************************************************************/
 
-public Contiguous!(S) copy(S) ( ref S src, ref Contiguous!(S) dst )
+version (D_Version2) // because `in ref` is not legal D1
```
gavin-norman-sociomantic: Where's `in ref`?

mathias-lang-sociomantic: `in ref` == `const ref`
Could it be `(ref Const!(S) src, ref Contiguous!(S) dst)` ?

gavin-norman-sociomantic: Right.

mihails-strasuns-sociomantic: @mathias-lang-sociomantic sadly it can't, because the function depends on type inference which `Const!(S)` breaks.

mihails-strasuns-sociomantic: `in S` would be usual workaround I do preserve type inference (`in` being same as `const` in D2) but D1 does not allow `in ref` combination, thus the `static if`.

mathias-lang-sociomantic (approved):

mihails-strasuns-sociomantic:
Ping @david-eckardt-sociomantic / @gavin-norman-sociomantic ? Last one for this point release

src/ocean/util/serialize/contiguous/Util.d
david-eckardt-sociomantic: Maybe update the comment, it mentions `in ref`, which isn't in the function signature, and doesn't mention that `Const!(S) src` breaks IFTI.

mihails-strasuns-sociomantic: Done + rebased on your serializer fixes

gavin-norman-sociomantic:
Alright then.

gavin-norman-sociomantic:
>Last one for this point release

There's also https://github.com/sociomantic-tsunami/ocean/pull/61.

mihails-strasuns-sociomantic:
Yeah I have move it to 2.6.5 it after this comment ;)

PR #78, opened 2017-04-18
Prevent clock drift in TimerExt by accounting for user delegate time.

scott-gibson-sociomantic:
Previously the timer would be registered using the interval without accounting for how long the user delegate took to complete. The extension now looks at how much time was taken to complete the user delegate and then adjusts the registration timer accordingly to get more consistant interval calls.

The current tests for the TimerExt are all disabled at the moment so i will look at seeing what i can do to re-enable them but adding the PR now.

### DIFF
--- a/relnotes/timerext.feature.md
+++ b/relnotes/timerext.feature.md
@@ -1,0 +1,7 @@
+* `ocean.util.app.ext.TimerExt`
+
+  `TimerExt` now accounts for the amount of time the user delegate took to
+  complete when scheduling the next timer call. If the user delegate takes
+  longer to complete than the interval, the next call will occur at an even
+  multiple of the specified period. (e.g. if the period is 5s and a call to
+  the delegate takes 7s, then the timer will fire 3s later.)
--- a/src/ocean/util/app/ext/TimerExt.d
+++ b/src/ocean/util/app/ext/TimerExt.d
@@ -108,6 +108,7 @@ public class TimerExt : IApplicationExtension
     import ocean.io.select.EpollSelectDispatcher;
     import ocean.io.select.client.TimerSet;
     import BucketElementFreeList = ocean.util.container.map.model.BucketElementFreeList;
+    import ocean.time.MicrosecondsClock;
     import ocean.time.timeout.TimeoutManager;
 
     /***************************************************************************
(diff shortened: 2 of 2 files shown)


src/ocean/util/app/ext/TimerExt.d
```diff
@@ -272,7 +277,14 @@ public class TimerExt : IApplicationExtension
         auto reregister = event.dg();
         if ( reregister )
         {
-            this.registerMicrosec(event.dg, event.repeat_microsec, event.repeat_microsec);
+            do
+            {
+                event.next_call += event.repeat_microsec;
+            }
+            while ( event.next_call < MicrosecondsClock.now_us() );
```
mathias-lang-sociomantic, line 286: Small style issue

gavin-norman-sociomantic: It's probably worth mentioning this behaviour in the method doc and the release notes.

scott-gibson-sociomantic: Description in release notes and method updated.

david-eckardt-sociomantic:
Hm, why does the `period_microsec` parameter of `registerMicrosec` not cover this case? In general, rather than calculating the next time of expiration everytime it makes more sense to setup the timer once to repeatedly expire with a given periodic interval.

relnotes/timerext.feature.md
```diff
@@ -0,0 +1,7 @@
+* `ocean.util.app.ext.TimerExt`
+
+  `TimerExt` now accounts for the amount of time the user delegate took to
+  complete when scheduling the next timer call. If the user delegate takes
+  longer to compelte than the interval, the calls for the intervals in the past
+  will be skipped.
```
gavin-norman-sociomantic: Typo "compelte".

Maybe this makes the behaviour a bit clearer?

"If the user delegate takes longer to complete than the interval, the next call will occur at an even multiple of the specified period. (e.g. if the period is 5s and a call to the delegate takes 7s, then the timer will fire 3s later.)"

scott-gibson-sociomantic:
Apologies for dropping the ball on this one, it appears i had a clock drift of myself. I have updated the release notes to fix the spelling and updated to gavin's suggestion.

scott-gibson-sociomantic:
I will look into improving the test coverage.

mihails-strasuns-sociomantic:
I don't think there will be any better test coverage considering most code is in app extension part.

@david-eckardt-sociomantic @gavin-norman-sociomantic does it look good?

gavin-norman-sociomantic (approved):

PR #80, opened 2017-04-19
Fix several Serializer and Deserializer bugs, especially const

### LINKED ISSUE #79: Deserializer trouble with const(T[])[] struct fields
`Deserializer` does not correctly deserialise `const(T[])[]` struct fields. The problem is a qualifier-incompatible way of checking for a dynamic array in [`sliceSubArrays`](https://github.com/sociomantic-tsunami/ocean/blob/v2.x.x/src/ocean/util/serialize/contiguous/Deserializer.d#L1050-L1052).
For the reference type qualifier inexperienced, this is how it fails:
```d
alias const(char[])[] T;
// Check if T is an array, and if so, derive the element type:
static if (is (T V : V[]))
{
    // V is const(char)
    // Now check if T is a dynamic array:
    static if (is(V[] == T)) // compares const(char)[] with const(char[]) -- not the same
    { /* ... */ }
    else
    { /* False negative, T is a dynamic array. */ }
}
```
So far so good, this can be fixed. But then another problem occurs: [`sliceSubArrays`](https://github.com/sociomantic-tsunami/ocean/blob/v2.x.x/src/ocean/util/serialize/contiguous/Deserializer.d#L1056) cannot call [`sliceArray`](https://github.com/sociomantic-tsunami/ocean/blob/v2.x.x/src/ocean/util/serialize/contiguous/Deserializer.d#L907) because the `out` parameter of `sliceArray` would need to be the array element type `const(char[])` -- which is illegal, an `out` parameter cannot be `const`.
This conflict is a symptom of a more fundamental problem with deserialising `const`. Given a struct
```d
struct S
{
    const(T[]) a;
}
```
What the deserialiser essentially does is the following:
```d
void[] input_data;
void[] slices_buf;
S dst;
dst.a = cast(const(T[]))(slices_buf[
    0 .. numberOfElements(input_data) * const(T[]).sizeof
]);
foreach (ref element; dst.a)
{
    element = deserialiseNextElement(data);
}
```
The problem here is that the assignment in the `foreach` loop is illegal because `typeof(element)` is `const(T[])` -- one cannot assign to `const`.
So what can we do here? The only options that come to my mind is to either ban such a `const` or cast it away when populating the array, considering `slices_buf` to be uninitialised memory.

Any ideas, anyone?

david-eckardt-sociomantic:
Fix #79 and a few more bugs I found on the way, also clean up the code a bit.
I made it so that the elements of a `const(T[])[]` array (which are of type `const(T[])`) will reside in the slices buffer which by itself is _not_ `const` but should never directly be written to anyway.

### DIFF
--- a/src/ocean/util/serialize/contiguous/Deserializer.d
+++ b/src/ocean/util/serialize/contiguous/Deserializer.d
@@ -552,7 +552,7 @@ struct Deserializer
 
 
 
-                static if (is (Element[] == Field))
+                static if (is (Element[] == RejectQualifier!(Field)))
                 {
                     // dynamic array
 
@@ -636,7 +636,7 @@ struct Deserializer
 
         This.e.enforceSizeLimit!(T[])(len, This.max_length);
 
-        static if (is (T U == U[]))
+        static if (is (RejectQualifier!(T) U == U[]))
         {
             /*
              * If array is an array of slices (dynamic arrays), obtain a data
@@ -729,7 +729,7 @@ struct Deserializer
         }
         else static if (is (T V : V[]))
         {
-            static if (is (V[] == T))
+            static if (is (V[] == RejectQualifier!(T)))
             {
                 for (size_t i = 0; i < len; i++)
                 {
@@ -866,8 +866,10 @@ struct Deserializer
 
                 pos += This.sliceArrays(*This.getField!(i, Field)(s), data[pos .. $], slices_buffer);
             }
-            else static if (is (Field Element : Element[]))
+            else static if (is (RejectQualifier!(Field) Element : Element[]))
             {
+                // To support const substitute Field with Unqual!(Field) in this
+                // scope.
                 static if (is (Element[] == Field))
                 {
                     This.e.enforceInputSize!(S)(data.length, pos);
@@ -944,7 +946,7 @@ struct Deserializer
 
         This.e.enforceSizeLimit!(T[])(len, This.max_length);
 
-        static if (is (T U == U[]))
+        static if (is (RejectQualifier!(T) U == U[]))
         {
             /*
              * If array is an array of slices (dynamic arrays), obtain a data
@@ -1047,8 +1049,10 @@ struct Deserializer
                 pos += This.sliceArrays(element, data[pos .. $], slices_buffer);
             }
         }
-        else static if (is (T V : V[]))
+        else static if (is (RejectQualifier!(T) V : V[]))
         {
+            // To support const substitute T with Unqual!(T) in this scope and
+            // cast(Unqual!(T)[])array.
             static if (is (V[] == T)) foreach (ref element; array)
             {
                 This.e.enforceInputSize!(T[])(data.length, pos);
@@ -1088,6 +1092,23 @@ struct Deserializer
     private static T* getField ( size_t i, T, S ) ( ref S s )
     {
         return cast (T*) ((cast (void*) &s) + S.tupleof[i].offsetof);
+    }
+
+    /**************************************************************************
+
+        `static assert`s that `T` has no qualifier.
+
+     **************************************************************************/
+
+    template RejectQualifier ( T )
+    {
+        version (D_Version2)
+            static if (!is(T == typeof(mixin("cast()T.init"))))
+            {
+                pragma(msg, "Warning: ", T,
+                " -- type qualifiers are not supported");
+            }
+        alias T RejectQualifier;
     }
 }
 
--- a/src/ocean/util/serialize/contiguous/Serializer.d
+++ b/src/ocean/util/serialize/contiguous/Serializer.d
@@ -417,7 +417,16 @@ struct Serializer
 
             foreach (subelement; element)
             {
-                len += This.countElementSize(subelement);
+                static if (is(Unqual!(Base) Sub == Sub[]))
+                {
+                    // subelement is a dynamic array
+                    len += This.countArraySize(subelement);
+                }
+                else
+                {
+                    // subelement is a value containing dynamic arrays
+                    len += This.countElementSize(subelement);
+                }
             }
 
             return len;
@@ -707,7 +716,7 @@ struct Serializer
 
             foreach (ref element; array)
             {
-                data = This.dumpElement(element, data);
+                data = This.dumpStaticArray(element[], data);
                 This.resetArrayReferences(element);
             }
         }
--- a/src/ocean/util/serialize/contiguous/Serializer.d
+++ b/src/ocean/util/serialize/contiguous/Serializer.d
@@ -289,11 +289,7 @@ struct Serializer
                     {
                         // Recurse into static array elements which contain a
                         // dynamic array.
-
-                        foreach (element; s.tupleof[i])
-                        {
-                            len += This.countArraySize(field);
-                        }
+                        len += This.countElementSize(field);
                     }
                 }
                 else static if (is (T == union))
--- a/src/ocean/util/serialize/contiguous/Serializer.d
+++ b/src/ocean/util/serialize/contiguous/Serializer.d
@@ -622,6 +622,7 @@ struct Serializer
                 {
                     // element is a dynamic array
                     data = This.dumpArray(element, data);
+                    *(cast(T*)&element) = null;
                 }
                 else
                 {
--- a/src/ocean/util/serialize/contiguous/Serializer.d
+++ b/src/ocean/util/serialize/contiguous/Serializer.d
@@ -696,7 +696,7 @@ struct Serializer
         }
         else static if (is (T Base : Base[]))
         {
-            static assert (!is (Base[] == T),
+            static assert (!is (Base[] == Unqual!(T)),
                "expected static, not dynamic array of " ~ T.stringof);
 
             debug (SerializationTrace)
--- a/src/ocean/util/serialize/contiguous/Serializer.d
+++ b/src/ocean/util/serialize/contiguous/Serializer.d
@@ -275,21 +275,26 @@ struct Serializer
 
                     len += This.countAllArraySize(field);
                 }
-                else static if (is (T Base == Base[]))
-                {
-                    // Dump dynamic array.
-
-                    len += This.countArraySize(field);
-                }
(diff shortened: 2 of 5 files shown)


gavin-norman-sociomantic:
If this is a bug fix, it should be based on v2.6.x.

gavin-norman-sociomantic:
>Use `ref` iteration over struct fields

Could you add an explanation of why this change is made, in the commit message?

src/ocean/util/serialize/contiguous/Serializer.d
```diff
@@ -290,7 +290,7 @@ struct Serializer
                         // Recurse into static array elements which contain a
                         // dynamic array.
 
-                        foreach (element; s.tupleof[i])
```
gavin-norman-sociomantic: This change (and other similar ones, below) looks unrelated to "Use `ref` iteration over struct fields".

gavin-norman-sociomantic:
>Deserializer: Fix deserialising qualified (i.e. const) arrays
>Serializer: Fix qualified dynamic arrays not detected

Would be nice to have unittests to confirm these fixes.

gavin-norman-sociomantic:
>Unqual whole array at once rather than each element at a time

Could you add an explanation of why this change is made, in the commit message?

src/ocean/util/serialize/contiguous/Serializer.d
david-eckardt-sociomantic: This commit replaces
```d
foreach (i, T; typeof(s.tupleof)) { /* use s.tupleof[i] to access the i-th field of s */ }
```
with
```d
foreach (i, ref field; s.tupleof) { /* use `field` to access the i-th field of s */ }
```
So it changes both the `foreach` iteration variable to `ref field` and `s.tupleof[i]` to `field` in the loop body.
This particular line here adjusts `s.tupleof[i]` -> `field` that seems to have been missed when the same change had been applied in a previous commit.
Or did you spot a change in this commit that does something different?

david-eckardt-sociomantic:
> Would be nice to have unittests to confirm these fixes.

I added it to the `ocean.util.serialize.contiguous.package_` test.

src/ocean/util/serialize/contiguous/Serializer.d
gavin-norman-sociomantic: Ah I see. I didn't get the connection between the use of `ref` and the change of `s.tupleof[i]` -> `field`. Maybe just add a bit of explanation in the commit message.

gavin-norman-sociomantic:
>I added it to the ocean.util.serialize.contiguous.package_ test.

In the same commits?

david-eckardt-sociomantic:
> In the same commits?

Ha, no. One commit too late.

gavin-norman-sociomantic:
Maybe put the unittests in the same commits. That'll make it clearer.

david-eckardt-sociomantic:
Updated:
- Addressed @gavin-norman-sociomantic's comments.
- Fixed more bugs, most importantly related to static arrays of dynamic arrays.
- Added support for `const(T[][])` and `const(T[])[n]` (which is the same as `const(T[][n])`).
- Fixed and extended `debug(De/SerializationTrace)` printouts.

I still have to look into `Serializer.resetArrayReferences`, which is not covered by the test and looks `const`-suspicious.

gavin-norman-sociomantic:
On a very quick look, it seems that this PR mixes bug-fixes and refactorings. Sorry -- I know it'll be hassle -- but these should be separated. We shouldn't merge refactorings into v2.6.x.

nemanja-boric-sociomantic:
Maybe it is worth running both D1 and D2 build locally (or on test server where we have RAM) with a profiler and run GetAll on a big channel and see if there's anything that stands out in D2 build. Given that most of the work is done inside the tokyocabinet, maybe it will be obvious what's the difference.

david-eckardt-sociomantic:
Fair enough.

david-eckardt-sociomantic:
Separating bugfixes and refactorings shouldn't be too difficult.

gavin-norman-sociomantic:
Ok great.

src/ocean/util/serialize/contiguous/Serializer.d
david-eckardt-sociomantic: Ironically this loop has always been broken in two ways: Doing the `foreach` loop is wrong, and in the loop `element` should be passed, not `field`.

david-eckardt-sociomantic:
Updated, reducing the changes to the bug fixes without unrelated refactoring.

src/ocean/util/serialize/contiguous/Serializer.d
```diff
@@ -622,6 +622,7 @@ struct Serializer
                 {
                     // element is a dynamic array
                     data = This.dumpArray(element, data);
+                    *(cast(U*)&element) = null;
```
mihails-strasuns-sociomantic, line 641: Out of curiosity - is there any point in doing so other than simplifying debugging? Deserializer doesn't care if pointers are null or garbage.

src/ocean/util/serialize/contiguous/package_.d
```diff
@@ -696,3 +697,46 @@ unittest
...
+}
+
+unittest
+{
+    void[] buffer;
+
+    ConstS s;
+
+    Serializer.serialize(s, buffer);
+    auto cont_S = Deserializer.deserialize!(ConstS)(buffer);
```
mihails-strasuns-sociomantic: As mentined in other issue, my intention was to not allow this, requiring target struct to be 100% mutable. Exactly because it simplifies (de)serializer code a lot while not being relevant to our typical use cases. What has caused you to pursue such struct support? Original plan was to go with using mutable structs internally and aplying `const` to them as a whole when accepting external params.

mihails-strasuns-sociomantic (commented):
Other than `const` concerns I have mentioned, I would really appreciate if every commit that fixes something included matching regression unittest. Serializer code base is rather fragile and convoluted, it would be risky to live even single line change to chance.

src/ocean/util/serialize/contiguous/package_.d
david-eckardt-sociomantic: The motivation is to support struct fields containing an array of swarm channels (i.e. strings) where the original channels, passed by the user to the swarm API, can be `immutable(char[])[]`.

david-eckardt-sociomantic: Sorry, I didn't read your comment very well :blush: -- let me check if there was a complication with the original plan.

mihails-strasuns-sociomantic: Please also check with @gavin-norman-sociomantic - we have just had a slack discussion about this and he went to try another approach in dhtproto. It looks like things got a bit of out of control because I expected deserializer to refuse const types and it silently ingored them instead :( Going to see if that alternative approach proves feasible before making decision about serializer.

david-eckardt-sociomantic:
>  I would really appreciate if every commit that fixes something included matching regression unittest.

A regression test is a test that will fail if a fixed bug is re-added again, correct?

mihails-strasuns-sociomantic:
> A regression test is a test that will fail if a fixed bug is re-added again, correct?

Yes. In this context that means that every commit that fixes something should include test that fails without that new fix.

david-eckardt-sociomantic:
There are still several dynamic array detections which incorrectly do not unqualify the input type so they should detecting `const(T[])` as a dynamic array, but surprisingly they succeed. This is my surprise:
```d
alias const(char[]) T;
static if (is(T U == U[]))
    static assert(is(T == U[])); // fails -- why is it true in the static if?
```
The [D2 specification](http://dlang.org/spec/expression.html#IsExpression) says
> The condition is satisfied if _Type_ [i.e. `T`] is semantically correct and is the same as _TypeSpecialization_ [i.e. `U[]`].

In the example the condition is satisfied but they are not the same as the `static assert` confirms. Is this a compiler bug? I couldn't find an issue on https://issues.dlang.org/ so far.

If it was really a compiler bug it would hide bugs in the `Serializer` code.

david-eckardt-sociomantic:
DMD bug reported: https://issues.dlang.org/show_bug.cgi?id=17353

david-eckardt-sociomantic:
Updated:
- Regression tests
- Added `Unqual` to dynamic array type checks including those that would work by accident due to DMD bug #17353.

src/ocean/util/serialize/contiguous/Serializer.d
```diff
@@ -622,6 +622,7 @@ struct Serializer
                 {
                     // element is a dynamic array
                     data = This.dumpArray(element, data);
+                    *(cast(T*)&element) = null;
```
mihails-strasuns-sociomantic: This is going to be a problem with https://github.com/sociomantic-tsunami/ocean/pull/88 though - or do we have a way to reset all pointers to null in-place?

mihails-strasuns-sociomantic:
@david-eckardt-sociomantic I see that it still has changes to support `const` in deserializer - could you please move them to separate PR so that non-controversial serializer fixes can be merged ASAP?

david-eckardt-sociomantic:
Updated, now rejecting type qualifiers in `Deserializer`. This changes only one commit, `De/Serializer: Reject and support const(T[]) and const(T[])[n] fields, resp`.

mihails-strasuns-sociomantic (approved):

src/ocean/util/serialize/contiguous/Deserializer.d
```diff
@@ -1089,6 +1093,20 @@ struct Deserializer
...
+    /**************************************************************************
+
+        `static assert`s that `T` has no qualifier.
+
+     **************************************************************************/
+
+    template RejectQualifier ( T )
+    {
+        version (D_Version2)
+            static assert(is(T == typeof(mixin("cast()T.init"))),
```
mihails-strasuns-sociomantic: As this is bug fix release, please make it `pragma` for now!

mihails-strasuns-sociomantic (changes requested):
One last request :)

mihails-strasuns-sociomantic (approved):

PR #81, opened 2017-04-21
Add missing break to prevent implicit fall-through

nemanja-boric-sociomantic:
As it is not allowed in D2.


gavin-norman-sociomantic (approved):

mathias-lang-sociomantic:
Why didn't the auto-tester catch this ? Which version of DMD2 are you using ?

PR #83, opened 2017-04-21
Fix documentation for sformat

joseph-wakeling-sociomantic:
These patches fix the incorrect documentation for the `ocean.text.convert.Formatter.sformat` overload that takes an `mstring buffer` as input, and rename it to `aformat`, leaving behind a deprecated wrapper function for backwards compatibility.  (Note, an alias is not possible because of the second `sformat` overload that takes a `Sink` as input.)

The original `sformat` documentation stated that the formatted string would be written _into_ the provided buffer, when in fact it is appended to it.  The choice of name seems to match better the original documentation than the correct documentation (by analogy to `snformat` in the same module, C's `sprintf`, and D2 phobos' `sformat`, all of which also write _into_ a buffer); the rename to `aformat` ('append format') is chosen to better distinguish the use-case.

Uses in ocean code have been switched to the new name, but a `deprecated unittest` block has been added to `ocean.text.convert.Formatter` to ensure that the behaviour of the old API continues to be validated.

### DIFF
--- a/src/ocean/text/convert/Formatter.d
+++ b/src/ocean/text/convert/Formatter.d
@@ -73,11 +73,11 @@ public istring format (Args...) (cstring fmt, Args args)
 
 /*******************************************************************************
 
-    Write the processed (formatted) input into a buffer
+    Append the processed (formatted) input onto the end of the provided buffer
 
     Params:
-        buffer  = The buffer to write the formatted string into, will be
-                  extended if needed
+        buffer  = The buffer to which to append the formatted string; its
+                  capacity will be increased if necessary
         fmt     = Format string to use
         args    = Variadic arguments to format according to `fmt`
 


src/ocean/text/convert/Formatter.d
```diff
@@ -86,7 +86,7 @@ public istring format (Args...) (cstring fmt, Args args)
 
 *******************************************************************************/
 
-public mstring sformat (Args...) (ref mstring buffer, cstring fmt, Args args)
+public mstring aformat (Args...) (ref mstring buffer, cstring fmt, Args args)
```
joseph-wakeling-sociomantic: Note, the original equivalent in `ocean.text.convert.Layout_tango` was `Layout.format`, so we could go back to using the `format` name instead of introducing this new one.

joseph-wakeling-sociomantic:
:police_car: I know this PR still needs release notes added, but I'd like to hold off writing them until it's decided whether this change is acceptable ;-)

mathias-lang-sociomantic:
The documentation fix is very welcome.

As for the rename, I think it's a high cost for no added value. This overload is the most used one in performance sensitive applications, so it would lead to a lot of work for downstream users at no benefit.

mihails-strasuns-sociomantic:
Why deprecate old one as opposed to having two where one does automatically reset the buffer and other doesn't?

joseph-wakeling-sociomantic:
> Why deprecate old one as opposed to having two where one does automatically reset the buffer and other doesn't?

How would you distinguish the two in terms of function signatures?

The intent of these changes is to pave the way towards introducing a new `sformat` in a later PR which follows the claim of the original docs (i.e. it writes into the buffer instead of appending to it).

joseph-wakeling-sociomantic:
> As for the rename, I think it's a high cost for no added value. This overload is the most used one in performance sensitive applications, so it would lead to a lot of work for downstream users at no benefit.

Point is, the name seems to have been chosen in line with the original, faulty, docs, and indeed is better suited to that use-case.  The cost for downstreams is pretty much just a single `sed` command or other search'n'replace.

mathias-lang-sociomantic:
> The cost for downstreams is pretty much just a single `sed` command or other search'n'replace.

No it's not, because we have many libraries using this functionality. So for example any usage of `sformat` in swarm v4.x.x will trigger this deprecation message unless special casing is introduced in swarm using the version of ocean.

joseph-wakeling-sociomantic:
> So for example any usage of sformat in swarm v4.x.x will trigger this deprecation message unless special casing is introduced in swarm using the version of ocean.

Sure, but let's distinguish between two different things: (i) the value of the change in its own right and (ii) the context in which to introduce it for minimal disturbance.

For example, it would be straightforward to introduce the rename + deprecation in the next major release of ocean.

joseph-wakeling-sociomantic:
> For example, it would be straightforward to introduce the rename + deprecation in the next major release of ocean.

Or, preferably, in the 3.x release prior to the next major swarm release.

gautam-kotian-sociomantic:
>let's distinguish between two different things: (i) the value of the change in its own right and (ii) the context in which to introduce it for minimal disturbance.

No idea about the context in which to introduce it, but I feel that renaming `sformat` to `aformat` is generally valuable.

>The intent of these changes is to pave the way towards introducing a new sformat in a later PR which follows the claim of the original docs (i.e. it writes into the buffer instead of appending to it).

That, I fear, will only introduce a lot of confusion. Especially when inspecting old code, people would need to be extra careful to see what `sformat` means at a particular old commit.

I'm just curious, what does the `s` in `sformat` stand for? I would propose having two functions `aformat` & `iformat` which I could think of as "append format" and "insert format" respectively -- think `vim` ;-)

nemanja-boric-sociomantic:
I assume `s` comes from `sprintf` which I always assumed comes from `string` (print formatted to string).

joseph-wakeling-sociomantic:
> That, I fear, will only introduce a lot of confusion. Especially when inspecting old code, people would need to be extra careful to see what sformat means at a particular old commit.

Yea, good point.

The rationale for repurposing the `sformat` name is that it brings it in line with the existing `snformat` method (and the `sformat` in phobos, which I guess is probably the naming influence here).  The consistency of meaning there is probably worth aiming for.  (And yes, as @nemanja-boric-sociomantic says, I assume this name essentially means 'string format'.)

I agree that it risks some problems interpreting old code, but at the end of the day, the old code comes with old submodules where the particular function in use can be validated.  I'd rather go for consistent naming patterns in the long run, than introduce a new naming convention to work around what was arguably a mistake in the original code.

src/ocean/text/convert/Formatter.d
llucax: That's what I was thinking, if there is no conflict with just `format()` maybe we should use that.

mathias-lang-sociomantic: We cannot use it because it's ambiguous. Considering the signatures:
```D
public istring format (Args...) (cstring fmt, Args args);
public mstring [s]format (Args...) (ref mstring buffer, cstring fmt, Args args);
```

The call is ambiguous in D1 in the following situation:
```D
istring fmt = "This is an {}";
format(fmt, "lvalue");
// ^ Any additional arguments are also ambiguous
```

It's also ambiguous with a mutable format string in D2.

andrew-stevenson-sociomantic:
> I assume s comes from sprintf which I always assumed comes from string (print formatted to string).

That's correct - but if we follow C naming schemes then 'a' implies it will allocate. I'm not sure we should restrict ourselves to such things though if it makes sense not to.

I use sformat to append in propulsor (and will use it more over time I expect). I was surprised that format appended but I don't really mind what names are chosen but perhaps there should be a table in the comments explaining your choices...e.g.

```
A number of permutations are available:

method | 1st param | usage
---------------------------
1format | mstring | format into an mstring, destroying existing contents and reallocating if required
2format | mstring | append to an mstring, preservering existing contents and reallocating if required
3format | mstring | format into an mstring, destroying existing contents. will truncate if dst is too small. will never reallocate.
4format | Stream | format and output to a Stream
```
etc. I just made those up - I forget what variations we actually have.

mihails-strasuns-sociomantic:
Sorry for the late response, I was thinking about the topic in a far background of my mind and this is what I want to propose:

1) Don't touch/deprecate existing format functions for now, I agree that the change is too intrusive and not important on its own to jumpt straight to deprecation
2) Introduce new `fmt` function which will be advertised as more principled replacement and which will step away from C style naming conventions completely

Something like this:

```D
    FormatWrapper!(T) fmt ( T... ) ( T args ) {
        return typeof(return)(args);
    }

    struct FormatWrapper ( T... ) {
        private T args;
        // format by allocating new string, can also be plain `toString()`
        mstring alloc ( );
        // append to existing aray/buffer
        ref mstring appendTo ( ref mstring );
        ref Buffer!(char) appendTo ( ref Buffer!(char) );
        // re-use existing array/buffer completely
        ref mstring overwrite ( ref mstring );
        ref Buffer!(char) overwrite ( ref Buffer!(char) );
    }
```
Envisioned usage example:

```D
    auto s = fmt("{}", 42).alloc();
    fmt("{}", 43).appendTo(s);
    fmt("done").overwrite(s);
```

What do you think?

gavin-norman-sociomantic:
We probably should have taken the opportunity (while writing the new formatter module) to move away from these cryptically named functions and toward something more obvious :/

mihails-strasuns-sociomantic:
Well, that opportunity was missed :( Still not too late.

joseph-wakeling-sociomantic:
@mihails-strasuns-sociomantic:
> 1. Don't touch/deprecate existing format functions for now, I agree that the change is too intrusive and not important on its own to jumpt straight to deprecation
> 2. Introduce new fmt function which will be advertised as more principled replacement and which will step away from C style naming conventions completely

Agree that deprecation right now is not the best option.  The thought of having an entirely new naming scheme separate from C conventions is an interesting one.

On that note: while I recognize that ocean is not phobos (and doesn't want to be), one motivation for wanting to change the name was to try to have a "least surprises" naming convention w.r.t. typical D2 code.  Phobos' `sformat` is similar to our `snformat` IIRC (there is no Phobos formatting function that I can think of that writes into anything other than a fixed-length buffer).

@andrew-stevenson-sociomantic 
> but if we follow C naming schemes then 'a' implies it will allocate.

Which, technically, it may (through extending the buffer), but I guess that might not be in line with the C expectation?

Would `xformat` (for 'extend') be better? :-)

joseph-wakeling-sociomantic:
Off-topic, but: I think it could be good to target a talk for next year's DConf on how we use process to manage breaking change ;-)

gavin-norman-sociomantic:
>Off-topic, but: I think it could be good to target a talk for next year's DConf on how we use process to manage breaking change ;-)

Heh heh. Yes!!

mathias-lang-sociomantic:
> Envisioned usage example:
>
> ```D
>    auto s = fmt("{}", 42).alloc();
>    fmt("{}", 43).appendTo(s);
>    fmt("done").overwrite(s);
> ```
>
> What do you think?

What's the driver / main motivation for that approach ? Does it allow a pattern that's not currently possible ? I'm probably a bit biased, but it seems much more complicated that what's currently available. Also I think the `sink` approach currently used is quite valuable, and this seems to discard it ?

andrew-stevenson-sociomantic:
> but I guess that might not be in line with the C expectation?

In the C world it would suggest it takes a pointer to a pointer and the format function will call malloc. You can read all the gory details in the man page... https://www.freebsd.org/cgi/man.cgi?query=asprintf&apropos=0&sektion=0&manpath=FreeBSD+11.0-RELEASE+and+Ports&arch=default&format=html

> Would xformat (for 'extend') be better? :-)

I'm sure whatever you all come up with will be fine :-)

joseph-wakeling-sociomantic:
> What's the driver / main motivation for that approach ? Does it allow a pattern that's not currently possible ? I'm probably a bit biased, but it seems much more complicated that what's currently available.

@mihails-strasuns-sociomantic's approach is nice inasmuch as it separates out _formatting_ from how the formatted string is _written_.

I guess one could object that it's a little verbose but I think that's a small price to pay for the clarity/lack of ambiguity.  We could also readily avoid that verbosity with a few common use-case overloads (e.g. a default `format` function that allocates, and which if given a sink or a buffer will write into that).

I do particularly like the clear separation of `appendTo` offered by this proposal.

joseph-wakeling-sociomantic:
> Also I think the sink approach currently used is quite valuable, and this seems to discard it ?

@mihails-strasuns-sociomantic's proposal seems compatible with sinks.  We can have a `writeInto` (probably more general than `overwrite`) that can take either a buffer or a sink as input.

mihails-strasuns-sociomantic:
@mathias-lang-sociomantic 

> What's the driver / main motivation for that approach ?

I had two rationales in mind when proposing it:

1) More clear/readable semantics of the function with little to no arcane naming conventions
2) It becomes safe and simple to extend `FormatWrapper` with more methods using UFCS (i.e. in stream module) as it will have own dedicated type that won't clash with anything else

> I'm probably a bit biased, but it seems much more complicated that what's currently available.

Why? It is literally the same code as righ now but with wrapping functions into one common struct. I even expect compiler to generate roughly the same code with `-inline`, optimizing extra stack allocation away.

> @mihails-strasuns-sociomantic's proposal seems compatible with sinks. We can have a writeInto (probably more general than overwrite) that can take either a buffer or a sink as input.

Indeed, I would envision `append` to be overloaded to work on any sink / output range (`overwrite` can't because one can't reset arbitrary output range)

mihails-strasuns-sociomantic:
btw @joseph-wakeling-sociomantic 

> there is no Phobos formatting function that I can think of that writes into anything other than a fixed-length buffer

You have missed the cornerstone of all Phobos formatting - `formattedWrite` (https://dlang.org/library/std/format/formatted_write.html) :) Which works on any output range including plain arrays by appending to them.

mathias-lang-sociomantic:
> @mihails-strasuns-sociomantic's approach is nice inasmuch as it separates out formatting from how the formatted string is written.

That's exactly why everything in the `Formatter` module is built on delegates. If you look closely, you'll see that `format`, `snformat` and the `sformat` overload taking a `ref mstring` all end up calling `sformat(sink, fmtString, args)`.
This is because arbitrary output logic can be expressed by using delegates: you can limit the output (`snformat`), choose to write to a GC-allocated string, choose to write to a `malloc`-allocated string, to a `Buffer`... Hence why I currently fail to see the advantage of such an approach, because everything can already be done via delegates.

mathias-lang-sociomantic:
By the way, what method would you expose to write to a static array ? `overwrite` is not suitable for the task as it takes a `ref`.

> More clear/readable semantics of the function with little to no arcane naming conventions

It's 3 (related) names to remember for any string formatting, I'd argue it's far from being complicated, even if it can surprise some people. What is intuitive to one might not be intuitive to another, and documentation should clear up any question, and in this case it didn't, which is how this issue started.

> It becomes safe and simple to extend `FormatWrapper` with more methods using UFCS (i.e. in stream module) as it will have own dedicated type that won't clash with anything else.

As far as I can see, writing to a stream using the formatter is already possible using the sink overload.
I don't see any D2-specific code coming in ocean this year, and while I agree we should be careful w.r.t. the future, whether or not UFCS will provide a more user-friendly interface is quite hard to tell at this point.

> You have missed the cornerstone of all Phobos formatting - formattedWrite (https://dlang.org/library/std/format/formatted_write.html) :) Which works on any output range including plain arrays by appending to them.

Yeah this is also something we could consider. Althought delegates and output ranges have a lot in common.

mihails-strasuns-sociomantic:
> By the way, what method would you expose to write to a static array ? overwrite is not suitable for the task as it takes a ref.

How is it not suitable? I have a feeling that you somehow misunderstand what I propose and assume it will be any different from current implementation.

```D
struct FormatWrapper (T...) {
   private T args

   public void overwrite ( ref char[] buffer ) {
       buffer.length = 0;
       enableStomping(buffer);
       .sformat(buffer, args);
   }
}
```

mihails-strasuns-sociomantic:
Let's also refresh the perspective a bit - you ask about what new API brings to the table, but I want to ask what does it _lose_ - to which answer is "nothing". API learning curve improvement with no functionality loss is already good enough per my standards to go for it as it is net positive.

mihails-strasuns-sociomantic:
> How is it not suitable?

Ah, sorry, I have only just got your question. Don't see a problem with dedicated `overwriteNoResize` method or any other method you may want. That is the very point of new API - to be able to add new helper methods arbitrarily as needed.

joseph-wakeling-sociomantic:
> API learning curve improvement with no functionality loss is already good enough per my standards to go for it as it is net positive.

Yup, code clarity is always a nice win.  It also makes ocean formatting functionality sufficiently different from Phobos (and C) that there's much less likelihood of confusion over what does what.

leandro-lucarella-sociomantic:
I will only make one useless comment. I find `fmt("{}", 42).alloc()` (and the rest) overly verbose. I know you people like `longJavaLikeNames` so I won't defend my posture, I just want to complain about it... :trollface:

joseph-wakeling-sociomantic:
> I find `fmt("{}", 42).alloc()` (and the rest) overly verbose.

It's a fair comment, even allowing for trollface ;-)  But I reckon we could have some easy and compact wrapper functions for common use cases.

mihails-strasuns-sociomantic:
That is why I have suggested `toString` as an option so that it becomes simply `fmt("{}", 42)` for quick and dirty scripting style ;) For production critical apps long names tend to be least of all possible problems.

gavin-norman-sociomantic:
Seems like the discussion has opened some larger questions here, so we'll delay this until v2.9.0.

mihails-strasuns-sociomantic:
Ping @sociomantic-tsunami/core-team - I would like some final decision to be made and get done with it, no point in keeping PR open and stalled.

mathias-lang-sociomantic:
I think the best course of action would be to strip this P.R. to only the first commit.

gavin-norman-sociomantic:
>I think the best course of action would be to strip this P.R. to only the first commit.

:+1: yeah, the second commit is where it gets controversial. This part could be replaced with another PR with more comprehensive name changes.

mihails-strasuns-sociomantic:
Well, considering such follow-up PR will certainly meet same indecision, it is hardly any different from "fix docs and don't change anything" decision :) I don't mind though.

gavin-norman-sociomantic:
>Well, considering such follow-up PR will certainly meet same indecision

How so? Presumably it would implement a completely new / different naming scheme to replace the C-style one?

mihails-strasuns-sociomantic:
Isn't 99% of discussion in this PR exactly about disagreement on naming scheme? :)

gavin-norman-sociomantic:
Yes. I would assume a better naming scheme would be agreed upon before someone would bother making a new PR.

mihails-strasuns-sociomantic:
Yes, and I proposed exactly to make that decision. Voting, flipping the coin, lengthy lead meeting debate - doesn't matter, let's just get done with it.

gavin-norman-sociomantic:
Why is this urgent now, though? Why not just do as Mathias suggested: merge the non-polemic part of this PR now and discuss the naming scheme as a separate thing?

mihails-strasuns-sociomantic:
It is not urgent but it will also never be urgent unless you say "ok, decision has to be done now or the issue closed". I don't object merging trivial part now but I don't want splitting another issues that will hang in issue tracker for years more because no one wants to touch controversial topics.

joseph-wakeling-sociomantic:
> I don't object merging trivial part now but I don't want splitting another issues that will hang in issue tracker for years more because no one wants to touch controversial topics.

When is the next major release of ocean planned?

I would suggest that we keep the second patch in mind to apply to 3.x.x once we're on the cusp of making a 4.0.0 release. That should give plenty of time for some broader discussion of what kind of formatting API we would like, but gives us a default fallback of just applying the name change as-is (which hopefully gives some impetus to actually examine the question in more detail, and hopefully come up with a better overall design).

mihails-strasuns-sociomantic:
"When it is necessary" ;)

Can you remove second commit for now? I'll merge the PR after.

joseph-wakeling-sociomantic:
OK.  Shall I also rebase on latest 2.x.x ... ?

joseph-wakeling-sociomantic:
Rebased on latest `v2.x.x` and dropped the rename patch.

joseph-wakeling-sociomantic:
> Codecov Report

Oh nice, I like this game!  Can we play it again? :-)

mihails-strasuns-sociomantic (approved):

PR #91, opened 2017-04-28
Rename constructor parameter of the IOException to error_num

nemanja-boric-sociomantic:
IOException has the member `error_num`, and it does the assignment
of it in the constructor: `this.error_num = error_num`. However, the
argument is called `error_code`, so the `this.error_num = error_num`
does nothing.


mihails-strasuns-sociomantic (approved):

nemanja-boric-sociomantic:
Green.

PR #93, opened 2017-04-29
Remove cyclic dependencies

Geod24:
Shows up with a recent druntime.

### DIFF
--- a/src/ocean/time/WallClock.d
+++ b/src/ocean/time/WallClock.d
@@ -167,16 +167,3 @@ struct WallClock
                 return fromDate(date) + TimeSpan(mod);
         }
 }
-
-
-version (Posix)
-{
-    version (darwin) {}
-    else
-    {
-        static this()
-        {
-            tzset();
-        }
-    }
-}


mihails-strasuns-sociomantic:
I don't like "fixing" cyclic imports by removing module constructors, it should be exactly other way around - cyclic import which is to be eliminated while module import being perfectly valid feature.

For example, I'd propose removing `DateTime_tango` import from `ocean.time.Time` and changing output of its `toString` to plain integer timestamp (breaking change), requiring all smart formatting to be done explicitly via layout/formatter if needed.

What do you think?

gavin-norman-sociomantic:
(We could also think about whether we want to keep maintaining all of this time stuff that no one in sociomantic uses.)

mathias-lang-sociomantic:
> I don't like "fixing" cyclic imports by removing module constructors, it should be exactly other way around - cyclic import which is to be eliminated while module import being perfectly valid feature.

While I generally agree that circular imports should be avoided, they shouldn't be avoided at all cost. The ability to do circular imports is consistent with the ability to have co-dependent symbols in a module.

mathias-lang-sociomantic:
> I don't like "fixing" cyclic imports by removing module constructors, it should be exactly other way around - cyclic import which is to be eliminated while module import being perfectly valid feature.

While I generally agree that circular imports should be avoided, they shouldn't be avoided at all cost. The ability to do circular imports is consistent with the ability to have co-dependent symbols in a module.

E.g. the following code works and is conceptually the same as a circular import:
```D
mixin template Bar ()
{
    typeof(Foo.i)* next;
}

class Foo
{
    mixin Bar!();

    int i;
}
```

The problem we are facing is that semantic analysis in the compiler is clever enough to figure out the dependency can be resolved (but would error out if it wasn't possible, e.g. by using `__traits(classInstanceSize, Foo)`), but the module ctors aren't.

I'm sorry but the following situation is just ridiculous:

```D
module a;
import b;
private int i;
static this () { i = 42; }
void main () {}
```

```D
module b;
import a;
private int i;
static this () { i = 42; }
```

```
% rdmd a.d                                                                                           :(
object.Error@src/rt/minfo.d(378): Cyclic dependency between module b and a
b* ->
a* ->
b*
```

And in this case, the smart formatting of `Time` is both used and making sense. What's the point of having a time type if it just behaves like an integer ?
I'll look into refactoring this nonetheless, but to me module ctors are so broken there has to be a very good reason to use them.

mihails-strasuns-sociomantic:
@gavin-norman-sociomantic I see plenty of imports of various `ocean.time` modules in our code base, what do you mean by "unused"?

gavin-norman-sociomantic:
I'd be surprised if  `src/ocean/time/chrono/Gregorian.d` is used. (I am open to being surprised, though ;)

mihails-strasuns-sociomantic:
@mathias-lang-sociomantic 

> While I generally agree that circular imports should be avoided, they shouldn't be avoided at all cost. The ability to do circular imports is consistent with the ability to have co-dependent symbols in a module.

Probably we have a different perspective on the matter. I view allowing of cyclic imports as purely rapid prototyping tool which makes it possible to write a valid working module hierarchy first and clean up any cycles later.

Co-dependent symbols in same module are not as bad because module is a minimal D unit of encapsulation, its internal symbol dependencies don't affect overall app/lib architecture. But any module cycle has essentially to be treated as one big meta-module with individual elements that can't be trivially split or replaced. I view any such case as a technical debt in the long run.

mihails-strasuns-sociomantic:
@gavin-norman-sociomantic check the cycle - it is used by all code using `Time` because it is the default calendar implementation used for date/time formatting.

gavin-norman-sociomantic:
Ok, Gregorian was a bad example. https://github.com/sociomantic-tsunami/ocean/blob/v2.x.x/src/ocean/time/chrono/Hebrew.d is probably more relevant to my point.

mihails-strasuns-sociomantic:
Don't touch my hebrew! 😠

david-eckardt-sociomantic:
I just noticed the East Asian calendars `Japanese`, `Korean`, `Taiwan` and `ThaiBuddhist` aren't implemented yet. Will submit a PR.

src/ocean/time/WallClock.d
```diff
@@ -167,16 +167,3 @@ struct WallClock
...
-
-
-version (Posix)
-{
-    version (darwin) {}
-    else
-    {
-        static this()
-        {
-            tzset();
```
llucax, line 179: Mmm, this looks a little scary. Even when the man says that is called by other functions using timezone data, the user can potentially use `tzname` and assume it is properly initialized without calling this function. I know that chances of breakage are very slim (and I'd really like to see this static ctor gone!), but technically this is a breaking change so I would delay it for v4, so users get a big notice that from now on they need to call it themselves (if needed).

llucax (commented):
Besides the *breaking change* comment, and this one is just stylistic, the commit message `Make time.chrono.Gregorian : generic a singleton` was quite confusing for me (since `:` is usually used to separate a prefix for the commit title saying which area was affected, my brain completely skipped the `Make` at the beginning and just read `module: generic a singleton`...

I suggest to use `Make time.chrono.Gregorian.generic a singleton`.

mathias-lang-sociomantic:
Rebased on v4.x.x

Geod24:
So, are we fine with this going into v4.x.x ?

mihails-strasuns-sociomantic:
Well I am personally against the changes proposed here for the reasons I have already explained. Would like to see more opinions.

gavin-norman-sociomantic:
I think it'd be beneficial to unravel these dependencies, if possible. It'd at least be worth looking into how complicated a task it'd be.

gavin-norman-sociomantic:
Not urgent. Can be addressed in v2.9.0.

leandro-lucarella-sociomantic:
> Not urgent. Can be addressed in v2.9.0.

Because of the `tzset()` change this was rebased against v4.x.x, so I guess it should be assigned to v4.0.0 instead.

gavin-norman-sociomantic:
Ah right. That too!

gavin-norman-sociomantic:
Well, though that change was debated. Anyway, when this is finalised, we can assign it to the correct milestone.

mathias-lang-sociomantic:
> Well I am personally against the changes proposed here for the reasons I have already explained. Would like to see more opinions.

So the cyclic dependencies have been sorted out another, and are no longer an issue.
I'm still inclined to kill those static ctor. Would that work for you ?

mathias-lang-sociomantic:
https://youtu.be/YQHsXMglC9A

mihails-strasuns-sociomantic:
As I have mentioned before, I don't accept current solution as legitimate. My own preferred approach would be to deprecate `toString` method in `Time` and suggest using `format(time_instance)` instead. Considering our plans for https://github.com/sociomantic-tsunami/ocean/issues/272 , this module is too fundamental to have fat dependencies needed for internal formatting.

mihails-strasuns-sociomantic:
I am going to submit competing proposal/PR in 1-2 days and we can debate them both to death ^_^

mihails-strasuns-sociomantic:
Please rebase. Now that circular deps are resolves, I would suggest to remove 415b5ee1bc2318aaac73c16b389b5db48e7aa04b as static ctor is legitimate there and keep only the tzset one (which I will happily merge! :))

mathias-lang-sociomantic:
Rebased

mihails-strasuns-sociomantic (approved):

PR #94, opened 2017-05-01
Temporarily disable test that check const fields

mihails-strasuns-sociomantic:
Because of deficiency in DMD implementation, combination of `pragma(msg)` and
`is(typeof())` will result in message printed even if instantiation is
speculative. As it is not possible to turn them into asserts yet, all tests that
trigger warning are disabled for now.

### DIFF
--- a/src/ocean/util/serialize/contiguous/package_.d
+++ b/src/ocean/util/serialize/contiguous/package_.d
@@ -782,10 +782,14 @@ version (D_Version2) unittest
     //Deserializer.deserialize!(IS)(buffer1);
     //Deserializer.deserialize!(II)(buffer2);
 
-    static assert(!is(typeof({Deserializer.deserialize!(IS)(buffer1);})),
-                  "Serializer should reject a struct with 'istring'");
-    static assert(!is(typeof({Deserializer.deserialize!(II)(buffer2);})),
-                  "Deserializer should reject a struct with 'immutable' element");
+    version (none)
+    {
+        // TODO: re-enable in v4.x.x
+        static assert(!is(typeof({Deserializer.deserialize!(IS)(buffer1);})),
+                      "Serializer should reject a struct with 'istring'");
+        static assert(!is(typeof({Deserializer.deserialize!(II)(buffer2);})),
+                      "Deserializer should reject a struct with 'immutable' element");
+    }
 }
 
 /******************************************************************************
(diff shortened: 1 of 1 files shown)


mihails-strasuns-sociomantic:
(we should release 2.6.6 with this ASAP to remove irrelevant warning spam in 2.6.5)

src/ocean/util/serialize/contiguous/package_.d
```diff
@@ -734,10 +734,10 @@ unittest
 {
     static struct CS
     {
-        cstring s;
+        mstring s;
```
gavin-norman-sociomantic, line 737: Will this change need to be reverted in v4.x.x as well?

src/ocean/util/serialize/contiguous/package_.d
```diff
@@ -782,10 +782,14 @@ version (D_Version2) unittest
     //Deserializer.deserialize!(IS)(buffer1);
     //Deserializer.deserialize!(II)(buffer2);
 
-    static assert(!is(typeof({Deserializer.deserialize!(IS)(buffer1);})),
-                  "Serializer should reject a struct with 'istring'");
-    static assert(!is(typeof({Deserializer.deserialize!(II)(buffer2);})),
-                  "Deserializer should reject a struct with 'immutable' element");
+    version (none)
```
gavin-norman-sociomantic, line 785: We should create an issue in v4.0.0 to reenable these tests.

src/ocean/util/serialize/contiguous/package_.d
mihails-strasuns-sociomantic: No, this will be a breaking change in `v4.x.x` - such struct will no longer deserialize (it was never intended to). I wonder though how will it affect bidding apps as I presume this test was added by @mathias-lang-sociomantic originally.

mihails-strasuns-sociomantic: Ok, we talked with @mathias-lang-sociomantic, looks like it was needed for one of work-in-progress ports and doesn't affect any merged code. We'll see if that code can be adjusted in a fashion simlar to swarm, for now this adjustments should be OK to merge.

src/ocean/util/serialize/contiguous/package_.d
mihails-strasuns-sociomantic: https://github.com/sociomantic-tsunami/ocean/issues/95

src/ocean/util/serialize/contiguous/package_.d
gavin-norman-sociomantic: Is this change unrelated to "Temporarily disable test that check const fields" then?

mihails-strasuns-sociomantic: A bit indirectly. I will split it into separate commit and explain there.

mihails-strasuns-sociomantic: Done

gavin-norman-sociomantic:
LGTM.

PR #98, opened 2017-05-03
Formatter: Add support for comma separator

oraby:
There were several points I had in mind in this PR:
- Choose a reasonable format specifier
   - [printf uses `'`](http://stackoverflow.com/questions/10602729/digit-grouping-in-cs-printf) to specify a locale, I didn't use it because currently the separators characters are hard-coded to `,` and I wanted to leave support for locale  to be added later.
     - I tried to make the current design easy to adapt for specifying locale separators later.
   - D already uses `_` as a number separator in code, so I chose that as a specifier 
- Provide a simple way to use comma-separator
  - My expectation is that this format-specifier will be used a lot. It follows the notation for formatter specifiers using  `{:_}`, however I also added special support for `{_}` to make it easier to remember and to use (and to look better).
- The formatter code is used a lot and can be considered hot-code for many applications. The existing integer-formatter and float-formatter code try to minimize branching and just use plain loops. I tried my best to follow the same design.

### DIFF
--- a/relnotes/formatter_comma_support.feature.md
+++ b/relnotes/formatter_comma_support.feature.md
@@ -1,0 +1,6 @@
+* `ocean.text.convert.Formatter`
+
+  Use `{_}` format specifier to print a comma `,` after every 3rd digit when
+  printing a decimal integers or a floating point number.  If combined with a
+  hex-number specifier `{:x_}` or with a binary number specifier  `{:b_}`, then
+  the `_` character is used and will be placed after every 4th digit instead.
--- a/src/ocean/text/convert/Float.d
+++ b/src/ocean/text/convert/Float.d
@@ -94,6 +94,30 @@ private enum
 
 /******************************************************************************
 
+  Defines information needed to format Commas and decimals.
+
+ ******************************************************************************/
+
+public struct DecimalMark
+{
+    /// Character used for a comma
+    char  comma_char;
+    /// Character used for a decimal
+    char  decimal_char;
+    /// Number of digits between commas. len = 0 means commas are not used.
+    ubyte len;
+}
+
+/******************************************************************************
+
+  Default separtor used by the format() function.
+
+ ******************************************************************************/
+
+private const DEFAULT_DECIMAL_MARK = DecimalMark(',', '.', 0);
+
+/******************************************************************************
+
   Convert a formatted string of digits to a floating-point
   number. Throws an exception where the input text is not
   parsable in its entirety.
@@ -247,6 +271,7 @@ public Const!(T)[] format (T, V) (T[] output, V v, in T[] fmt)
     int dec = Dec;
     int exp = Exp;
     bool pad = true;
+    DecimalMark separator = DEFAULT_DECIMAL_MARK;
 
     for (auto p = fmt.ptr, e = p + fmt.length; p < e; ++p)
         switch (*p)
@@ -257,6 +282,9 @@ public Const!(T)[] format (T, V) (T[] output, V v, in T[] fmt)
         case 'e':
         case 'E':
             exp = 0;
+            break;
+        case '_':
+            separator = DecimalMark(',', '.', 3);
             break;
         default:
             Unqual!(T) c = *p;
@@ -269,7 +297,7 @@ public Const!(T)[] format (T, V) (T[] output, V v, in T[] fmt)
             break;
         }
 
-    return format!(T)(output, v, dec, exp, pad);
+    return format!(T)(output, v, dec, exp, pad, separator);
 }
 
 
@@ -286,6 +314,26 @@ unittest
     // Unlike Layout.floater, 'x' and 'X' aren't handled.
     //assert(format(buff, 8400.0, "X") == "0X40C0680000000000");
     assert(format(buff, 8400.0, "X") == "8400.00");
+
+    assert(format(buff, 1_234_567.89, "_") == "1,234,567.89");
+    assert(format(buff, -1_234_567.89, "_") == "-1,234,567.89");
+    assert(format(buff, 999.0, "_") == "999.00");
+    assert(format(buff, -999.0, "_") == "-999.00");
+    assert(format(buff, 1, "_") == "1.00");
+    assert(format(buff, -1, "_") == "-1.00");
+    assert(format(buff, 0f, "_") == "0.00");
+    assert(format(buff, 0.0001, "_.4") == "0.0001");
+    assert(format(buff, 0.0001, "_e4") == "1.0000e-04");
+    assert(format(buff, 1_234.567, "_.3") == "1,234.567");
+    assert(format(buff, 1_234.0, "_.") == "1,234");
+    assert(format(buff, float.nan, "_.") == "nan");
+    assert(format(buff, float.infinity, "_.") == "inf");
+
+    alias format!(char) formatT;
+    auto separator = DecimalMark('.', ',', 3); // German locale example
+    assert(formatT(buff, 1_234.56, Dec, Exp, Pad, separator) == "1.234,56");
+    separator = DecimalMark('_', '&', 2);
+    assert(formatT(buff, 1_23_45.67, Dec, Exp, Pad, separator) == "1_23_45&67");
 }
 
 
@@ -304,7 +352,8 @@ unittest
 
  ******************************************************************************/
 
-T[] format(T) (T[] dst, NumType x, int decimals=Dec, int e=Exp, bool pad=Pad)
+T[] format(T) (T[] dst, NumType x, int decimals=Dec, int e=Exp, bool pad=Pad,
+    DecimalMark separator = DEFAULT_DECIMAL_MARK)
 {
     Const!(char)*  end, str;
     int       exp,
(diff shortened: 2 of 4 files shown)


src/ocean/text/convert/Integer_tango.d
```diff
@@ -257,6 +257,14 @@ unittest
     assert (s == "43");
 }
 
+/*
```
gavin-norman-sociomantic: Why the non-standard comment format?

relnotes/formatter_comma_support.feature.md
```diff
@@ -0,0 +1,6 @@
+* `ocean.text.convert.Formatter`
+
+  Use `{_}` format specifier to print a comma `,` after every 3rd digit when
+  printing a decimal integers or a floating point number.  If combined with a
```
gavin-norman-sociomantic, line 4: Typo "integers" (should be singular).

src/ocean/text/convert/Float.d
```diff
@@ -94,6 +94,30 @@ private enum
...
 
+  Defines information needed to format Commas and decimals.
+
+ ******************************************************************************/
+
+public struct DecimalMark
+{
+    /// Character used for a comma
+    char  comma_char;
+    /// Character used for a decimal
```
gavin-norman-sociomantic, line 105: The word "decimal" here is a bit ambiguous. It could mean a decimal point or a number in decimal base.

I'd suggest calling it `decimal_mark_char` (https://en.wikipedia.org/wiki/Decimal_mark).

relnotes/formatter_comma_support.feature.md
```diff
@@ -0,0 +1,6 @@
...
+  hex-number specifier `{:x_}` or with a binary number specifier  `{:b_}`, then
```
gavin-norman-sociomantic, line 5: In hex, it could make sense to separate every 2 digits = 1 byte.

src/ocean/text/convert/Float.d
```diff
@@ -94,6 +94,30 @@ private enum
 
 /******************************************************************************
 
+  Defines information needed to format Commas and decimals.
```
gavin-norman-sociomantic: I think this is more informative: "Defines information needed to format numbers with digit grouping and/or decimal marks."

src/ocean/text/convert/Float.d
```diff
@@ -94,6 +94,30 @@ private enum
...
+
+ ******************************************************************************/
+
+public struct DecimalMark
+{
+    /// Character used for a comma
+    char  comma_char;
```
gavin-norman-sociomantic, line 104: Suggest calling this `digit_grouping_char`.

src/ocean/text/convert/Float.d
```diff
@@ -94,6 +94,30 @@ private enum
 
 /******************************************************************************
 
+  Defines information needed to format Commas and decimals.
+
+ ******************************************************************************/
+
+public struct DecimalMark
```
gavin-norman-sociomantic, line 101: `DecimalMarkConfig`?

src/ocean/text/convert/Float.d
```diff
@@ -94,6 +94,30 @@ private enum
...
+{
+    /// Character used for a comma
+    char  comma_char;
+    /// Character used for a decimal
+    char  decimal_char;
+    /// Number of digits between commas. len = 0 means commas are not used.
+    ubyte len;
```
gavin-norman-sociomantic, line 108: `digit_grouping_size`?

src/ocean/text/convert/Float.d
```diff
@@ -304,7 +352,8 @@ unittest
 
  ******************************************************************************/
 
-T[] format(T) (T[] dst, NumType x, int decimals=Dec, int e=Exp, bool pad=Pad)
+T[] format(T) (T[] dst, NumType x, int decimals=Dec, int e=Exp, bool pad=Pad,
+    DecimalMark separator = DEFAULT_DECIMAL_MARK)
```
gavin-norman-sociomantic, line 356: Params doc not updated.

src/ocean/text/convert/Integer_tango.d
```diff
@@ -265,20 +266,31 @@ unittest
         pre = set to string prefix
         width = set to the padding space that the number should occupy
 */
-private void decode(T) (T[] fmt, ref char type, out char pre, out int width)
+private void decode(T) (T[] fmt, out char type, out char pre, out int width,
+    out bool use_separators)
```
gavin-norman-sociomantic: Params doc not updated.

src/ocean/text/convert/Integer_tango.d
```diff
@@ -291,9 +303,18 @@ private struct _FormatterInfo(T)
     byte    radix;
     T[]     prefix;
     T[]     numbers;
+    DecimalMark decimal_mark;
 }
 
-Const!(T)[] formatter(T, N) (T[] dst, N i_, char type, char pre, int width)
+private struct DecimalMark
```
gavin-norman-sociomantic, line 309: Hm am I understanding correctly that you're adding this feature to the old tango formatter too? (That seems unnecessary, if that's the case.)

gavin-norman-sociomantic (commented):
Mostly comments on doc/naming.

relnotes/formatter_comma_support.feature.md
llucax: I think it might make sense to separate  every 2 digits for hex and 8 for binary (and 4 for octal), and maybe not using a comma for them, but a space or a dash maybe.

mathias-lang-sociomantic:
I'm not a big fan of adding features inside the `Formatter` itself.
The way it is designed, it should be fairly trivial to build something on top that send the `int` wrapped in a `struct` with a `toString` overload which does the right thing.

Pseudo code:
```D
istring localeFormatter (Args...) (cstring fmt, Args args)
{
   // Turns any `int` argument into `DecimalFormatter`
   auto args_ = args.mapAny!((int c) => DecimalFormatter(c, formattingOptions));
   format(fmt, args_);
}

struct DecimalFormatter
{
   this (int value, FormattingOptions opts); // Just store them

  void toString (scope Sink sink)
  {
      // Format `this.value` according to `this.options`.
  }
}
```

src/ocean/text/convert/Float.d
```diff
@@ -375,11 +424,26 @@ T[] format(T) (T[] dst, NumType x, int decimals=Dec, int e=Exp, bool pad=Pad)
             if (exp <= 0)
                 *p++ = '0';
             else
+            {
+                if (separator.len)
+                {
+                    // +1 to not print comma after the last 3 digits
+                    while (exp >= separator.len + 1)
```
david-eckardt-sociomantic, line 431:
```d
while (exp > separator.len)
```

gavin-norman-sociomantic:
>I'm not a big fan of adding features inside the Formatter itself.
The way it is designed, it should be fairly trivial to build something on top that send the int wrapped in a struct with a toString overload which does the right thing.

The existing formatting code doesn't work like this, so why should extensions?

mathias-lang-sociomantic:
> The existing formatting code doesn't work like this, so why should extensions?

I'm not sure I follow. Could you elaborate on what is "this" ?

gavin-norman-sociomantic:
You say that the proposed extension (comma separation via `{_}`) should be implemented via a `toString` of a wrapper struct. As far as I can see, none of the existing formatter features (e.g. `{:x}` for hex number formatting) are implemented in this way (via a `toString` of a wrapper struct), so my question is: why start implementing new features like that?

I'm not opposed to what you're suggesting, just interested why some features should be handled differently to others.

mathias-lang-sociomantic:
Ah, I see, very good point!

The reason is that this idea came after the `Formatter` was written. At the time I wrote it, I was not sure what people would want from the formatter, so it was intentionally restricted to what Tango supported (with a couple of minor additions which did not impact the design, like structs).

A year (?) after, I can see a couple of issues that needs addressing:
- `size_t` delegate should be `void`
- `struct` and other aggregates have no way of getting formatting informations: you either get the default formatting for this aggregate, or the one from `toString`. The user have no way to control it.
- Basic types suffer from the same issue. Ultimately it would be nice to be able to externally define a way to format any type.
- Some people even want a different default for some types.

So the idea to solve this was to add an optional `cstring fmt` parameter to the `toString` overload, so aggregate could choose how to format themselve according to their own format string. After discussing with Hathem, this approach seems generic enough to cover his cases, and all formatting needs.
The only two things I'm wary of is the amount of templates that'll be instantiated and the effect on performance, which I'll need to measure.

leandro-lucarella-sociomantic:
I have to confess this was making noise in my head too because of the lack of extensibility. If every time someone has a need to add a new way to format strings we need to keep adding complexity to the formatter, that doesn't seem scalable.

I also think having `toString()` methods that accept a `string fmt` string that the user can use to get custom formatting seems like the best solution. So the formatter basically delegates the details of the format specification to the object if it has a `toString(string fmt)`. Python does this (don't hate me!) and it works really well.

Example:
```py
class Country:
    def __init__(self, name, iso):
        self.name, self.iso = name, iso

    def __format__(self, spec):
        if spec == 'short':
            return self.iso
        return self.name

country = Country("Austria", "AUT")

print("{}".format(country))
print("{:short}".format(country))
```

gavin-norman-sociomantic:
Sounds good to me. It would be nice to adapt the currently built-in formatter rules to work in the same way, as far as possible (out of the scope of this PR, of course).

gavin-norman-sociomantic:
What's the status of this now?

hatem-oraby-sociomantic:
I haven't found time yet to get back to this yet, I'm hoping to get sometime this weekend.

mihails-strasuns-sociomantic:
Casual ping :)

mathias-lang-sociomantic:
It's more a ping to me, actually, since I have to implement the enhancement I mentioned in the Formatter :P

gavin-norman-sociomantic:
Do we have an issue for that?

gavin-norman-sociomantic:
I wonder if we should close this. The two main people involved don't work here any more :D

mihails-strasuns-sociomantic:
:okay:

Issue #84, opened 2017-04-21
Provide an ocean-d2 repo


hatem-oraby-sociomantic:
Currently if you have a D2 only application and you decided to use ocean then you have to do the following:
- Make sure that you have [d1to2fix](https://github.com/sociomantic-tsunami/d1to2fix/) installed on your system
- Add Ocean as a submodule
- Go to the ocean submodule directory and checkout the desired branch
- Add [Makd](https://github.com/sociomantic-tsunami/makd/) as a submodule to ocean
- Run `make d2conv`
- If your project is under version-control, then you need to commit the converted code so you can have none-dirty commits.
- And only at that point you are ready to use ocean.

The above procedure may deter a lot of D2 users who would like to test Ocean. It would be much more convenient to have a new ocean-d2 repo that contains converted snapshots of the released tags of this repo.
Syncing the d2 repo can be automatically maintained by Travis.
Also, I don't know if current ocean is in dub, but certainly the converted Ocean can live there.

fyi @leandro-lucarella-sociomantic 

---
**Disclaimer:** I'm a direct beneficiary of this issue. I will give a (2x-slot-)lightning talk about Ocean in DConf. I'd like the ease of using Ocean to be one of the selling points.

leandro-lucarella-sociomantic:
Yeah, I think simplifying access to D2 users is key for people to even dare to have an eye on Ocean, so I think a plan like this could be a very good start. We still have the major issue of not being able to use upstream compilers yet, but hopefully that will be sorted out not too far in the future if the DIP about `@future` gets accepted and implemented soon so we can solve the `Throwable.message()` issue.

mihails-strasuns-sociomantic:
Last time it was discussed, automatic generation of branches with `-d2` suffix in the same repo was the intended approach. I think it works better than dedicated repo because it prevents confusion where to create issues/PRs and makes it more obvious that content is automatically generated.

gavin-norman-sociomantic:
>I will give a (2x-slot-)lightning talk about Ocean in DConf. I'd like the ease of using Ocean to be one of the selling points.

Oh that's great!

hatem-oraby-sociomantic:
> Last time it was discussed, automatic generation of branches with -d2 suffix in the same repo was the intended approach. I think it works better than dedicated repo because it prevents confusion where to create issues/PRs and makes it more obvious that content is automatically generated.

Both would work fine, however there are few pros and cons to each side:
- Pros for a separate D2 repo:
[...]

PR #100, opened 2017-05-05
travis: Push converted tags to ocean-d2 repo

llucax:


oraby:
LGTM, I've tested it on a local repo and it works as expected.
Let's wait for a bit to see if we should include dub information as well or we'd leave it for a later time.

On unrelated note, I was reading about [SemVer](http://semver.org/) to see why you used `+d2.auto` and it turned out that you are adding metdata, so all is good. However, it mentions that  the hyphen character  should  be added for pre-releases. but we use it for `-preview` which is actually more of "end-of-life" support. Made we should use `+preview` in the future instead.

oraby:
> On unrelated note, I was reading about SemVer to see why you used +d2.auto and it turned out that you are adding metdata, so all is good. However, it mentions that the hyphen character should be added for pre-releases. but we use it for -preview which is actually more of "end-of-life" support. Made we should use +preview in the future instead.

Oh crap, I should read your message to the end in the future :p, you've just mentioned something similar in #84

oraby:
An extra thought: if we want to add dub and modified README then one way to do it is still to add a a sync branch in the d2 repo and we can change the `after-success.sh` to basically download the latest version of convert code as in [here](https://github.com/oraby/ocean/blob/727d6eea53ccb2fbb16dc8eea44f16a993ca528c/sync_d2_start.sh) for example.
If we'd do that then we'd add the sync branch commit hash in the commit message.

llucax:
> However, it mentions that the hyphen character should be added for pre-releases. but we use it for -preview which is actually more of "end-of-life" support. Made we should use +preview in the future instead.

`-preview` was only done when this wasn't the real upstream repo for the main development. The tag was fine because the idea was to mark this as a pre-release, since nobody was actually using these tags. Now that the development was moved here, we should completely forget about `-preview` tags.

llucax:
> An extra thought: if we want to add dub and modified README then one way to do it is still to add a a sync branch in the d2 repo and we can change the after-success.sh to basically download the latest version of convert code as in here for example.

For me this is too much complexity. If we want to add a dub file (I wouldn't add a readme file), we can just include it as part of the script.

llucax:
Updated to use a more complex but secure method for authentication (if the user/password are included in the URL, any errors or messages from Git will reveal the credentials).

llucax:
I did a quick try and it seems to work, we'll just need to test the credentials + travis integration in practice though.

oraby:
Here's my go at `dub.json`:
```json
{
    "name": "ocean-d2",
    "description": "Auto convert of ocean-d1 library. Ocean is general purpose library with a focus on supporting the development of high-performance, real-time applications.",
    "authors": ["Sociomantic Labs GmbH "],
    "homepage": "https://github.com/sociomantic-tsunami/ocean",
    "license": "BSD 3-clause or Check Ocean License",
    "copyright": "Copyright (c) 2004-2017 Sociomantic Labs GmbH and Tango contributors.\nAll rights reserved.",
    "systemDependencies": "Depending on the component used you might need one of the following libraries:\n-lglib-2.0 -lpcre -lxml2 -lxslt -lebtree -lreadline -lhistory -llzo2 -lbz2 -lz -ldl -lgcrypt -lgpg-error -lrt.\nNotice that the used -lebtree comes from https://github.com/sociomantic-tsunami/ebtree",
    "importPaths": ["src/"],
    "targetType": "sourceLibrary"
}

```

llucax:
> "name": "ocean-d2",

Any reason not to use just "ocean"?

> "homepage": "https://github.com/sociomantic-tsunami/ocean",

And here use ocean-d2, as it is from where the code really comes?

oraby:
> Any reason not to use just "ocean"?

Basically I want to reserve the `ocean` name for when we move this (ocean-d1) repo to be full d2 repo.

> "homepage": "https://github.com/sociomantic-tsunami/ocean",

As far as I understand, that needs to be the project homepage. That can be the read-the-docs or something. So I thought to add the mian-repo as the homepage as it's where all the development happens.

llucax:
> Basically I want to reserve the ocean name for when we move this (ocean-d1) repo to be full d2 repo.

👍 

> As far as I understand, that needs to be the project homepage. That can be the read-the-docs or something. So I thought to add the mian-repo as the homepage as it's where all the development happens.

I think this might be a bit confusing, it should be clearer to the user to send them to the page that explains they are using an automated conversion of a project, not really the project itself.

llucax:
I just pushed an update adding the dub file, I converted it to `.sdl` and changed the URL, but we can keep discussing the URL :)

ci/ocean-d2-dub.sdl
```diff
@@ -0,0 +1,9 @@
+name "ocean-d2"
+description "Auto convert of ocean-d1 library. Ocean is general purpose library with a focus on supporting the development of high-performance, real-time applications."
+authors "Sociomantic Labs GmbH "
```
oraby: extra space after GmbH

llucax:
Pushed new updated, I reverted the URL because DUB already shows the repo URL, so putting the original project seems reasonable.

llucax:
OK, we already sort of pushed this online for Hatem's lighting talk, so I will merge. Let's see how it works next time we tag something.

FYI @gavin-norman-sociomantic @mihails-strasuns-sociomantic

PR #101, opened 2017-05-05
Implement `ReduceType` and start new `ocean.meta`

mihails-strasuns-sociomantic:
I have finally managed to fix my `ReduceType` implementation to not crash the compiler but trying to integrate it into existing ocean code has ended up with me starting new top-level package :)

There are two reason for that:

1) Purist one - this is something I was tempted to do long time ago because existing `ocean.core.Traits` in an unstructured kitchen sink containing many utilities that are not even traits. I propose more abstract `ocean.meta` name ("metaprogramming") with 3 distinct packages-categories - traits, type operations and code generation utils. This PR defines few building blocks for the former two.

2) Backwards compatibility concern - I wanted to keep resulting code as clean as possible and switch existing utilities to use new `ReduceType` as a backend but it is quite scary considering how much different convoluted code uses those utilities. Starting with independent package hierarchy allows to switch to new stuff on demand.

Time to destroy!


mihails-strasuns-sociomantic:
(misses release notes for now as I foresee some discussion)

gavin-norman-sociomantic:
`ocean.meta` is a good idea.

src/ocean/meta/types/package_.d
```diff
@@ -0,0 +1,11 @@
+/*******************************************************************************
+
+    This package contains various template utilities that deduce types or values
+    by reflecting on supplied type arguments and doing some compile-time
+    manipulations on them.
+
```
gavin-norman-sociomantic, line 6: Every module should have the copyright / license, right?

src/ocean/meta/traits/Basic.d
```diff
@@ -0,0 +1,424 @@
...
+
+    Params:
+        T = type to check
+
+    Returns:
+        `true` if `T` is a primitive type
+
+*******************************************************************************/
+
+///
```
gavin-norman-sociomantic: Spurious?

src/ocean/meta/traits/Basic.d
```diff
@@ -0,0 +1,424 @@
...
+
+*******************************************************************************/
+
+module ocean.meta.traits.Basic;
+
+import ocean.meta.types.Qualifiers : Unqual, Const;
+
+/*******************************************************************************
+
+    Types are the types from which one or multiple other types cannot be
```
gavin-norman-sociomantic: Odd wording: "Types are the types"

src/ocean/meta/traits/Basic.d
```diff
@@ -0,0 +1,424 @@
...
+}
+
+/*******************************************************************************
+
+    Used by other traits to distinguish between dynamic and static arrays
+    instead of plain `bool` values.
+
+*******************************************************************************/
+
+public enum ArrayKind
```
gavin-norman-sociomantic: :+1: good idea.

src/ocean/meta/traits/Basic.d
```diff
@@ -0,0 +1,424 @@
...
+{
+    NotArray    = 0,
```
gavin-norman-sociomantic, line 286: Maybe note: `// defined as 0 so that the value of isArrayType!() can be used as a bool`.

src/ocean/meta/traits/Basic.d
```diff
@@ -0,0 +1,424 @@
...
+}
+
+/*******************************************************************************
+
+    Evaluates to true if T is a function, function pointer, delegate, or
+    callable object.
+
+*******************************************************************************/
+
+template isFunctionType ( T )
```
gavin-norman-sociomantic: `isCallableType`?

src/ocean/meta/traits/package_.d
```diff
@@ -0,0 +1,11 @@
+/*******************************************************************************
+
+    This package contains various template utilities that allow to reason about
+    types in some way. Most traits answer simple boolean question about
+    parameter type having certain qualities ("traits") thus the name.
+
```
gavin-norman-sociomantic, line 6: License / copyright.

src/ocean/meta/types/ReduceType.d
```diff
@@ -0,0 +1,199 @@
+/*******************************************************************************
+
+    Generic utility allowing to recursively visit an arbitrary type and reduce
+    its definition to some compile-time value. It is intended to be used as
+    implementation cornerstone for complex type traits to address recursive
```
gavin-norman-sociomantic: A bit clearer: "...implementation cornerstone for complex type traits to avoid having to rewrite the recursive type reflection boilerplate wherever it's needed."

src/ocean/meta/types/ReduceType.d
```diff
@@ -0,0 +1,199 @@
...
+        Reducer = struct type conforming `assertValidReducer` requirements used
+            to do actual calculations/checks on the type
+
+    Example:
+
+    Example reducer that counts amount of integer typed entities accessible
+    though the base types (i.e. fields and referenced types):
+
+    ---
+    static struct ExampleReducer
```
gavin-norman-sociomantic, line 47: Any reason not to do this as a documented unittest?

src/ocean/meta/types/ReduceType.d
```diff
@@ -0,0 +1,199 @@
...
+        is(Reducer.Result),
+        Reducer.stringof ~  " must define `Result` type alias"
+    );
+    static assert (
+        is(typeof(Reducer.seed) == Reducer.Result),
+        Reducer.stringof ~ " must define `seed` constant of type `Result`"
+    );
+    static assert (
+        is(typeof(Reducer.accumulate)),
+        Reducer.stringof ~ " must define `accumluate` CTFE function"
```
gavin-norman-sociomantic: "of type ..."

src/ocean/meta/types/ReduceType.d
```diff
@@ -0,0 +1,199 @@
...
+    );
+    static assert (
+        is(typeof(Reducer.visit!(int))),
+        Reducer.stringof ~ " must define `visit` templated function"
```
gavin-norman-sociomantic: "of type ..."

src/ocean/meta/traits/Indirections.d
```diff
@@ -0,0 +1,172 @@
...
+    {
+        return accum || next;
+    }
+
+    static Result visit ( T ) ( )
+    {
+        return (isArrayType!(T) == ArrayKind.Dynamic)
+            || (isArrayType!(T) == ArrayKind.Associative)
+            || isPointerType!(T)
+            || is(T == class)
```
gavin-norman-sociomantic: How about interfaces?

src/ocean/meta/traits/Indirections.d
```diff
@@ -0,0 +1,172 @@
...
+    Params:
+        T = type to check
+
+    Returns:
+        true if T or any of its subtypes is a multi-dimensional dynamic array or
+        false otherwise.
+
+*******************************************************************************/
+
+public template containsMultiDimensionalDynamicArrays ( T )
```
gavin-norman-sociomantic, line 125: `has` or `contains`?

src/ocean/meta/traits/Indirections.d
```diff
@@ -0,0 +1,172 @@
...
+            || is(T == class)
+            || isFunctionType!(T);
+    }
+}
+
+/*******************************************************************************
+
+    Checks if T or any of its subtypes is a multi-dimensional dynamic array.
+
+    T and all of its subtypes, if any, are expected to be
```
gavin-norman-sociomantic: What happens if these criteria are not met?

gavin-norman-sociomantic (commented):
This looks excellent! Probably @david-eckardt-sociomantic will come up with hundreds of corner cases and spoil your fun ;)

src/ocean/meta/types/package_.d
david-eckardt-sociomantic:
```d
import("licence.txt");
```

src/ocean/meta/types/Qualifiers.d
```diff
@@ -0,0 +1,199 @@
+/*******************************************************************************
+
+    Templates to define types with modified qualifiers based on some input
+    types.
+
```
david-eckardt-sociomantic, line 5: Perhaps mention that this is D2 transition related? (Well, mostly... `Unqual` is needed in D2 as well.)

src/ocean/meta/traits/Basic.d
```diff
@@ -0,0 +1,424 @@
...
+
+    Returns:
+        `ArrayKind` value indicating if `T` is an array and if it is, static or
+        dynamic
+
+*******************************************************************************/
+
+public template isArrayType ( T )
+{
+    static if (is(T U == U[]))
```
david-eckardt-sociomantic: `Unqual!(T)`!!! ****THIS IS THE CARDINAL ERROR I'VE BEEN HUNTING FOR WEEKS NOW!!!****

src/ocean/meta/traits/Basic.d
```diff
@@ -0,0 +1,424 @@
...
+        T = type to check
+
+    Returns:
+        `true` if `T` is a pointer type.
+
+*******************************************************************************/
+
+template isPointerType ( T )
+{
+    static if (is(T U == U*))
```
david-eckardt-sociomantic: `Unqual!(T)`

src/ocean/meta/traits/Basic.d
```diff
@@ -0,0 +1,424 @@
...
+        T = type to check
+
+    Returns:
+        `true` if `T` is a struct, class or union
+
+*******************************************************************************/
+
+template isAggregateType ( T )
+{
+    const isAggregateType = is(T == class) || is(T == struct) || is(T == union);
```
david-eckardt-sociomantic: `interface`?

src/ocean/meta/traits/Basic.d
mihails-strasuns-sociomantic: I don't see how it is applicable. Your case was about nested `static if` and re-assembling original type. Here I don't care at all if `T` is `const` or not - only that it matches `U[]` pattern (with `U` not being used).

src/ocean/meta/traits/Basic.d
mihails-strasuns-sociomantic: Same comment as for array.

src/ocean/meta/traits/Basic.d
david-eckardt-sociomantic: See https://issues.dlang.org/show_bug.cgi?id=17353.

mihails-strasuns-sociomantic: So, your point is that eventually this issue will be "fixed"? I rather doubt so, if anything is to happen, stripping of top level const will be simply documented as it is now.

In the meanwhile, with existing semantics the usage is correct.

david-eckardt-sociomantic: For the record:
http://dlang.org/spec/expression.html#IsExpression
> `is ( Type Identifier == TypeSpecialization )`
>  The condition is satisfied if `Type` is semantically correct and is the same as `TypeSpecialization`. The Identifier is declared to be either an alias of the `TypeSpecialization` or, if `TypeSpecialization` is dependent on `Identifier`, the deduced type. 

With `T == const(U[])` the specialisation is `const(U)[]`, which is _not the same_ as `T` so the `is` should
yield false. Still it yields true, which is a contradiction to the D2 specification, so it is a compiler bug hiding a missing `Unqual`.

david-eckardt-sociomantic: My point is that relying on a compiler bug may potentially be a suboptimal design decision. Do you prefer code that relies on compiler bugs for convenience?
Maybe this bug will never be fixed, maybe it will. Maybe it will declared to be a feature as you predict, but that would at least cause more `is` expression inconsistencies with type qualifiers. A perfectly consistent solution would be to just ignore all qualifiers in all `is(A == B)` and `is(A == B C)` expressions, which would make me very happy.

src/ocean/meta/types/ReduceType.d
```diff
@@ -0,0 +1,199 @@
...
+            // do nothing, already processed
+        }
+        else static if (isAggregateType!(T))
+        {
+            foreach (TElem; typeof(T.init.tupleof))
+                accumulate(result, reduce!(TElem)());
+        }
+        else static if (isArrayType!(T))
+        {
+            static if (is(T U : U[]))
```
david-eckardt-sociomantic: Suggest using the return value of the nice new `isArrayType` template which already tells you what kind of array it is.

src/ocean/meta/traits/Basic.d
```diff
@@ -0,0 +1,424 @@
...
+        `ArrayKind` value indicating if `T` is an array and if it is, static or
+        dynamic
+
+*******************************************************************************/
+
+public template isArrayType ( T )
+{
+    static if (is(T U == U[]))
+        const isArrayType = ArrayKind.Dynamic;
+    else static if (is(T U : U[]))
```
david-eckardt-sociomantic: What do you think about using `__traits(isStaticArray, T)` and `__traits(isAssociativeArray, T)` in D2 here?

src/ocean/meta/types/ReduceType.d
```diff
@@ -0,0 +1,199 @@
...
+            else
+            {
+                // associative
+                accumulate(result, reduce!(typeof(T.init.keys[0]))());
+                accumulate(result, reduce!(typeof(T.init.values[0]))());
+            }
+        }
+        else static if (isPointerType!(T))
+        {
+            accumulate(result, reduce!(typeof(*T.init))());
```
david-eckardt-sociomantic: - Suggest `is(Unqual!(T) U == U*)` since you need the pointer base type `U` for `reduce`.
- `isPointerType!(T)` (as well as `is(Unqual!(T) U == U*)`) yields `true` if `T` is a function pointer so the ` static if (isFunctionType!(T))` should be prior to this branch.

src/ocean/meta/traits/Indirections.d
david-eckardt-sociomantic: `isReferenceType`?

src/ocean/meta/traits/Indirections.d
```diff
@@ -0,0 +1,172 @@
...
+    const seed = false;
+
+    static Result accumulate ( Result accum, Result next )
+    {
+        return accum || next;
+    }
+
+    static Result visit ( T ) ( )
+    {
+        static if (is(T U == U[]))
```
david-eckardt-sociomantic: `Unqual!(T)`

src/ocean/meta/traits/Indirections.d
```diff
@@ -0,0 +1,172 @@
...
+
+    static assert(containsMultiDimensionalDynamicArrays!(A));
+
+    struct C
+    {
+        int x;
+        float[][3][] y;
+        char[] z;
+    }
+
```
david-eckardt-sociomantic: What about qualified array/element types? Should they be supported or not?

src/ocean/meta/traits/Basic.d
mihails-strasuns-sociomantic: Depends on kind of the bug. If bug is unlikely to be fixed and workaround complicates code in many places, I'd totally consider relying on it like I did here. Rationale is that if it is going to be fixed, such intrusive change would require deprecations anyway and we will be able to adjust. If it is not going to be fixed, we pollute the code for no reason.

src/ocean/meta/traits/Indirections.d
mihails-strasuns-sociomantic: Totally should but I haven't added any tests of my own so far, only copied old ones. Will try adding some new

david-eckardt-sociomantic:
Pretty cool! Apart from type qualifiers no unhandled corner cares spring to my mind right now (however, experience shows we may encounter them when trying to use it in swarm neo :trollface:).

src/ocean/meta/traits/Basic.d
mihails-strasuns-sociomantic: Nope, I have intentionally not included objects with `opCall` in this list. `isCallableType` would be a superset from `isFunctionType`.

src/ocean/meta/traits/Basic.d
david-eckardt-sociomantic: I disagree, but at least add a comment here. Some day someone with a `const` mindset will maintain, extend or refactor this code and have a hard time understanding how it works with a missing `Unqual` -- or add `static assert` consistency checks which mysteriously fail etc.

gavin-norman-sociomantic: I agree with David. At very least, the bug should be mentioned here with an explanation of why the code currently works.

src/ocean/meta/traits/Basic.d
```diff
@@ -0,0 +1,424 @@
...
+///
+unittest
+{
+    void foo1() { }
+    auto foo2 = () { };
+    static void foo3() { }
+
+    static assert (isFunctionType!(typeof(&foo1)));
+    static assert (isFunctionType!(typeof(foo2)));
+    static assert (isFunctionType!(typeof(&foo3)));
```
david-eckardt-sociomantic, line 490: Looks like `typeof(foo3)` should be tested, too.

src/ocean/meta/traits/Basic.d
```diff
@@ -0,0 +1,424 @@
...
+    static assert (!isCharType!(byte));
+}
+
+/*******************************************************************************
+
+    Params:
+        T = type to check
+
+    Returns:
+        `true` if `T` is a signed integer type
```
don-clugston-sociomantic: The docs for all of these traits should say something like `built-in signed integer type`, so that it's clear that they don't include synthetic types like `BigInt`, `CustomFloat`.

src/ocean/meta/types/ReduceType.d
mihails-strasuns-sociomantic: Yep, I can't have nested struct with templated method in D1.

mihails-strasuns-sociomantic: (and declaring it outside of test block kills the documentation purpose)

src/ocean/meta/types/ReduceType.d
mihails-strasuns-sociomantic: Any reasons to not use `typeof(*T.init)`? I find it both more reliable and readable than pattern matching.

src/ocean/meta/traits/Indirections.d
mihails-strasuns-sociomantic: Original name used `has` but I think `contains` is more straightforward.

src/ocean/meta/traits/Indirections.d
mihails-strasuns-sociomantic: Was copied from old doc. With new implementation there are no restrictions on T at all ;)

mihails-strasuns-sociomantic:
Major update:

- addressed review comments
- enhanced some tests
- added `ocean.meta.traits.Basic.isCallableType` and `ocean.meta.traits.Basic.isReferenceType`
- added `ocean.meta.types.Arrays.ElementTypeOf`
- tweaked `ReduceType` implementation to allow non-static `visit` and accumulate - makes stateful reducers possible if needed
- tweaked `ReduceType` so that if `accumulate`/`seed` are not present, it tries to default to logical OR based one
- release notes

mihails-strasuns-sociomantic:
Have stumbled upon one major limitation with this approach - it dies with any type with recursive definition, i.e. `struct S { S* ptr; }`. Any ideas?

mihails-strasuns-sociomantic:
Adding simple cache fixed infinite recursion but is inherently unreliable as we can't get qualified names of types in D1:

```diff
@@ -145,6 +145,9 @@ private struct ReduceTypeImpl ( Reducer )
     // internal state
     Reducer reducer;
 
+    // To avoid infinite CTFE recursion with recursive types
+    Reducer.Result[string] cache;
+
     // Helper to calculate new value and update accumulator in one step
     private void accumulate ( T ) ( ref T accum, T next )
     {
@@ -178,7 +181,13 @@ private struct ReduceTypeImpl ( Reducer )
             auto result = false;
         }
 
-        accumulate(result, reducer.visit!(T)());
+        auto cached = T.stringof in cache;
+        if (cached)
+            return *cached;
+
+        auto value = reducer.visit!(T)();
+        cache[T.stringof] = value;
+        accumulate(result, value);
 
         static if (isPrimitiveType!(T))
         {
```

src/ocean/meta/types/ReduceType.d
gavin-norman-sociomantic: Could be a D2 only unittest...?

gavin-norman-sociomantic:
As verbally discussed, the limitation relating to recursive definitions should be mentioned in the release notes and doc. We can enhance this in the future.

src/ocean/meta/traits/Basic.d
```diff
@@ -0,0 +1,508 @@
...
+        void* a;
+        uint  b;
+    }
+
+    static assert (!isPointerType!(Eggs));
+    static assert ( isPointerType!(Immut!(Eggs*)));
+
+    struct Bacon { }
+
+    static assert (!isPointerType!(Bacon));
```
david-eckardt-sociomantic, line 384: Python style?

src/ocean/meta/traits/Basic.d
```diff
@@ -0,0 +1,508 @@
...
+        `true` if T is a function, function pointer or delegate or callable
+        aggregate
+
+*******************************************************************************/
+
+template isCallableType ( T )
+{
+    const bool isCallableType =
+           isFunctionType!(T)
+        || is(typeof(T.opCall));
```
david-eckardt-sociomantic: Minor point/out of interest: Wouldn't it be better to check if `typeof(&T.init.opCall)` is a function or delegate? (I could imagine that could be a hassle, though.)

src/ocean/meta/types/Arrays.d
```diff
@@ -0,0 +1,70 @@
...
+
+public template ElementTypeOf ( T )
+{
+    static if (isArrayType!(T) == ArrayKind.Dynamic
+            || isArrayType!(T) == ArrayKind.Static)
+    {
+        static if (is(T U : U[]))
+        {
+            alias U ElementTypeOf;
+        }
```
david-eckardt-sociomantic, line 40: No need for
```d
isArrayType!(T) == ArrayKind.Dynamic || isArrayType!(T) == ArrayKind.Static
```
because
```d
is(T U : U[])
```
yields the same result.

src/ocean/meta/types/ReduceType.d
```diff
@@ -0,0 +1,208 @@
...
+        }
+        else static if (is(T U == enum))
+        {
+            accumulate(result, reduce!(U)());
+        }
+        else
+        {
+            static assert (false,
+                "Unexpected type kind during recursive iteration" ~ T.stringof);
```
david-eckardt-sociomantic: Append a colon and a space to the message string.

src/ocean/meta/traits/Basic.d
```diff
@@ -400,6 +400,43 @@ unittest
 
 *******************************************************************************/
 
+template isReferenceType ( T )
+{
+    const isReferenceType =
+           isPointerType!(T)
+        || isFunctionType!(T)
```
david-eckardt-sociomantic: Only function *pointers* are reference types, not functions, so this would yield a false positive if a function type is passed. (I can't think of a reasonable use case for this, but it can at least hide a bug in the code using this template.)

(Personal note: Function types are the most obscure and unnecessary D types I know -- neither reference nor value types -- and should IMHO be removed from the language because while being useless they need special handling like this.)

david-eckardt-sociomantic: Just thought about it: `isPointerType` will detect function pointers so `is(T == delegate)` should be sufficient here.

src/ocean/meta/traits/Indirections.d
david-eckardt-sociomantic: I like `contains`.

david-eckardt-sociomantic (changes requested):
Added a few more requests for changes; the following are older and still need to be addressed:

https://github.com/sociomantic-tsunami/ocean/pull/101/commits/b301965f87908a8ba537e3bc9f6a55fbd5ee84f7#r115250581

https://github.com/sociomantic-tsunami/ocean/pull/101/commits/b301965f87908a8ba537e3bc9f6a55fbd5ee84f7#r115250977

src/ocean/meta/traits/Basic.d
mihails-strasuns-sociomantic: What is a python style?

src/ocean/meta/types/Arrays.d
mihails-strasuns-sociomantic: You asked me to do exactly other way around for the sake of readability last time ;)

src/ocean/meta/traits/Basic.d
mihails-strasuns-sociomantic: You are 100% right.

mihails-strasuns-sociomantic:
>  the following are older and still need to be addressed

Could you please copy comments here? GitHub refuses to link them :(

src/ocean/meta/types/Qualifiers.d
mihails-strasuns-sociomantic: I added mention that some of templates are transition related but the module as a whole is not.

src/ocean/meta/types/ReduceType.d
mihails-strasuns-sociomantic: There is already unittest with same code, I copied it here for the purpose of doc rendering only.

mihails-strasuns-sociomantic:
Updated with documentation and release mention for recursive type problem. Also some fixes for @david-eckardt-sociomantic :)

src/ocean/meta/types/Arrays.d
david-eckardt-sociomantic: Oops, really?! Fair enough then...

src/ocean/meta/traits/Basic.d
david-eckardt-sociomantic: IIRC using `ham, eggs, bacon` as example symbols is Python style.

david-eckardt-sociomantic:
- `Unqual!(T)` request in `ocean.meta.traits.Indirections:116` https://github.com/sociomantic-tsunami/ocean/pull/101/commits/b9da44c8099422a24b4832f29133b0337c10e26f#r115250581
- `containsMultiDimensionalDynamicArrays`: Question whether to support and test qualified types in `ocean.meta.traits.Indirections:214` https://github.com/sociomantic-tsunami/ocean/pull/101/commits/b9da44c8099422a24b4832f29133b0337c10e26f#r115250977

src/ocean/meta/traits/Basic.d
```diff
@@ -0,0 +1,508 @@
...
+
+    Params:
+        T = type to check
+
+    Returns:
+        `true` if `T` is a built-in complex floating point type
+
+*******************************************************************************/
+
+public template isComplexType( T )
```
mathias-lang-sociomantic, line 206: Do we need this (and `isImaginaryType`) given they're deprecated in D2 and not used anywhere ?

src/ocean/meta/traits/Basic.d
```diff
@@ -0,0 +1,508 @@
...
+
+public template isArrayType ( T )
+{
+    static if (is(Unqual!(T) U == U[]))
+        const isArrayType = ArrayKind.Dynamic;
+    else static if (is(Unqual!(T) U : U[]))
+        const isArrayType = ArrayKind.Static;
+    else static if (is(typeof(T.init.values[0])[typeof(T.init.keys[0])] ==
+            Unqual!(T)))
+        const isArrayType = ArrayKind.Associative;
```
mathias-lang-sociomantic, line 313: This syntax is really ugly.
Can't we reuse the much nicer https://github.com/sociomantic-tsunami/ocean/blob/8f7d8925ae0d1890e404c27b077a6f9f356400eb/src/ocean/core/Traits.d#L1431-L1497 instead ?

src/ocean/meta/traits/Basic.d
```diff
@@ -0,0 +1,508 @@
...
+
+*******************************************************************************/
+
+template isFunctionType ( T )
+{
+    const bool isFunctionType =
+           is(T == function)
+        || is(typeof(*T.init) == function)
+        || is(T == delegate);
+}
```
mathias-lang-sociomantic, line 479: I guess we completely ignore `@property` for now ?

src/ocean/meta/traits/Basic.d
mihails-strasuns-sociomantic: This test block was copied from `ocean.core.Traits` blindly :)

mihails-strasuns-sociomantic:
Updated:
* Address b9da44c#r115250581

b9da44c#r115250977 should be already addressed - I added quite some qualifiers to tests of `containsMultiDimensionalDynamicArrays` last time.

david-eckardt-sociomantic (approved):

src/ocean/meta/types/Arrays.d
```diff
@@ -0,0 +1,70 @@
...
+    }
+    else static if (isArrayType!(T) == ArrayKind.Associative)
+    {
+        alias AAElementType!(typeof(T.init.keys[0]), typeof(T.init.values[0]))
+            ElementTypeOf;
```
mathias-lang-sociomantic, line 45: Ditto about the ugly syntax

src/ocean/meta/traits/Basic.d
mihails-strasuns-sociomantic: It is used via `isPrimitiveType` trait.

src/ocean/meta/traits/Basic.d
mihails-strasuns-sociomantic: The syntax you link can only be used as part of template declaration in D1 and not in `is` expression. I don't want to add define new templates just to make a single line that no one will ever look at prettier ;)

src/ocean/meta/traits/Basic.d
mihails-strasuns-sociomantic: Yep, and I don't see why we would ever want to pretend it exists.

PR #102, opened 2017-05-05
travis: Build matrix for F,DC,DIST combinations

llucax:
This commit configures Travis to use matrix builds for all the combinations of *F={production,devel}* (flavours), *DC={dmd1,dmd-transitional}* (compiler, in the future upstream DMD could be added as a "allow_failures" combination) and *DIST={trusty,xenial}* (Ubuntu version).

For now the Dockerfile is sort of duplicated, but since it's minimal, it shouldn't be a problem anyway.


llucax:
@mihails-strasuns-sociomantic I think you will appreciate this ❤️

llucax:
BTW, this is going to be moved to cachalot, but I wanted to have a working prototype before abstracting it away.

.travis.yml
```diff
@@ -7,6 +7,16 @@ services:
...
 
+env:
+    - DC=dmd1 DIST=trusty F=production
+    - DC=dmd1 DIST=trusty F=devel
+    - DC=dmd1 DIST=xenial F=production
+    - DC=dmd1 DIST=xenial F=devel
+    - DC=dmd-transitional DIST=trusty F=production
+    - DC=dmd-transitional DIST=trusty F=devel
+    - DC=dmd-transitional DIST=xenial F=production
+    - DC=dmd-transitional DIST=xenial F=devel
```
mihails-strasuns-sociomantic, line 17: One thing I wonder about is if would be better to use specific versions (`1.080.0`, `2.070.2.s10`) in matrix, so that it can be naturally extended with more.

llucax: This will be more complex to achieve, since we need to rethink the docker images for that. Right now we are using global system installations of `dmd`, so we can only install `dmd1`, `dmd-transitional` and `dmd` because they use different config files. If we want to use multiple versions for each, we should move to another scheme, probably providing the compiler in an isolated directory with it's config file, and add the directory to the `PATH` when needed.

mihails-strasuns-sociomantic: I was envisioning simply using docker tags for base images that depend on compiler version and pretty much keeping existing scripts as they are apart from that. it is more of a matter of getting more readable test report matrix in travis than actual infrastructure.

llucax: I'm not convinced that approach will be easier and it might be less efficient, and explode the image tags (I guess we'll have to use something like `v1-dmd-1.080-trusty` for example, or do you think we can completely remove the "image version" (v1) in this case, but then how this interact with the base image?). Changing the way we install the compiler seems trivial and adding more compilers once we do that it's just a matter of calling the dmd building script again with a different version.

Since this is not trivial, maybe we should move the discussion to a cachalot issue. Can you create an issue there with a detailed proposal?

mihails-strasuns-sociomantic:
You want to move ahead with this without waiting for cachalot?

leandro-lucarella-sociomantic:
Yes, I will move it afterwards. I prefer to have some use cases before generalizing.

PR #103, opened 2017-05-06
[WIP] travis: Cache docker images

llucax:


mihails-strasuns-sociomantic:
So this relies on the fact that docker will check if `Dockerfile` has changed itself and reuse intermediate images if available locally?

llucax:
> So this relies on the fact that docker will check if `Dockerfile` has changed itself and reuse intermediate images if available locally?

Yes, I had troubles before trying to use the docker hub to store the images, so this is just another wild try.

llucax:
```
+docker load
invalid argument
```

😒

ci/travis-install.sh
```diff
@@ -1,5 +1,12 @@
 #!/bin/sh
 set -xe
 
+mkdir -p /tmp/docker-cache
+if test -f /tmp/docker-cache/ocean.tar.gz; then
+	zcat /tmp/docker-cache/ocean.tar.gz | docker load
```
mihails-strasuns-sociomantic: TIL: `zcat` :)

mihails-strasuns-sociomantic (approved):
Looks good and green. Feel free to auto-merge.

llucax:
Still not caching anything unfortunately :(

I need to keep investigating why Travis hates me.

llucax:
OK, I finally got some time to get to the Travis debug console and even when I don't fully understand why this works as it does, it seems like I made it work by saving not only the finally produced image but all the images in docker. I will try if it's enough to save the `FROM` image or if the intermediate images are needed too to get the build cache to work, but with the latest update, the build cache is picking the pre-built image.

Another point to improve is to not re-save the image if it wasn't rebuilt, but this is at least moving somewhere now. 🎉

mihails-strasuns-sociomantic:
I was going to propose to implement simple caching model on script side, i.e. store `docker/` folder hash next to the saved image and if it matches current one during rebuild, skip `docker build` step completely.

leandro-lucarella-sociomantic:
Yeah, I think to do this efficiently we need to hash the docker contents ourselves, otherwise we'll always have to fetch and save the images even if they won't be built because of the cache.

leandro-lucarella-sociomantic:
The problem with this is we either have to rely on convention or parse the Dockerfile to see what's really used there.

leandro-lucarella-sociomantic:
OK, I implemented this and it seems to work (at least when no docker files were changed, I didn't test the changing part) but even then, loading the image from the cache takes the same amount of time as building the image, so I guess this doesn't make any sense unfortunately (because the cache is not really local to the VM where stuff runs, it has to be copied over the network before being used, and then there is the load time for the docker image itself).

I will probably still move the scripts to cachalot because I would assume that there is threshold for which this is useful (I guess if there is little download but a lot of processing to build the image, like compiling something).

leandro-lucarella-sociomantic:
Just for the records, some time information:
* avg time to build this without cache: 5.5m
* with cache after it's cached: 6.5m
* with cache the first time (build and save the cache): 9m

mihails-strasuns-sociomantic:
And how does that compare to using completely pre-built image from the registry?

leandro-lucarella-sociomantic:
That didn't work because it seems that pushing the image to the registry doesn't push all the intermediate layers like it happened when only saving the resulting image. I didn't try to push all the intermediate images to the registry because AFAIK to do that I will have to tag them and add a lot of noise, but in principle I don't think it will help. The same process would take place AFAIK (when pulling images they are first downloaded and then imported somehow) and I expect Travis cache downloading should be faster than downloading from DockerHub, but I'm not sure.

In any case, there is very little evidence that this could be faster, so I don't think spending more time on this issue is justified. I will close it when the code is added to Cachalot, because for some project with little download and a lot of building, this caching could still be beneficial, but it doesn't seems to be the case for this project.

mihails-strasuns-sociomantic:
Sorry for not being clear - I was referring to the fact that in my simple experiment using 100% cached `sociomantictsunami/ocean` image was 3x faster than full build script approach we had before and I was wondering how both of those points fit into the curve.

llucax:
> Sorry for not being clear - I was referring to the fact that in my simple experiment using 100% cached sociomantictsunami/ocean image was 3x faster than full build script approach we had before and I was wondering how both of those points fit into the curve.

Yeah, the thing is all the heavy processing (compilation) was moved to the base image, if we were to build the image from the scratch (starting at a base Ubuntu image) then for sure this caching would help a lot. Although because of how caching works in Travis I think the first run will always be slow, because the cache is per PR/branch/matrix cell. AFAIK there is no global cache for the project.

llucax:
In that case pushing to the registry (or some other cache that is more under our control) could mitigate that.

mihails-strasuns-sociomantic:
What is the plan here?

leandro-lucarella-sociomantic:
I will include this in the travis scripts (new repo is coming for this). I prefer to keep this open until the code is moved there, if you don't mind.

leandro-lucarella-sociomantic:
I labeled this as wontfix to make it clear that it will not be merged.

mihails-strasuns-sociomantic:
Fine by me, I will modify milestone too in that case.

mihails-strasuns-sociomantic:
@leandro-lucarella-sociomantic was it included in beaver already?

leandro-lucarella-sociomantic:
No, but I have it in a local branch, and I doubting more and more about its usefulness, so closing.

PR #105, opened 2017-05-07
Compile with dmd v2.070.2

daniel-zullo:
Defines the extenal functions and macros definitionto be able to compile with dmd v2.070.2

### DIFF
--- a/src/ocean/core/UnitTestRunner.d
+++ b/src/ocean/core/UnitTestRunner.d
@@ -58,8 +58,29 @@ import ocean.transition;
 
 import ocean.stdc.string: strdup, strlen, strncmp;
 import core.sys.posix.unistd: unlink;
-import core.sys.posix.libgen: basename;
-import core.sys.posix.sys.time: gettimeofday, timeval, timersub;
+import core.sys.posix.sys.time: gettimeofday, timeval;
+
+static if (__VERSION__ >= 2000 && __VERSION__ < 2073)
+{
+    extern(C) char* basename (char*);
+
+    void timersub (in timeval* a, in timeval* b, timeval *result)
+    {
+        result.tv_sec = a.tv_sec - b.tv_sec;
+        result.tv_usec = a.tv_usec - b.tv_usec;
+        if (result.tv_usec < 0)
+        {
+            --result.tv_sec;
+            result.tv_usec += 1_000_000;
+        }
+    }
+}
+else
+{
+    import core.sys.posix.libgen: basename;
+    import core.sys.posix.sys.time: timersub;
+}
+
 import ocean.core.Runtime: Runtime;
 import ocean.core.Exception_tango : AssertException;
 import ocean.io.Stdout_tango: Stdout, Stderr;
--- a/src/ocean/io/select/protocol/task/TaskSelectTransceiver.d
+++ b/src/ocean/io/select/protocol/task/TaskSelectTransceiver.d
@@ -35,7 +35,11 @@ class TaskSelectTransceiver
     import ocean.stdc.posix.sys.uio: iovec, readv;
     import ocean.stdc.posix.sys.socket: setsockopt;
     import ocean.stdc.posix.netinet.in_: IPPROTO_TCP;
-    import core.sys.linux.sys.netinet.tcp: TCP_CORK;
+
+    static if (__VERSION__ >= 2000 && __VERSION__ < 2073)
+        enum { TCP_CORK = 3 }
+    else
+        import core.sys.linux.sys.netinet.tcp: TCP_CORK;
 
     import ocean.sys.Epoll: epoll_event_t;
     alias epoll_event_t.Event Event;
--- a/src/ocean/sys/GetIfAddrs.d
+++ b/src/ocean/sys/GetIfAddrs.d
@@ -28,7 +28,44 @@ import core.stdc.string;
 import ocean.stdc.posix.arpa.inet;
 import ocean.stdc.posix.netinet.in_: sockaddr_in, sockaddr_in6;
 import ocean.stdc.posix.sys.socket;
-import core.sys.linux.ifaddrs;
+
+static if (__VERSION__ >= 2000 && __VERSION__ < 2073)
+{
+    extern (C)
+    {
+        struct ifaddrs
+        {
+            /// Next item in the list
+            ifaddrs*         ifa_next;
+            /// Name of the interface
+            char*            ifa_name;
+            /// Flags from SIOCGIFFLAGS
+            uint      ifa_flags;
+            /// Address of interface
+            sockaddr* ifa_addr;
+            /// Netmask of interface
+            sockaddr* ifa_netmask;
+
+            union
+            {
+                /// Broadcast address of the interface
+                sockaddr* ifu_broadaddr;
+                /// Point-to-point destination addresss
+                sockaddr* if_dstaddr;
+            }
+
+            /// Address specific data
+            void* ifa_data;
+        };
+
+        int getifaddrs (ifaddrs**);
+
+        void freeifaddrs (ifaddrs*);
+    }
+}
+else
+    import core.sys.linux.ifaddrs;
+
 import core.sys.posix.sys.socket: AF_INET, AF_INET6;
 
 import ocean.core.Test;
--- a/src/ocean/text/util/StringEncode.d
+++ b/src/ocean/text/util/StringEncode.d
@@ -50,7 +50,22 @@ module ocean.text.util.StringEncode;
 
 import ocean.transition;
 
-import core.sys.posix.iconv;
+static if (__VERSION__ >= 2000 && __VERSION__ < 2073)
+{
+    extern (C)
+    {
+        alias void* iconv_t;
+
+        iconv_t iconv_open (in char* tocode, in char* fromcode);
+
+        size_t iconv (iconv_t cd, char** inbuf, size_t* inbytesleft,
+            char** outbuf, size_t* outbytesleft);
+
+        int iconv_close (iconv_t cd);
+    }
+}
+else
+    import core.sys.posix.iconv;
 
 import core.stdc.errno;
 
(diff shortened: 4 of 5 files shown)


daniel-zullo:
FYI @mathias-lang-sociomantic @leandro-lucarella-sociomantic

llucax:
This is, with the upstream DMD2.070.2.

BTW, this should be ideally based on v2.x.x.

src/ocean/core/UnitTestRunner.d
```diff
@@ -58,8 +58,30 @@ import ocean.transition;
...
-import core.sys.posix.libgen: basename;
-import core.sys.posix.sys.time: gettimeofday, timeval, timersub;
+
+import core.sys.posix.sys.time: gettimeofday, timeval;
+
+static if (__VERSION__ >= 2000 && __VERSION__ < 2073)
+{
+    extern(C) char* basename(char*);
+
+    extern (D): pure: @safe: @nogc: nothrow:
```
llucax: Are you sure this only applies to the *body* of the `static if` and not all the rest of the module? Also, you can put only one `:` at the end, you don't have to put one after each attribute.

Also, `extern (D)` is redundant here, the previous `extern (C)` only applies to the symbol you are defining.

src/ocean/sys/GetIfAddrs.d
```diff
@@ -28,7 +28,43 @@ import core.stdc.string;
 import ocean.stdc.posix.arpa.inet;
 import ocean.stdc.posix.netinet.in_: sockaddr_in, sockaddr_in6;
 import ocean.stdc.posix.sys.socket;
-import core.sys.linux.ifaddrs;
+
+
+static if (__VERSION__ >= 2000 && __VERSION__ < 2073)
+{
+    extern (C): nothrow: @nogc:
```
llucax: Same here.

src/ocean/text/util/StringEncode.d
```diff
@@ -50,7 +50,21 @@ module ocean.text.util.StringEncode;
 
 import ocean.transition;
 
-import core.sys.posix.iconv;
+static if (__VERSION__ >= 2000 && __VERSION__ < 2073)
+{
+    extern (C): nothrow: @nogc:
```
llucax: And here.

llucax (commented):
Other than that, LGTM.

BTW, this still doesn't make Ocean compile with upstream because of the `Throwable.message` issue, but is a good step forward.

daniel-zullo:
> this should be ideally based on v2.x.x.

Now it is based on v2.x.x

src/ocean/core/UnitTestRunner.d
daniel-zullo: You are right, it'll be applied to the whole module in the current state. I'll write it within a scope. And we actually won't need to define the D2 attributes here.

indeed `extern(D)` can be removed.

daniel-zullo:
Updated based on suggestions from Luca and Mathias L.

src/ocean/sys/GetIfAddrs.d
```diff
@@ -28,7 +28,44 @@ import core.stdc.string;
 import ocean.stdc.posix.arpa.inet;
 import ocean.stdc.posix.netinet.in_: sockaddr_in, sockaddr_in6;
 import ocean.stdc.posix.sys.socket;
-import core.sys.linux.ifaddrs;
+
+static if (__VERSION__ >= 2000 && __VERSION__ < 2073)
```
Burgos, line 32: Ha, yes, shame we don't have a way to officially backport at least the bindings in the previous version - they are too tightly coupled with compiler releases IMHO. Perhaps TODO here? "TODO: remove this when drop support for 2073" or similar.

mihails-strasuns-sociomantic:
@leandro-lucarella-sociomantic 

> BTW, this still doesn't make Ocean compile with upstream because of the Throwable.message issue, but is a good step forward.

It should compile as we use `getMsg` wrapper to hide it away, but the tests checking error message content will fail.

mihails-strasuns-sociomantic (changes requested):
I would prefer to have single `compat` module/package defining such stuff as it is easier to maintain those during ocmpiler upgrades this way.

I think it is also time to introduce new standard `version` active for `dmd-transitional`, i.e. `version(SociomanticDmdTransitional)`, so that backported bits don't get ignored because of version checks added here.

gavin-norman-sociomantic:
>I think it is also time to introduce new standard version active for dmd-transitional, i.e. version(SociomanticDmdTransitional), so that backported bits don't get ignored because of version checks added here.

:+1:

llucax:
I think it is implied that these checks should be removed when we drop support for a version. All should fit better together if we start testing multiple DMD versions in a matrix, then it is obvious which versions are supported in tests too.

leandro-lucarella-sociomantic:
> I would prefer to have single compat module/package defining such stuff as it is easier to maintain those during ocmpiler upgrades this way.

I think is a very rare and thin layer to justify a compat module, or maybe I'm not seeing what you suggest. How will your compat module look like?

leandro-lucarella-sociomantic:
> I think it is also time to introduce new standard version active for dmd-transitional, i.e. version(SociomanticDmdTransitional), so that backported bits don't get ignored because of version checks added here.

For the changes introduced here there is no loss at all to use the backported bit. Even more, I would prefer not to backport stuff that we can workaround, otherwise we keep departing more and more from upstream, making it harder to catch up when we can. I would completely stop backporting C lib additions to druntime to dmd-transitional.

mihails-strasuns-sociomantic:
> I think is a very rare and thin layer to justify a compat module, or maybe I'm not seeing what you suggest. How will your compat module look like?

Single module containing all bindings currently spread over the code base in this PR. In my opinion it is easier to maintain it this way and remember to review all compatibility bits each time compiler is upgraded.

> For the changes introduced here there is no loss at all to use the backported bit. 

I am not very excited about it - that would mean that patches bundled with dmd-transitional may get out of sync. And putting own copy of such bindings in each repo which uses them is rather bad hack for impractical upstream release model.

daniel-zullo:
> Single module containing all bindings currently spread over the code base in this PR. In my opinion it is easier to maintain it this way and remember to review all compatibility bits each time compiler is upgraded.

If we agree on this approach then I guess we will need to use selective imports to avoid importing the whole module.

Anyway I'll wait for @leandro-lucarella-sociomantic and @mihails-strasuns-sociomantic to agree on the changes to do and then I'll address them

src/ocean/sys/GetIfAddrs.d
daniel-zullo: If we keep the current workaround then I'll add the TODO

leandro-lucarella-sociomantic:
> Single module containing all bindings currently spread over the code base in this PR. In my opinion it is easier to maintain it this way and remember to review all compatibility bits each time compiler is upgraded.

I don't like this approach, it makes very local hacks global. I don't see any advantages to it really. To remove the compatibility hacks is enough to grep for `2073` when dropping for a compiler (maybe `__VERSION__` if you want to be more conservative).

> I am not very excited about it - that would mean that patches bundled with dmd-transitional may get out of sync. And putting own copy of such bindings in each repo which uses them is rather bad hack for impractical upstream release model.

As I said, I wouldn't keep backporting C binding additions to druntime in dmd-transitional. I want for our custom compiler to differ as little as possible from upstream, adding missing bindings to dmd-transitional is just making us depart more from upstream, I want to depart as little as possible, so catching up will be easier in the future.

If we are having too many of these missing things, I'm fine to put it in a module (or just add temporarily a `src/core/whatever` that is just a copy of the missing file) until we drop support for version missing these bindings.

leandro-lucarella-sociomantic:
And this should probably be rebased to v2.6.x so it's included in all supported branches.

mihails-strasuns-sociomantic:
I disagree but not strongly eough to argue :) Let's move forward with it.

@daniel-zullo-sociomantic could you please rebase?

daniel-zullo:
Rebased to v2.6.x

daniel-zullo:
Also I added a temporal fixup commit to add the `TODO` message suggested. It has to be squashed if you are happy with it, otherwise I'll have to remove it.

mihails-strasuns-sociomantic (approved):

src/ocean/io/select/protocol/task/TaskSelectTransceiver.d
```diff
@@ -35,7 +35,11 @@ class TaskSelectTransceiver
     import ocean.stdc.posix.sys.uio: iovec, readv;
     import ocean.stdc.posix.sys.socket: setsockopt;
     import ocean.stdc.posix.netinet.in_: IPPROTO_TCP;
-    import core.sys.linux.sys.netinet.tcp: TCP_CORK;
+
+    static if (__VERSION__ >= 2000 && __VERSION__ < 2073)
```
leandro-lucarella-sociomantic, line 39: You missed this one, and if you are going to amend this, I think it will make more sense to use the "normal" compiler version in the comment (2.073), otherwise it makes no difference if you have to grep for `2073`.

daniel-zullo: Right, the I think it should be `DMD 2.070.2`
I'll fix it right now

daniel-zullo:
Updated based on a suggestion from Luca

src/ocean/io/select/protocol/task/TaskSelectTransceiver.d
leandro-lucarella-sociomantic: I think 2.070 is better because there could be a new .3 release for example (not under the current release model of DMD but still). In any case 2.072.2 contains 2.072 so for grepping is fine.

The problem is, this should be when we drop support for versions older than 2.073 actually, as the code says. If we drop support for 2.070.2 but we still suppot 2.071 and 2.072, this has to stay.

leandro-lucarella-sociomantic: This is why the comment for me it's really redundant, we should go to the source and just grep for `__VERSION__` when we drop support for a compiler and see them case by case, comments are not really reliable...

daniel-zullo: I see your point. Then I'll remove the TODO comments

daniel-zullo:
TODO comments have been removed.

leandro-lucarella-sociomantic:
Thanks!

PR #42, opened 2017-03-15
Make StringC.toDString more D2 friendly

mathias-lang-sociomantic:
CC @jens-mueller-sociomantic

PR #109, opened 2017-05-10
Deprecate ocean.util.stringz : fromStringz

mathias-lang-sociomantic:
While porting one of our app, I came accross this code and had a [deja vu](https://github.com/sociomantic-tsunami/ocean/pull/42).

### DIFF
--- a/relnotes/stringz.deprecation.md
+++ b/relnotes/stringz.deprecation.md
@@ -1,0 +1,17 @@
+* `ocean.stdc.stringz`
+
+  This module and all functions inside of it have been deprecated,
+  as they duplicates already existing functionalities in ocean.text.util.StringC,
+  which is more idiomatic and actively maintained.
+  In place of using static buffer, it is recommended to use a dynamic buffer,
+  `copy` and `StringC.toCString` instead. Example:
+  ```D
+  char[64] buffer;
+  open(toStringz("/etc/limits.conf", buffer));
+  ```
+  turns into:
+  ```D
+  static mstring buffer;
+  buffer.copy("/etc/limits.conf");
+  open(StringC.toCString(buffer));
+  ```
--- a/src/ocean/net/device/Berkeley.d
+++ b/src/ocean/net/device/Berkeley.d
@@ -16,14 +16,15 @@ import ocean.transition;
 
 import ocean.sys.Common;
 
+import ocean.core.array.Mutation;
 import ocean.core.Exception_tango;
 
 import ocean.stdc.string : strlen;
-import ocean.stdc.stringz;
 
 import ocean.stdc.posix.netdb;
 import ocean.stdc.posix.sys.socket;
 import ocean.stdc.posix.netinet.in_; // : sockaddr_in, sockaddr_in6
+import ocean.text.util.StringC;
 
 version (UnitTest)
 {
@@ -1033,7 +1034,9 @@ deprecated public abstract class Address
                 hints.ai_flags = flags;
                 hints.ai_family = (flags & AIFlags.PASSIVE && af == AddressFamily.UNSPEC) ? AddressFamily.INET6 : af;
                 hints.ai_socktype = SocketType.STREAM;
-                int error = getaddrinfo(toStringz(host), service.length == 0 ? null : toStringz(service), &hints, &info);
+                int error = getaddrinfo((host ~ "\0").ptr,
+                                        service.length == 0 ? null : (service ~ "\0").ptr,
+                                        &hints, &info);
                 if (error != 0)
                     throw new AddressException("couldn't resolve " ~ host);
 
@@ -1092,7 +1095,7 @@ deprecated public abstract class Address
                 // Getting name info. Don't look up hostname, returns
                 // numeric name. (NIFlags.NUMERICHOST)
                 getnameinfo (name, nameLen, host.ptr, host.length, null, 0, NIFlags.NUMERICHOST);
-                return idup(fromStringz (host.ptr));
+                return idup(StringC.toDString(host.ptr));
         }
 
         /***********************************************************************
@@ -1320,7 +1323,7 @@ deprecated public class IPv4Address : Address
         override istring toAddrString()
         {
                 char[16] buff = 0;
-                return idup(fromStringz(inet_ntop(AddressFamily.INET,
+                return idup(StringC.toDString(inet_ntop(AddressFamily.INET,
                     &sin.sin_addr.s_addr, buff.ptr, 16)));
         }
 
@@ -1352,8 +1355,9 @@ deprecated public class IPv4Address : Address
 
         static uint parse(istring addr)
         {
-                char[64] tmp;
-                return ntohl(inet_addr(toStringz(addr, tmp)));
+            static mstring tmp;
+            tmp.copy(addr);
+            return ntohl(inet_addr(StringC.toCString(tmp)));
         }
 }
 
@@ -1602,7 +1606,7 @@ protected:
         {
 
                 char[100] buff = 0;
-                return idup(fromStringz(inet_ntop(AddressFamily.INET6, &sin.sin6_addr, buff.ptr, 100)));
+                return idup(StringC.toDString(inet_ntop(AddressFamily.INET6, &sin.sin6_addr, buff.ptr, 100)));
         }
 
         /***********************************************************************
@@ -1685,7 +1689,7 @@ deprecated public class NetHost
                 int i;
                 char* p;
 
-                auto mkey = fromStringz(he.h_name);
+                auto mkey = StringC.toDString(he.h_name);
                 name = assumeUnique(mkey);
 
                 for (i = 0;; i++)
@@ -1700,7 +1704,7 @@ deprecated public class NetHost
                    aliases = new istring[i];
                    for (i = 0; i != aliases.length; i++)
                    {
-                        auto malias = fromStringz(he.h_aliases[i]);
+                        auto malias = StringC.toDString(he.h_aliases[i]);
                         aliases[i] = assumeUnique(malias);
                    }
                    }
@@ -1730,9 +1734,10 @@ deprecated public class NetHost
 
         bool getHostByName(istring name)
         {
-                char[1024] tmp;
+                static mstring tmp;
 
-                auto he = gethostbyname(toStringz(name, tmp));
+                tmp.copy(name);
+                auto he = gethostbyname(StringC.toCString(tmp));
                 if(!he)
                    return false;
                 validHostent(he);
@@ -1762,9 +1767,10 @@ deprecated public class NetHost
         //shortcut
         bool getHostByAddr(istring addr)
         {
-                char[64] tmp;
+                static mstring tmp;
 
-                uint x = inet_addr(toStringz(addr, tmp));
+                tmp.copy(addr);
+                uint x = inet_addr(StringC.toCString(tmp));
                 auto he = gethostbyaddr(&x, 4, cast(int)AddressFamily.INET);
                 if(!he)
                     return false;
--- a/src/ocean/net/device/Berkeley.d
+++ b/src/ocean/net/device/Berkeley.d
@@ -2076,5 +2076,3 @@ deprecated public class SocketSet
                 return select (checkRead, checkWrite, checkError, null);
         }
 }
-
-
--- a/src/ocean/sys/ErrnoException.d
+++ b/src/ocean/sys/ErrnoException.d
@@ -37,12 +37,11 @@ version (UnitTest)
 
 public class ErrnoException : Exception
 {
+    import ocean.core.Exception : ReusableExceptionImplementation;
+    import ocean.core.Traits : ReturnTypeOf, identifier;
     import ocean.stdc.string;
     import ocean.stdc.stringz;
-    import ocean.core.Traits : ReturnTypeOf;
 
-    import ocean.core.Exception : ReusableExceptionImplementation;
-    import ocean.core.Traits : identifier;
 
     /**************************************************************************
 
--- a/src/ocean/sys/ErrnoException.d
+++ b/src/ocean/sys/ErrnoException.d
@@ -40,7 +40,7 @@ public class ErrnoException : Exception
     import ocean.core.Exception : ReusableExceptionImplementation;
     import ocean.core.Traits : ReturnTypeOf, identifier;
     import ocean.stdc.string;
-    import ocean.stdc.stringz;
+    import ocean.text.util.StringC;
 
 
     /**************************************************************************
@@ -334,7 +334,7 @@ public class ErrnoException : Exception
             return this.append("Expected non-zero errno after failure");
 
         char[256] buf;
-        auto errmsg = fromStringz(strerror_r(err_num, buf.ptr, buf.length));
+        auto errmsg = StringC.toDString(strerror_r(err_num, buf.ptr, buf.length));
         return this.ReusableImpl.append(errmsg);
     }
 
--- a/Build.mak
+++ b/Build.mak
@@ -92,6 +92,7 @@ TEST_FILTER_OUT += \
 	$C/src/ocean/stdc/stdint.d \
 	$C/src/ocean/stdc/stdio.d \
 	$C/src/ocean/stdc/stdlib.d \
+	$C/src/ocean/stdc/stringz.d \
 	$C/src/ocean/stdc/tgmath.d \
 	$C/src/ocean/stdc/time.d \
 	$C/src/ocean/stdc/wctype.d \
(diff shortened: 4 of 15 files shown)


mihails-strasuns-sociomantic:
AFAIR for `toStringX` functions suggested deprecation is not truly suitable as `ocean.text.util` version does not support memory friendly buffer overload.

mathias-lang-sociomantic:
Hence why I didn't deprecate the UTF-8 overload ;)

mihails-strasuns-sociomantic:
How about unifying it all in one place and leaving deprecated aliases? Not very good to have single functionality spread over.

Geod24:
> How about unifying it all in one place and leaving deprecated aliases? Not very good to have single functionality spread over.

I knew you'd say that.

> AFAIR for toStringX functions suggested deprecation is not truly suitable as ocean.text.util version does not support memory friendly buffer overload.

Actually it's the other way around. `StringC` uses `ref`, `stringz` allocates. So what we need in `StringC` is the easy to use, non overload version. Any preference on the name ? `newCString` maybe ?

mihails-strasuns-sociomantic:
> I knew you'd say that

Well, if you want to touch it, better to do it right ;)

> Any preference on the name ?

Does the name have to be different? AFAIR they have different argument count, haven't they?

Geod24:
Not quite: [stringz.fromStringz](https://github.com/sociomantic-tsunami/ocean/blob/cbb07a095d41e6971c6b007a284dcc6fa48cb98a/src/ocean/stdc/stringz.d#L28-L53) vs [StringC.toCString](https://github.com/sociomantic-tsunami/ocean/blob/cbb07a095d41e6971c6b007a284dcc6fa48cb98a/src/ocean/text/util/StringC.d#L88-L96).

IIRC having the ability to quickly get a C string, even with allocation, has been asked internally.

mihails-strasuns-sociomantic:
I see. How about `addZeroTerminator` vs `toCString`?

mihails-strasuns-sociomantic:
CC @sociomantic-tsunami/core-team

gavin-norman-sociomantic:
Another way would be just to deprecate [stringz.fromStringz](https://github.com/sociomantic-tsunami/ocean/blob/cbb07a095d41e6971c6b007a284dcc6fa48cb98a/src/ocean/stdc/stringz.d#L28-L53). There's no real advantage to using over the version that reuses the same buffer.

Geod24:
It's a matter or (minor) convenience for applications with less performance
requirements or exploration of C libraries. I think
@jens-mueller-sociomantic uses it.

gavin-norman-sociomantic:
It doesn't seem worth introducing it as a (slight) variant in `StringC`, if we want to combine everything in one place. (Which seems like a good aim, to me.)

jens-mueller-sociomantic:
Yes. I'll use `fromStringz` and `toStringz`. I used to use `StringC`. But when converting to a C string you have to pass a mutable string, which forced me to add `dup` on a new line because it is passed by `ref`. In summary I prefer the interface of `ocean.stdc.stringz`. If I ever run into memory issues I can move forward by providing a buffer.

jens-mueller-sociomantic:
That said I move my code as needed. Just want to point out that I'd like to see an interface that is easy to use and especially does not force me to `dup`. It can do that internally, when I don't provide a buffer.

gavin-norman-sociomantic:
My perspective is that functions which deliberately allocate a new array are highly non-idiomatic in ocean, so should be avoided/removed, where possible. I feel like using `.dup` _is_ appropriate, in your use case -- it's a built-in language feature which explicitly declares what you're doing.

mathias-lang-sociomantic:
Note that in your case, instead of doing:
```
void foo (cstring whatever)
{
   mstring x = whatever.dup;
   StringC.toCString(whatever, x);
   myCFunction(x);
}
```

You can just do:
```
void foo (cstring whatever)
{
   myCFunction(whatever ~ "\0");
}
```

As, IIUC, you are not reusing the buffer.

jens-mueller-sociomantic:
I'll guess that might work (including D2). But then I wonder what's the use of the library function. To me it should cover all use cases.

jens-mueller-sociomantic:
> I feel like using `.dup` is appropriate, in your use case -- it's a built-in language feature which explicitly declares what you're doing.

I find nice about `dup` that it explicitly points out allocations, like `new`. But then I think ocean should not restrict itself, if it can cover more use cases easily. If my usage is really special, then I'm okay with using `dup` explicitly but it makes the code unnecessary complex.

jens-mueller-sociomantic:
```
myCFunction(whatever ~ "\0");
```

should be

```
myCFunction((whatever ~ "\0").ptr);
```

But I see the point.

gavin-norman-sociomantic:
>I find nice about dup that it explicitly points out allocations, like new.

Exactly.

>But then I think ocean should not restrict itself, if it can cover more use cases easily. If my usage is really special, then I'm okay with using dup explicitly but it makes the code unnecessary complex.

Honestly, I feel this is such a minor and obscure usage that it's not worth adding a new function for it (or avoiding deprecating a function that otherwise would be). (Especially when it's absolutely trivial to do what you want either using `StringC` + `.dup` or just appending `\0`.)

david-eckardt-sociomantic:
I never understood the point of `toStringz` -- why not simply append `\0`? (Likewise `fromStringz` which is just `strlen`).

gavin-norman-sociomantic:
Ditto.

jens-mueller-sociomantic:
I wonder whether append `\0` is a good idea. With a function call you can get additional checking. For example when converting a D string to a C string I would expect it to check that it does not contain `\0`.

gavin-norman-sociomantic:
You're right.

jens-mueller-sociomantic:
So the recommendation is to use `StringC` and `dup`?

gavin-norman-sociomantic:
That would be my preference.

jens-mueller-sociomantic:
I'd like to change to `toStringz(str.dup)`, if you can make it work. Note though that if I don't surround it with `version (D_Version2)` it will always allocate. I wonder whether this is good for ocean in the long run. I think this is why I find `ocean.stdc.stringz` much nicer to use.

gavin-norman-sociomantic:
>Note though that if I don't surround it with version (D_Version2) it will always allocate.

I don't understand what you mean here.

jens-mueller-sociomantic:
Since in D1 strings are always mutable you might get away without allocating. I mean is it possible that there is already allocated space at the end that can be used to place the `\0`. This does not usually work for D2 since you work with mostly with `const` or `immutable` strings.

gavin-norman-sociomantic:
Right, but you're saying that your current usage always allocates anyway (and that's not a problem), or?

jens-mueller-sociomantic:
Yes. Currently it is not a for me problem. But if there was a function in ocean that takes care of these issues under the hood it'll be better. You could even have both. A more automatic way of doing things calling the low level functionality. But both functions are public. Also note currently `toStringz(str.dup)` does not work. You have to write `auto tmp = str.dup; toStringz(tmp)`. This gets annoying quickly, especially when you have more than one string that needs to be converted. These are all minor issues. What looks important to me is that `ocean.util.stringz` seemed more user friendly compared to `ocean.text.util.StringC`. You might get both (user friendliness and efficiency). Note there is also the special case of string literals in D2, which already include `\0`. Here you can just pass the directly the pointer.

gavin-norman-sociomantic:
>Also note currently toStringz(str.dup) does not work. You have to write auto tmp = str.dup; toStringz(tmp).

Why is that?

jens-mueller-sociomantic:
I think it is because you pass by `ref`. Apparently you cannot pass the return value of a function by `ref`, because you cannot form an address here. Technically it should be possible I think but it does not work.

gavin-norman-sociomantic:
Ah yes, I see. That's annoying.

mathias-lang-sociomantic:
> Since in D1 strings are always mutable you might get away without allocating. I mean is it possible that there is allocated space at the end that can be used to place the \0. This does not usually work for D2 since you work with mostly with const or immutable strings.

Nope. If you use `str ~ "\0"` you will always have an allocation, because it's appending (`~`), not append-and-assign (`~=`).

Frankly, if you do not care about memory allocation, just use `(str ~ "\0").ptr`. If you do care, use a buffer. Anything in between will be introducing complication for little to no benefit.

So, @gavin-norman-sociomantic , I'll follow up on your idea and deprecate the whole module.

jens-mueller-sociomantic:
I never said that the function should use `~` to append `\0`. I said that a function might check for a way to write a `\0` without allocation. So the function optimizes the allocation away when it can. We also pointed out why doing `~ "\0"` is a bad idea. I prefer a single complication inside a library over the same complications over and over in applications.

mihails-strasuns-sociomantic:
I disagree and vote for @gavin-norman-sociomantic proposal too. Existing `toStringz` function doesn't do any checking of premature `\0` so it is strictly equivalent to `s ~ '\0'` when you don't care about allocations. Simplicity trumps variety in this case in my opinion.

mathias-lang-sociomantic:
Well, the existing function [in StringC](https://github.com/sociomantic-tsunami/ocean/blob/be5fc69980e3d9e1583969317083a45b0c9b528d/src/ocean/text/util/StringC.d#L90) checks for the existence of a null terminator, and so does [~`fromStringz`~ (EDIT: `toStringz`)](https://github.com/sociomantic-tsunami/ocean/blob/be5fc69980e3d9e1583969317083a45b0c9b528d/src/ocean/stdc/stringz.d#L39).
But if you don't care about allocation, it's not an optimization worth looking into.

@jens-mueller-sociomantic :
> Note there is also the special case of string literals in D2, which already include \0. Here you can just pass the directly the pointer.

It is also the case in D1, string literal are null terminated. However the null terminator is *not* part of the slice, so in most use case you loose it. You cannot access it later either, as it would require you to try an out-of-range access (potentially reading garbage).

mihails-strasuns-sociomantic:
But `toStringz` doesn't which is what we talk about here. Of course `fromStringz` does, it has no other way.

mathias-lang-sociomantic:
> But toStringz doesn't which is what we talk about here. Of course fromStringz does, it has no other way.

You should have clicked the link :P
My comment mentioned `fromStringz`, but I was linking to `toStringz`.

mihails-strasuns-sociomantic:
The check you have linked is a different one from what @jens-mueller-sociomantic was talking about. He referred to essentially a contract that ensures D string has no `\0` in the middle of the buffer.

mihails-strasuns-sociomantic:
So, how about proceeding with only buffer-based function and documentation suggesting to concat `\0` when one doesn't care?

mihails-strasuns-sociomantic:

nemanja-boric-sociomantic:
> So, how about proceeding with only buffer-based function and documentation suggesting to concat \0 when one doesn't care?

Although I found places where the behavior of checking for NUL is useful (in case of the fixed-length strings), I agree with this one, going from the @mathias-lang-sociomantic comment:

> Frankly, if you do not care about memory allocation, just use (str ~ "\0").ptr. If you do care, use a buffer. Anything in between will be introducing complication for little to no benefit.

mathias-lang-sociomantic:
I have a commit to clean that up locally, but it's more work than expected. Also I always considered this kind of low prio, but it's definitely on my TODO. Will finalize it before next week's friday.

jens-mueller-sociomantic:
@mathias-lang-sociomantic What's the status of this? I like to see this finished because then I know what changes are needed for the code I'm working on.

mathias-lang-sociomantic:
Deprecated the whole module and rebased.

mihails-strasuns-sociomantic (approved):
LGTM
That is plenty of modules changes so another eye from @sociomantic-tsunami/core-team would be nice

src/ocean/sys/ErrnoException.d
```diff
@@ -37,12 +37,11 @@ version (UnitTest)
 
 public class ErrnoException : Exception
 {
+    import ocean.core.Exception : ReusableExceptionImplementation;
```
gavin-norman-sociomantic, line 40: Can you remove this change (and the other "sorting imports" changes)? Unnecessary changes like this have caused problems with merging recently.

src/ocean/net/device/Berkeley.d
```diff
@@ -1033,7 +1033,9 @@ deprecated public abstract class Address
                 hints.ai_flags = flags;
                 hints.ai_family = (flags & AIFlags.PASSIVE && af == AddressFamily.UNSPEC) ? AddressFamily.INET6 : af;
                 hints.ai_socktype = SocketType.STREAM;
-                int error = getaddrinfo(toStringz(host), service.length == 0 ? null : toStringz(service), &hints, &info);
+                int error = getaddrinfo((host ~ "\0").ptr,
```
gavin-norman-sociomantic, line 1037: In the places where you're simply appending `\0`, what if the string already ends with that character? This doesn't seem like a direct replacement.

src/ocean/net/device/Berkeley.d
```diff
@@ -1353,7 +1355,14 @@ deprecated public class IPv4Address : Address
         static uint parse(istring addr)
         {
                 char[64] tmp;
-                return ntohl(inet_addr(toStringz(addr, tmp)));
+                if ((addr.length + 1) > tmp.length)
```
gavin-norman-sociomantic: Is there really not a proper replacement of `toStringz` that checks for these kinds of conditions? I don't really like moving logic like this out of a helper function into the call site, just for the sake of deprecating the helper function.

gavin-norman-sociomantic (commented):
A couple of questions.

src/ocean/net/device/Berkeley.d
mathias-lang-sociomantic: [We discussed it already](https://github.com/sociomantic-tsunami/ocean/pull/109#issuecomment-302082401).

src/ocean/net/device/Berkeley.d
mathias-lang-sociomantic: I am a bit dubious myself TBH. But 90% of the affected code is old (everything but `DropPriviledgesExt`, which we could look into deprecating, is Tango AFAICS).

src/ocean/net/device/Berkeley.d
gavin-norman-sociomantic: Fair enough. For places where allocation doesn't matter anyway, I guess it's not a big deal.

src/ocean/sys/ErrnoException.d
mathias-lang-sociomantic: :(
I feel that the reason you're having the conflicts is that imports get added randomly, instead of a defined order. But will do.

src/ocean/net/device/Berkeley.d
gavin-norman-sociomantic: Yeah... It shows that `ocean.text.util.StringC` doesn't really provide a convenient replacement for all usages of `ocean.stdc.stringz`.

This pattern is repeated 3 times in this module and I see similarly awkward code replacing a simple call in several other modules in this PR. I dunno. I don't really like it, whether the modules in question could be deprecated or not. Presumably the same problem could occur in application code.

gavin-norman-sociomantic: You know what I mean? It doesn't seem very cool to me to say to users: "This function is deprecated. Instead, please just copy the bits of its internal logic that you need into your application code." (Where some of the internal logic is more than just a simple matter of `~ '\0'`.)

mathias-lang-sociomantic: CC @mihails-strasuns-sociomantic as he was the one that asked for this change.

mihails-strasuns-sociomantic: Why not simply replace stack-allocated `char[64] tmp` with `static char[] tmp` and use `copy` + `ocean.text.utill.StringC` instead? That was what I have anticipated as a migration instruction, though in this specific case I liked preserving exact old semantics because it is untested scary legacy code.

gavin-norman-sociomantic: Sounds fine. As long as there's _some_ guidance to users.

mathias-lang-sociomantic:
Updated with an hopefully better approach

relnotes/stringz.deprecation.md
```diff
@@ -0,0 +1,5 @@
+* `ocean.stdc.stringz`
+
+  This module and all functions inside of it have been deprecated,
+  as they duplicates already existing functionalities in ocean.text.util.StringC,
+  which is more idiomatic and actively maintained.
```
mihails-strasuns-sociomantic, line 5: Maybe add a comment here suggesting to switch to `static mstring buffer` when necessary? I would expect more questions like @gavin-norman-sociomantic had would arise.

mathias-lang-sociomantic: Expanded

mihails-strasuns-sociomantic (approved):

mathias-lang-sociomantic:
Pingitty-ping

gavin-norman-sociomantic (approved):
Updated release notes look good :+1:

PR #110, opened 2017-05-12
Fix implicit string concatenation in StructConverter and Contiguous

joseph-wakeling-sociomantic:
Implicit concatenation is deprecated in D2, so replacing these with explicit concatenations will save us some compiler warnings in the future.


mihails-strasuns-sociomantic (approved):

mathias-lang-sociomantic:
There are also some on the deprecation side. I'd say we can make those just one long line and save us some pain. Thoughts ?

PR #112, opened 2017-05-14
Formatter: Deprecate the size_t returning delegate

### LINKED ISSUE #41: Add void sink delegate support to Formatter
The return value of the sink delegate (`size_t delegate ( cstring chunk )`) is currently unused. For more convenient usage, it'd be nice for `Formatter` to support void sink delegates.

Geod24:
The return value is not used and the compatibility with Layout_tango is not needed.
Using void as a return type simplifies user code.

Fix #41 
CC @gavin-norman-sociomantic

### DIFF
--- a/src/ocean/text/convert/Formatter.d
+++ b/src/ocean/text/convert/Formatter.d
@@ -36,10 +36,15 @@ version (UnitTest)
 
 /*******************************************************************************
 
-    Type of 'sink' that can be passed to `format`, matches `Layout.SizeSink`
+    Type of 'sink' that can be passed to `format`, that will just format a
+    string into the provided sink.
 
 *******************************************************************************/
 
+public alias void delegate(cstring) FormatterSink;
+
+/// Deprecated as the `size_t` return is not used. Use `FormatterSink` instead.
+deprecated("Use FormatterSink instead, Sink's return value is never used")
 public alias size_t delegate(cstring) Sink;
 
 
@@ -55,7 +60,7 @@ public alias size_t delegate(cstring) Sink;
 
 *******************************************************************************/
 
-private alias size_t delegate(cstring, ref Const!(FormatInfo)) ElemSink;
+private alias void delegate(cstring, ref Const!(FormatInfo)) ElemSink;
 
 /// This was accidentally made private so needs to be deprecated
 deprecated("This alias shouldn't be used from outside the Formatter")
@@ -79,10 +84,9 @@ public istring format (Args...) (cstring fmt, Args args)
 {
     mstring buffer;
 
-    scope Sink sink = (cstring s)
+    scope FormatterSink sink = (cstring s)
     {
         buffer ~= s;
-        return s.length;
     };
 
     sformat(sink, fmt, args);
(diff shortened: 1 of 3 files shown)


mihails-strasuns-sociomantic:
Red

src/ocean/text/convert/Formatter.d
```diff
@@ -36,15 +36,39 @@ version (UnitTest)
...
+    of cropping and width
+
+    This sink expects to receive a full element as the first parameter,
+    in other words, the full chunk of text that needs a fixed size.
+    This is why one cannot format a whole aggregate (struct, arrays etc.), but
+    only individual elements.
+
+*******************************************************************************/
+
+private alias size_t delegate(cstring, ref Const!(FormatInfo)) PElementSink;
```
gavin-norman-sociomantic: `ElemSink`? That `P` is pretty cryptic.

src/ocean/text/convert/Formatter.d
```diff
@@ -36,15 +36,39 @@ version (UnitTest)
...
+
+/// This was accidentally made private so needs to be deprecated
+deprecated("This alias shouldn't be used from outside the Formatter")
```
gavin-norman-sociomantic: Is this deprecation a separate change to "Formatter: Deprecate the size_t returning delegate"?

src/ocean/text/convert/Formatter.d
```diff
@@ -242,7 +254,7 @@ public alias size_t delegate(cstring, ref Const!(FormatInfo)) ElementSink;
 
 *******************************************************************************/
 
-private size_t widthSink (Sink sink, cstring str, ref Const!(FormatInfo) f)
+private void widthSink (FormatterSink sink, cstring str, ref Const!(FormatInfo) f)
```
gavin-norman-sociomantic: Returns doc not updated.

src/ocean/text/convert/Formatter.d
mathias-lang-sociomantic: `P` is just for private (I needed a separate name). Since this is not public I wasn't too bothered about the name (but can change it if you prefer).

gavin-norman-sociomantic: Yeah, it took me a min to figure out what the `P` meant ;)

src/ocean/text/convert/Formatter.d
mathias-lang-sociomantic: Not really. The sink type needed to be updated to have `void` return type instead of `size_t`.
I just forgot to `git add` it in a previous version of this commit :blush:

mathias-lang-sociomantic:
Fixed everything (green now!)

src/ocean/text/convert/Formatter.d
gavin-norman-sociomantic: Would be nice to split it into a separate commit then.

mathias-lang-sociomantic: Done

gavin-norman-sociomantic (approved):

PR #113, opened 2017-05-15
Move error-suppressing static asserts to unittest

mathias-lang-sociomantic:
The is no advantage in having them within the body, are unittests for ocean (and other applications) are always run.
However it does suppress error during development and lead to extra steps while debugging some core code (e.g. formatter).


mihails-strasuns-sociomantic:
I think I am actually going to remove this completely as all this stuff originated in over-generalization attempt when it seemed possible to unite all serializers we have under a single API.

mihails-strasuns-sociomantic (approved):

PR #114, opened 2017-05-16
Deprecate contiguous serializer traits

mihails-strasuns-sociomantic:
Those originated in over-generalized assumption that new framework will
be use to adjust all over serializers but it proved to be impactical for
the time.

At the same time extra compile-time reflection complicated code
structure consierably and resulted in suppression of relevant compiler
messages. It is better to deprecate/remove it for now and reconsider if
standardization of serializer is to happen again.

### DIFF
--- a/relnotes/serializer-traits.deprecation.md
+++ b/relnotes/serializer-traits.deprecation.md
@@ -1,0 +1,9 @@
+* `ocean.util.serializer.model.Traits`
+
+  The traits that check serializer/deserializer interface via structural typing
+  have been deprecated and current serializer/decorator implementation adapted
+  to not use them.
+
+  Those originated in over-generalized assumption that new framework will
+  be use to adjust all over serializers but it proved to be impactical for
+  the time. This has proven to not be the case and only complicated development.
--- a/src/ocean/util/serialize/model/Traits.d
+++ b/src/ocean/util/serialize/model/Traits.d
@@ -47,7 +47,7 @@ private struct _Dummy
 
 *******************************************************************************/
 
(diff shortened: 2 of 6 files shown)


relnotes/serializer-traits.deprecation.md
```diff
@@ -0,0 +1,9 @@
+* `ocean.util.serializer.model.Traits`
+
+  The traits that check serializer/deserializer interface via duck typing have
```
mathias-lang-sociomantic: Nitpick: `structural typing` (duck typing is RT only).

src/ocean/util/serialize/model/Traits.d
```diff
@@ -47,7 +47,7 @@ private struct _Dummy
 
 *******************************************************************************/
 
-template isSerializer(T)
+deprecated template isSerializer(T)
```
mathias-lang-sociomantic, line 50: Mark the module as deprecated as well so we don't have lingering imports.

mathias-lang-sociomantic (approved):
Happy to see those going away.

src/ocean/util/serialize/model/Traits.d
mihails-strasuns-sociomantic: Tricky bit here - the module is publicly imported via `package_.d` and can't be marked deprecated wtihout removing from the imports, which would be a breaking change on its own. I have no better idea how to fix that other than just kill it in next major.

mihails-strasuns-sociomantic:
Updated: `duck typing` -> `structural typing`

PR #115, opened 2017-05-17
Deprecate ocean.io.device.{SerialPort,ProgressFile}

mathias-lang-sociomantic:
Because of EUSERNOTFOUND


gavin-norman-sociomantic:
Thanks!

PR #120, opened 2017-05-18
Formatter: toString overloads are not recognized anymore

mathias-lang-sociomantic:
This regression was introduced in #112. While changing the type internally used by the formatter,
it affected which toString overload can be used.
We now correctly recognize both toString overloads and issue a 'deprecation' message when size_t is encountered.

### DIFF
--- a/src/ocean/text/convert/Formatter.d
+++ b/src/ocean/text/convert/Formatter.d
@@ -385,7 +385,14 @@ private void handle (T) (T v, FormatInfo f, FormatterSink sf, ElemSink se)
     // toString hook: Give priority to the non-allocating one
     // Note: sink `toString` overload should take a `scope` delegate
     else static if (is(typeof(v.toString(sf))))
-        v.toString((cstring e) { return se(e, f); });
+        v.toString((cstring e) { se(e, f); });
+    else static if (is(typeof(v.toString((size_t delegate(cstring)).init))))
+    {
+        pragma(msg, "Deprecation: Please change toString to accept ",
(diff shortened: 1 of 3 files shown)


mihails-strasuns-sociomantic:
Regression test case would be really appreciated ;)

src/ocean/text/convert/Formatter.d
```diff
@@ -385,7 +385,14 @@ private void handle (T) (T v, FormatInfo f, FormatterSink sf, ElemSink se)
     // toString hook: Give priority to the non-allocating one
     // Note: sink `toString` overload should take a `scope` delegate
     else static if (is(typeof(v.toString(sf))))
+        v.toString((cstring e) { se(e, f); });
+    else static if (is(typeof(v.toString((size_t delegate(cstring)).init))))
+    {
+        pragma(msg, "Deprecation: Please change stringOf to accept ",
```
mihails-strasuns-sociomantic: `stringOf` -> `toString`?

mathias-lang-sociomantic:
> Regression test case would be really appreciated ;)

The issue is that it'll print a message to stdout during unittests :(

mihails-strasuns-sociomantic:
Not if you add it as `something_test.d` module.

Geod24:
> Not if you add it as something_test.d module.

That's an excellent idea. Separating the tests from the Formatter could also allow us to solve the license issue (</noteToSelf>).
Updated.

mihails-strasuns-sociomantic (approved):

mihails-strasuns-sociomantic:
Thanks!

PR #123, opened 2017-05-22
Enable code coverage checks

mihails-strasuns-sociomantic:


mihails-strasuns-sociomantic:
I think that should be it, @leandro-lucarella-sociomantic can you give a sanity check?

https://github.com/codecov/browser-extension is also recommended now to be able to see various coverage bits in github UI directly

Build.mak
```diff
@@ -1,6 +1,3 @@
-# This will make D2 unittests fail if stomping prevention is triggered
-export ASSERT_ON_STOMPING_PREVENTION=1
-
```
mathias-lang-sociomantic: Why ?

mihails-strasuns-sociomantic: Guess what - `-cov` implementation in runtime triggers the assertion and there is no way to change it apart from tweaking runtime :)

Though I had to disable D2 `-cov` bit because of another bug anyway, will restore that now.

mihails-strasuns-sociomantic:
`coment.layout` config doesn't seem to work sadly :/

Build.mak
mathias-lang-sociomantic: Won't that mean that `unittest` won't catch mistaken allocationss now ?

mihails-strasuns-sociomantic: I have restored this bit, at least until D2 `-cov -inline` bug is fixed and we can test that too

mathias-lang-sociomantic (approved):
LGTM

leandro-lucarella-sociomantic:
> coment.layout config doesn't seem to work sadly :/

maybe it's missing an `m` (`comment` :D)?

.codecov.yml
```diff
@@ -0,0 +1,18 @@
...
+            default:
+                if_no_uploads: success
+                if_not_found: success
+        changes:
+            default:
+                if_no_uploads: success
+                if_not_found: success
+
+comment:
+    layout: "diff"
```
leandro-lucarella-sociomantic, line 18: Do we really want comments at all? It seems quite annoying, I guess you can find the details going to the commit status link anyway?

Build.mak
```diff
@@ -9,10 +9,12 @@ TEST_RUNNER_MODULE := ocean.core.UnitTestRunner
 
 ifeq ($(DVER),1)
 override DFLAGS := $(filter-out -di,$(DFLAGS)) -v2 -v2=-static-arr-params -v2=-volatile
+COVFLAG:=-cov
 else
 # Open source Makd uses dmd by default
 DC ?= dmd
 override DFLAGS += -de
+COVFLAG:=
```
leandro-lucarella-sociomantic, line 17: Why no coverage in D2?

ci/travis-after-success.sh
```diff
@@ -1,6 +1,9 @@
 #!/bin/sh
 set -xe
 
+# Submit code coverage report
+curl -s https://codecov.io/bash | bash
```
leandro-lucarella-sociomantic: Oh, this is nasty. I don't like it at all, is there anything that doesn't involve executing a remote script to upload the coverage report?

leandro-lucarella-sociomantic (commented):
Well, the links seem broken for now, but I guess it's just because it's the first time this is used, right?

ci/travis-after-success.sh
mihails-strasuns-sociomantic: Do you consider "navigating through 1500 lines of convoluted bash code to hacking own replacement" a viable alternative? This is the only "official" way.

To be honest I don't see much harm in it as it is a public project tested in vitual environment - what can possibly happen?

Build.mak
mihails-strasuns-sociomantic: Needs more work:

1) Triggers stomping prevention assert
2) Wrong code gen bug (reducing now)

.codecov.yml
mihails-strasuns-sociomantic: It is better than it may seem as it will only post one comment and keep updating it on pushes instead of placing more comments. I intentionally reduced the layout to minimal useful detail (small text-only table with % stats).

mihails-strasuns-sociomantic:
> maybe it's missing an m (comment :D)?

It is correct in code. I googled a bit and the issue seems that it uses layout configuration from base branch, not PR one. So it may fix itself after merging.

.codecov.yml
leandro-lucarella-sociomantic: Mmm, we can try it out, but still it looks a bit too verbosy...

ci/travis-after-success.sh
leandro-lucarella-sociomantic: They could steal credentials to push to ocean D2 (that, thanks to GitHub great permissions granularity, right now it means having access to ALL our code).

leandro-lucarella-sociomantic: We could download the script once and use that until it breaks, that would be a step forward. We could also isolate this by running it in a separate (empty) docker container.

leandro-lucarella-sociomantic: Probably running in an empty docker container (only mapping the coverage report file to it) should be the easiest choice.

leandro-lucarella-sociomantic: OK, the script is opensource and stored in GitHub. I would prefer to use a GitHub URL using a particular hash to download it or even just shipping the script as part of the `sociomantictsunami/base` cachalot image, so it is available to all projects (although for now downloading might be better until we decide to go this route or not).

But I was just looking at the scripts and they are really awful. Is not making me trust codecov much really, in the python upload script they parse the yaml manually through regex... WTF. Also they do all sort of guessing for all the types or CIs out there and searching for all type of coverage reports out there, when all needed should be just a simple REST call to a website to send a report, like any other sane service out there...

I really don't understand why they do this and I don't like it at all...

leandro-lucarella-sociomantic: Could you change this to run it using `bash -x` instead to see what's the exact query sent at the end. I would really like to avoid running thousands of lines of code just to do a `curl -F report URL` at the end...

mihails-strasuns-sociomantic: Will try `bash -x`. Note though that credential issue is a major vulnerability in all public CI services - anyone can make PR against your repo with arbitrary modifications to `.travis.yml` and unless you manage to stop the build before it gets tested, the script will run with your credentials. I am not aware of any solutions to that (though of course it is not as subtle as changing remote bash script).

.codecov.yml
mihails-strasuns-sociomantic: We can easily disable it later if it proves annoying (`comment: false`)

ci/travis-after-success.sh
mihails-strasuns-sociomantic: @leandro-lucarella-sociomantic I have reduced it to essentially a loop over all report files that concatenates them into single uploaded file and then does few curl queries - one against codecov API to get a token and later to Amazon S3 to actually upload report.

With some trial and error we could probably reproduce it but it seems somewhat fragile to me with all hard-coded urls. What do you think about simply storing the whole script inside ocean repo? It doesn't download / execute any other scripts internally AFAICS so that would work as sort of approved pinning.

leandro-lucarella-sociomantic: > Will try bash -x. Note though that credential issue is a major vulnerability in all public CI services - anyone can make PR against your repo with arbitrary modifications to .travis.yml and unless you manage to stop the build before it gets tested, the script will run with your credentials. I am not aware of any solutions to that (though of course it is not as subtle as changing remote bash script).

Yes, I know, but as you said is quite more complicated than just changing a remote script. I want to create bot accounts, trying to get approval for that to avoid having super-powered tokens out there.

leandro-lucarella-sociomantic: > With some trial and error we could probably reproduce it but it seems somewhat fragile to me with all hard-coded urls. What do you think about simply storing the whole script inside ocean repo? It doesn't download / execute any other scripts internally AFAICS so that would work as sort of approved pinning.

Yeah, thinking about it, probably running the script with `docker run` (but without sharing the whole build dir, only the reports, and all the env vars with the container) it should be safe and convenient enough, even if we keep using the remote script (although I would download it from GitHub, not the codecov.io URL).

ci/travis-after-success.sh
```diff
@@ -1,6 +1,9 @@
 #!/bin/sh
 set -xe
 
+# Submit code coverage report
+bash ci/codecov.sh
```
leandro-lucarella-sociomantic: mmm, this after success script is not running in a container at all :thinking: (beaver should take care of this so for now I will not focus on improving this), but I would put here:

```sh
curl -s https://codecov.io/bash | docker run -ti --rm -v PATH_TO_REPORTS ocean bash
```

leandro-lucarella-sociomantic: Or with the GitHub URL

leandro-lucarella-sociomantic: Oh, we probably need to export some `TRAVIS_` variables so codecov can detect which CI we are using, but we can probably tell it that too via command-line options: https://github.com/codecov/codecov-bash/blob/master/codecov#L78-L176

mihails-strasuns-sociomantic:
Updated:
- download pinned version via github link
- use docker container to run script
- pass travis env vars to docker container
- disable bunch of script options we don't need

mihails-strasuns-sociomantic:
It seems to work for most part but custom config yml is ignored because stupid bash script tries to find it via `git`/`hg` commands. Will either need to give up on running it in separate folder or give up on downloading script from upstream repo so that we can patch it.

ci/travis-after-success.sh
```diff
@@ -1,6 +1,19 @@
...
+mkdir coverage
+cp .codecov.yml coverage/
+cp ci/codecov.sh coverage/
+find . -maxdepth 1 -type f -name "*.lst" -exec mv '{}' 'coverage/' \;
+CODECOVCMD="codecov.sh -y .codecov.yml -R ."
+docker run -ti --rm -v $PWD/coverage:/docker -w /docker \
+    -e CI \
+    -e TRAVIS -e TRAVIS_BRANCH -e TRAVIS_COMMIT -e TRAVIS_JOB_NUMBER \
+    -e TRAVIS_PULL_REQUEST -e TRAVIS_JOB_ID -e TRAVIS_REPO_SLUG -e TRAVIS_TAG \
+    -e TRAVIS_OS_NAME \
```
leandro-lucarella-sociomantic: Not a huge deal, but I think it would be easier to just pass these as [arguments to the script](https://github.com/codecov/codecov-bash/blob/d82a80bc465343ed367fa0788aff312ce5582a61/codecov#L149-L156), the less the script discovers by itself, the better IMHO.

leandro-lucarella-sociomantic (approved):
Other than that, LGTM.

leandro-lucarella-sociomantic:
Feel free to merge as is or after that's changed.

ci/travis-after-success.sh
mihails-strasuns-sociomantic: I considered that but it actually uses more than just values exposes via cmdline args - https://github.com/codecov/codecov-bash/blob/d82a80bc465343ed367fa0788aff312ce5582a61/codecov#L431-L440 thus I decided it is safer to just rely on env vars and expose everything.

PR #124, opened 2017-05-22
Make ocean.core.Traits : isTypedef, StripTypedef useful in D2

mathias-lang-sociomantic:
It's been a common problem during porting that some meta code needed to special-case Typedef but couldn't.
The code was originally trying to ensure that no typedef special casing happened, but it proved impossible in some cases,
which led to libraries / utilities implementing their own StripTypedef / IsTypedef (e.g. the Formatter).
This fixes the problem at the root, by exposing same behavior between D1 and D2.

### DIFF
--- a/src/ocean/core/Traits.d
+++ b/src/ocean/core/Traits.d
@@ -849,10 +849,7 @@ unittest
 
     Determines if T is a typedef.
 
-    Typedef has been removed in D2 and this template will always evaluate to
-    false if compiled with version = D_Version2.
-
-    Template_Params:
+    Params:
         T = type to check
 
     Evaluates to:
@@ -864,7 +861,7 @@ version (D_Version2)
 {
     public template isTypedef (T)
     {
-        const bool isTypedef = false;
+        const bool isTypedef = is(T.IsTypedef == void);
     }
 }
 else
@@ -895,25 +892,14 @@ else
 unittest
 {
     mixin(Typedef!(int, "MyInt"));
-
-    version (D_Version2)
-    {
-        static assert (!isTypedef!(MyInt)); // just a struct
-    }
-    else
-    {
-        static assert ( isTypedef!(MyInt));
-    }
+    static assert (isTypedef!(MyInt));
 }
 
 /*******************************************************************************
 
     Strips the typedef off T.
 
-    Typedef has been removed in D2 and this template is a no-op if compiled
-    with version = D_Version2.
-
-    Template_Params:
+    Params:
         T = type to strip of typedef
 
     Evaluates to:
@@ -925,7 +911,10 @@ version (D_Version2)
 {
     public template StripTypedef (T)
     {
-        alias T StripTypedef;
+        static if (isTypedef!(T))
+            alias T.TypedefBaseType StripTypedef;
+        else
+            alias T StripTypedef;
     }
 }
 else
@@ -956,15 +945,7 @@ else
 unittest
 {
     mixin(Typedef!(int, "MyInt"));
-
-    version (D_Version2)
-    {
-        static assert (is(StripTypedef!(MyInt) == MyInt));
-    }
-    else
-    {
-        static assert (is(StripTypedef!(MyInt) == int));
-    }
+    static assert (is(StripTypedef!(MyInt) == int));
 }
 
 /******************************************************************************
@@ -991,14 +972,8 @@ template ContainsDynamicArray ( T ... )
     {
         static if (isTypedef!(T[0]))
         {
-            mixin(`
-            static if (is (T[0] Base == typedef))
-            {
-                // Recurse into typedef.
-
-                const ContainsDynamicArray = ContainsDynamicArray!(Base, T[1 .. $]);
-            }
-            `);
+            const ContainsDynamicArray = ContainsDynamicArray!(
+                StripTypedef!(T[0]), T[1 .. $]);
         }
         else static if (is (T[0] == struct) || is (T[0] == union))
         {
@@ -1081,10 +1056,7 @@ template ReturnAndArgumentTypesOf ( T )
 {
     static if (isTypedef!(T))
     {
-        mixin(`
-        static if (is(T F == typedef))
-            alias ReturnAndArgumentTypesOf!(F) ReturnAndArgumentTypesOf;
-        `);
+        alias ReturnAndArgumentTypesOf!(StripTypedef!(T)) ReturnAndArgumentTypesOf;
     }
     else static if (is(T Args == function) && is(T Return == return))
     {
--- a/src/ocean/io/serialize/StructSerializer.d
+++ b/src/ocean/io/serialize/StructSerializer.d
@@ -575,12 +575,8 @@ struct StructSerializer ( bool AllowUnions = false )
 
                 static if (isTypedef!(T))
                 {
-                    mixin(`
-                    static if ( is(T B == typedef) )
-                    {
-                        serializer.serialize(data, cast(B)(field), field_name);
-                    }
-                    `);
+                    serializer.serialize(data, cast(StripTypedef!(T))(field),
+                                         field_name);
                 }
                 else static if ( is(T B == enum) )
                 {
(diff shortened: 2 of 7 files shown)


mihails-strasuns-sociomantic:
If you insist on going this path I would like to make it more generic via some convention so that arbitrary struct can enable typedef-like treatment. Something like `is(S.TypedefBaseType) && is(typeof({ S.TypedefBaseType base = S.init; }))`

mathias-lang-sociomantic:
> If you insist on going this path I would like to make it more generic via some convention so that arbitrary struct can enable typedef-like treatment.

Do we really want to go this way ? My intent is not to promote `typedef`. This feature was killed. The suggested replacement does not cover all the use case there is, because `typedef` is a way to subtype, but `alias this` is not. It does 95% of the job, but *cannot* do all of it.
However, we *do* rely on this feature in our code. With D2 abilities we might be able to transition to a more lax approach, but for the moment we need one way to emulate `Typedef`, and that's it. I just found another case where typedef has to be handled explicitly.

mathias-lang-sociomantic:
My previous patch was quite incomplete, and there's some code which should have broke which did not.
In any case, I added several more fixes.

mihails-strasuns-sociomantic:
>  My intent is not to promote typedef. This feature was killed.

It is exactly the reasoning of my proposal. I don't want to introduce trait which sole purpose is to support legacy concept, it only introduce more technical debt in any code that uses it. Instead it would be much more sustainable to introduce concept saying "A is supertype of A.Type and is implicitly convertible to A.Type" and implement all generic algorithms in these terms.

For example, I would totally like formatter to print `Contiguous!(S)` structs same as `S` in D2 when `alias ptr this` will be introduced. With my proposal it would be simply a matter of defining `alias SubType = S;`, with yours it would require explicit formatter support or wrapper `toString` that simply forwards to `ptr.toString()` - and add more such workaround for each purpose that treats `typedef` specially right now. I don't think it scales at all.

mathias-lang-sociomantic:
```diff
diff --git a/src/ocean/core/Traits.d b/src/ocean/core/Traits.d
index 1419fe1..7e01e23 100644
--- a/src/ocean/core/Traits.d
+++ b/src/ocean/core/Traits.d
@@ -912,7 +912,7 @@ version (D_Version2)
     public template StripTypedef (T)
     {
         static if (isTypedef!(T))
-            alias typeof(T.value) StripTypedef;
+            alias T.TypedefBaseType StripTypedef;
         else
             alias T StripTypedef;
     }
diff --git a/src/ocean/transition.d b/src/ocean/transition.d
index c8d223f..f2607a4 100644
--- a/src/ocean/transition.d
+++ b/src/ocean/transition.d
@@ -101,6 +101,7 @@ template Typedef(T, istring name, T initval)
                 ("static struct " ~ name ~
                 "{ " ~
                 "alias IsTypedef = void;" ~
+                "alias TypedefBaseType = " ~ T.stringof ~ ";" ~
                 T.stringof ~ " value = " ~ initval.stringof ~ ";" ~
                 "alias value this;" ~
                 "this(" ~ T.stringof ~ " rhs) { this.value = rhs; }" ~
@@ -127,6 +128,7 @@ template Typedef(T, istring name)
                 ("static struct " ~ name ~
                 "{ " ~
                 "alias IsTypedef = void;" ~
+                "alias TypedefBaseType = " ~ T.stringof ~ ";" ~
                 T.stringof ~ " value; " ~
                 "alias value this;" ~
                 "this(" ~ T.stringof ~ " rhs) { this.value = rhs; }" ~
```

mathias-lang-sociomantic:
I've no idea why the coverage seems to go down. It points to `Convert.d` for example, but I've removed an unused branch. But since the code is not compiled in, DMD doesn't care...

mihails-strasuns-sociomantic:
It even reports some unchanged module. I wonder who is to blame here - dmd or codecov x_X

mathias-lang-sociomantic:
Well this module is a random engine, so it might be neither of them :)

mathias-lang-sociomantic:
Ping @mihails-strasuns-sociomantic

src/ocean/io/serialize/StructSerializer.d
```diff
@@ -575,12 +575,8 @@ struct StructSerializer ( bool AllowUnions = false )
 
                 static if (isTypedef!(T))
                 {
-                    mixin(`
-                    static if ( is(T B == typedef) )
-                    {
-                        serializer.serialize(data, cast(B)(field), field_name);
-                    }
-                    `);
+                    serializer.serialize(data, cast(StripTypedef!(T))(field),
```
mihails-strasuns-sociomantic, line 578: So this is exactly why I don't want to allow silent `StripTypedef` application to D2 structs. Serializer must not reduce an entity to the base type but serialize it as actual struct. For current `Typedef` implementation ABI is compatible as deserialization will not show any immediate issues but it is an undefined behaviour time bomb.

mihails-strasuns-sociomantic (changes requested):
Sorry for long response, had some wild medical adventures, now catching back.

Sadly, this PR still works in "typedef" terms and this is not acceptable (see one specific line comment for issue example). My proposal was to do something conceptually different:

1) Deprecate `isTypedef` / `StripTypedef` completely
2) Renamed mixed in field in `Typedef` to `BaseType`
3) Introduce new `hasBaseType` and `toBaseType` functions with implementation similar to the one for `StripTypedef` in the current PR
4) Evaluate each current usage place of `isTypedef` individually, checking if it indeed to makes sense to replace one type with possibly completely different type in that context. Among places highlighted by this PR, formatter definitely can and should do it, but serializer is a no go.

src/ocean/io/serialize/StructSerializer.d
mathias-lang-sociomantic: ABI **has** to be compatible. If not, D2 apps won't be able to deserialize profiles that contain `typedef` which were serialized in D1.

mihails-strasuns-sociomantic: Yes, but intended usage of this feature is not limited to legacy `Typedef`. As I have mentioned previously, after we drop D1 support in ocean I want to modify `Contiguous` to say `ref S value(); alias value this; alias BaseType = S;` and get it processed by `Formatter` like `S` for free. There are many more places where one may want well-defined subtyping relation but if your utilities operate in typedef terms, it becomes applicable only for typedef-like entities.

mihails-strasuns-sociomantic: To re-iterate previous context - I don't want any non-deprecate symbols in ocean that support typedef as a special concept in D2 mode. That thing is gone and we should focus more on future proof concepts than on legacy support.

src/ocean/core/Traits.d
```diff
@@ -991,14 +972,8 @@ template ContainsDynamicArray ( T ... )
...
-            mixin(`
-            static if (is (T[0] Base == typedef))
-            {
-                // Recurse into typedef.
-
-                const ContainsDynamicArray = ContainsDynamicArray!(Base, T[1 .. $]);
-            }
-            `);
+            const ContainsDynamicArray = ContainsDynamicArray!(
+                StripTypedef!(T[0]), T[1 .. $]);
```
mihails-strasuns-sociomantic, line 976: Here `StripTypedef` is not needed in D2 either, the field will be processed as any struct field anyway.

mihails-strasuns-sociomantic:
So I have just checked again and looks like formatter is the only piece of ocean which uses `StripTypedef` legitimately here.

mathias-lang-sociomantic:
> So I have just checked again and looks like formatter is the only piece of ocean which uses StripTypedef legitimately here.

https://github.com/sociomantic-tsunami/ocean/pull/138

mihails-strasuns-sociomantic:
"here" means "in this PR" - #138 is much more tricky case and I would discuss it separately (using `StripTypedef` there is *wrong* but necessary to maintain comptibility between D1 and D2 file formats)

mathias-lang-sociomantic:
Closing as per lack of agreement and since #138 is now using its own copy of `IsTypedef`  / `DropTypedef`.

PR #125, opened 2017-05-22
D2: Use hasIndirections to determine AppendBuffer's parameter type

mathias-lang-sociomantic:
Before only class and interface was listen, but the issue is not actually that they are reference type,
but that they are aggregate. Before this patch, AppendBuffer would not instantiate with T = struct with
 mutable indirection.


mihails-strasuns-sociomantic:
`hasIndirections` from traits?

mathias-lang-sociomantic:
:man_facepalming: 

Updated with `hasIndirections` + moar tests

mihails-strasuns-sociomantic:
Red

mathias-lang-sociomantic:
Woops, fixed

PR #121, opened 2017-05-19
Several integration test fixes in preparation for https://github.com/sociomantic-tsunami/makd/issues/41

leandro-lucarella-sociomantic:

PR #126, opened 2017-05-22
Several integration test fixes in preparation for #41

leandro-lucarella-sociomantic:
Re-do of #121.

### DIFF
--- a/test/filesystemevent/main.d
+++ b/test/filesystemevent/main.d
@@ -48,135 +48,6 @@ import ocean.task.Task;
 
 import ocean.task.Scheduler;
 
-deprecated class FileModificationTestTask: Task
-{
-    /***************************************************************************
-
-        File operations to be checked
-
-    ***************************************************************************/
-
-    private bool modified = false;
-    private bool deleted  = false;
-    private bool closed   = false;
-
-    /***************************************************************************
-
-        Name of the created file.
-
-    ***************************************************************************/
-
-    private cstring created_name;
-
-    /***************************************************************************
-
-        Path to the monitored file.
-
-    ***************************************************************************/
-
-    private FilePath temp_path;
-
-    /***************************************************************************
-
-        Variable to control/test the order of the file operations
-
-    ***************************************************************************/
-
-    private int operation_order = 0;
-
-    /***************************************************************************
-
-        Tested FileSystemEvent instance.
-
-    ***************************************************************************/
-
-    private FileSystemEvent inotifier;
-
-    /***************************************************************************
-
-        Test entry point. Prepares environment and tests the FileSystemEvent.
-
-    ***************************************************************************/
-
-    override public void run ( )
-    {
-        auto temp_file = new TempFile(TempFile.Permanent);
-        this.temp_path = FilePath(temp_file.toString());
-
-        this.inotifier  = new FileSystemEvent(&this.fileSystemHandler);
-        inotifier.watch(cast(char[]) temp_file.toString(),
-                       FileEventsEnum.IN_MODIFY | FileEventsEnum.IN_DELETE_SELF
-                     | FileEventsEnum.IN_CLOSE_WRITE );
-
-        theScheduler.epoll.register(inotifier);
-
-        temp_file.write("something");
-        temp_file.close;
-        temp_path.remove();
-
-        this.suspend();
-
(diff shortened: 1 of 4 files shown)


test/filesystemevent/main.d
```diff
@@ -48,135 +48,6 @@ import ocean.task.Task;
 
 import ocean.task.Scheduler;
 
-deprecated class FileModificationTestTask: Task
```
mathias-lang-sociomantic, line 51: Doesn't that imply we loose ability to test deprecated symbols ?

mihails-strasuns-sociomantic: You can suppress deprecations via `DFLAGS` when building specific higher level test binary.

mihails-strasuns-sociomantic:
As I have said before, I would prefer to merge that in 2.x.x in one PR with makd upgrade to get a CI confirmation that changes indeed fix all issues. Bug or not, it has no impact on ocean users and there is no reason to rush it into bug fix release.

test/filesystemevent/main.d
mathias-lang-sociomantic: Then shouldn't it be done in this P.R. instead of removing the test ?

mihails-strasuns-sociomantic: I recommended @leandro-lucarella-sociomantic to nother bother and simply delete these one as it is rather old and going away soon'ish :) Don't mind either way tbh, just didn't seem important.

mathias-lang-sociomantic: Fine by me :+1:

leandro-lucarella-sociomantic: Yeah, I wouldn't pay much attention on testing deprecated modules, most modules are just added to the ignore list anyway. Is not the ideal but as long as we don't have a better method to test deprecated stuff, I wouldn't bother much.

leandro-lucarella-sociomantic:
> As I have said before, I would prefer to merge that in 2.x.x in one PR with makd upgrade to get a CI confirmation that changes indeed fix all issues. Bug or not, it has no impact on ocean users and there is no reason to rush it into bug fix release.

This is related but independent of makd changes on how to test integration tests. I really don't see a point on delaying this trivial unittest-only fix.

mihails-strasuns-sociomantic:
> This is related but independent of makd changes on how to test integration tests. I really don't see a point on delaying this trivial unittest-only fix.

I see no point in merging it faster :) With current makd version ocean has _no_ test issues that need fixing - this change is related to upcoming changes in makd. As such, it should be done in same chunk as makd upgrade to both ensure it truly addresses the issue and make history more coupled.

leandro-lucarella-sociomantic:
> I see no point in merging it faster :) With current makd version ocean has no test issues that need fixing - this change is related to upcoming changes in makd. As such, it should be done in same chunk as makd upgrade to both ensure it truly addresses the issue and make history more coupled.

I feel like we are going in rounds here, but I see it the exact opposite. This is an old bug which wasn't exposed before but it was conceptually wrong anyway. People writing new integration tests could take it as an example and make new wrong integration tests. Also, having the PR opened but not merged is an unnecessary maintenance hassle (for me that I will have to rebase it in the future and for people looking at open PRs that will see this open PR as noise).

I really think it is nonsense to keep this PR open...

mihails-strasuns-sociomantic:
> This is an old bug which wasn't exposed before but it was conceptually wrong anyway

Again, I see no bug in current ocean state. Needing to hack our code to adjust for compiler breakage is not a bug in ocean, it is maintenance/infrastructure effort.

leandro-lucarella-sociomantic:
Well, it's pointing to v2.x.x, so it would be considered a small refactoring if you will. Its OK if you don't want to make it part of a patch release, but I don't see any reason not to merge now in a minor branch.

leandro-lucarella-sociomantic:
@gavin-norman-sociomantic can you maybe intercede as a mediator and untie this pointless discussion? :)

gavin-norman-sociomantic:
I'm fine with merging this in v2.x.x. Whatever the background to the reasons for wanting to make this change now, I see it as a simple, non-urgent code improvement (i.e. perfect for a minor release).

leandro-lucarella-sociomantic:
Can we merge it then? :)

gavin-norman-sociomantic:
Sure. I was just waiting to see if there were any more comments.

PR #128, opened 2017-05-24
Cannot use ocean.core.array.Mutation : sort, reverse on arrays of static arrays

mathias-lang-sociomantic:
It fails in D1, saying one cannot assign to a static array.


mihails-strasuns-sociomantic (approved):
Dirty but I see no other option sadly :(

PR #131, opened 2017-05-29
Correct HTTP status phrase for 200

mathias-lang-sociomantic:
As requested by @federico-cuello-sociomantic

### DIFF
--- a/relnotes/bitflip-gone-wrong.migration.md
+++ b/relnotes/bitflip-gone-wrong.migration.md
@@ -1,0 +1,5 @@
+* `ocean.net.http.consts.StatusCodes : StatusPhrase.StatusReasonPhrases`
+
+  The status reason was changed from `"Ok"` to `"OK"`.
+  While the previous version wasn't strictly against the RFC defining HTTP 1.1,
+  it consistently uses `"OK"` through the RFC.
--- a/src/ocean/net/http/HttpException.d
+++ b/src/ocean/net/http/HttpException.d
@@ -125,7 +125,7 @@ class HttpException : HttpServerException
         catch (Exception)
         {
             test!("==")(e.status, HttpResponseCode.OK);
-            test!("==")(getMsg(e), "Ok Invalid resource");
+            test!("==")(getMsg(e), "OK Invalid resource");
         }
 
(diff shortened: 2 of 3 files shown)


mihails-strasuns-sociomantic:
Jokes aside I would be really curious to check it with some of our internal HTTP projects first to see if some weird legacy module no one has touched for years breaks.

leandro-lucarella-sociomantic (commented):
Same as @mihails-strasuns-sociomantic, also, I would like to see some sort of reasoning behind this besides being recommended by Fede. Why did he recommended this? Was it violating a RFC? Etc.

leandro-lucarella-sociomantic:
I'm tempted to mark this as a breaking change. Is one of those things that in theory is **should** be harmless but that then triggers all sorts of obscure bugs :)

Unless this is a bug fix, and even then if there is no urgency for this, I would put it in the next major, just for potential risk and being totally unexpected.

federico-cuello-sociomantic:
:man_shrugging: ok, I saw it when checking a response from jet (it just looked "wrong") and the RFC is using all upper case but it's not *against* the RFC per se.

https://tools.ietf.org/html/rfc7231#section-6.3.1

mihails-strasuns-sociomantic:
Yeah I agree with @leandro-lucarella-sociomantic , better move it to next major in that case

mathias-lang-sociomantic:
Rebased on v4.x.x

mihails-strasuns-sociomantic (approved):

mihails-strasuns-sociomantic (changes requested):
Release notes :)

mathias-lang-sociomantic:
Done. Note that some files in `relnotes` should probably be deleted for v4 :)

mihails-strasuns-sociomantic:
What do you mean? AFAICS https://github.com/sociomantic-tsunami/ocean/tree/v4.x.x/relnotes has only one file and it is indeed for v4.x.x

mathias-lang-sociomantic:
My bad, my v4.x.x was badly out of sync.

mihails-strasuns-sociomantic (approved):

PR #134, opened 2017-05-30
Fix code coverage report generation

mihails-strasuns-sociomantic:
Seems like https://github.com/sociomantic-tsunami/ocean/pull/123 was not sufficient despite being green and reporting no errors in log. Investigating.

### DIFF
--- a/ci/codecov.sh
+++ b/ci/codecov.sh
@@ -38,7 +38,6 @@ ft_xcodeplist="0"
 
 _git_root=$(git rev-parse --show-toplevel 2>/dev/null || hg root 2>/dev/null || echo $PWD)
 git_root="$_git_root"
-codecov_yml=""
 remote_addr=""
 if [ "$git_root" = "$PWD" ];
 then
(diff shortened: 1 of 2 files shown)


mihails-strasuns-sociomantic:
@leandro-lucarella-sociomantic works again - sadly it seems necessary for docker container to be mounted in repo root as codecov bash script makes heavy use of `git` commands. Adding workarounds for that should be possible but I would prefer to simply get it working for now. Do we expose anything sensitive this way?

leandro-lucarella-sociomantic:
I think we can leave it like this for now, nothing occurs to me that could leak, it's just general distrust given how bad the codecov script it and how badly maintained the repo for it is too.

leandro-lucarella-sociomantic:
So it says we are at 60% line coverage, but the circle's graph doesn't look like it at all... is that correct?

https://codecov.io/gh/sociomantic-tsunami/ocean

leandro-lucarella-sociomantic:
Ahhh, 90% is still barely green (which makes sense), that's why it's all red... :man_facepalming:

mathias-lang-sociomantic:
I just looked at the report for the formatter and it looks like there are a couple of bugs on coverage as well :(

PR #136, opened 2017-05-31
Add utility function to set the length of a string

gkotian:
The function can be used to shorten/lengthen a string as necessary to get it to a desired length.

### DIFF
--- a/src/ocean/text/Util.d
+++ b/src/ocean/text/Util.d
@@ -106,6 +106,9 @@
 
 module ocean.text.Util;
 
+import ocean.core.Array : copy;
+import ocean.core.Enforce;
+import ocean.text.convert.Formatter;
 import ocean.transition;
 
 version (UnitTest) import ocean.core.Test;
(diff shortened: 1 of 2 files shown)


src/ocean/text/Util.d
```diff
@@ -785,6 +788,136 @@ bool isSpace(T) (T c)
...
+/*******************************************************************************
+
+    Shortens/Lengthens a string as necessary to get it to a desired length.
+
+    Note 1:
+        If the string is shortened, an ellipsis (3 dots) will be appended to
+        indicate intentional truncation.
+
+    Note 2:
+        The length specified must be at least 4. If this constraint were absent
```
gavin-norman-sociomantic, line 814: You could use the ellipsis character (`…`) instead of `...`?

src/ocean/text/Util.d
```diff
@@ -785,6 +788,136 @@ bool isSpace(T) (T c)
...
+        and a length of 3 or lower was specified, then the input string would be
+        converted simply to an ellipsis, losing all information about the
+        string.
+
+    Note 3:
+        If the input string needs to be lengthened and padding on both sides
```
gavin-norman-sociomantic, line 820: I dunno if this is really important enough to bother noting.

src/ocean/text/Util.d
```diff
@@ -785,6 +788,136 @@ bool isSpace(T) (T c)
...
+        pad_type = defines how the string should be padded (applicable only if
+            the string needs to be lengthened, defaults to padding being added
+            after the original string)
+
+    Returns:
+        The transformed string with the desired length
+
+*******************************************************************************/
+
+public mstring setLength (cstring orig_str, size_t final_length,
```
gavin-norman-sociomantic, line 838: It's more typical for such functions in ocean to accept a `ref mstring` and mutate it in-place. What was the reason for rendering into a separate buffer?

src/ocean/text/Util.d
```diff
@@ -785,6 +788,136 @@ bool isSpace(T) (T c)
...
+public enum PadType : ubyte
+{
+    PadPre,    // Padding is only applied before the string
+    PadPost,   // Padding is only applied after the string
+    PadAround  // Padding is applied on both sides of the string
+}
+
+/*******************************************************************************
+
+    Shortens/Lengthens a string as necessary to get it to a desired length.
```
gavin-norman-sociomantic, line 807: I wonder about splitting the functionality into:
1. a function to truncate a string to the specified length,
2. a function to pad a string the the specified length

to make it more modular.

Your usage would then look like `pad(truncate(str, desried_length), desired_length)`. (Or you could add a helper/wrapper function `padOrTruncate`?)

src/ocean/text/Util.d
gavin-norman-sociomantic: I find the name `setLength` a little deceptive. It sounds like a simple `array.length = X;` wrapper.

gavin-norman-sociomantic (commented):
Some initial thoughts on the concept / API.

src/ocean/text/Util.d
gkotian: Hmm, from a usage point of view, isn't it more convenient to have a single function which automatically takes care of either padding or truncating as necessary without the programmer having to bother about which one to use?

I could of course have private functions `pad()` & `truncate()`, and call them from the convenience wrapper, but I feel they'll simply be superfluous. The meat of this function really is the lengthening of the original string (because you need to check whether you want to increase the length from the left or the right or on both sides). Also, given the reasonable size of the function as it currently is, splitting it further wouldn't add much value imo.

gavin-norman-sociomantic: I think, in a library, it's always nicer to have smaller building blocks. We can have a helper/wrapper, too, that combines them in the way which you envisage using them.

src/ocean/text/Util.d
gkotian: Adding the `...` is [a tango thing](https://github.com/sociomantic-tsunami/ocean/blob/v2.x.x/src/ocean/text/convert/Layout_tango.d#L1138-L1141). We just use the existing chopping functionality here.

gavin-norman-sociomantic: Oh, interesting. I never knew that exists! :D

src/ocean/text/Util.d
gkotian: It is here because I got a review comment in the past to note this. I should add here that this function is not really new. It's been running in some apps (closed-source unfortunately) since as far back as July 2014. The only major change I've made here is to use the new ocean formatter.

gavin-norman-sociomantic: Fair enough. I can't imagine a situation where anyone would care where the extra space is...?

src/ocean/text/Util.d
gkotian: >It's more typical for such functions in ocean to accept a ref mstring and mutate it in-place. What was the reason for rendering into a separate buffer?

The main reason is that it makes the calling code simpler, as the caller doesn't need to copy the string into a mutable buffer first, and only needs to provide a mutable buffer. The other thing is that if we need to lengthen the string by adding the padding characters _before_ it, then mutating it in place would be a bit unwieldy and you'd probably need to use a temporary static buffer to make this happen.

>I find the name setLength a little deceptive. It sounds like a simple `array.length = X;` wrapper.

Yes, I agree. How about `strSetLength()` or `setStrLength()`? I'm open to other suggestions as well.

src/ocean/text/Util.d
gavin-norman-sociomantic: >I should add here that this function is not really new. It's been running in some apps (closed-source unfortunately) since as far back as July 2014.

By the way, it's typical that when code is moved from a project to a library, its structure is reassessed. The requirements of library code are (obviously) quiet different.

src/ocean/text/Util.d
gavin-norman-sociomantic: >The main reason is that it makes the calling code simpler, as the caller doesn't need to copy the string into a mutable buffer first, and only needs to provide a mutable buffer.

Assuming the user wants to keep the original string around. Though I can understand that in the described use case (formatting something to be displayed in a table) that's likely.

>The other thing is that if we need to lengthen the string by adding the padding characters before it, then mutating it in place would be a bit unwieldy and you'd probably need to use a temporary static buffer to make this happen.

I don't think so. You'd extend the array length, move the beginning to the middle, then set the padding characters at the start and end.

gavin-norman-sociomantic:
Some discussion has led to the idea that this feature might be implemented via the `Formatter` refactoring proposed in https://github.com/sociomantic-tsunami/ocean/pull/98#issuecomment-300120760. Blocked until @mathias-lang-sociomantic has time to assess that.

PR #138, opened 2017-05-31
Fix TypeId with Typedef and improve code coverage

mathias-lang-sociomantic:
This is based on #124 

`TypeId` was not tested, and was in fact not working with `Typedef` in D2.
As it is used by the mapserializer, it caused applications using maps to fail to load their maps, as the hash was different. This should fix it.

### DIFF
--- a/src/ocean/io/serialize/TypeId.d
+++ b/src/ocean/io/serialize/TypeId.d
@@ -9,63 +9,8 @@
     unions and function/delegate parameter lists and using the base type of
     enums and typedefs.
 
-    Example:
-
-    ---
-        struct S
-        {
-            typedef int Spam;
-
-            struct T
-            {
-                enum Eggs : ushort
-                {
-                    Ham = 7
-                }
-
-                Eggs eggs;                              // at offset 0
-
-                char[] str;                             // at offset 8
-            }
-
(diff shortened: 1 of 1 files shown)


mathias-lang-sociomantic:
Tested with the app, it works

relnotes/typedef-no-more.feature.md
```diff
@@ -0,0 +1,6 @@
+* `ocean.core.Traits : isTypedef, StripTypedef`
```
mihails-strasuns-sociomantic: Release notes from other pr, missing rebase?

mihails-strasuns-sociomantic (changes requested):

relnotes/typedef-no-more.feature.md
mathias-lang-sociomantic: > This is based on #124

mihails-strasuns-sociomantic: Sorry, missed that one :)

mathias-lang-sociomantic:
Changed it to not rely on #124
Also changed it to target `v2.7.x`

mihails-strasuns-sociomantic (approved):
Merging now to unblock fixing downstream while long-term approach is discussed

PR #139, opened 2017-06-01
Simple wide integer implementation

mihails-strasuns-sociomantic:
Bare minimum I could test that does anything useful, available for early feedback.

### DIFF
--- a/relnotes/bigint.deprecation.md
+++ b/relnotes/bigint.deprecation.md
@@ -1,0 +1,10 @@
+* `ocean.math.BigInt`
+
+  This module is deprecated in favor of `ocean.math.WideUint` because its
+  implementation (immutable value that gets re-allocated on most operations)
+  makes it unsuitable for Sociomantic projects. At the same time its main
+  feature (unbound maximum integer size) is not of any use to Sociomantic
+  projects.
+
+  If you are interested in actual unbound big integer implementation, please
+  switch to a different third-party library.
--- a/relnotes/wideuint.feature.md
+++ b/relnotes/wideuint.feature.md
@@ -1,0 +1,8 @@
+* `ocean.math.WideUInt`
+
+  New module that implements non-negative integer number of a fixed size. It is
+  intended as `BigInt` replacement that works as pure value type and thus fits
+  Sociomantic memory usage patterns.
+
+  Initial implementation provides bare minimum of functionality. More API
+  boilerplate (like various operator overloads) will be added later on demand.
--- a/src/ocean/math/WideUInt.d
+++ b/src/ocean/math/WideUInt.d
@@ -1,0 +1,625 @@
+/*******************************************************************************
+
+    Struct emulating a large (bigger than ulong) non-negative integer of fixed
+    range (defined via template argument).  `ulong` but still fixed in size
+    (defined via template argument).
+
+    Copyright:
+        Copyright (c) 2017 sociomantic Labs GmbH.
+        All rights reserved.
+
+    License:
+        Boost Software License Version 1.0. See LICENSE_BOOST.txt for details.
+        Alternatively, this file may be distributed under the terms of the Tango
+        3-Clause BSD License (see LICENSE_BSD.txt for details).
+
+*******************************************************************************/
+
+module ocean.math.WideUInt;
+
+import ocean.transition;
+import ocean.core.Test;
+import ocean.core.Enforce;
+
+import core.stdc.math;
+import ocean.math.IEEE : feqrel;
+
+/*******************************************************************************
+
+    Struct emulating a large (bigger than ulong) non-negative integer of fixed
+    range (defined via template argument).  `ulong` but still fixed in size
+    (defined via template argument).
+
+    Such struct is a value type with stable binary layout for a given size which
+    is the primary goal of this implementation. Performance may be eventually
+    improved if necessity arises, right now implementation is intentionally
+    simplistic and naive.
+
+    Internally wide integer is represented by a static array of `uint` values
+    such that concatenating their binary representation together results in long
+    binary sequence representing actual stored number.
+
+    Params:
+        N = amount of uint-size words to use as a backing storage for the
+            number
+
+*******************************************************************************/
+
+public struct WideUInt ( size_t N )
+{
+    static assert (
+        N > 2,
+        "For 'N <= 2' using 'uint' or 'ulong' directly is suggested"
+    );
+
+    /***************************************************************************
+
+       Binary representation of the number can be formed by putting binary
+       representation of each individual word in `payload` side by
+       side starting from the `payload[$-1]`.
+
+       `uint` is chosen as individual word type because multiplying two `uint`
+       words will result in `ulong` at most, making possible to use basic
+       integer arithmetic to implement most of wide integer arithmetic.
+
+    ***************************************************************************/
+
+    private uint[N] payload;
+
+    version (D_Version2)
+    {
+        /***********************************************************************
+
+            Constructor from a regular ulong
+
+        ***********************************************************************/
+
+        mixin("
+        this ( ulong value )
+        {
+            this.assign(value);
+        }
+        ");
+    }
+
+    /***************************************************************************
+
+        Returns:
+            amount of meaningful decimal digits in currently stored value
+
+    ***************************************************************************/
+
+    public size_t decimal_digits ( )
+    {
+        if (this.opEquals(0))
+            return 1;
+
+        WideUInt copy = *this;
+
+        size_t count;
+        while (copy != 0)
+        {
+            copy.divideBy(10);
+            ++count;
+        }
+        return count;
+    }
+
+    unittest
+    {
+        WideUInt num = 0;
+        test!("==")(num.decimal_digits(), 1);
+
+        num = 42;
+        test!("==")(num.decimal_digits(), 2);
+
+        num = ulong.max;
+        test!("==")(num.decimal_digits(), 20);
+
+        num.multiplyBy(100);
+        test!("==")(num.decimal_digits(), 22);
+    }
+
+    /***************************************************************************
+
+        Inefficient allocating string conversion useful for tests and
+        prototyping.
+
+        Returns:
+            string representation
+
+    ***************************************************************************/
+
+    public istring toString ( )
+    {
+        mstring result;
+        this.toString((cstring s) { result ~= s; });
+        return assumeUnique(result);
+    }
+
+    unittest
+    {
+        WideUInt num = 0;
+        test!("==")(num.toString(), "0");
+
+        num = 42;
+        test!("==")(num.toString(), "42");
+
+        num.payload[2] = 1;
+        test!("==")(num.toString(), "18446744073709551658");
+
+        num = 0;
+        num.payload[2] = 1;
+        test!("==")(num.toString(), "18446744073709551616");
+    }
+
+    /***************************************************************************
+
+        Sink based string conversion
+
+        Params:
+            sink = delegate to call with resulting string
+
+    ***************************************************************************/
+
+    public void toString ( void delegate (cstring) sink )
+    {
+        auto n = this.decimal_digits();
+        static mstring buffer;
+        buffer.length = n;
+        enableStomping(buffer);
+
+        WideUInt copy = *this;
+        for (ptrdiff_t idx = n-1; idx >= 0; --idx)
+        {
+            auto remainder = copy.divideBy(10);
+            buffer[idx] = cast(char) ('0' + remainder);
+        }
+
+        sink(buffer);
+    }
+
+    /***************************************************************************
+
+        Enables assignment from a plain integer type
+
+        Params:
+            rhs = value to assign
+
+    ***************************************************************************/
+
+    public void opAssign ( ulong rhs )
+    {
+        this.assign(rhs);
+    }
+
+    unittest
+    {
+        WideUInt i;
+        i = 42;
+        test!("==")(i.payload[0], 42);
+        i = ulong.max;
+        test!("==")(i.payload[0], uint.max);
+        test!("==")(i.payload[1], uint.max);
+    }
+
+    unittest
+    {
+        WideUInt a = 42;
+        WideUInt b = 43;
+        a = b;
+        test!("==")(a, 43);
+    }
+
+    /***************************************************************************
+
+        Enables equality comparison with a plain integer type
+
+        Params:
+            rhs = value to compare to
+
+        Returns:
+            value as defined by `opEquals` spec
+
+    ***************************************************************************/
+
+    public equals_t opEquals ( ulong rhs )
+    {
+        if (this.payload[0] != (rhs & uint.max))
+            return false;
+
+        if (this.payload[1] != (rhs >> 32))
+            return false;
+
+        foreach (ref word; this.payload[2 .. N])
+        {
+            if (word != 0)
+                return false;
+        }
+
+        return true;
+    }
+
+    unittest
+    {
+        WideUInt i;
+
+        i = 42;
+        test!("==")(i, 42);
+
+        i = ulong.max / 2;
+        test!("==")(i, ulong.max / 2);
+        test!("!=")(i, ulong.max);
+
+        i.payload[$-1] = 42;
+        test!("!=")(i, 42);
+
+        i.payload[0] = 0;
+        test!("!=")(i, 42);
+    }
+
+    /***************************************************************************
+
+        Enables equality comparison with WideUInt of the same size
+
+        Params:
+            rhs = value to compare against
+
+        Returns:
+            value as defined by `opEquals` spec
+
+    ***************************************************************************/
+
+    public equals_t opEquals ( WideUInt rhs )
+    {
+        return this.payload[] == rhs.payload[];
+    }
+
+    unittest
+    {
+        WideUInt a, b;
+        test!("==")(a, b);
+        foreach (size_t i, ref elem; a.payload)
+        {
+            a.payload[i] = 1337;
+            b.payload[i] = 1337;
+        }
+        test!("==")(a, b);
+        a.payload[$-1] = 0;
+        test!("!=")(a, b);
+    }
+
+    /***************************************************************************
+
+        Enables ordering comparison with WideUInt of the same size
+
+        Params:
+            rhs = value to compare against
+
+        Returns:
+            value as defined by `opCmp` spec
+
+    ***************************************************************************/
+
+    mixin(genOpCmp("
+    {
+        ptrdiff_t idx = N-1;
+        while (idx > 0)
+        {
+            if (this.payload[idx] != 0 || rhs.payload[idx] != 0)
+                break;
+            --idx;
+        }
+
+        auto a = this.payload[idx];
+        auto b = rhs.payload[idx];
+        return a < b ? -1 : (a > b ? 1 : 0);
+    }
+    "));
+
+    unittest
+    {
+        WideUInt num1, num2;
+        num1.payload[1] = 1;
+        num2.payload[2] = 1;
+        test(num1 < num2);
+        test(num2 > num1);
+    }
+
+    /***************************************************************************
+
+        Increment current value by one fitting in uint
+
+        Params:
+            rhs = value to increment with. Limited to `uint` for now to simplify
+                internal arithmetic.
+
+        Throws:
+            WideUIntRangeException if this WideUInt value range was to be
+            exceeded
+
+    ***************************************************************************/
+
+    public void opAddAssign ( uint rhs )
+    {
+        if (add(this.payload[0], rhs))
+            enforce(.wideint_exception, this.checkAndInc(1));
+    }
+
+    unittest
+    {
+        WideUInt num = 0;
+        for (auto i = 0; i < 1000; ++i)
+            num += (1 << 31);
+        test!("==")(num.payload[0], 0);
+        test!("==")(num.payload[1], 500);
+    }
+
+    /***************************************************************************
+
+        Mutates current number in-place, dividing it by `rhs` and calculating
+        remainder.
+
+        Params:
+            rhs = number to divide by
+
+        Returns:
+            remainder of division
+
+    ***************************************************************************/
+
+    public uint divideBy ( uint rhs )
+    out (remainder)
+    {
+        assert(remainder < rhs);
+    }
+    body
+    {
+        ulong remainder = 0;
+
+        for (ptrdiff_t idx = this.payload.length - 1; idx >= 0; --idx)
+        {
+            remainder = (remainder << 32) + this.payload[idx];
+            ulong result = remainder / rhs;
+            remainder -= rhs * result;
+            assert(result <= uint.max);
+            this.payload[idx] = cast(uint) result;
+        }
+
+        assert(remainder <= uint.max);
+        return cast(uint) remainder;
+    }
+
+    unittest
+    {
+        WideUInt num = 56 * 47;
+        test!("==")(0, num.divideBy(47));
+        test!("==")(num.payload[0], 56);
+
+        num.payload[0] = 1337;
+        num.payload[1] = 100;
+        test!("==")(1337, num.divideBy(1 << 31));
+        test!("==")(num.payload[0], 200);
+    }
+
+    /***************************************************************************
+
+        Mutates this number in place multiplying it by a regular integer
+
+        Params:
+            rhs = value to multiply by
+
+        Throws:
+            WideUIntRangeException if resulting value can't fit in current
+            WideUInt value range
+
+    ***************************************************************************/
+
+    public void multiplyBy ( uint rhs )
+    {
+        ulong overflow = 0;
+
+        for (size_t i = 0; i < this.payload.length; ++i)
+        {
+            overflow += cast(ulong)(this.payload[i]) * rhs;
+            this.payload[i] = cast(uint) (overflow & uint.max);
+            overflow >>= 32;
+        }
+
+        enforce(.wideint_exception, overflow == 0);
+    }
+
+    unittest
+    {
+        WideUInt num = 1_000_000_000;
+        num.multiplyBy(1_000_000_000);
+        test!("==")(num.toString(), "1000000000000000000");
+    }
+
+    /***************************************************************************
+
+        Returns:
+            Double precision floating point value which is nearest to this
+            number.  Uses the same semantics as conversion from ulong to double.
+
+    ***************************************************************************/
+
+    public double toDouble ( )
+    {
+        // find most significant word with non-zero value
+        int idx = this.payload.length-1;
+        while (this.payload[idx] == 0 && idx > 0)
+            --idx;
+
+        // if stored value <= ulong.max, just use plain cast
+        if (idx < 2)
+        {
+            ulong value = this.payload[1];
+            value <<= 32;
+            value |= this.payload[0];
+            return cast(double) value;
+        }
+
+        // else calculate floating point value from 3 most significant words
+        double MAXWORD = pow(2.0, 32.0);
+        return ((this.payload[idx] * MAXWORD + this.payload[idx-1])
+                * MAXWORD + this.payload[idx-2])
+            * ldexp(1.0, (idx-2)*32);
+    }
+
+    unittest
+    {
+        WideUInt num = 123456789;
+        test!("is")(num.toDouble(), 123_456_789.0);
+
+        for (int i = 0; i < 3; ++i)
+            num.multiplyBy(1_000_000_000);
+
+        test(feqrel(num.toDouble(), 123_456_789e+27) == double.mant_dig);
+    }
+
+    /***************************************************************************
+
+        Helper utility to increment word under index `idx`, calling itself
+        recursively for following words if wraparound happens.
+
+        Params:
+            idx = index of word in the internal payload to increment by 1
+
+        Returns:
+            'false' if whole WideUInt has overflown (== error)
+
+    ***************************************************************************/
+
+    private bool checkAndInc ( size_t idx )
+    {
+        if (idx >= N)
+            return false;
+
+        uint after = ++this.payload[idx];
+        if (after == 0)
+            return this.checkAndInc(idx+1);
+
+        return true;
+    }
+
+    unittest
+    {
+        WideUInt num = uint.max;
+        test(num.checkAndInc(0));
+        test(num.payload[0] == 0);
+        test(num.payload[1] == 1);
+
+        num.payload[$-1] = uint.max;
+        test(!num.checkAndInc(num.payload.length - 1));
+    }
+
+    /***************************************************************************
+
+        Trivial helper to initialize to least significant words of the number
+        from ulong.
+
+    ***************************************************************************/
+
+    private void assign ( ulong value )
+    {
+        this.payload[] = 0;
+        this.payload[0] = cast(uint) value;
+        this.payload[1] = cast(uint) (value >> 32);
+    }
+
+    /***************************************************************************
+
+        Helper utility that adds two uint integers checking for wraparound
+
+        Params:
+            a = reference to integer to increment
+            b = number to add
+
+        Returns:
+            'true' if wraparound/overflow has happenned, 'false' otherwise
+
+    ***************************************************************************/
+
+    static private bool add ( ref uint a, uint b )
+    {
+        bool flag;
+
+        version (D_InlineAsm_X86_64)
+        {
+            asm
+            {
+                add   dword ptr [RSI], EDI;
+                setb  flag;
+            }
+        }
+        else
+        {
+            uint result = a + b;
+            flag = result < a;
+            a = result;
+        }
+
+        return flag;
+    }
+
+    unittest
+    {
+        uint a = uint.max;
+        test( add(a, 1));
+        test!("==")(a, 0);
+
+        a = 1000;
+        test(!add(a, 1337));
+        test!("==")(a, 2337UL);
+
+        a = 0;
+        test(!add(a, uint.max));
+        test!("==")(a, uint.max);
+    }
+}
+
+///
+unittest
+{
+    WideUInt!(4) counter;
+    counter = 10000000000000000000U;
+    counter.multiplyBy(10);
+    ++counter;
+    test!("==")(counter.toString(), "100000000000000000001");
+}
+
+unittest
+{
+    WideUInt!(4) num = ulong.max;
+    test!("==")(num.toString(), "18446744073709551615");
+    num.multiplyBy(2);
+    test!("==")(num.toString(), "36893488147419103230");
+    num.divideBy(2);
+    test!("==")(num.toString(), "18446744073709551615");
+
+    num = 0;
+    num.payload[2] = 1; // ulong.max + 1
+    test!("==")(num.toString(), "18446744073709551616");
+}
+
+/*******************************************************************************
+
+    Exception of this class is thrown when WideUInt overflow is about to happen
+
+*******************************************************************************/
+
+public class WideUIntRangeException : Exception
+{
+    import ocean.core.Exception : DefaultExceptionCtor;
+    mixin DefaultExceptionCtor;
+}
+
+// shared between different WideUInt!(N) instances
+private static WideUIntRangeException wideint_exception;
+
+static this ( )
+{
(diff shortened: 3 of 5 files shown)


src/ocean/math/WideInt.d
```diff
@@ -0,0 +1,474 @@
...
+
+        Returns:
+            double precision floating point representation of this number
+
+    ***************************************************************************/
+
+    public double toDouble ( )
+    {
+        // TODO: for widenumbers with value > ulong.max need to calculate
+        // mantiss and exponent manually by counting bitshift
```
mihails-strasuns-sociomantic: @don-clugston-sociomantic do we have anything in `ocean.math` to use for this?

src/ocean/math/WideInt.d
```diff
@@ -0,0 +1,474 @@
...
+
+    private uint[N] payload;
+
+    /***************************************************************************
+
+        Constructor from a regular ulong
+
+    ***************************************************************************/
+
+    this ( ulong value )
```
gavin-norman-sociomantic: Huh? I've never seen this in D1!

src/ocean/math/WideInt.d
```diff
@@ -0,0 +1,474 @@
...
+    {
+        this.payload[0] = cast(uint) value;
+        this.payload[1] = value >> 32;
```
joseph-wakeling-sociomantic: This is implicitly taking advantage of the fact that `payload[i]` will be default-initialized to 0 for i > 1, right?

src/ocean/math/WideInt.d
```diff
@@ -0,0 +1,474 @@
...
+    }
+
+    /***************************************************************************
+
+        Returns:
+            amount of meaninfful decimal digits in currently stored value
```
joseph-wakeling-sociomantic: `meaningful`

src/ocean/math/WideInt.d
```diff
@@ -0,0 +1,474 @@
...
+
+        Params:
+            rhs = value to assign
+
+    ***************************************************************************/
+
+    public void opAssign ( ulong rhs )
+    {
+        this.payload[] = 0;
+        this.payload[0] = cast(uint) (rhs & uint.max);
```
joseph-wakeling-sociomantic: This is different to the constructor's assignment for `this.payload[0]`.  Is one of them wrong?

src/ocean/math/WideInt.d
```diff
@@ -0,0 +1,474 @@
...
+
+/// ditto
+public struct WideInt ( size_t N )
+{
+    static assert (
+        N > 2,
+        "For 'N <= 2' using 'uint' or 'ulong' directly is suggested"
+    );
+
+    private uint[N] payload;
```
gavin-norman-sociomantic: Please document the ordering. It appears to be in ascending order of significance.

src/ocean/math/WideInt.d
```diff
@@ -0,0 +1,474 @@
...
+
+    /***************************************************************************
+
+        Constructor from a regular ulong
+
+    ***************************************************************************/
+
+    this ( ulong value )
+    {
+        this.payload[0] = cast(uint) value;
```
joseph-wakeling-sociomantic: This is different to how `opAssign ( ulong )` handles `this.payload[0]`.  Is one of them wrong?

src/ocean/math/WideInt.d
```diff
@@ -0,0 +1,474 @@
...
+
+        Params:
+            rhs = number to divide by
+
+        Returns:
+            remainder of division
+
+    ***************************************************************************/
+
+    public uint divideBy ( uint rhs )
```
joseph-wakeling-sociomantic: Having the remainder is nice, but is there a reason to not implement `opDivAssign`?

joseph-wakeling-sociomantic: In addition, is there any reason not to implement division by another `WideInt`?

src/ocean/math/WideInt.d
```diff
@@ -0,0 +1,474 @@
...
+        Throws:
+            WideIntRangeException if this WideInt value range was to be
+            exceeded
+
+    ***************************************************************************/
+
+    public void opAddAssign ( uint rhs )
+    {
+        if (add(this.payload[0], rhs))
+            enforce(.wideint_exception, checkAndInc(1));
```
gavin-norman-sociomantic: `this.checkAndInc`

src/ocean/math/WideInt.d
```diff
@@ -0,0 +1,474 @@
...
+        Params:
+            rhs = value to multiply by
+
+        Throws:
+            WideIntRangeException if resulting value can't fit in current
+            WideInt value range
+
+    ***************************************************************************/
+
+    public void multiplyBy ( uint rhs )
```
joseph-wakeling-sociomantic: Is there a reason to not implement this as `opMulAssign`?  And is there a reason to not have an overload taking a `ulong`, or another `WideInt`?

(I appreciate that last one opens a can of worms in terms of, what _size_ of `WideInt`, but we could presumably consider it in terms of `WideInt` of same size to start with, and follow up with support for multiple different sizes.)

src/ocean/math/WideInt.d
```diff
@@ -0,0 +1,474 @@
...
+        }
+
+        enforce(.wideint_exception, overflow == 0);
+    }
+
+    unittest
+    {
+        WideInt num = 1_000_000_000;
+        num.multiplyBy(1_000_000_000);
+        test!("==")(num.toString(), "1000000000000000000");
```
joseph-wakeling-sociomantic: Test the overflow exception, maybe?

src/ocean/math/WideInt.d
```diff
@@ -0,0 +1,474 @@
...
+*******************************************************************************/
+
+module ocean.math.WideInt;
+
+import ocean.transition;
+import ocean.core.Test;
+import ocean.core.Enforce;
+
+/// ditto
+public struct WideInt ( size_t N )
```
joseph-wakeling-sociomantic: With no signedness in place, isn't this strictly speaking a `WideUInt`?

src/ocean/math/WideInt.d
```diff
@@ -0,0 +1,474 @@
+/*******************************************************************************
+
+    Implementation of integer-like struct with arbitrary but fixed size, defined
```
joseph-wakeling-sociomantic: `unsigned integer-like`, going by the current implementation ... ?

src/ocean/math/WideInt.d
```diff
@@ -0,0 +1,474 @@
...
+        ulong overflow = 0;
+
+        for (size_t i = 0; i < this.payload.length; ++i)
+        {
+            overflow += cast(ulong)(this.payload[i]) * rhs;
+            this.payload[i] = cast(uint) (overflow & uint.max);
+            overflow >>= 32;
+        }
+
+        enforce(.wideint_exception, overflow == 0);
```
joseph-wakeling-sociomantic: Given that this is a wide unsigned integer, do we want this exception-on-overflow behaviour, or do we want unsigned integer-style wraparound (for consistency)?

The current behaviour looks simpler to implement, so I'm inclined to go with this personally.

src/ocean/math/WideInt.d
joseph-wakeling-sociomantic: Note, we need this `TODO` to be addressed before this satisfies our planned use-case (although, see above regarding the idea of allowing different word sizes).

src/ocean/math/WideInt.d
joseph-wakeling-sociomantic: Do we want to insist on a `uint` wordsize, or also make that a template parameter of `WideInt` (perhaps defaulting to `uint` but not constrained to it)?

Motivation: assuming we want a 128-bit unsigned integer, wouldn't a `WideInt!(2, ulong)` be a more performant way of doing this (at least on a 64-bit system)?

gavin-norman-sociomantic (commented):
Couple of small noticings after a quick review.

src/ocean/math/WideInt.d
```diff
@@ -0,0 +1,474 @@
...
+            sink = delegate to call with resulting string
+
+    ***************************************************************************/
+
+    public void toString ( void delegate (cstring) sink )
+    {
+        auto n = this.digits();
+        static mstring buffer;
+        buffer.length = n;
+        enableStomping(buffer);
```
joseph-wakeling-sociomantic: This looks like a workaround for `sink` not accepting data one `char` at a time?  Or is that considered sufficiently inefficient that you want to capture the whole number in a single string _before_ passing it on to the sink?

What concerns me here is that this implementation eagerly allocates an entire array for formatting the number, rather than allowing `sink` to decide the allocation strategy.

src/ocean/math/WideInt.d
```diff
@@ -0,0 +1,474 @@
...
+
+        Throws:
+            WideIntRangeException if this WideInt value range was to be
+            exceeded
+
+    ***************************************************************************/
+
+    public void opAddAssign ( uint rhs )
+    {
+        if (add(this.payload[0], rhs))
```
joseph-wakeling-sociomantic: I'm not sure I quite follow how this `add` function is actually working in practice.  Can you explain the magic? :-)

joseph-wakeling-sociomantic (commented):
Looks pretty nice!  I have a few thoughts/remarks/questions ...

src/ocean/math/WideInt.d
mihails-strasuns-sociomantic: You don't use structs a lot :) They do have constructors in D1, only destructors are missing.

src/ocean/math/WideInt.d
mihails-strasuns-sociomantic: Yes, I omitted it for aesthetical purpose. Do you think it is confusing?

src/ocean/math/WideInt.d
joseph-wakeling-sociomantic: Damn, this is news to me too.  I think I'd have used struct constructors more, if I'd known ... :-P

src/ocean/math/WideInt.d
mihails-strasuns-sociomantic: AFAICS `uint` is the only option that makes sense. It allows to do multiplication of words by promoting them to `ulong` and handling overflow using regular arithmetic, with `ulong` you will effectively have to emulate CPU arithemtic unit yourself, hardly an efficient approach. Thus for 64bit `uint` is optimal unless you have hardware-optimized `cent` arithmetic available.

src/ocean/math/WideInt.d
joseph-wakeling-sociomantic: I think it'd be better to be precise (especially as we might want to later add a genuine `WideInt` with signedness included; cf. how `BigUInt` and `BigInt` relate, although in that case the former is really treated more as an implementation detail of the latter.)

src/ocean/math/WideInt.d
mihails-strasuns-sociomantic: No, just different ways to write the same operation. I will sync them.

src/ocean/math/WideInt.d
mihails-strasuns-sociomantic: Yes

src/ocean/math/WideInt.d
mihails-strasuns-sociomantic: No, it is because dividing by 10 will give your individual digits in reverse order compared to one you want to print. Maybe there are some algorithms to get digits starting with most significant one but I haven't bothered to search right now.

src/ocean/math/WideInt.d
joseph-wakeling-sociomantic: I guess it is more convenient to do it this way, but it does occur to me that we've got an interesting mix of big- and little-endian in terms of how this data is structured :-)

joseph-wakeling-sociomantic: OK, makes sense.  Thanks for clarifying.  Might be worth noting in the documentation somewhere?

src/ocean/math/WideInt.d
mihails-strasuns-sociomantic: It adds `rhs` to `this.payload[0]` (with regular unsigned int wraparound semantics) and returns `true` if wraparound has happened - which means higher order digit needs to be incremented.

src/ocean/math/WideInt.d
joseph-wakeling-sociomantic: Might be worth implementing some little internal set-the-payload function to DRY up things here?

src/ocean/math/WideInt.d
joseph-wakeling-sociomantic: Ouch.  OK, fair enough; let's roll with this for now and we can follow up later if we identify a better way to do it.

src/ocean/math/WideInt.d
mihails-strasuns-sociomantic: No, this is simply a most basic functional primitive to build other API on top

As for `WideInt rhs` overloads - do we actually need them now? I assumed it will be used as sort of aggregator value which only works with regular `ulong`/`uint` values individually but may internally grow over their size limit.

My intention is to implement minimal amount of functionality that fits current purpose and leave rest to PRs of interested parties ;)

src/ocean/math/WideInt.d
mihails-strasuns-sociomantic: I am quite confident we want exception in this case. Native integers have wraparound because it is how hardware works and it is fast. In almost all `WideInt` use cases you assume that it is "big enough" for all your values, essentually using it as `BigInt` with fixed size. If such assumption is violated, you want to know about it.

src/ocean/math/WideInt.d
```diff
@@ -0,0 +1,474 @@
...
+
+    ***************************************************************************/
+
+    public equals_t opEquals ( ulong rhs )
+    {
+        if (this.payload[0] != (rhs & uint.max))
+            return false;
+
+        if (this.payload[1] != (rhs >> 32))
+            return false;
```
joseph-wakeling-sociomantic: This is where I miss the opportunity to return a `uint[2]` from some helper function.  We could encapsulate everything needed to assign from `ulong`, and also generate a single point of comparison for `this.payload[0 .. 2]` ... :-P

src/ocean/math/WideInt.d
mihails-strasuns-sociomantic: Acknowledged, I only skipped it for now to see if I can learn from Don wisdom before getting into manual bit tinkering :)

src/ocean/math/WideInt.d
joseph-wakeling-sociomantic: Oh, I see.  I think I just didn't look properly at what `checkAndInc` was doing before, so didn't follow either the consequences or the reason for the `enforce`.  Thanks again for clarifying.

src/ocean/math/WideInt.d
joseph-wakeling-sociomantic: I thoroughly endorse this approach :-)

src/ocean/math/WideInt.d
mihails-strasuns-sociomantic: Hm, actually travis build rejects it. I may have some hacked dmd1 build locally :oops:

src/ocean/math/WideInt.d
joseph-wakeling-sociomantic: As long as it's sufficient for `BidYieldProfile`'s needs (which I imagine you've already looked at in more fine detail than me:-), all good.

joseph-wakeling-sociomantic: As long as it's sufficient for the use-case I described to you (whose fine details I imagine by now you know better than I do:-), all good.

src/ocean/math/WideInt.d
david-eckardt-sociomantic: http://digitalmars.com/d/1.0/struct.html claims D1 structs don't have constructors (I wish they would).

src/ocean/math/WideInt.d
david-eckardt-sociomantic: So the reason to use 32-bit words on 64-bit hardware is to handle carry information?

src/ocean/math/WideInt.d
```diff
@@ -0,0 +1,474 @@
...
+        static mstring buffer;
+        buffer.length = n;
+        enableStomping(buffer);
+
+        WideInt copy = *this;
+        for (ptrdiff_t idx = n-1; idx >= 0; --idx)
+        {
+            auto remainder = copy.divideBy(10);
+            buffer[idx] = cast(char) ('0' + remainder);
+        }
```
david-eckardt-sociomantic: Sorry, but I can't help myself:
```d
foreach_reverse (ref c; buffer)
{
    c = ...
}
```
is just easier to read than
```d
for (ptrdiff_t idx = n-1; idx >= 0; --idx)
{
    buffer[idx] = ...
}
```
and inherently safe from off-by-one and index overflow errors.

src/ocean/math/WideInt.d
david-eckardt-sociomantic: Well, hardware reports wrap-around as well, though the carry and overflow flags, but D swallows this exception.

src/ocean/math/WideInt.d
david-eckardt-sociomantic: There is [`ldexp(x, y) := x * 2^y`](http://pubs.opengroup.org/onlinepubs/9699919799/functions/ldexp.html) which you can probably use like this:
```d
double result = 0;
foreach (i, word; this.payload)
    result += ldexp(word, i * 32);
```

src/ocean/math/WideInt.d
```diff
@@ -0,0 +1,474 @@
...
+        last <<= 32;
+        last += this.payload[0];
+        return cast(double) last;
+    }
+
+    unittest
+    {
+        WideInt num = 1_222_333_444;
+        num.multiplyBy(1_000_000_000);
+        test!("is")(num.toDouble(), 1_222_333_444e+9);
```
david-eckardt-sociomantic: You are sure you won't get in trouble with limited floating-point precision, right? :trollface:

src/ocean/math/WideInt.d
david-eckardt-sociomantic: I also suggest using `real`.

src/ocean/math/WideInt.d
mihails-strasuns-sociomantic: No, it is not about carry (you may notice below that I have addition function that does report carry flag), it is about multiplication. When you multiply `ulong` by `ulong` your result can be up to 128bit large.

src/ocean/math/WideInt.d
mihails-strasuns-sociomantic: My understanding is that `real` is to be avoided as much as possible unless one has a case with no other choice. We should ask FP master @don-clugston-sociomantic

src/ocean/math/WideInt.d
mihails-strasuns-sociomantic: https://github.com/sociomantic-tsunami/ocean/pull/139#discussion_r119687711

src/ocean/math/WideInt.d
```diff
@@ -0,0 +1,474 @@
...
+    }
+
+    /***************************************************************************
+
+        Returns:
+            amount of meaninfful decimal digits in currently stored value
+
+    ***************************************************************************/
+
+    public size_t digits ( )
```
david-eckardt-sociomantic: Better: `decimal_digits`

src/ocean/math/WideInt.d
mihails-strasuns-sociomantic: Airthmetic here macthes internal floating point representation exactly (fits the mantiss and only changes exponent) so I would expect result to be binary identical (thus `is`).

src/ocean/math/WideInt.d
david-eckardt-sociomantic: But that's overflow then, which is still reported by hardware but not D.

mihails-strasuns-sociomantic: AFAIK hardware reports that it has happened, not overflown amount. That is sufficient for addition (where amount always == 1) but not multiplication.

src/ocean/math/WideInt.d
david-eckardt-sociomantic: Do you know that for sure? 1,222,333,444,000,000,000 is divisible by 2048 so this test will still pass if you lose 11 mantissa bits and may fail for 1,222,333,444,000,000,001 or even 1,222,333,444,100,000,000.
Ok, you *are* right, but maybe verify this by something like
```d
static assert(1_222_333_444_000_000_000 > (1 << double.mant_dig));
```

src/ocean/math/WideInt.d
david-eckardt-sociomantic: That's right.

mihails-strasuns-sociomantic: So you do agree that using `uint` for a single word is necessary? :)

src/ocean/math/WideInt.d
gavin-norman-sociomantic: I do use structs a lot! Haha. D2 only feature ;)

src/ocean/math/WideInt.d
hatem-oraby-sociomantic: would make sense to do it as: `size_t digits (base = 10)` to make it more generic?

src/ocean/math/WideInt.d
david-eckardt-sociomantic: Oh yes.

src/ocean/math/WideInt.d
david-eckardt-sociomantic: That is tempting (Tango has such string formatting) but I don't think it's worth it. Except for decimal I've only ever seen and used bases of powers of 2, and they come down to `lzcnt`.

src/ocean/math/WideInt.d
david-eckardt-sociomantic: > Airthmetic here macthes internal floating point representation exactly (fits the mantiss and only changes exponent

Oh no. You specify a _decimal_ exponent here, but IEEE754 floating-point values use _binary_ exponents: The exact FP representation is `mantissa * 2^exponent` (where `mantissa` is a fixed-point fractional number between `1.000...` and `1.111...`).
There are two ways (I'm aware of) of constructing exact FP values:
- at compile time: the [hexadecimal floating-point number literal](http://dlang.org/spec/lex.html#HexFloat)
- at run time: [`ldexp(a, b)`](http://pubs.opengroup.org/onlinepubs/9699919799/functions/ldexp.html) where `a` is an integer less than `2^double.mant_dig`.

mihails-strasuns-sociomantic: You are right, I have been thinking about it as if it has `0b1000_ ...`. Will need to rework this test case completely.

david-eckardt-sociomantic: To clarify: The decimal `{x}.{y}e{z}` notation is for convenience only and unrelated to the actual (binary exponential) representation of the resulting value.

src/ocean/math/WideInt.d
don-clugston-sociomantic: To me, it looks like implementation of positive integers, not unsigned. It doesn't use wraparound semantics.

src/ocean/math/WideInt.d
don-clugston-sociomantic: `BigUint` is a misnomer. Fortunately it is not exposed.  I'd rather not perpetuate my earlier mistake.

src/ocean/math/WideInt.d
don-clugston-sociomantic: AFAIK in general there aren't any good algorithms for getting the digits in left-to-right order. The most efficient algorithms use even powers of 10, eg first divide by 100000000, then by 10000, then by 100. You can use that to give you the digits in a funny order rather than pure right-to-left, but it doesn't reduce the size of your buffer very much.

The best you can do is divide by 10^19 first (since that tightly fits into a `ulong`). Then you can produce the output in chunks of 19 digits, instead of 1 character at a time. If your base type is `uint` then you can only use 10^9.
That reduces the required buffer space by a factor of more than 2. But since it does not eliminate it, it's rather marginal benefit.

src/ocean/math/WideInt.d
don-clugston-sociomantic: If you know that the payload length is always small, you can just use the brute force approach.

```d
double result = 0;
foreach_reverse (word, this.payload )
{
   result *= 2^^32; // this is an exactly representable number, it's an error-free calculation
   result += word;
}
```

If that's too slow you can speed it up by scanning for the first non-zero payload word. There are only 63 bits of mantissa in `real` so additions after the first three have no effect and can be skipped. (We need 3 words, not 2, because we might only get 1 bit from the first payload word).
If `p` is the index of the first non-zero word, then

```d
const double MAXWORD = 2.0^^32; // so we get floating-point multiplies. Could also use real instead`
return (((this.payload[p]*MAXWORD +(this.payload[p-1]))*MAXWORD)+this.payload[p-2])* ldexp(1.0, (p-2)*32);
```

Of course you need to worry about array bounds errors if p is less than 2. But you can set `p = min(p, 2)` because the formula is still valid even if the high payload values are zero.

src/ocean/math/WideInt.d
don-clugston-sociomantic: FYI: Some compilers have intrinsics to support these operations. Here are the gcc ones:
https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html
D should really add something like this as well.

src/ocean/math/WideInt.d
nemanja-boric-sociomantic: This method could use an unittest.

src/ocean/math/WideInt.d
```diff
@@ -0,0 +1,474 @@
...
+    {
+        WideInt copy = *this;
+        size_t count;
+        while (copy != 0)
```
nemanja-boric-sociomantic: This will fail for `WideInt num = 0`; it will report 0 digits, but it has 1 digit.

src/ocean/math/WideInt.d
david-eckardt-sociomantic: It should be mentioned that `this.payload[p]*MAXWORD` is equivalent to `ldexp(this.payload[p], 32)`, but multiplication is faster than `ldexp`.

don-clugston-sociomantic: Clarification, after discussing with David:
Note that `ldexp` is a slow operation, that is why I've only used it once in the code above.

mihails-strasuns-sociomantic:
Updated with documentation fixes, renamed to `WideUInt` and lot of test improvements.
`toDouble` now uses implementation suggested by @don-clugston-sociomantic though it could use test case suggestions too ;)

mihails-strasuns-sociomantic:
== I think this is ready for review/merge as initial prototype at least

src/ocean/math/WideUInt.d
```diff
@@ -0,0 +1,591 @@
...
+
+       `uint` is chosen as individual word type because multiplying two `uint`
+       words will result in `ulong` at most, making possible to use basic
+       integer arithmetic to implement most of wide integer arithmetic.
+
+    ***************************************************************************/
+
+    private uint[N] payload;
+
+    version (D_Version2)
```
gavin-norman-sociomantic, line 69: D2-only behaviour?

mihails-strasuns-sociomantic: Nope, this is required for compatibility between D1 and D2. WideUInt num = 42 will result in opAssign being called in D1 but in D2 it requires constructor.

gavin-norman-sociomantic: Fair enough.

mihails-strasuns-sociomantic:
Ping

joseph-wakeling-sociomantic:
Amusing coincidence:
https://forum.dlang.org/thread/fqkwcjhlnhrdgtikwhue@forum.dlang.org

(There are a couple of different implementations that other people are doing linked in there)

src/ocean/math/WideUInt.d
```diff
@@ -0,0 +1,591 @@
...
+    {
+        bool flag;
+
+        version (D_InlineAsm_X86_64)
+        {
+            // When adding support for other compilers, optimized machine
+            // code output should be checked to see if assembly block can be
+            // reliably replaced with plain D arithmetics
+
+            asm
```
nemanja-boric-sociomantic: I would guess this would be pretty close:

```
static uint add ( uint a, uint b, out bool overflow )
{
    ulong res = cast(ulong)(a) + cast(ulong)(b);
    overflow = res >> 0x20;
    return cast(uint)res;
}
```

david-eckardt-sociomantic: I like using the CPU's built-in overflow detection.

nemanja-boric-sociomantic: Most (all?) of the modern CPUs do have barrel shifter with which this operation costs nothing, and I would guess (though it's just my gut feeling, I have no numbers) it's not slower than using non-portable (not even across the compilers) CPU instructions.

mihails-strasuns-sociomantic: I played a bit with d.godbolt.org and it looks like both gdc and ldc generate suboptimal code for this add: https://godbolt.org/g/07RZO2

However, there is alternative implementation which result in exactly same assembly:

```D

uint add2 ( uint a, uint b, out bool overflow )
{
    uint result = a + b;
    overflow = result < a;
    return result;
}
```
I will use it as a fallback.

mihails-strasuns-sociomantic:
Updated with D code fallback for assembly block

mihails-strasuns-sociomantic:
Ping

relnotes/wideuint.feature.md
```diff
@@ -0,0 +1,8 @@
+* `ocean.math.WideUint`
+
+  New module that implements positive integer number of a fixed size. It is
```
nemanja-boric-sociomantic: `s/positive/non-negative`?

relnotes/bigint.deprecation.md
```diff
@@ -0,0 +1,7 @@
+* `ocean.math.BigInt`
+
+  This module is deprecated in favor of `ocean.math.WideUint` because its
```
nemanja-boric-sociomantic, line 3: Given the open source position on this, this release note seems a little bit too brutal. For example, `WideUint` is not a replacement for a `BigInt`. I'm not sure we can do anything about it, though.

src/ocean/math/WideUInt.d
```diff
@@ -0,0 +1,591 @@
...
+    is the primary goal of this implementation. Performance may be eventually
+    improved if necessity arises, right now implementation is intentionally
+    simplistic and naive.
+
+    Internally wide integer is represented by a static array of `uint` values
+    such that concatenating their binary representation together results in long
+    binary sequence representing actual stored number.
+
+    Copyright:
+        Copyright (c) 2017 Sociomantic Labs GmbH.
```
nemanja-boric-sociomantic: lowercase `s` in `sociomantic` Sorry :-)

src/ocean/math/WideUInt.d
```diff
@@ -0,0 +1,591 @@
...
+
+import ocean.transition;
+import ocean.core.Test;
+import ocean.core.Enforce;
+
+import core.stdc.math;
+import ocean.math.IEEE : feqrel;
+
+/// ditto
+public struct WideUInt ( size_t N )
```
nemanja-boric-sociomantic: Missing documentation for `N`

nemanja-boric-sociomantic (commented):
Generally LGTM. @david-eckardt-sociomantic, @gavin-norman-sociomantic ?

gavin-norman-sociomantic:
You can probably update the commit message from "Start of WideUInt implementation" now ;)

relnotes/wideuint.feature.md
```diff
@@ -0,0 +1,8 @@
+* `ocean.math.WideUint`
```
gavin-norman-sociomantic: `WideUInt` or `WideUint`?

src/ocean/math/WideUInt.d
```diff
@@ -0,0 +1,591 @@
+/*******************************************************************************
+
+    Implementation struct emulating positive number with value range exceeding
```
gavin-norman-sociomantic: This is clearer:
"Struct emulating a large (bigger than ulong) non-negative integer of fixed range (defined via template argument)."

src/ocean/math/WideUInt.d
```diff
@@ -0,0 +1,591 @@
...
+        Params:
+            rhs = number to divide by
+
+        Returns:
+            remainder of division
+
+    ***************************************************************************/
+
+    public uint divideBy ( uint rhs )
+    {
```
gavin-norman-sociomantic: You could add an out contract that `remainder < rhs`.

src/ocean/math/WideUInt.d
```diff
@@ -0,0 +1,591 @@
...
+        test!("==")(a, 43);
+    }
+
+    /***************************************************************************
+
+        Enables equality comparison with a plain integer type
+
+        Params:
+            rhs = value to compare to
+
```
gavin-norman-sociomantic, line 220: Returns:

src/ocean/math/WideUInt.d
```diff
@@ -0,0 +1,591 @@
...
+        {
+            a.payload[i] = 1337;
+            b.payload[i] = 1337;
+        }
+        test!("==")(a, b);
+        a.payload[$-1] = 0;
+        test!("!=")(a, b);
+    }
+
+    /**************************************************************************/
```
gavin-norman-sociomantic: Why no doc?

src/ocean/math/WideUInt.d
```diff
@@ -0,0 +1,591 @@
...
+            ulong result = remainder / rhs;
+            remainder -= rhs * result;
+            assert(result <= uint.max);
+            this.payload[idx] = cast(uint) result;
+        }
+
+        assert(remainder <= uint.max);
+        return cast(uint) remainder;
+    }
+
```
gavin-norman-sociomantic, line 392: `alias opDivAssign divideBy;`?

src/ocean/math/WideUInt.d
```diff
@@ -0,0 +1,591 @@
...
+        for (size_t i = 0; i < this.payload.length; ++i)
+        {
+            overflow += cast(ulong)(this.payload[i]) * rhs;
+            this.payload[i] = cast(uint) (overflow & uint.max);
+            overflow >>= 32;
+        }
+
+        enforce(.wideint_exception, overflow == 0);
+    }
+
```
gavin-norman-sociomantic, line 431: `alias opMulAssign multiplyBy;`?

src/ocean/math/WideUInt.d
```diff
@@ -0,0 +1,591 @@
...
+        WideUInt num = 1_000_000_000;
+        num.multiplyBy(1_000_000_000);
+        test!("==")(num.toString(), "1000000000000000000");
+    }
+
+    /***************************************************************************
+
+        Returns:
+            double precision floating point representation of this number
+
```
gavin-norman-sociomantic: What happens if the integer is larger than can be represented by a double? (The maximum double value is 10^308, which could be stored in a 1024-bit unsigned-integer, so a very small `WideUInt`.)

gavin-norman-sociomantic (commented):
Focusing on doc and API...

src/ocean/math/WideUInt.d
don-clugston-sociomantic: It will be double.infinity. How about:
```
Returns:
double precision floating point value which is nearest to this number.
Uses the same semantics as conversion from ulong to double.
```

src/ocean/math/WideUInt.d
mihails-strasuns-sociomantic: Omitted intentionally for now. I would prefer to add "meat to bones" (various opeator overloads) separately from core impl.

src/ocean/math/WideUInt.d
mihails-strasuns-sociomantic: Good wording!

mihails-strasuns-sociomantic:
Updated.

src/ocean/math/WideUInt.d
gavin-norman-sociomantic: Fair enough.

gavin-norman-sociomantic (approved):

Issue #37, opened 2017-03-13
Allow to reconfigure loggers live from UnixSocket


mathias-lang-sociomantic:
Now that we have UnixSocket support in DaemonApp, it would be nice to get a set of basic commands in (which can be overriden / skipped if the dev wants to).

One of those would be the ability to change logger configurations live, so one can temporarily enable / disable some loggers for debugging purpose.

gavin-norman-sociomantic:
Huh cool idea!

mathias-lang-sociomantic:
Started working on this

PR #140, opened 2017-06-01
UnixSocketExt: Do not pass empty strings as arguments to the handler

mathias-lang-sociomantic:
No test for it in this P.R., but I have an integration test for https://github.com/sociomantic-tsunami/ocean/issues/37 which will ensure there's no regression.

### DIFF
--- a/src/ocean/util/app/ext/UnixSocketExt.d
+++ b/src/ocean/util/app/ext/UnixSocketExt.d
@@ -37,7 +37,7 @@ import ocean.util.app.ext.model.IConfigExtExtension;
 public class UnixSocketExt : IApplicationExtension, IConfigExtExtension
 {
     import ocean.core.Buffer;
-    import ocean.core.array.Transformation: split;
+    import ocean.core.array.Transformation: filter, split;
     import ocean.io.select.EpollSelectDispatcher;
     import ocean.net.server.unix.UnixListener;
     import ocean.util.config.ConfigParser;
@@ -112,7 +112,9 @@ public class UnixSocketExt : IApplicationExtension, IConfigExtExtension
         if (auto handler = command in this.handlers)
         {
             split(args, " ", this.args_buf);
-            (*handler)(this.args_buf[], send_response);
+            scope predicate = (cstring v) { return v.length; };
+            auto arguments = filter(this.args_buf[], predicate, this.args_buf);
+            (*handler)(arguments, send_response);
         }
         else
         {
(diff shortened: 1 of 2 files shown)


nemanja-boric-sociomantic:
I would push this for a patch release even, as the previous behavior on this level smells buggy and unexpected to me

mathias-lang-sociomantic:
Rebased / targeted

nemanja-boric-sociomantic:
I've removed feature label from this PR and changed the milestone.

src/ocean/util/app/ext/UnixSocketExt.d
```diff
@@ -109,10 +109,13 @@ public class UnixSocketExt : IApplicationExtension, IConfigExtExtension
     public void handle ( cstring command, cstring args,
                          void delegate (cstring) send_response )
     {
+        scope pred = (cstring v) { return v.length; };
```
nemanja-boric-sociomantic: There's no way to inline this into the `filter` call, right? (due to `scope` I assume), since it would make the existence of this clearer.

mathias-lang-sociomantic: It's possible, I just found it uglier (same reason why we have `const NAME = "App"`).

I'm not sure I understand what you mean by:

> since it would make the existence of this clearer.

nemanja-boric-sociomantic: I mean that the definition of the `pred` doesn't carry any info of where and for what it will be used (it's not declared even in the same scope). Even moving it to just above the filter will help the reader see that this is the predicate for the `filter`, but I don't see this as a big deal, more a nit-pick, I'm fine to leave it as is, if you insist.

mathias-lang-sociomantic: Oh, indeed. How about:

```diff
diff --git a/src/ocean/util/app/ext/UnixSocketExt.d b/src/ocean/util/app/ext/UnixSocketExt.d
index d0c0bce..3e05c84 100644
--- a/src/ocean/util/app/ext/UnixSocketExt.d
+++ b/src/ocean/util/app/ext/UnixSocketExt.d
@@ -109,12 +109,11 @@ public class UnixSocketExt : IApplicationExtension, IConfigExtExtension
     public void handle ( cstring command, cstring args,
                          void delegate (cstring) send_response )
     {
-        scope pred = (cstring v) { return v.length; };
-
         if (auto handler = command in this.handlers)
         {
             split(args, " ", this.args_buf);
-            auto arguments = filter(this.args_buf[], pred, this.args_buf);
+            scope predicate = (cstring v) { return v.length; };
+            auto arguments = filter(this.args_buf[], predicate, this.args_buf);
             (*handler)(arguments, send_response);
         }
         else
```

?

nemanja-boric-sociomantic: Much better, thanks!

nemanja-boric-sociomantic (approved):

PR #142, opened 2017-06-01
[WIP] (Re)Configurable loggers

### LINKED ISSUE #37: Allow to reconfigure loggers live from UnixSocket

mathias-lang-sociomantic:
```
This commit add a default handler for the UnixSocket extension,
which allow one to reconfigure the loggers without restarting the app.
Connecting to the unix socket (through e.g. `nc`), one can send
commands such as `SetLogger set fully.qualified.name level=[Info,Fatal,...]`.

Potential enhancements being considered:
- Currently we only support 'propagate' and 'level'.
  What other properties would be interesting ? 'file' / 'layout' ?
- Currently we only support the 'set' command.
  One of the original idea is to also support 'reset'.
  But this would require to store the original config within the logger (arguably doesn't scale).
  Another option would be to add the ability to re-parse the config file(s).

Fixes #37
```

Just here so we don't loose track of it. The integration test passes (also tested with ocean v3, found a bug that I'll file separately). Will test with a real app soon :tm:

### DIFF
--- a/src/ocean/util/app/DaemonApp.d
+++ b/src/ocean/util/app/DaemonApp.d
@@ -397,9 +397,11 @@ public abstract class DaemonApp : Application,
         this.config_ext.registerExtension(this.pidlock_ext);
         this.registerExtension(this.pidlock_ext);
 
+        // Unix socket extension and default commands
         this.unix_socket_ext = new UnixSocketExt();
         this.config_ext.registerExtension(this.unix_socket_ext);
         this.registerExtension(this.unix_socket_ext);
+        this.unix_socket_ext.addHandler("SetLogger", &this.log_ext.unixSocketHandler);
 
         if (settings.use_task_ext)
         {
--- a/src/ocean/util/app/ext/LogExt.d
+++ b/src/ocean/util/app/ext/LogExt.d
@@ -17,6 +17,9 @@ module ocean.util.app.ext.LogExt;
 
 import ocean.core.TypeConvert;
 import ocean.io.device.File;
+import ocean.stdc.string;
+import ocean.text.convert.Formatter;
+import ocean.text.util.SplitIterator;
 import ocean.transition;
 import ocean.util.app.Application;
 import ocean.util.app.ext.ConfigExt;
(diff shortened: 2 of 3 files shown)


mathias-lang-sociomantic:
@mihails-strasuns-sociomantic : I assume coverage report doesn't include result of integration tests, only unittests ?

mihails-strasuns-sociomantic:
@mathias-lang-sociomantic correct. I don't know if there is any way/tool to merge coverage reports from multiple runs so if we allowed `-cov` in integration tests, those would overwrite previous ones.

mihails-strasuns-sociomantic:
Release notes, just in case ;)

mathias-lang-sociomantic:
I still need to test this in an application, and see what other people's needs are.
Also I'm starting to consider making it a new logger only feature, which will simplify merging for you. But first we need to deprecate the old loggers.

mathias-lang-sociomantic:
I removed the ability to configure old loggers, since there is now a P.R. to deprecate them, and made this target v3.x.x
This is now based / blocked on https://github.com/sociomantic-tsunami/ocean/pull/173

mathias-lang-sociomantic:
Rebased on v3.x.x

Does someone have an interest in this feature and would like to give me pointer to the needed features ?
CC @daniel-zullo-sociomantic  @joseph-wakeling-sociomantic @andrew-stevenson-sociomantic

andrew-stevenson-sociomantic:
We are a way off being able to make use of it but I can see that being able to set file would be useful if not too difficult. Before upping the level of a particular logger its likely we would want to go to it's own file

mihails-strasuns-sociomantic:
@mathias-lang-sociomantic what is the current state of this? I am for last v2 (v2.10.0) release in about two weeks, wonder if this PR can be ready by that time.

mathias-lang-sociomantic:
>  I am for last v2 (v2.10.0) release in about two weeks, wonder if this PR can be ready by that time.

I don't know if it'll be ready, but this is targetting v3.x.x nowadays (the milestone wasn't changed)

mihails-strasuns-sociomantic:
Oops! Thanks :)

gavin-norman-sociomantic:
Not ready for v3.5.0. Moved to future. Can be moved into a release milestone again when ready.

daniel-zullo:
I think we should support enabling/disabling `console`

> Another option would be to add the ability to re-parse the config file(s).

IMO this is a useful feature to have.

mathias-lang-sociomantic:
I think this will need a bit of love from someone else.

Issue #111, opened 2017-05-14
Issue 16521 will prevent us from upgrading DMD to 2.074


mathias-lang-sociomantic:
https://issues.dlang.org/show_bug.cgi?id=16521

The issue itself is worked around in ocean, but with 2.074 it generates X deprecation messages for each instantiation of `sformat` - where X is the number of arguments. That makes deprecations completely unusable for us starting from 2.074.

PR #144, opened 2017-06-02
Formatter: Do the static foreach over the types, not the value

mathias-lang-sociomantic:
We cannot access the value anyway, so all we need is a static index and an unrolled loop.
This will shut up a wrong DMD deprecation message generated by 2.074.

Fixes sociomantic-tsunami/ocean#111


mathias-lang-sociomantic:
Ping, that one should be a no-brainer.

mihails-strasuns-sociomantic (approved):

PR #146, opened 2017-06-07
Introduce ocean.core.array.Iteration.reverse

mihails-strasuns-sociomantic:


mihails-strasuns-sociomantic:
I hope @david-eckardt-sociomantic will appreciate it :)

relnotes/reverse.feature.md
```diff
@@ -0,0 +1,5 @@
+* `ocean.core.array.Iteration`
+
+  New module containing single function for now - `reverse`, which allows to
+  iterate given slice in reverse order without any allocations (similar to
+  built-in `foreach_reverse`).
```
gavin-norman-sociomantic, line 5: I guess I'm missing some background here (which should probably be mentioned in the release notes / doc). Why not just use `foreach_reverse`?

mihails-strasuns-sociomantic: While not being deprecated in D2, it is considered extremely bad style to use it, and for a good reason - it is a leaky abstraction which wastes a keyword for trivial library utility. We talked about it before in some other issue which I can't currently find :(

gavin-norman-sociomantic: Yeah... I dunno. Whatever you think of `foreach_reverse`, it exists. My feeling is, why not use it? I don't see a great deal of sense in replicating built-in language features in library code, just out of some principle.

mihails-strasuns-sociomantic: Because library version is better? For example, it only works on arrays and does not silently accept `opApply` struct pretending to reverse it when it doesn't. I'd like to introduce a static lint check one day that prohibits usage of `foreach_reverse` completely tbh.

gavin-norman-sociomantic: Ok, if there are actual practical downsides to `foreach_reverse`, that's a different matter.

mihails-strasuns-sociomantic: Most of its downsides is that it doesn't to any type control, happily resorting to plain `foreach` semantics for anything that isn't an array. That makes it rather error-prone in scope of refactorings especially.

gavin-norman-sociomantic: Right, yes that is dodgy (and weird).

david-eckardt-sociomantic: I agree with @gavin-norman-sociomantic.
`foreach_reverse` is clumsy-pragmatic: It works only with a subset of objects one can iterate over, support for `foreach` doesn't imply `foreach_reverse`, and it's probably problematic with ranges. But there are cases where `foreach_reverse` is applicable, most importantly arrays. In these cases it is simply useful. If `foreach_reverse` wouldn't exist I wouldn't request it, but given that it does exist and works perfectly fine for arrays I don't think it makes sense to boycot and work around it.

david-eckardt-sociomantic: > Most of its downsides is that it doesn't to any type control, happily resorting to plain `foreach` semantics for anything that isn't an array. That makes it rather error-prone in scope of refactorings especially.

That may be true but sounds a bit too speculative to me. I've never seen that happen.
Also, when using `foreach_reverse`, plain `foreach` semantics don't apply to _everything_ that is not an array: `foreach_reverse` over a class or struct that implements `opApply` causes a compile error.

mihails-strasuns-sociomantic: I can't understand why you are so eager to use badly designed utility simply because it is built-in into the language. If it was just another library function which worked like that, I doubt anyone would object deprecating it and replacing with a more reliable version. How is language builti-in any better?

david-eckardt-sociomantic: The reason is just pragmatic: It gets a job done, works perfectly where applicable, is simple code and doesn't involve additional code complexity (which library functions do).

mihails-strasuns-sociomantic: "it gets a job done" - same as library replacement
"worse perfectly where applicable" - worse than library replacement
"additional code complexity" - additional language feature to learn, library version is trivially simple

I don't see any pragmatism here, only "it worked good enough for me" argument.

david-eckardt-sociomantic: Yes, it worked good enough for me, and it's less source code. I like that and take the liberty of boldly claiming it works good enough for every reverse array iteration.

PR #147, opened 2017-06-07
Revisit typedef handling approach in ocean

mihails-strasuns-sociomantic:
Attempt to make library code which has to handle old D1 typedef more
concise while clearly distinguishing places where typedef-specific will
need to remain even after dropping D1 from those where such code
branches can immediately be safely removed after dropping D1.

### DIFF
--- a/relnotes/typedef.deprecation.md
+++ b/relnotes/typedef.deprecation.md
@@ -1,0 +1,7 @@
+* `ocean.core.Traits`
+
+  `StripTypedef` and `isTypedef` templates are deprecated with intention to
+  revisit every place where those have been used and evaluate if such custom
+  handling is still needed in D2 builds. New `ocean.transition.isD1Typedef`
+  and `ocean.transition.TypedefBaseType` templates were added to clearly
+  differentiate such places after decision will be made.
--- a/src/ocean/core/Traits.d
+++ b/src/ocean/core/Traits.d
@@ -243,7 +243,10 @@ private bool hasIndirectionsImpl ( T... )()
     }
     else
     {
-        alias StripEnum!(StripTypedef!(Unqual!(T[0]))) Type;
+        static if (isD1Typedef!(T[0]))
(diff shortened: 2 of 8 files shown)


mihails-strasuns-sociomantic:
@mathias-lang-sociomantic I think this should satisfy both your and mine goals as discussed in chat but please double sanity check :)

mathias-lang-sociomantic:
I don't see how that's an enhancement. It's just shifting what's in traits to transition. It makes it even less clear that `isD1Typedef` will only work in D1 (the name is only explicit when you already know what to expect...). And I see it covers neither Formatter or TypeId case.

mihails-strasuns-sociomantic:
But those were exactly my goals:

1) Clearly differentiate `isD1Typedef` case from "is something typedef-like". The way code currently is adjusted, you can completely remove `isD1Typedef` branches in D2 build from all utils with no second thoughts.
2) Move stuff from traits to transition module to make it more clear that this is migration-related utility and is not encouraged for new code.

It doesn't currently touch formatter because it wasn't using any of deprecated symbols. `typeid` module was adjusted though.

mathias-lang-sociomantic:
What I dislike about this:
- Different branches between D1 and D2, and it's even less obvious than before. At least with string mixin and usage of the `typedef` keyword, it gave a strong hint.
- It doesn't help with my issues, where Typedef is IMO too eagerly killed.
  As I mentioned before, I would have liked things to work the way they did before, and that's all. What to do with it can be discussed with other stakeholder, not unilaterally as part of the D2 transition.

The original P.R. was in order to simplify https://github.com/sociomantic-tsunami/ocean/pull/138
Since the current iteration of #138 doesn't use that anymore, if you are happy with it, I'll just kill the initial P.R. and I'll keep on copy pasting those helpers whenever I encounter a place that does need special Typedef.

mihails-strasuns-sociomantic:
Closing for now per lack of agreement on course of action to take.

PR #148, opened 2017-06-07
Use docker image tag when reporting coverage

mihails-strasuns-sociomantic:
Mergin code coverage bit script with matrix build changes has broken the
former because `ocean:latest` is not available anymore.

### DIFF
--- a/ci/travis-after-success.sh
+++ b/ci/travis-after-success.sh
@@ -7,7 +7,7 @@ docker run -ti --rm -v $PWD:/docker -w /docker \
     -e TRAVIS -e TRAVIS_BRANCH -e TRAVIS_COMMIT -e TRAVIS_JOB_NUMBER \
     -e TRAVIS_PULL_REQUEST -e TRAVIS_JOB_ID -e TRAVIS_REPO_SLUG -e TRAVIS_TAG \
     -e TRAVIS_OS_NAME -e TRAVIS_PULL_REQUEST_BRANCH -e TRAVIS_PULL_REQUEST_SHA \
-    ocean bash ci/codecov.sh
+    ocean:$DIST bash ci/codecov.sh
 
 # If this is a tag, convert and push to ocean-d2 repo
 if test -n "$TRAVIS_TAG"
--- a/ci/travis-after-success.sh
+++ b/ci/travis-after-success.sh
@@ -1,6 +1,8 @@
 #!/bin/sh
 set -xe
 
+test "$AFTER_SCRIPT" = "1" || exit 0
+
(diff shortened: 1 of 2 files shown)


mihails-strasuns-sociomantic:
Will auto-merge now to fix coverage reports in other PRs, @leandro-lucarella-sociomantic please poke me if anything is wrong with it

mihails-strasuns-sociomantic:
Hm, @leandro-lucarella-sociomantic do you know if there is any way to run `after_script` part only once instead of per each matrix row?

leandro-lucarella-sociomantic:
> Hm, @leandro-lucarella-sociomantic do you know if there is any way to run after_script part only once instead of per each matrix row?

Nope. And I don't think it's possible, because each matrix build runs in a different VM, and AFAIK you can't pass artefacts from one VM to another (that is not using your own file server or something), so even if you could have a final step that runs only once (I think we could maybe do that with stages), then you wouldn't have the coverage files because they are not in the new VM. I think the only way to hack this is to hace some `if test "$DIST" = xenial; then codecov; fi`.

leandro-lucarella-sociomantic:
There seems to be some spurious network failure in one matrix job, I'm re-running it. Will merge if it's OK.

mihails-strasuns-sociomantic:
How about we just set `AFTER_SCRIPT=1` flag for one of matrix rows and run post-scripts from that only? D2 conversion will have the same issue.

mihails-strasuns-sociomantic:
Updated with such change.

ci/travis-after-success.sh
```diff
@@ -1,6 +1,8 @@
 #!/bin/sh
 set -xe
 
+test "$AFTER_SCRIPT" = "1" || exit 0
```
leandro-lucarella-sociomantic, line 4: Mmm, this will make the script exit with a non 0 status if the test fail, because `-e` will make the shell exit as soon as the `test` command fails, so `exit 0` is never executed. At least in theory, but it seems like tests passed, I don't know why.

leandro-lucarella-sociomantic: Mmm, it seems like it's not how I remembered it. I know there is a catch in using a plain `test` when `set -e` is used, but it seems like it's not this one. It looks like the shell run the `exit 0` (which I guess it makes sense). I think the problem is only with `test ... && something`, because if the first test fails, the script stops right there.

Anyway, nevermind.

mihails-strasuns-sociomantic: I will keep in mind that it is something to double-check, thanks :)

PR #150, opened 2017-06-09
Replace explicit cast when deserializing

daniel-zullo:
It is safer to define the data type to deserialize rather
than casting it to `void[]`, otherwise the compiler wouldn't
complain about it if the data type changes to something that
cannot be implicitly converted to void[].

### DIFF
--- a/src/ocean/util/container/queue/NotifyingQueue.d
+++ b/src/ocean/util/container/queue/NotifyingQueue.d
@@ -108,6 +108,8 @@ import ocean.util.serialize.contiguous.Deserializer;
 
 import ocean.core.Array;
 
+import ocean.transition;
+
 import ocean.util.container.AppendBuffer;
 
 import ocean.io.serialize.StructSerializer;
@@ -617,7 +619,7 @@ class NotifyingQueue ( T ) : NotifyingByteQueue
                 return null;
             }
 
-            auto void_buffer = cast(void[]) data;
+            Const!(void[]) void_buffer = data;
 
             Deserializer.deserialize(void_buffer, cont_buffer);
 
--- a/src/ocean/util/container/queue/NotifyingQueue.d
+++ b/src/ocean/util/container/queue/NotifyingQueue.d
@@ -619,7 +619,7 @@ class NotifyingQueue ( T ) : NotifyingByteQueue
 
             auto void_buffer = cast(void[]) data;
 
-            Deserializer.deserialize!(T)(void_buffer, cont_buffer);
+            Deserializer.deserialize(void_buffer, cont_buffer);
 
             return cont_buffer.ptr;
         }
--- a/src/ocean/util/serialize/contiguous/Deserializer.d
+++ b/src/ocean/util/serialize/contiguous/Deserializer.d
@@ -1131,5 +1131,5 @@ unittest
     auto r1 = Deserializer.deserialize!(Trivial)(buf);
 
     Contiguous!(Trivial) copy_dst;
-    auto r2 = Deserializer.deserialize!(Trivial)(buf, copy_dst);
+    auto r2 = Deserializer.deserialize(buf, copy_dst);
 }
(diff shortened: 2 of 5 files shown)


src/ocean/util/serialize/contiguous/Deserializer.d
```diff
@@ -1127,7 +1127,7 @@ unittest
         int a,b;
     }
 
-    void[] buf = cast(void[]) [ 42, 43 ];
+    Const!(void[]) buf = [ 42, 43 ];
```
mathias-lang-sociomantic: `void[] buf = ...`
This uses the in place deserialization

daniel-zullo: Yup, I'll revert this line

daniel-zullo: It is also used in the second example where deserialization is not in place.

src/ocean/util/container/queue/NotifyingQueue.d
```diff
@@ -617,9 +619,9 @@ class NotifyingQueue ( T ) : NotifyingByteQueue
                 return null;
             }
 
-            auto void_buffer = cast(void[]) data;
+            Const!(void[]) void_buffer = data;
 
-            Deserializer.deserialize!(T)(void_buffer, cont_buffer);
+            Deserializer.deserialize(void_buffer, cont_buffer);
```
mihails-strasuns-sociomantic, line 624: `Deserializer.deserialize!(T)(data, cont_buffer)`?

mihails-strasuns-sociomantic (commented):
Sadly explicit mention of type for `deserialize` was intentional here - D2 compiler complains about overload ambiguity otherwise

daniel-zullo:
> Sadly explicit mention of type for deserialize was intentional here - D2 compiler complains about overload ambiguity otherwise

I understood it is only needed using in place `deserialize()`. If the buffer to store deserialized data is passed then the compiler could deduce the type.
Now  I wonder why all tests passed. (I mean none of the compilers complained about it)

src/ocean/util/container/queue/NotifyingQueue.d
daniel-zullo: D1 will complain about using `Deserializer.deserialize!(T)(data, cont_buffer);` (without explicit cast)
```
./src/ocean/util/container/queue/NotifyingQueue.d(622): Error: template ocean.util.serialize.contiguous.Deserializer.Deserializer.deserialize(S) does not match any function template declaration
./src/ocean/util/serialize/contiguous/Deserializer.d(237): Error: template ocean.util.serialize.contiguous.Deserializer.Deserializer.deserialize cannot deduce template function from argument types !(S)(ubyte[],Contiguous!(S))
./src/ocean/util/container/queue/NotifyingQueue.d(704): Error: template instance ocean.util.container.queue.NotifyingQueue.NotifyingQueue!(S) error instantiating
```

mihails-strasuns-sociomantic:
Hm, maybe my memory is from time when we used 2.066 as base compiler (ocean was one of first ported projects). Great it works!

mihails-strasuns-sociomantic (approved):

src/ocean/util/container/queue/NotifyingQueue.d
david-eckardt-sociomantic: It's a bit silly: in `Deserializer` there are
```d
static Contiguous!(S) deserialize ( S ) ( ref void[] src )
static Contiguous!(S) deserialize ( S ) ( in void[] src, ref Contiguous!(S) dst )
```
Commenting out the first, which doesn't work with `ubyte[] src`, fixes this compile error and works with `ubyte[] src`.

mihails-strasuns-sociomantic: Yeah it seems to be D1 compiler bug.

daniel-zullo:
Cool, thanks!

PR #152, opened 2017-06-11
Re-enable deserialization of const fields

mihails-strasuns-sociomantic:
Trying to upgrade downstream libraries has unconvered several cases
where such deserialization was necessary to serialize data coming from
string literals and it was not possible to easily change it because of
impact on public API.

### DIFF
--- a/src/ocean/util/serialize/contiguous/Deserializer.d
+++ b/src/ocean/util/serialize/contiguous/Deserializer.d
@@ -523,15 +523,9 @@ struct Deserializer
 
         size_t pos = 0;
 
-        foreach (i, Field; typeof (S.tupleof))
+        foreach (i, QualField; typeof (S.tupleof))
         {
-            // Mixin to avoid -v2 warning about immutable being a D2 keyword
-            version (D_Version2)
-                mixin(`
-                static assert (!is(Field == immutable),
-                    "Cannot deserialize field " ~  FieldName!(i, S)
-                    ~ " of struct " ~ S.stringof ~ " because it is immutable");
-                `);
+            alias StripQualifier!(QualField) Field;
 
             static if (is (Field == struct))
             {
@@ -539,20 +533,11 @@ struct Deserializer
 
                 pos += This.countStructArraySizes!(Field)(data[pos .. $], extra_bytes);
             }
-            else static if (is (Field Element : Element[]))
+            else static if (is (Field QualElement : QualElement[]))
             {
-                // Mixin to avoid -v2 warning about immutable being a D2 keyword
-                version (D_Version2)
-                    mixin(`
-                    static assert (!is(Element == immutable),
-                        "Cannot deserialize field " ~ FieldName!(i, S)
-                        ~ " of struct " ~ S.stringof
-                        ~ " because it has an array of immutable elements");
-                    `);
+                alias StripQualifier!(QualElement) Element;
 
-
-
-                static if (is (Element[] == RejectQualifier!(Field)))
+                static if (is (QualElement[] == Field))
                 {
                     // dynamic array
 
@@ -636,7 +621,7 @@ struct Deserializer
 
         This.e.enforceSizeLimit!(T[])(len, This.max_length);
 
-        static if (is (RejectQualifier!(T) U == U[]))
+        static if (is (StripQualifier!(T) U == U[]))
         {
             /*
              * If array is an array of slices (dynamic arrays), obtain a data
@@ -729,7 +714,7 @@ struct Deserializer
         }
         else static if (is (T V : V[]))
         {
-            static if (is (V[] == RejectQualifier!(T)))
+            static if (is (V[] == StripQualifier!(T)))
             {
                 for (size_t i = 0; i < len; i++)
                 {
@@ -858,18 +843,18 @@ struct Deserializer
 
         size_t pos = 0;
 
-        foreach (i, Field; typeof (S.tupleof))
+        foreach (i, QualField; typeof (S.tupleof))
         {
+            alias StripQualifier!(QualField) Field;
+
             static if (is (Field == struct))
             {
                 This.e.enforceInputSize!(S)(data.length, pos);
 
                 pos += This.sliceArrays(*This.getField!(i, Field)(s), data[pos .. $], slices_buffer);
             }
-            else static if (is (RejectQualifier!(Field) Element : Element[]))
+            else static if (is (Field Element : Element[]))
             {
-                // To support const substitute Field with Unqual!(Field) in this
-                // scope.
                 static if (is (Element[] == Field))
                 {
                     This.e.enforceInputSize!(S)(data.length, pos);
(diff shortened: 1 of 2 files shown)


nemanja-boric-sociomantic:
Should this follow with a issue or PR against v4.x.x with disabling this again?

mihails-strasuns-sociomantic:
No, I intend to keep it for now. Requiring downstream library to bend _public_ API for the sake of not hitting this case is too much in my opinion.

david-eckardt-sociomantic:
The test doesn't use a `const` field: A `cstring` field is mutable. In my previous tests, when I implemented deserialising `const` fields and nested arrays, I found two issues with `const` fields:
1. Incorrect detection of arrays i.e. `const` needs to be stripped from `T` before `static if (is(T U == U[]))`. This PR fixes that.
2. IIRC compile errors for
  - `const(T[][])`
  - `const(T[])[]`
  - static array of dynamic arrays `const(T[])[n]` -- same as `const(T[][n])`
  - the aforementioned three cases when replacing `T[]` with a struct `S` that contains dynamic arrays
IIRC the problem is in the `foreach (ref element; array)` loops in `sliceSubArrays` where `typeof(element) is T`, which can be `const`, so you cannot assign to `element` but need to cast `const` away from `T`. I'm not sure if this PR addresses that or if it was already fixed somehow in my previous PR. A thorough test would show it.

In `ocean.util.serialize.contiguous.package_` there are several tests that serialise structs with `const` fields, then deserialise the corresponding unqualified structs. A thorough `const`-deserialiser test should include those `const` structs.

Admittedly getting the nesting right is a bit of a hassle. However, I found that the actual code changes are only a few and rather small -- just finding the spots that need to be adapted involves a lot of `pragma(msg)` discovery using complicated nested `const` array/struct test cases.

Unfortunately my old commits with the `const` deserialiser implementation and tests seem to be gone.

mihails-strasuns-sociomantic:
I will try to find some of those old tests or will add some of my own otherwise ;)

mihails-strasuns-sociomantic:
Updated:
- adjusted one of old tests to do deserialization of branched const arrays
- fixed deserializer for those tests to pass :)

nemanja-boric-sociomantic:
> fixed deserializer for those tests to pass :)

lol :-)

src/ocean/util/serialize/contiguous/Deserializer.d
```diff
@@ -397,15 +397,15 @@ struct Deserializer
     {
         debug (DeserializationTrace)
         {
-            Stdout.formatln("< countStructArraySizes!({})({}) : {}", S.stringof,
+            Stdout.formatln("< countRequiredSize!({})({}) : {}\n", S.stringof,
```
mathias-lang-sociomantic: Why the newline ?

mathias-lang-sociomantic (approved):
LGTM

src/ocean/util/serialize/contiguous/Deserializer.d
mihails-strasuns-sociomantic: No idea. Removed.

src/ocean/util/serialize/contiguous/Deserializer.d
```diff
@@ -397,7 +397,7 @@ struct Deserializer
     {
         debug (DeserializationTrace)
         {
-            Stdout.formatln("< countRequiredSize!({})({}) : {}\n", S.stringof,
+            Stdout.formatln("< countRequiredSize!({})({}) : {}", S.stringof,
```
david-eckardt-sociomantic: This looks like it should be in the "Fix function name in DeserializationTrace " commit.

src/ocean/util/serialize/contiguous/Deserializer.d
```diff
@@ -523,44 +523,29 @@ struct Deserializer
...
+                static if (is (_Element[] == Field))
                 {
                     // dynamic array
 
                     This.e.enforceInputSize!(S)(data.length, pos);
 
                     pos += This.countDynamicArraySize!(Element)(data[pos .. $], extra_bytes);
                 }
-                else static if (hasIndirections!(Element))
+                else static if (hasIndirections!(_Element))
```
david-eckardt-sociomantic: You can keep using `Element` here.

src/ocean/util/serialize/contiguous/Deserializer.d
```diff
@@ -523,44 +523,29 @@ struct Deserializer
...
+            alias StripQualifier!(_Field) Field;
 
             static if (is (Field == struct))
             {
                 This.e.enforceInputSize!(S)(data.length, pos);
 
                 pos += This.countStructArraySizes!(Field)(data[pos .. $], extra_bytes);
             }
-            else static if (is (Field Element : Element[]))
+            else static if (is (Field _Element : _Element[]))
```
david-eckardt-sociomantic: Perhaps use something like `QualElement`? With `_Element` and `Element` it is easy to confuse which is unqualified.

src/ocean/util/serialize/contiguous/Deserializer.d
```diff
@@ -1100,15 +1088,11 @@ struct Deserializer
...
     {
         version (D_Version2)
-            static if (!is(T == typeof(mixin("cast()T.init"))))
-            {
-                pragma(msg, "Warning: ", T,
-                " -- type qualifiers are not supported");
-            }
-        alias T RejectQualifier;
+            mixin("static assert (!is(T == immutable));");
+        alias Unqual!(T) StripQualifier;
```
david-eckardt-sociomantic, line 1095: Given that there are more qualifiers than `const` and `immutable`, would it be better to only accept `const` rather than only reject `immutable`?

mihails-strasuns-sociomantic: `immutable` is the only one where writing something over would be really bad by spec, why bother?

src/ocean/util/serialize/contiguous/Deserializer.d
mihails-strasuns-sociomantic: Done

src/ocean/util/serialize/contiguous/Deserializer.d
mihails-strasuns-sociomantic: Done

src/ocean/util/serialize/contiguous/Deserializer.d
mihails-strasuns-sociomantic: Done

src/ocean/util/serialize/contiguous/Deserializer.d
david-eckardt-sociomantic: Yeah, seems so. I thought `shared` might somehow cause trouble, but I'm not sure.

mihails-strasuns-sociomantic: No one still even knows what _exactly_ `shared` means ;)

PR #154, opened 2017-06-14
Deprecate Logger's LayoutChainsaw

mathias-lang-sociomantic:
Should be a no op for everyone.

### DIFF
--- a/relnotes/chainsaw.deprecation.md
+++ b/relnotes/chainsaw.deprecation.md
@@ -1,0 +1,9 @@
+* `ocean.util.log.LayoutChainsaw`
+
+  This module, and the eponymous class it contains, were using private details
+  of the old logger implementation, and were not used in any project,
+  and thus have been deprecated.
+  As a result, it is not configurable by the `LogExt` anymore.
+  If really needed, this self contained module can trivially be copied,
+  and configuration is possible via the `makeLayout` parameter in
+  `configure{Old,New}Loggers`.
--- a/src/ocean/util/log/LayoutChainsaw.d
+++ b/src/ocean/util/log/LayoutChainsaw.d
@@ -15,7 +15,7 @@
 
 *******************************************************************************/
 
-module ocean.util.log.LayoutChainsaw;
+deprecated module ocean.util.log.LayoutChainsaw;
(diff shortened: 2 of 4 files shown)


mathias-lang-sociomantic:
The only person that could remotely consider using it is @leandro-lucarella-sociomantic . So is it fine for you ? :P

src/ocean/util/log/LayoutChainsaw.d
```diff
@@ -31,6 +31,7 @@ import ocean.util.log.Log;
 
 *******************************************************************************/
 
+deprecated("This class is believe to be unused and isn't maintained")
```
leandro-lucarella-sociomantic: I think this is a poor deprecation message. There is no solution offered to potential users. Now that Ocean is public we can't assume nobody is using. I would just say that the class is unmaintained and as a solution I would suggest users to copy the code in their projects if they need to keep using it.

leandro-lucarella-sociomantic (changes requested):
Oh, no! XML log format!

Just a small comment on the deprecation message / migration instructions. Other than that (and that Travis is red), LGTM.

mathias-lang-sociomantic:
Fixed according to @leandro-lucarella-sociomantic 's feedback.

relnotes/chainsaw.deprecation.md
```diff
@@ -0,0 +1,8 @@
+* `ocean.util.log.LayoutChainsaw`
+
+  This module, and the eponymous class it contains, are not used in any project,
+  and thus were deprecated.
+  As a result, it is not configurable by the `LogExt` anymore.
+  If really needed, this self contained module can trivially be copied,
+  and configuration is possible via the `makeLayout` parameter in
+  `configure{Old,New}Loggers`.
```
leandro-lucarella-sociomantic: Please say here too why was it deprecated. As a user that was using the module, reading the release notes it will be my first question. "But why?!?!?" :)

mathias-lang-sociomantic:
Changed the release note to mention the reason of deprecation.

mihails-strasuns-sociomantic (approved):

PR #156, opened 2017-06-15
Decode the base64 with the missing padding

### LINKED ISSUE #155: Base64 decoding of unpadded data leads to incomplete decoded data.
Extending the following unit test to check for the decoding of padded data shows that the decoding is broken.

https://github.com/sociomantic-tsunami/ocean/blob/v2.x.x/src/ocean/util/encode/Base64.d#L514
```
// Encode without padding
unittest
{
    istring str = "Hello, how are you today?";
    Const!(ubyte)[] payload = cast(Const!(ubyte)[]) str;
    cstring result = encode(payload, false);
    test!("==")(result, "SGVsbG8sIGhvdyBhcmUgeW91IHRvZGF5Pw");

  
    auto decoded = cast(cstring)decode(result);
    test!("==")(decoded, str);
}
```
results in 
```
./src/ocean/util/encode/Base64.d:521: test error: expression 'Hello, how are you today == Hello, how are you today?' evaluates to false
```

nemanja-boric-sociomantic:
This makes base64 work when there not all padding bytes needed are in
the input string. It decodes as many whole non-padded quartets as it
can, and then it tries to decode the last remaining bytes into a pair,
or a single character.

Fixes #155

### DIFF
--- a/src/ocean/util/encode/Base64.d
+++ b/src/ocean/util/encode/Base64.d
@@ -317,6 +317,12 @@ body
     unique. This allows the use of decode tables with alternatives for certain
     characters.
 
+    Since the padding is required only between multiple base64 strings (for the
+    decoder not to interpret the start of the next string as the end of the
+    previous one), data may or may not contain the trailing padding, and it will
+    still be decoded correctly.
+
+
     Params:
       Table = The decode table to use. Variadic template parameter is used to
               allow passing it as an expression in D1, but a single `ubyte[256]`
@@ -357,6 +363,11 @@ body
     Note that the entries in the provided decode table are not required to be
     unique. This allows the use of decode tables with alternatives for certain
     characters.
+
+    Since the padding is required only between multiple base64 strings (for the
+    decoder not to interpret the start of the next string as the end of the
+    previous one), data may or may not contain the trailing padding, and it will
+    still be decoded correctly.
 
     Params:
       Table = The decode table to use. Variadic template parameter is used to
@@ -422,8 +433,8 @@ body
 
         if (padCount > 2)
             throw new Exception("Improperly terminated base64 string. Base64 pad character (=) found where there shouldn't be one.");
-        if (padCount == 0)
-            paddedPos = 0;
+        if (padCount < 2)
+            paddedPos = cast(ubyte)((data.length-padCount) % 4);
 
         auto nonPadded = data[0..($ - paddedPos)];
         foreach(piece; nonPadded)
@@ -446,12 +457,17 @@ body
         if (paddedPos)
         {
             auto padded = data[($ - paddedPos) .. $];
-            foreach(char piece; padded)
+            assert (padded.length <= 4);
+
+            // Fill the missing paddings if needed
+            base64Quad[padded.length .. $] = BASE64_PAD;
+
+            foreach(i, char piece; padded)
             {
                 ubyte next = table[piece];
                 if (next || piece == 'A')
                     *quadPtr++ = next;
-                if (quadPtr is endPtr)
+                if (i == padded.length - 1)
                 {
                     *rtnPt++ = cast(ubyte) (((base64Quad[0] << 2) | (base64Quad[1]) >> 4));
                     if (base64Quad[2] != BASE64_PAD)
@@ -507,6 +523,35 @@ unittest
         test!("==")(result, payload);
         result = decode("SGVsbG8sIGhvdyBhcmUgeW91IHRvZGF5Pw==");
         test!("==")(result, payload);
+        result = decode("SGVsbG8sIGhvdyBhcmUgeW91IHRvZGF5Pw=");
+        test!("==")(result, payload);
+        result = decode("SGVsbG8sIGhvdyBhcmUgeW91IHRvZGF5Pw");
+        test!("==")(result, payload);
+    }
+
+    // single padding test
+    {
+        // try the string with the single `=` padding, and without it
+        auto result = cast(Const!(ubyte)[])"any carnal pleasure.";
+        test!("==")(result, decode("YW55IGNhcm5hbCBwbGVhc3VyZS4="));
+        test!("==")(result, decode("YW55IGNhcm5hbCBwbGVhc3VyZS4"));
+    }
+
+    // short decode test, double-= padding
+    {
+        auto result = cast(Const!(ubyte)[])"M";
+
+        test!("==")(result, decode("TQ"));
+        test!("==")(result, decode("TQ="));
+        test!("==")(result, decode("TQ=="));
+    }
+    // ditto, with a single = required for padding
+    {
+        auto result = cast(Const!(ubyte)[])"Ma";
+
+        test!("==")(result, decode("TWE"));
+        test!("==")(result, decode("TWE="));
+        test!("!=")(result, decode(""));
     }
 }
 
@@ -517,6 +562,9 @@ unittest
     Const!(ubyte)[] payload = cast(Const!(ubyte)[]) str;
     cstring result = encode(payload, false);
     test!("==")(result, "SGVsbG8sIGhvdyBhcmUgeW91IHRvZGF5Pw");
+
+    auto decoded = cast(cstring)decode(result);
+    test!("==")(decoded, str);
 }
 
 


nemanja-boric-sociomantic:
@scott-gibson-sociomantic could you try this in your use case?

scott-gibson-sociomantic:
Confirming that changes are working with my use case.

src/ocean/util/encode/Base64.d
```diff
@@ -503,10 +508,38 @@ unittest
     // decodeTest
     {
         ubyte[1024] decoded;
-        ubyte[] result = decode("SGVsbG8sIGhvdyBhcmUgeW91IHRvZGF5Pw==", decoded);
+        Const!(ubyte)[] result = decode("SGVsbG8sIGhvdyBhcmUgeW91IHRvZGF5Pw==", decoded);
```
mathias-lang-sociomantic: Why this change ?

src/ocean/util/encode/Base64.d
```diff
@@ -446,12 +446,17 @@ body
...
+            // Fill the missing paddings if needed
+            base64Quad[padded.length .. $] = BASE64_PAD;
+
+            foreach(i, char piece; padded)
             {
                 ubyte next = table[piece];
                 if (next || piece == 'A')
                     *quadPtr++ = next;
-                if (quadPtr is endPtr)
+                if (i == padded.length - 1)
```
mathias-lang-sociomantic, line 470: Can `padded.length` ever be `0` ?

nemanja-boric-sociomantic: No, that would require that `paddedPos` is zero, so it would not enter this branch. Additionally, the `foreach` wouldn't iterate over `padded`.

src/ocean/util/encode/Base64.d
nemanja-boric-sociomantic: because of the reassigning it here: https://github.com/sociomantic-tsunami/ocean/pull/156/files/7ba30bf6d7030f4220b2eeaeaf443ad324444622#diff-183a2b58e80d01cce23dd21bd51c86ceR521

Not terribly important, though, I can try to please you - I was just fixing the d2 failure.

src/ocean/util/encode/Base64.d
mathias-lang-sociomantic: > Additionally, the foreach wouldn't iterate over padded.

:man_facepalming:

src/ocean/util/encode/Base64.d
mathias-lang-sociomantic: I'd just like to have a test that ensure we're getting a mutable buffer. Would it be possible to use another variable instead ? Actually it could be its own unittest block

src/ocean/util/encode/Base64.d
```diff
@@ -503,10 +508,38 @@ unittest
...
         result = decode("SGVsbG8sIGhvdyBhcmUgeW91IHRvZGF5Pw==");
         test!("==")(result, payload);
+        result = decode("SGVsbG8sIGhvdyBhcmUgeW91IHRvZGF5Pw=");
+        test!("==")(result, payload);
+        result = decode("SGVsbG8sIGhvdyBhcmUgeW91IHRvZGF5Pw");
+        test!("==")(result, payload);
+
+        // try the string with the single `=` padding, and without it
+        result = cast(Const!(ubyte)[])"any carnal pleasure.";
+        test!("==")(cast(char[])result, cast(char[])decode("YW55IGNhcm5hbCBwbGVhc3VyZS4="));
```
mathias-lang-sociomantic: No casting of `istring` to `mstring` please :(

src/ocean/util/encode/Base64.d
nemanja-boric-sociomantic: Sure, good point.

src/ocean/util/encode/Base64.d
nemanja-boric-sociomantic: Yup, it was the debuging aid, removed.

nemanja-boric-sociomantic:
Updated with the comments from @mathias-lang-sociomantic

mathias-lang-sociomantic (approved):

src/ocean/util/encode/Base64.d
```diff
@@ -507,6 +512,37 @@ unittest
...
+    {
+        // try the string with the single `=` padding, and without it
+        auto result = cast(Const!(ubyte)[])"any carnal pleasure.";
+        test!("==")(result, decode("YW55IGNhcm5hbCBwbGVhc3VyZS4="));
+        test!("==")(result, decode("YW55IGNhcm5hbCBwbGVhc3VyZS4"));
+    }
+
+    // short decode test
+    {
+        ubyte[1024] decoded;
```
david-eckardt-sociomantic: - Unused: Should it be passed to `decode`?
- What is the reason for this length, could it as well be `result.length`?

david-eckardt-sociomantic:
It would be good to document this extension. Just supporting missing padding bytes makes me wonder:
- What is the point of padding bytes if they can be omitted?
- Can they always be omitted, or are there cases where some or all padding bytes are required?
- Given a Base-64 string that normally has padding bytes, can removing one or all padding bytes change the decoder output?

src/ocean/util/encode/Base64.d
nemanja-boric-sociomantic: Ah, its a copy/paste error. I will remove. `decode` has overflow which will allocate its own buffer, I think that's fine for this unittest.

nemanja-boric-sociomantic:
> What is the point of padding bytes if they can be omitted?

The reason where you need padding bytes is when you need concatenate multiple base64 strings. If there are no padding bytes at the end of the first base64 string, decoder will chop off missing bytes from the second string, and interpret them as a part of the first string.

> Can they always be omitted, or are there cases where some or all padding bytes are required?

See above.


> Given a Base-64 string that normally has padding bytes, can removing one or all padding bytes change the decoder output?

In a case of a single base64 string (not series of it), removing one or all padding bytes can't change the decoder output. This is the case here, as our decoder doesn't support decoding multiple byte64 sequences. I'm investigating if that's a bug.

In any case, this is very good point, I will improve the documentation.

nemanja-boric-sociomantic:
> I'm investigating if that's a bug.

This seems to be done by design. I'll submit another PR based on v2.x.x which adds support for this.

nemanja-boric-sociomantic:
Updated with the removed `ubyte` array and some more documentation on the padding.

nemanja-boric-sociomantic:
Feature for decoding multiple strings: #158

src/ocean/util/encode/Base64.d
```diff
@@ -317,6 +317,12 @@ body
     unique. This allows the use of decode tables with alternatives for certain
     characters.
 
+    Since the padding is required only between multiple base64 strings (for the
```
gavin-norman-sociomantic, line 320: Unmatched (

src/ocean/util/encode/Base64.d
```diff
@@ -358,6 +364,11 @@ body
...
```
gavin-norman-sociomantic, line 367: Unmatched (

src/ocean/util/encode/Base64.d
```diff
@@ -507,6 +523,35 @@ unittest
...
+    // short decode test
+    {
+        auto result = cast(Const!(ubyte)[])"M";
+
+        test!("==")(result, decode("TQ"));
+        test!("==")(result, decode("TQ="));
+        test!("==")(result, decode("TQ=="));
+    }
+    // ditto
+    {
```
gavin-norman-sociomantic: What's this test checking that the preceding one ("short decode test") doesn't?

nemanja-boric-sociomantic: It tests the case where you get less than four characters, and +/- padding. I'll set the documentation straight.

nemanja-boric-sociomantic: ah, actually it was single `=` and double `==` required for padding.

nemanja-boric-sociomantic:
Updated.

gavin-norman-sociomantic (approved):

PR #157, opened 2017-06-15
Allow users to choose HTTP response buffer size

andrew-stevenson-sociomantic:
In order to minimise GC activity it is useful to pre-allocate buffers to a size that is a reasonable estimate of the largest size that will be seen. This change allows clients to do just that without changing the behaviour of existing clients.

### DIFF
--- a/src/ocean/net/Uri.d
+++ b/src/ocean/net/Uri.d
@@ -225,12 +225,17 @@ class Uri : UriView
 
                 Create an empty Uri
 
+                Params:
+                    initial_buffer_size = the initial amount of memory to
+                        allocate to hold the URI-decoded URI. Will be extended
+                        if necessary.
+
         ***********************************************************************/
 
-        this ()
+        this ( uint initial_buffer_size = 512 )
         {
                 port_ = InvalidPort;
-                decoded.expand (512);
+                decoded.expand (initial_buffer_size);
         }
 
         /***********************************************************************
--- a/src/ocean/net/http/HttpRequest.d
+++ b/src/ocean/net/http/HttpRequest.d
@@ -167,17 +167,22 @@ class HttpRequest : HttpHeader
                                        header fields as well
             msg_body_prealloc_length = expected message body length for
                                        preallocation;
+            uri_prealloc_length      = the initial amount of memory the
+                                       contained Uri object will be asked to
+                                       pre-allocate to hold the URI-decoded
+                                       URI of a request.
 
      **************************************************************************/
 
-    public this ( bool add_entity_headers = false, size_t msg_body_prealloc_length = 0 )
+    public this ( bool add_entity_headers = false, size_t msg_body_prealloc_length = 0,
+            uint uri_prealloc_length = 512)
     {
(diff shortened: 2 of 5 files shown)


mihails-strasuns-sociomantic (approved):
Looks good on its own though I wonder if others want to pursue this path as until now "reuse & grow automatically" seemed to be a most dominant pattern.

andrew-stevenson-sociomantic:
> "reuse & grow automatically"

that still works and we still use it but if you know there will be a grow within the first seconds...

gavin-norman-sociomantic:
One question about wording: you use the word "clients", but we don't have an HTTP client in ocean. What are you referring to exactly (this seems to all be HTTP server-side code)?

nemanja-boric-sociomantic:
I assumed the users of this facility.

gavin-norman-sociomantic:
>I assumed the users of this facility.

I guess so too, but using the word "client" in the context of HTTP is kind of loaded ;)

andrew-stevenson-sociomantic:
Yes Nemanja is correct, a client is a user of provided feature. Is it not the ordinary English meaning of client?

gavin-norman-sociomantic:
It's not the usual library terminology. We usually refer to "users" of library code. (Plus the confusion over HTTP clients.)

gavin-norman-sociomantic (commented):
Change "clients" to "users".

andrew-stevenson-sociomantic:
I have made the change - though I would have thought any possible ambiguity with client must also exist with user.

nemanja-boric-sociomantic:
Oh, the last thing. This should be based on `v2.x.x` and not on `v3.x.x`. You can change the target branch above, but the rebase could also be due.

nemanja-boric-sociomantic (approved):

gavin-norman-sociomantic:
>I have made the change - though I would have thought any possible ambiguity with client must also exist with user.

Who's ever heard of an "HTTP user"? ;)

src/ocean/net/Uri.d
```diff
@@ -227,10 +227,10 @@ class Uri : UriView
 
         ***********************************************************************/
 
-        this ()
+        this ( uint initial_buffer_size = 512 )
```
gavin-norman-sociomantic: Method params doc should be updated.

src/ocean/net/http/HttpRequest.d
```diff
@@ -170,14 +170,15 @@ class HttpRequest : HttpHeader
 
      **************************************************************************/
 
-    public this ( bool add_entity_headers = false, size_t msg_body_prealloc_length = 0 )
+    public this ( bool add_entity_headers = false, size_t msg_body_prealloc_length = 0,
```
gavin-norman-sociomantic, line 177: Method params doc should be updated.

gavin-norman-sociomantic (commented):
Tiny doc comment.

andrew-stevenson-sociomantic:
Updated comments as per Gavin's request. Fixed some of the existing wording whilst I was there.

gavin-norman-sociomantic (approved):
Thanks!

PR #158, opened 2017-06-16
Allow Base64's decode to decode multiple concatenated strings.

nemanja-boric-sociomantic:
`decode` method is now able to decode multiple concatenated base64 strings.
Note that for this to work, padding must be correctly applied on every string,
but the last one.

### DIFF
--- a/relnotes/base64_multiple.feature.md
+++ b/relnotes/base64_multiple.feature.md
@@ -1,0 +1,5 @@
+* `ocean.util.encode.Base64`
+
+  `decode` method is now able to decode multiple concatenated base64 strings.
+  Note that for this to work, padding must be correctly applied on every string,
+  but the last one.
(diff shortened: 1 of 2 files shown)


nemanja-boric-sociomantic:
Blocked on #156

mathias-lang-sociomantic:
I'm not a fan of this feature TBH. The decoder becomes more and more complex. Could we just make the core decode a range, and build the support for this feature on top ?

nemanja-boric-sociomantic:
Yeah, I think that should be possible, and sounds like a good idea. I would go with the slightly different API, then, proposing something as `decodeMulti`, which would chunk the input base64 string on the separators (if any) and call `decode` for each chunk.

nemanja-boric-sociomantic:
I'm not sure new API is needed though. Will try to make it work.

mihails-strasuns-sociomantic:
Probably not blocked anymore?

nemanja-boric-sociomantic:
You're right, I'll try to tidy up this today.

nemanja-boric-sociomantic:
This is fine to stay closed for the time being. I don't know if this is useful to anybody, anyway.

PR #163, opened 2017-06-19
Allow interactive session with UnixSocketExt/UnixConnectionHandler

### LINKED ISSUE #52: UnixSocket: Let handlers interact with users
Working on https://github.com/sociomantic-tsunami/ocean/issues/37 I found myself wishing a way to interact with the user. Currently commands are one-shot: One send a command, an action is performed and one or multiple messages are sent back.

However, certain actions are complex or dangerous, in which case the handler might want to interact with the user. For example, here's an example of using `adduser`:

```
% sudo adduser test
Adding user `test' ...
Adding new group `test' (1002) ...
Adding new user `test' (1001) with group `test' ...
The home directory `/home/test' already exists.  Not copying from `/etc/skel'.
Enter new UNIX password: 
Retype new UNIX password: 
passwd: password updated successfully
Changing the user information for test
Enter the new value, or press ENTER for the default
	Full Name []: Test
	Room Number []: 42   
	Work Phone []: 
	Home Phone []: 
	Other []: 
Is the information correct? [Y/n] Y
```

In the case of reconfiguring loggers it's even more sensible, because one might just want to change the logger level. However, the config currently allow to:
- Change the level
- Toggle write to console
- Change the file the logger is writing to
- Change the console layout
- Change the file layout
- Propagate settings

While it is possible to handle this via a variety of method, a mean to interact with the user would be welcome. However, AFAICS it would require some redesign of `UnixSocketHandler`.

nemanja-boric-sociomantic:
UnixListener now accepts a handlers with an additional delegate,
get_response, which can be used inside a handler for an interactive
session.

Fixes #52

### DIFF
--- a/test/unixlistener/main.d
+++ b/test/unixlistener/main.d
@@ -16,6 +16,14 @@
 import ocean.transition;
 
 import core.thread;
+import core.stdc.errno;
+import core.stdc.stdio;
+import core.sys.posix.stdlib;
+import core.sys.posix.semaphore;
+import core.sys.posix.sys.mman;
+import core.sys.posix.unistd;
+import core.sys.posix.sys.wait;
+
 import ocean.core.Test;
 import ocean.core.Enforce;
 import ocean.core.Time;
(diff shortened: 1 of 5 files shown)


test/unixlistener/main.d
```diff
@@ -67,46 +141,105 @@ void writeToClient( cstring socket_path, cstring command )
...
-    writeToClient(local_address, "increment 2\n");
-    writeToClient(local_address, "increment 1\n");
-    writeToClient(local_address, "shutdown\n");
-
-    // Spin the server
-    epoll.eventLoop();
-
-    // This will be reached only if "shutdown" command was successful.
-    test!("==")(expected_value, 3);
+    auto local_address = "/tmp/ocean_unixsocket_test.socket";
```
leandro-lucarella-sociomantic: I suggest using a temporary directory created via `mkdtemp` just in case someone want's to run the same test concurrently. I know it sounds strange, but when building matrixes in automated builds it might happen.

leandro-lucarella-sociomantic (commented):
I'm not crazy about the `get_response`, `send_response` names. For me it makes the dialogue confusing.

`get_response` actually sends a reponse and then wait for a reply, right? I guess `send_response` only sends a response?

If so, what about leaving `send_response` as the only way to send a response and then have a separate call called something like `wait_reply`? So instead of `get_response("are you there?", response);` this would be `send_response("are you there?"); wait_reply(reply);`

test/unixlistener/main.d
nemanja-boric-sociomantic: Yeah, good point. I've just adapted the existing test.

nemanja-boric-sociomantic:
> get_response actually sends a reponse and then wait for a reply, right? I guess send_response only sends a response?

Yes, this is the left over from the previous system, where one was only able to send the response to the user from the command handler.

> > If so, what about leaving send_response as the only way to send a response and then have a separate call called something like wait_reply? So instead of get_response("are you there?", response) this would be send_response("are you there?"); wait_reply(reply);

Sounds good :+1:

nemanja-boric-sociomantic:
Huh, I forgot about this one. Will address the naming suggestions @leandro-lucarella-sociomantic

nemanja-boric-sociomantic:
Fixed bugs in `UnixConnectionHandler.getNextLine` and hardened the test case. Will address the API now.

nemanja-boric-sociomantic:
Updated with the API change suggested by @leandro-lucarella-sociomantic .

nemanja-boric-sociomantic:
Updated the test case a little bit. This is again ready for the review.

nemanja-boric-sociomantic:
Fixed D2 failure.

leandro-lucarella-sociomantic (approved):
LGTM

mihails-strasuns-sociomantic:
Please rebase

nemanja-boric-sociomantic:
Rebased.

mihails-strasuns-sociomantic (approved):

PR #169, opened 2017-06-26
Implement method to parse floating point strings to int

### LINKED ISSUE #99: Imported symbols cannot be exposed through aggregate
Thanks to the new import rules in 2.071.0+, the design of `LibFeatures` doesn't work.

The definition is currently:
```D
module ocean.LibFeatures;

static immutable has_features_2_7 = true;
static immutable has_features_2_6 = true;
static immutable has_features_2_5 = true;
static immutable has_features_2_4 = true;
static immutable has_features_2_3 = true;
static immutable has_features_2_1 = true;
static immutable has_features_2_0 = true;
```

And the usage is:
```D
private bool hasFeaturesFromFunc ( istring libname, ulong major, ulong minor ) ()
{
    struct Library
    {
        mixin("import " ~ libname ~ ".LibFeatures;");
    }

   // Generates `return is(typeof(Library.has_features_2_0));` for example
    mixin("return is(typeof(Library.has_features_"
        ~ ctfe_i2a(major) ~ "_" ~ ctfe_i2a(minor) ~ "));");
}
```

However this doesn't work with new import rules, which make import scope not accessible from outside of an aggregate.

mathias-baumann-sociomantic:


mathias-baumann-sociomantic:
@reporting-team fyi

mathias-baumann-sociomantic:
hmm that 'mention' didn't work

andrej-mitrovic-sociomantic:
I don't follow, how is https://github.com/sociomantic-tsunami/ocean/issues/99 connected to this?

src/ocean/text/convert/Integer.d
```diff
@@ -30,6 +30,133 @@ import ocean.transition;
...
+        num_parts[1].length = decimal_points;
+
+    // Fill with zeros if too short
+    if (num_parts[1].length < decimal_points)
+    {
+        zeros_suffix_buf[0 .. num_parts[1].length] = num_parts[1];
+        num_parts[1] = zeros_suffix_buf;
+    }
+
+    ulong tryParse ( cstring str )
```
andrej-mitrovic-sociomantic: I think nested methods which access the context pointer allocate in D2. Might be a good idea to mark this one static and pass `exception` explicitly? CC @mihails-strasuns-sociomantic

andrej-mitrovic-sociomantic: At least, I saw a warning once when I ran `-vgc` with D2 and it complained about these things..

mathias-baumann-sociomantic: What you say makes sense

mihails-strasuns-sociomantic: They don't allocate on its own but will allocate as soon as they get passed as arguments or returned from a function (usual delegate -> closure rule).

src/ocean/text/convert/Integer.d
```diff
@@ -30,6 +30,133 @@ import ocean.transition;
...
+        T              = type of the integer
+        value          = floating point number string to parse
+        exception      = optional, a reusable exception instance to throw
+
+    Returns:
+        the resulting integer number
+
+*******************************************************************************/
+
+public T floatStringToInt ( size_t decimal_points, T = ulong )
```
mihails-strasuns-sociomantic: Why decimal points is template argument? You don't need it at CT and this creates template bloat.

mihails-strasuns-sociomantic (commented):
Isn't it essentially `cast(long) toFloat(str) * 10^^decimal_points` ?

src/ocean/text/convert/Integer.d
mathias-baumann-sociomantic: Well, I figured it is a compile time optimization because that way the `pow()` line is also done at compile time and only the multiplication with the price needs to be done at runtime

mihails-strasuns-sociomantic: I'd say this is a premature optimization with a high chance of actually degrading performance in practice (because of extra bloat, unless everything is force inlined). Unless you had specific case where it helped to get required performance?

mathias-baumann-sociomantic: Well, it is used like that in reef/coral already where this function is called about 7k times/second..

mihails-strasuns-sociomantic: So what? Unless you benchmarked both cases explicitly and/or investiagted assembly it doesn't mean anything but the fact that whatever performance it has is sufficient.

mihails-strasuns-sociomantic: To rephrase: my opinion is that one should use straighforward function definitions unless importance of non-common optimizations is exlicitly proved. Default option should always be the boring one.

mathias-baumann-sociomantic:
> Isn't it essentially cast(long) toFloat(str) * 10^^decimal_points

"essentially" it is. Practically it's avoid the float part as float can be imprecise and dangerous when becoming values like inf or nan.

src/ocean/text/convert/Integer.d
mathias-baumann-sociomantic: There is also the stack allocated array of zeros...

mathias-baumann-sociomantic: A benchmark showed the following for processing 19m entries:
Runtime: 2.34 - 2.38 seconds
Compiletime: 2.26 - 2.28 seconds

mathias-baumann-sociomantic:
I pushed a less templated version of the method

mathias-baumann-sociomantic:
I pushed again to cover that last line in the covtest ^^

mathias-baumann-sociomantic:
Ping

src/ocean/text/convert/Integer.d
```diff
@@ -30,6 +30,138 @@ import ocean.transition;
...
+    "0.01"  ->   10  (decimal_points = 3)
+
+    Any characters longer than the requested amount of decimal points will be
+    cut off:
+
+    "1.2345"  ->  123 (decimal_points = 2)
+    "10.2030" -> 1020 (decimal_points = 2)
+
+    Params:
+        decimal_points = amount of decimal points to consider
```
gavin-norman-sociomantic: If `decimal_points` is 0, this will just discard the fractional part, right? If that's correct, how about moving that argument to second place and giving it a default value of 0?
```D
public T floatStringToInt ( T = ulong ) ( cstring value, size_t decimal_points = 0, ReusableException exception = null )
```
The reason for this suggestion is that the main point of the function (as the name says) is to convert a float string to an integer. The intuitively obvious way of doing this is to discard the fractional part. The logic for (essentially) moving the decimal point before doing the conversion seems like an optional extra.

src/ocean/text/convert/Integer.d
```diff
@@ -30,6 +30,138 @@ import ocean.transition;
...
 import ocean.core.Traits;
 
+import ocean.core.array.Search : find;
+
+import ocean.util.ReusableException;
+
+import ocean.math.Math;
+
+
+/*******************************************************************************
```
gavin-norman-sociomantic: I don't think this function should go first in the module. It's less commonly useful than `toInteger`, say.

src/ocean/text/convert/Integer.d
```diff
@@ -30,6 +30,138 @@ import ocean.transition;
...
+        T              = type of the integer
+        value          = floating point number string to parse
+        exception      = optional, a reusable exception instance to throw
+
+    Returns:
+        the resulting integer number
+
+*******************************************************************************/
+
+public T floatStringToInt ( T = ulong ) ( size_t decimal_points, cstring value, ReusableException exception = null )
```
gavin-norman-sociomantic: I notice the API of this function is at variance with all the other functions in this module -- it returns the converted value and throws on error. All the other functions return a bool indicating parsing success and have an out value to return the converted integer. It would seem cleaner to follow the same pattern with this new function.

gavin-norman-sociomantic (commented):
This gives me a sort of deja vu feeling ;)

gavin-norman-sociomantic:
Oh, needs release notes, too.

mathias-baumann-sociomantic:
Those are good points. Now fixed

src/ocean/text/convert/Integer.d
```diff
@@ -170,6 +176,140 @@ public bool toUlong ( T ) ( T[] digits, out ulong value, uint radix = 0 )
...
+}
+
+version ( UnitTest )
+{
+    import ocean.core.Test;
+}
+
+unittest
+{
+    auto exc = new ReusableException;
```
mihails-strasuns-sociomantic: not used anymore

mathias-baumann-sociomantic:
Removed unused code line

src/ocean/text/convert/Integer.d
```diff
@@ -30,6 +30,12 @@ import ocean.transition;
 
 import ocean.core.Traits;
 
+import ocean.core.array.Search : find;
+
+import ocean.util.ReusableException;
```
mihails-strasuns-sociomantic: Unused import :)

mathias-baumann-sociomantic: ffs

mathias-baumann-sociomantic:
Removed unused import

gavin-norman-sociomantic (approved):

mihails-strasuns-sociomantic (approved):

PR #172, opened 2017-07-03
Reduce dependency on old loggers, move some functionality in ILogger

mathias-lang-sociomantic:
I am currently working on deprecating the old loggers. I have a set of patches (almost) working on top of v3.x.x, but some of the changes can be done in v2.x.x, so I cherry-picked them here.
The 3rd patch (Trivial: Fix ILogger code style and comments) might create conflicts because the logging primitives were removed in v3.x.x. This should be trivial to fix, but feel free to ping me if it isn't.
Additionally, since those patches were authored agaisnt v3.x.x, the documentation mention they are removed. I'm not sure it's worth the trouble of writing "correct" documentation for v2.x.x and then changing it in v3.x.x...

### DIFF
--- a/src/ocean/util/log/model/ILogger.d
+++ b/src/ocean/util/log/model/ILogger.d
@@ -1,17 +1,23 @@
 /*******************************************************************************
 
-        Copyright:
-            Copyright (c) 2004 Kris Bell.
-            Some parts copyright (c) 2009-2016 Sociomantic Labs GmbH.
-            All rights reserved.
+    Base interface for Loggers implementation
 
-        License:
-            Tango Dual License: 3-Clause BSD License / Academic Free License v3.0.
-            See LICENSE_TANGO.txt for details.
+    Note:
+        The formatting primitives (error, info, warn...) are not part of the
+        interface anymore, as they can be templated functions.
 
-        Version: Initial release: May 2004
+    Copyright:
+        Copyright (c) 2004 Kris Bell.
+        Some parts copyright (c) 2009-2016 Sociomantic Labs GmbH.
+        All rights reserved.
 
-        Authors: Kris
+    License:
+        Tango Dual License: 3-Clause BSD License / Academic Free License v3.0.
+        See LICENSE_TANGO.txt for details.
+
+    Version: Initial release: May 2004
+
+    Authors: Kris
 
 *******************************************************************************/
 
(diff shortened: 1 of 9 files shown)


src/ocean/util/log/model/ILogger.d
```diff
@@ -96,52 +205,94 @@ interface ILogger
...
+    test!("==")(ILogger.convert("Info"), ILogger.Level.Info);
+    test!("==")(ILogger.convert("INFO"), ILogger.Level.Info);
+    test!("==")(ILogger.convert("FATAL"), ILogger.Level.Fatal);
+    // Use the default value
+    test!("==")(ILogger.convert("Info!"), ILogger.Level.Trace);
+    test!("==")(ILogger.convert("Baguette", ILogger.Level.Warn),
+                ILogger.Level.Warn);
+    // The first entry in the array
+    test!("==")(ILogger.convert("trace", ILogger.Level.Error),
+                ILogger.Level.Trace);
```
mathias-lang-sociomantic, line 296: Those unittests are already present in the new logger (as well as the method pasted here).

mihails-strasuns-sociomantic (approved):

src/ocean/util/log/model/ILogger.d
```diff
@@ -96,52 +129,69 @@ interface ILogger
 
         void fatal (cstring fmt, ...);
 
-        /***********************************************************************
+    /***************************************************************************
```
mathias-baumann-sociomantic, line 207: Shouldn't the functions above here not also be moved one indent to the left?

src/ocean/util/log/model/ILogger.d
```diff
@@ -1,60 +1,93 @@
...
 
-        ***********************************************************************/
+        Params:
+            `Level` to test for, defaults to `Level.Fatal`.
+
+    ***************************************************************************/
+
+    public bool enabled (Level level = Level.Fatal);
+
+
```
mathias-baumann-sociomantic: You introduced two blank lines here where as the rest of the functions have a blank-line distance of one.4
Please keep it consistent :)

mathias-baumann-sociomantic (changes requested):

src/ocean/util/log/model/ILogger.d
mathias-baumann-sociomantic: I suppose those are the functions that would be removed later? In that case you can keep it like this I guess :)-

mathias-lang-sociomantic: Correct

src/ocean/util/log/model/ILogger.d
mathias-lang-sociomantic: Removed extra line

mathias-baumann-sociomantic (approved):

PR #173, opened 2017-07-04
Deprecate old loggers

mathias-lang-sociomantic:
This set of patch deprecate the old loggers in favor of the new ones.
It means we can remove the old loggers in v4.x.x, and are one big step closer to remove Tango's Layout.

### DIFF
--- a/src/ocean/util/app/ext/LogExt.d
+++ b/src/ocean/util/app/ext/LogExt.d
@@ -33,6 +33,7 @@ import ocean.util.app.ext.ConfigExt;
 
 import ocean.util.app.ext.ReopenableFilesExt;
 
+import ocean.util.config.ConfigFiller;
 import ocean.util.config.ConfigParser;
 import LogUtil = ocean.util.log.Config;
 import ConfigFiller = ocean.util.config.ConfigFiller;
@@ -54,8 +55,6 @@ import ocean.util.log.Appender;
 
 class LogExt : IConfigExtExtension
 {
-    import ocean.util.config.ConfigFiller;
-
     /***************************************************************************
 
         Adds a list of extensions (this.extensions) and methods to handle them.
@@ -171,7 +170,7 @@ class LogExt : IConfigExtExtension
 
         enable_loose_parsing(conf_ext.loose_config_parsing);
 
-        LogUtil.configureOldLoggers(log_config, log_meta_config, &appender,
+        configureOldLoggersLogExtHack(log_config, log_meta_config, &appender,
             this.layout_maker, this.use_insert_appender);
 
         LogUtil.configureNewLoggers(log_config, log_meta_config, &appender,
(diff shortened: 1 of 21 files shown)


src/ocean/util/app/ext/LogExt.d
```diff
@@ -249,3 +248,17 @@ class LogExt : IConfigExtExtension
...
+    ClassIterator!(LogUtil.Config, ConfigParser) config,
+    LogUtil.MetaConfig m_config,
+    AppenderExternD file_appender, MakeLayoutExternD makeLayout,
+    bool use_insert_appender = false);
+/// Hack because DMD1 applies extern(C) to the delegate type as well...
+private alias extern(D) Appender delegate (istring, Appender.Layout)
+    AppenderExternD;
+/// Ditto
+private alias extern(D) Appender.Layout delegate (cstring)
+    MakeLayoutExternD;
```
mathias-lang-sociomantic, line 265: This is a huge hack. Maybe we should raise a DIP to allow for deprecated blocks. @mihails-strasuns-sociomantic what do you think ?

mihails-strasuns-sociomantic: Interesting idea but I am not sure about its merit. It is not possible to enable it in library code without also enabling it in app code and latter opens up abuse possibility that kind of kills `deprecated` guarantees.

Your hack is pretty cool in that regard actually - it is convoluted enough to only be used if really commited/desperate but at the same time completely solves the problem from the practical PoV.

mathias-lang-sociomantic: Actually there's an even simpler way: https://issues.dlang.org/show_bug.cgi?id=7322
Should it be used instead ? :)

mihails-strasuns-sociomantic: Simple one is a bug which may get fixed one day, you current one is an honest abuse of language feature ;)

mihails-strasuns-sociomantic: Only one request - can you add a comment here with sub-string "deprecated" so that it pops up on regular cleanup when releasing 4.0.0 ?

mathias-lang-sociomantic:
```diff
diff --git a/src/ocean/util/app/ext/LogExt.d b/src/ocean/util/app/ext/LogExt.d
index 1359757..2ca9049 100644
--- a/src/ocean/util/app/ext/LogExt.d
+++ b/src/ocean/util/app/ext/LogExt.d
@@ -251,6 +251,7 @@ class LogExt : IConfigExtExtension
 
 
 /// Hack to allow to call a deprecated function without deprecations
+/// TODO: Remove for v4.0.0 release with other deprecated things
 private extern extern(C) void configureOldLoggersLogExtHack (
     ClassIterator!(LogUtil.Config, ConfigParser) config,
     LogUtil.MetaConfig m_config,
```

mihails-strasuns-sociomantic (approved):

mathias-lang-sociomantic:
Ping

PR #177, opened 2017-07-06
Add missing const and inout annotations to gnu.string

mathias-lang-sociomantic:
Reported by @don-clugston-sociomantic


mihails-strasuns-sociomantic (approved):
Nice!

PR #178, opened 2017-07-07
Various enhancements of `ocean.meta`

mihails-strasuns-sociomantic:
Last week I have been trying to implement system to generate string
representation of types (and `DeepUnqual` based on it). Unfortunately, this
initial goal has proved to be challenging to the point of probably impossible :(
However I did plenty of modifications/enhancements in `ocean.meta` while hacking
it and those are pretty good to go.


mihails-strasuns-sociomantic:
@mathias-lang-sociomantic you may like this version of typedef utilities more as it is pretty close to your original request.

src/ocean/meta/codegen/Identifier.d
```diff
@@ -0,0 +1,110 @@
...
+    assert (identifier!(ClassName.fieldName) == "fieldName");
+
+    static assert (identifier!(funcName) == "funcName");
+    assert (identifier!(funcName) == "funcName");
+
+    static assert (identifier!(funcNameArgs) == "funcNameArgs");
+    assert (identifier!(funcNameArgs) == "funcNameArgs");
+}
+
+private istring funcIdentifierHack(alias Sym)()
```
gavin-norman-sociomantic: Normal doc would be nice, even if it's private.

gavin-norman-sociomantic (commented):
Looks fine. This PR adds quite a few duplicates (or near duplicates) of functionality in `ocean.core.Traits`. What's the intention / plan there?

mihails-strasuns-sociomantic:
> What's the intention / plan there?

I am going to eventually deprecate `ocean.core.Traits` in favor of new one but only after trying to switch ocean internals to it and ensuring it works with no nasty surprises. That means several minor releases at least.

gavin-norman-sociomantic:
>I am going to eventually deprecate ocean.core.Traits in favor of new one but only after trying to switch ocean internals to it and ensuring it works with no nasty surprises. That means several minor releases at least.

Ok. Could you maybe create an issue for this and make a note in the header for `ocean.core.Traits`?

src/ocean/meta/types/ReduceType.d
```diff
@@ -188,6 +161,10 @@ private struct ReduceTypeImpl ( Reducer )
         {
             // do nothing, already processed
         }
+        else static if (isTypedef!(T))
```
mathias-lang-sociomantic, line 164: Huge :+1:

relnotes/meta-istring.feature.md
```diff
@@ -1,6 +0,0 @@
-* `ocean.meta.types.Qualifiers`
-
-  Moved `istring`/`cstring`/`mstring` definitions from `ocean.transition`
-  module, keeping them available in original module via non-deprecated alias.
-  This is done as part of splitting `ocean.transition` into smaller parts and
-  thus simplifying dependency chains.
```
mathias-lang-sociomantic: It's a bit weird to add files after files only to remove them in a later commit... Can't the text just be appended to meta.feature.md ?

src/ocean/meta/AliasSeq.d
```diff
@@ -0,0 +1,20 @@
...
+    License:
+        Boost Software License Version 1.0. See LICENSE_BOOST.txt for details.
+        Alternatively, this file may be distributed under the terms of the Tango
+        3-Clause BSD License (see LICENSE_BSD.txt for details).
+
+*******************************************************************************/
+
+module ocean.meta.AliasSeq;
+
+template AliasSeq (T...)
```
mathias-lang-sociomantic, line 17: `/// Ditto` + access qualifier. And `AliasSeq` is extremely ugly, but I will not get into that debate ;)

src/ocean/meta/traits/Basic.d
```diff
@@ -338,6 +338,26 @@ unittest
...
+*******************************************************************************/
+
+public template staticArrayLength ( T : U[Dim], U, size_t Dim )
+{
+    const staticArrayLength = Dim;
+}
+
+unittest
+{
+    static assert (staticArrayLength!(int[][5]) == 5);
```
mathias-lang-sociomantic, line 355: Could you add UT with other types, for completeness ?

src/ocean/meta/codegen/CTFE.d
```diff
@@ -0,0 +1,94 @@
...
+/*******************************************************************************
+
+    Converts integer to string
+
+    Params:
+        i = integer to convert
+
+*******************************************************************************/
+
+istring toString ( long i )
```
mathias-lang-sociomantic, line 28: We already have https://github.com/sociomantic-tsunami/ocean/blob/bcd88397f484ca4821ceed942b518f05c2e2274f/src/ocean/core/Traits_tango.d#L707-L805 . If you want to move it there, those should be deprecated. Also, why only the signed version is provided ?

src/ocean/meta/codegen/CTFE.d
```diff
@@ -0,0 +1,94 @@
...
+
+unittest
+{
+    static assert (toString(42) == "42");
+    static assert (toString(-42L) == "-42");
+    static assert (toString(31) == "31");
+    static assert (toString(-31) == "-31");
+    static assert (toString(14u) == "14");
+    static assert (toString(14L) == "14");
+    static assert (toString(14UL )== "14");
```
mathias-lang-sociomantic: `long.max`, `long.min`, `ulong.max` test cases are needed.

src/ocean/meta/codegen/CTFE.d
```diff
@@ -0,0 +1,94 @@
...
+    if (idx == name.length)
+        return name;
+    else
+        return name[idx + 1 .. $];
+}
+
+unittest
+{
+    static assert (stripQualifiedPrefix("aa.bb") == "bb");
+    static assert (stripQualifiedPrefix("aabb") == "aabb");
```
mathias-lang-sociomantic: Needing a test case with more than one dot (and a test case with a trailing dot).

src/ocean/meta/codegen/Identifier.d
```diff
@@ -0,0 +1,110 @@
...
+
+/*******************************************************************************
+
+    Template to get the name of the ith member of a struct / class.
+
+    Used over plain `identifier` when iterating over aggregate fields with
+    `.tupleof` as D1 compiler refuses to pass such field as template alias
+    parameter.
+
+    Template parameter:
```
mathias-lang-sociomantic: `Params`

src/ocean/meta/codegen/Identifier.d
```diff
@@ -0,0 +1,110 @@
...
+        i = index of member to get
+        T = type of compound to get member name from
+
+    Evaluates to:
```
mathias-lang-sociomantic: Returns

src/ocean/meta/AliasSeq.d
mihails-strasuns-sociomantic: I agree that it is ugly but decided that matching Phobos naming pattern here (and not clashing with `std.typecons.Tuple`) will help us a lot with the learning curve in the long term.

src/ocean/meta/codegen/Identifier.d
```diff
@@ -0,0 +1,110 @@
...
+unittest
+{
+    static struct TestStruct
+    {
+        int a;
+        double b;
+    }
+
+    static assert (fieldIdentifier!(TestStruct, 0) == "a");
+    static assert (fieldIdentifier!(TestStruct, 1) == "b");
```
mathias-lang-sociomantic, line 120: Can you add a test with:

```D
struct Foo { AliasSeq!(int, double, Object) fields; }
```

src/ocean/meta/codegen/Identifier.d
```diff
@@ -0,0 +1,110 @@
...
+}
+
+/*******************************************************************************
+
+    Returns "name" (identifier) of a given symbol as string
+
+    When used with aggregate type field, will return only name of field itself,
+    without any qualification.
+
+    Template_Params:
```
mathias-lang-sociomantic: `Params`

src/ocean/meta/codegen/CTFE.d
```diff
@@ -0,0 +1,94 @@
...
+    Params:
+        i = integer to convert
+
+*******************************************************************************/
+
+istring toString ( long i )
+{
+    istring digit="0123456789";
+
+    if (i==0)
```
mathias-lang-sociomantic: This function really needs spacing

src/ocean/meta/traits/Function.d
```diff
@@ -0,0 +1,50 @@
...
+
+unittest
+{
+    void foo(int x, ref double y, char[] z) {}
+    alias ParametersOf!(typeof(foo)) Params;
+
+    static assert (Params.length == 3);
+    static assert (is(Params[0] == int));
+    static assert (is(Params[1] == double));
+    static assert (is(Params[2] == char[]));
```
mathias-lang-sociomantic, line 50: Can you add a test with a delegate ? What about callable types ?

mathias-lang-sociomantic (changes requested):

src/ocean/meta/codegen/CTFE.d
mihails-strasuns-sociomantic: It is cleaned up copy of those. As I have answered previously to Gavin, intention it to deprecate all symbols in `ocean.core.Traits` in one go when replacement package is sufficiently battle tested, that will be both more reliable and easier migration-wise (one big sed script to run once).

For now only signed long version was copied as I expect other case be of unnecessary complexity in CTFE (How often do you format numbers > long.max in that context?). May copy second overload later.

mathias-lang-sociomantic: Well the `ulong` version is not really hard to implement given your `long` implementation...

src/ocean/meta/traits/Function.d
mihails-strasuns-sociomantic: Delegate tests added, generic callable types are not supported right now.

mihails-strasuns-sociomantic:
Updated, should be all @mathias-lang-sociomantic comments addressed.

src/ocean/meta/traits/Function.d
```diff
@@ -0,0 +1,73 @@
...
+
+*******************************************************************************/
+
+template ParametersOf ( Fn )
+{
+    static if (is( Fn Params == function ))
+        alias AliasSeq!(Params) ParametersOf;
+    else static if (is( Fn Params == delegate ))
+        alias AliasSeq!(ParametersOf!(Params)) ParametersOf;
+    else static if (is( Fn Params == Params* ))
```
david-eckardt-sociomantic: Suggest replacing `Params` with a name that describes what it really is:
- `is( Fn Params == delegate )`: not `Params` but `DgFuncptr`
- `is( Fn Params == Params* )`: not `Params` but `Function`

src/ocean/meta/codegen/CTFE.d
david-eckardt-sociomantic: 0 is needed, too.

src/ocean/meta/codegen/CTFE.d
david-eckardt-sociomantic: Please add a `ulong` version. Even if it's not used very often it's bad to not have it.

src/ocean/meta/codegen/CTFE.d
david-eckardt-sociomantic:
```d
foreach_reverse (i, c; name)
    if (c == '.')
        return name[i + 1 .. $];

return name;
```
:trollface:

nemanja-boric-sociomantic: haha

src/ocean/meta/codegen/Identifier.d
```diff
@@ -0,0 +1,122 @@
...
+    else
+        const identifier = Sym.stringof;
+}
+
+///
+unittest
+{
+    class ClassName { int fieldName; }
+    void funcName ( ) { }
+    extern(C) void funcNameArgs ( int a, double b ) { }
```
david-eckardt-sociomantic: `static`? Not sure if it makes a difference here, nested `extern(C)` functions are unusual (if at all really possible in the sense that one can call them from an extern C program).

src/ocean/meta/codegen/CTFE.d
mihails-strasuns-sociomantic: You rejected a PR that would allow to write it in a reasonable way so I have no other choice.

src/ocean/meta/codegen/Identifier.d
mihails-strasuns-sociomantic: `static` makes no difference here. `extern(C)` is onlt relevant here because it is reflected in `.stringof` which utility relies on.

mihails-strasuns-sociomantic:
Updated:
- CTFE `toString` for ulong + more tests
- Better type names inside `ParametersOf`

david-eckardt-sociomantic (approved):

leandro-lucarella-sociomantic (commented):
Conceptual LGTM (I didn't check the implementations, only gave a quick glance at the commit messages and changelog entries mostly)

PR #181, opened 2017-07-09
Better const compatibility of Map

mihails-strasuns-sociomantic:
For some of Map methods it is fine to accept const qualified version of
the key as it is neither stored nor mutated.

Has to be based against v4.x.x because it anything that overrides `toHash`
method will need small signature adjustment to keep compiling with D2 compiler.

### DIFF
--- a/src/ocean/util/container/map/HashMap.d
+++ b/src/ocean/util/container/map/HashMap.d
@@ -104,6 +104,10 @@ public class HashMap ( V ) : Map!(V, hash_t)
         Calculates the hash value from key. Uses the identity since key is
         expected to be a suitable hash value.
 
+        This method has to accept `const(hash_t)` because it overrides templated
+        base `toHash (K) (in K key)` method and DMD demands exact qualifier
+        match for arguments.
+
         Params:
             key = key to hash
 
@@ -112,7 +116,7 @@ public class HashMap ( V ) : Map!(V, hash_t)
 
     ***************************************************************************/
 
-    public override hash_t toHash ( hash_t key )
+    public override hash_t toHash ( in hash_t key )
     {
         return key;
     }
--- a/src/ocean/util/container/map/Map.d
+++ b/src/ocean/util/container/map/Map.d
@@ -467,7 +467,7 @@ public abstract class Map ( V, K ) : BucketSet!(V.sizeof, K)
 
     ***************************************************************************/
 
-    public V* opIn_r ( K key )
+    public V* opIn_r ( in K key )
     {
         auto element = this.get_(key);
 
@@ -494,7 +494,7 @@ public abstract class Map ( V, K ) : BucketSet!(V.sizeof, K)
 
     ***************************************************************************/
 
-    public V* get ( K key )
+    public V* get ( in K key )
     out (val)
     {
         assert (val !is null);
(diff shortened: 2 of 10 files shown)


gavin-norman-sociomantic (approved):

src/ocean/util/container/map/HashMap.d
```diff
@@ -112,7 +112,7 @@ public class HashMap ( V ) : Map!(V, hash_t)
 
     ***************************************************************************/
 
-    public override hash_t toHash ( hash_t key )
+    public override hash_t toHash ( in hash_t key )
```
mathias-lang-sociomantic, line 119: Heh, why ?

mihails-strasuns-sociomantic: Good question!

It is result of a bit awkward series of events. Root abstract `toHash` comes from `BucketSet` and is templated on key type. And some keys are non-value types so `const` becomes necessary. But when checking overrides D compiler ensures exact qualifier compatibility even for value types, so you can't override `toHash ( K ) ( in K )` with plain `toHash ( hash_t )`.

nemanja-boric-sociomantic: Oh...

mathias-lang-sociomantic: Of course :man_facepalming: 
Could you add a comment about this ?

mihails-strasuns-sociomantic: Sure, in which place? (`toHash` changes affect bunch of modules)

mathias-lang-sociomantic: I'd say just the hash_t overload, since they are the ones that will come as surprising to the reader.

src/ocean/util/container/map/Map.d
```diff
@@ -1228,6 +1228,14 @@ unittest
     test_val!(long);
 }
 
+unittest
+{
+    // ensure opIn works with const keys
+    auto map = new StandardKeyHashingMap!(int, mstring)(5);
+    auto v = "something" in map;
+    test!("is")(v, null);
+}
```
mathias-lang-sociomantic, line 1237: In issue 1628, there is the following test case:
```D
void main ()
{
    StandardKeyHashingMap!(istring, istring) map;
    *map.put("Foo") = "Bar";
    cstring k = "Foo";
    assert(k in map);
}
```

I assume this will also handle it ?

mihails-strasuns-sociomantic: Sadly no, and I don't know if this can be fixed trivially:

```
 Error: function ocean.util.container.map.Map.Map!(string, string).Map.opIn_r (const(string) key) is not callable using argument types (const(char)[])
```

Problem is that `istring` requires strictly `istring` as input, not some `cstring`. We know that in this case it doesn't matter and could redefine it as `opIn_r ( in Unqual!(ElementTypeOf!(T))[] )` but as base class is generic for any element/key types, it will quickly get messy :/

mihails-strasuns-sociomantic:
Updated with extra comments

mathias-lang-sociomantic (approved):
LGTM but not tested

mihails-strasuns-sociomantic:

mihails-strasuns-sociomantic:
Rebased

PR #183, opened 2017-07-11
StringC: Append null terminator even if the string is empty

mathias-lang-sociomantic:


mihails-strasuns-sociomantic (commented):
Looks like a bug fix to me tbh (== v2.8.x base material).

But of course there always can be someone relying on the fact that for some cases `toCString` does not return a c string.. ;)

mathias-lang-sociomantic:
Good point, changed base

relnotes/stringc-nullterm.migration.md
```diff
@@ -0,0 +1,3 @@
+* `ocean.text.util.StringC : StringC.toCString`
+
+  This function will now properly append a null terminator, even for empty string.
```
mihails-strasuns-sociomantic: Don't need release notes now that it is based on point branch

mathias-lang-sociomantic:
Removed relnotes

mihails-strasuns-sociomantic (approved):

Issue #171, opened 2017-06-28
Formatter must never format arrays as `null`


mihails-strasuns-sociomantic:
```D
void main ()
{
    int[] arr;
    Stdout.print(format("{}\n", arr));

    arr = new int[5];
    arr.length = 0;
    Stdout.print(format("{}\n", arr));
}
```

This code must print `[]` twice. Right now it has special case for `arr.ptr is null` and will print `null` for a first call.

As this is a subtle breaking change, scheduling it to v4.0.0 for now.

mihails-strasuns-sociomantic:
FYI @mathias-lang-sociomantic

andrej-mitrovic-sociomantic:
For `char[]` I expect it to print ` `, as in, nothing. That's the behavior of the old code.

mathias-lang-sociomantic:
> This code must print `[]` twice. 

Why?

mihails-strasuns-sociomantic:
Because D spec / implementation has no such notion as "null array", content of pointer is purely implementation detail. We know of some ways to guarantee it will ne null or non-null but it is not defined in general - this issue, for example, was brought to me by @andrej-mitrovic-sociomantic because of tests starting to fail depending on how variable is declared as they expect specific consistent formatting.

And, in general, I can't imagine why anyone would want to have an array printed as a pointer. In the code snippet above I create an empty array. I explicitly initialize it with `[]` literal. It is rather surprising to get anything else as a string representation. In my opinion `null` must be only used when printing pointers/classes and never with arrays/strings.

mihails-strasuns-sociomantic:
NB: Original snipet used to start with `int[] arr = [ ]` but I forgot the literal when writing it here ;)

mathias-lang-sociomantic:
> Because D spec / implementation has no such notion as "null array", content of pointer is purely implementation detail.

This is wrong. There's a difference between empty slices and null slices.
```D
void main ()
{
    int[] arr1 = [42];
    int[] arr2;
    foo(arr1[0..0]); // Pass
    foo(arr2);       // Fails
}

void foo (int[] arr)
{
    assert(arr);
}
```

And when a change to this was attempted, Andrei was explicitly against it and it was reverted.
Not to defend the spec in that regard - I am not a fan of this behavior - but the difference is clear and (nowadays) quite widely known.

> And, in general, I can't imagine why anyone would want to have an array printed as a pointer.

[...]

PR #185, opened 2017-07-11
Formatter: Treat null and empty array the same

mathias-lang-sociomantic:
This was originally different but it proved to be more of a problem
than an help. Reverting to the old Tango Layout's behavior.

Fixes sociomantic-tsunami/ocean#171

### DIFF
--- a/src/ocean/text/convert/Formatter.d
+++ b/src/ocean/text/convert/Formatter.d
@@ -323,16 +323,6 @@ private void widthSink (FormatterSink sink, cstring str, ref Const!(FormatInfo) 
 
 private void handle (T) (T v, FormatInfo f, FormatterSink sf, ElemSink se)
 {
-    // Handle ref types explicitly
-    static if (is (typeof(v is null)))
-    {
-        if (v is null)
-        {
-            se("null", f);
-            return;
-        }
-    }
-
     /** The order in which the following conditions are applied matters.
      * Explicit type checks (e.g. associative array, or `is(T == V)`)
      * should go first as they are unambiguous.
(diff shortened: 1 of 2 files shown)


src/ocean/text/convert/Formatter.d
```diff
@@ -610,6 +604,34 @@ private template DropTypedef (T)
...
+    Before calling `toString`, a `null` check should be performed.
+    However, it shouldn't always be performed, as `T` might not be a ref type.
+
+    Params:
+        v = Object to check, might be a reference type
+        expr = Expression to call if `v` is not null or not a ref type
+        onNull = Expression to call if `v` is null
+
+    Returns:
+        Return type of `expr`.
```
mihails-strasuns-sociomantic: I don't think D functions return types ;)

mihails-strasuns-sociomantic (approved):
Looks sane, would be nice to get confirmation from @don-clugston-sociomantic that it matches old format.

Big question is target branch though. Technically it is a breaking change from current head state. But current state is a bug for certain projects.

mathias-lang-sociomantic:
> Big question is target branch though. Technically it is a breaking change from current head state.

I was wondering the same. Making it target v4.x.x will put it too far in the future. And it's too big of a change to go in a patch release.

src/ocean/text/convert/Formatter.d
mathias-lang-sociomantic: Corrected

don-clugston-sociomantic:
> Looks sane, would be nice to get confirmation from @don-clugston-sociomantic that it matches old format.

Confirmed with ocean 3.1.7 since it seems be the newest ocean release that all of our other libs support without generating warnings/deprecation messages.
Without this issue fixed, I am stuck in ocean 2.x.

mihails-strasuns-sociomantic:
Ok, I think I will merge it into v2.x.x in that case and we will resort to tagging it as `v2.10.0+breaking` (CC @leandro-lucarella-sociomantic ) to warn about exception as discussed ages ago in some other context :)

don-clugston-sociomantic:
Can we have this merged into a ocean v3 release? For D2 conversion I cannot use anything earlier than ocean v3.1, but this issue prevents upgrading to v3.

It may be a breaking change with regard to ``Format``, but it causes a regression for logging.

nemanja-boric-sociomantic:
This is already released as v3.2.5, @don-clugston-sociomantic .

nemanja-boric-sociomantic:
Ignore my previous comment, I thought I was on the different PR.

mihails-strasuns-sociomantic:
@don-clugston-sociomantic it is already in v3.x.x - https://github.com/sociomantic-tsunami/ocean/commits/v3.x.x , I merged head of v2.x.x into it few days ago. Not in any released version yet though.

PR #187, opened 2017-07-11
Re-apply lost commits from legacy ocean repo

mihails-strasuns-sociomantic:


mihails-strasuns-sociomantic:
I haven't yet checked which parts affect API and may need to be moved to v4.x.x - just picked old commits and fixes conflicts/tests.

mathias-lang-sociomantic:
First 2 commits LGTM, last one needs a bit of adjustment.

src/ocean/text/convert/Memory.d
```diff
@@ -87,7 +87,7 @@ public char[] memoryToHexAscii ( void[] mem, char[] output = null )
                 c = '.';
             ascii[idx] = c;
         }
-        Format.format(output, "{}\n", ascii[0 .. idx]);
```
mathias-lang-sociomantic: You can revert this whole file, it's deprecated (https://github.com/sociomantic-tsunami/ocean/pull/174)

mathias-lang-sociomantic:
Ping

nemanja-boric-sociomantic:
D2 build is also red.

mathias-lang-sociomantic:
> D2 build is also red.

Yeah, reverting changes to `ocean.text.convert.Memory` should fix it.

mihails-strasuns-sociomantic:
Fixed.

But after re-examining actual patches I don't think we can merge it into v3.x.x anymore :( Those are risky changes that intended to be in 3.0.0 and that train is gone.

Instead I would propose to re-purpose most of this PR against v4.x.x and aim for releasing v4.0.0 roughly at the same time as v2.x.x goes out of support. What do you thing @sociomantic-tsunami/core-team @mathias-lang-sociomantic ?

mathias-lang-sociomantic:
> But after re-examining actual patches I don't think we can merge it into v3.x.x anymore

I think first and last commit can definitely go in. The second would need some adjustment, like turning `Snoop` removal into a deprecation and reverting the changes that affect the API (`AppStatus`).
Now that `""` and `null` are treated the same I don't think anything will break (famous last words).

mihails-strasuns-sociomantic:
> Now that "" and null are treated the same I don't think anything will break (famous last words)

That is pretty much the reason I am worried to just let it go in minor release - these last words are just _too famous_ ;)

mihails-strasuns-sociomantic:
Have split https://github.com/sociomantic-tsunami/ocean/pull/203 into separate commit and turned snoop into deprecation.

mihails-strasuns-sociomantic:

src/ocean/io/console/StructTable.d
```diff
@@ -274,7 +274,7 @@ public class StructTable ( S )
     private char[] defaultFieldString ( T ) ( T* field )
     {
         this.format_buffer.length = 0;
-        Format.format(this.format_buffer, "{}", *field);
+        sformat(this.format_buffer, "{}", *field);
```
mathias-lang-sociomantic: Missing `enableStomping` here as well ?

src/ocean/io/select/client/TimerSet.d
```diff
@@ -233,7 +232,7 @@ public class TimerSet ( EventData ) : TimerEventTimeoutManager
             protected cstring id ( )
             {
                 this.id_buf.length = 0;
```
mathias-lang-sociomantic, line 234: Ditto `enableStomping` ?

src/ocean/text/util/DigitGrouping.d
```diff
@@ -278,9 +277,12 @@ public class BitGrouping
         output.length = 0;
         enableStomping(output);
 
+        if (unit.length == 0)
+            unit = "";
```
mathias-lang-sociomantic: I don't think that's necessary anymore

leandro-lucarella-sociomantic (approved):
I wanted to give my LGTM yesterday just when the :unicorn: madness was unleashed.

src/ocean/io/console/StructTable.d
nemanja-boric-sociomantic: ping @mihails-strasuns-sociomantic ?

src/ocean/io/select/client/TimerSet.d
nemanja-boric-sociomantic: ping here as well.

mihails-strasuns-sociomantic:
Updated:
- more `enableStomping`
- removed unnecessary null array special case
- rebased

mathias-lang-sociomantic (approved):
LGTM

nemanja-boric-sociomantic (approved):
LGTM

PR #191, opened 2017-07-14
Allow usage of Buffer!(char) and Buffer!(Wchar) with StringC.toStringC

nemanja-boric-sociomantic:
`StringC`'s `toCString` overloads now support `Buffer!(char)` and
`Buffer!(Wchar)` as well as the arrays.

### DIFF
--- a/relnotes/stringcbuffer.feature.md
+++ b/relnotes/stringcbuffer.feature.md
@@ -1,0 +1,4 @@
+* `ocean.text.util.StringC`
+
+  `StringC`'s `toCString` overloads now support `Buffer!(char)` as well, so
+  it can work with `Buffer!(char)`.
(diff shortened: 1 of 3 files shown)


mihails-strasuns-sociomantic:
Is the `wchar` overload really necessary? I thought concensus was that we never use those and should slowly strip ocean code base from such overloads.

nemanja-boric-sociomantic:
I'm not sure, but it's certainly a pain to maintain it and it makes the entire thing less readable. I would then go and remove that overload and create an issue to deprecate it in v4.x.x?

mihails-strasuns-sociomantic:
Yeah I'd suggest to only support `Buffer!char` and deprecate `Wchar[]`.

nemanja-boric-sociomantic:
Updated with `Buffer!char`. What do you feel - should we deprecate `Wchar[]` in v2.x.x, v3.x.x or v4.x.x. I feel deprecating in v2.x.x at the end of its lifetime could be a little bit too early?

mihails-strasuns-sociomantic:
As far as I know we don't use any of those overloads anyway (which is why they are being deprecated in the first place) so v2.x.x is fine.

nemanja-boric-sociomantic:
Deprecated.

mihails-strasuns-sociomantic (approved):

PR #192, opened 2017-07-16
Install relevant compiler package based on matrix

mihails-strasuns-sociomantic:
Preparation to make possible to test ocean with multiple compiler versions and not just installed dmd1/dmd-transitional ones.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,17 +19,24 @@ env:
         # Make sure beaver is in the PATH
         - PATH="$(git config -f .gitmodules submodule.beaver.path)/bin:$PATH"
     matrix:
-        - DC=dmd1 DIST=trusty F=production
-        - DC=dmd1 DIST=trusty F=devel
-        - DC=dmd1 DIST=xenial F=production AFTER_SCRIPT=1
-        - DC=dmd1 DIST=xenial F=devel
-        - DC=dmd-transitional DIST=trusty F=production
-        - DC=dmd-transitional DIST=trusty F=devel
-        - DC=dmd-transitional DIST=xenial F=production
-        - DC=dmd-transitional DIST=xenial F=devel
+        - DMD=1.080.0 DIST=trusty F=production
+        - DMD=1.080.0 DIST=trusty F=devel
+        - DMD=1.080.0 DIST=xenial F=production AFTER_SCRIPT=1
+        - DMD=1.080.0 DIST=xenial F=devel
+        - DMD=2.070.2.s12 DIST=trusty F=production
+        - DMD=2.070.2.s12 DIST=trusty F=devel
+        - DMD=2.070.2.s12 DIST=xenial F=production
+        - DMD=2.070.2.s12 DIST=xenial F=devel
 
-install: beaver install
+install:
+    - beaver docker build --build-arg "DMD=$DMD" --build-arg "DC=$(ci/DC.sh)" -f "Dockerfile${DIST:+.$DIST}" .
 
-script: BEAVER_DOCKER_VARS="DC" beaver run ci/travis.sh
+script:
+    - beaver run ci/test.sh
 
-after_success: ci/travis-after-success.sh
+after_success:
+    # run only once among matrix rows, workaround for Travis lack of support
+    # for sequential pipelines
+    - test "$AFTER_SCRIPT" || exit 0
+    - beaver run ci/codecov.sh
+    - beaver run ci/d2-convert-tag.sh
--- a/ci/d2-convert-tag.sh
+++ b/ci/d2-convert-tag.sh
@@ -1,11 +1,6 @@
 #!/bin/sh
 set -xe
 
-test "$AFTER_SCRIPT" = "1" || exit 0
-
-# Submit code coverage report
-beaver run ci/codecov.sh
-
 # If this is a tag, convert and push to ocean-d2 repo
 if test -n "$TRAVIS_TAG"
 then
--- a/ci/d2-convert-tag.sh
+++ b/ci/d2-convert-tag.sh
@@ -24,5 +24,5 @@ then
 	set +x
 	git -c "credential.https://github.com.username=${OCEAN_D2_USER}" \
 		-c "core.askPass=./ci/askpass.sh" \
-		push "https://github.com/sociomantic-tsunami/ocean-d2.git" "$d2tag"
+		push "https://github.com/sociomantic-tsunami/ocean.git" "$d2tag"
 fi
--- a/Dockerfile.trusty
+++ b/Dockerfile.trusty
@@ -1,3 +1,6 @@
 FROM sociomantictsunami/dlang:trusty-v2
+ARG DMD
+ARG DC
+ENV DMD="$DMD" DC="$DC"
 COPY docker/build /
 RUN /build && rm /build
--- a/Dockerfile.xenial
+++ b/Dockerfile.xenial
@@ -1,3 +1,6 @@
 FROM sociomantictsunami/dlang:xenial-v2
+ARG DMD
+ARG DC
+ENV DMD="$DMD" DC="$DC"
 COPY docker/build /
 RUN /build && rm /build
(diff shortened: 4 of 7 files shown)


ci/ci.sh
```diff
@@ -3,9 +3,17 @@ set -xe
...
 DIST=${DIST:-xenial}
 
+# Binary/package name deduced based on supplied DMD version
+case "$DMD" in
+    1.*    ) DC=dmd1 ;;
+    2.*.s* ) DC=dmd-transitional ;;
+    *      ) DC=dmd ;;
+esac
+
+apt install -y $DC=$DMD-$DIST
```
mihails-strasuns-sociomantic: @leandro-lucarella-sociomantic how do you envision doing this correctly? (I can't `apt install` when script is executed and there are no env vars to check during `docker build` AFAICS)

leandro-lucarella-sociomantic: Yeah, good point, we need some support in the beaver side. I will have a look.

leandro-lucarella-sociomantic: For now, as a workaround, you can do the following (eventually I will integrate this into beaver, but we can give it a go first here):

```yml
install: beaver docker build --build-arg "DMD=$DMD" -f "Dockerfile${DIST:+.$DIST}" .
```

(this is basically what `beaver install` does plus passing the `DMD` argument to the build command).

Then in the `Dockerfile`:
```
ARG DMD
ENV DMD=$DMD
```

Then you can use the `DMD` env variable from inside the build script.

leandro-lucarella-sociomantic: Although with your current script I think the default would be broken, I think it might be better to default to something that makes sense (here the installed dmd1).

leandro-lucarella-sociomantic: Build args reference: https://docs.docker.com/engine/reference/builder/#arg

leandro-lucarella-sociomantic: Also, please create an issue in beaver about this (if there isn't one yet, there might be).

mihails-strasuns-sociomantic: > Although with your current script I think the default would be broken

I actually wanted to raise the issue about current usage of defaults because it causes matrix builds to silently succeed (with default values) if something gets broken in env var passing. To me it is more important than minor convenience of running scripts manually.

Will try `--build-arg` approach tomorrow.

leandro-lucarella-sociomantic: Mmm, good point. I just like/find convenient to be able to use the Dockerfile as is because I do it regularly. Maybe we can have some sort of utility script to define defaults for manual use or something.

mihails-strasuns-sociomantic:
@leandro-lucarella-sociomantic OK, using your suggestion as a base I have managed to make it work. Would be nice to get some advices to make it pretty now (so that it can be used as a reference for beaver PR)

leandro-lucarella-sociomantic (commented):
OK, looks good, I just added a new commit with some cleaning, let me know what you think...

mihails-strasuns-sociomantic:
@leandro-lucarella-sociomantic I have squashed your suggestions and did some more refactoring of my own on top. It looks quite different because I have been trying to pursue KISS approach and minimize amount of intermediate scripts, hope you won't mind.

Also added one commit to start pushing D2 converted tags in the same repo because why not.

.travis.yml
```diff
@@ -19,17 +19,24 @@ env:
...
-script: BEAVER_DOCKER_VARS="DC" beaver run ci/travis.sh
+script:
+    - beaver run ci/test.sh
 
-after_success: ci/travis-after-success.sh
+after_success:
+    # run only once among matrix rows, workaround for Travis lack of support
+    # for sequential pipelines
+    - test "$AFTER_SCRIPT" || exit 0
+    - beaver run ci/codecov.sh
```
leandro-lucarella-sociomantic, line 41: The problem with this is it will share the credentials with codecov (`beaver run` shares Bintray credential with the docker container, this should probably done only with `beaver bintray ...`, but still there might be other credential sharing in the future with `beaver run`), which is what we wanted to avoid in the first place by running it inside a docker container. Ideally we should move the codecov magic to beaver so we don't have to care about these details in the projects, but for now we should call docker differently to run codecov.

.travis.yml
```diff
@@ -19,17 +19,24 @@ env:
...
+    - beaver run ci/d2-convert-tag.sh
```
leandro-lucarella-sociomantic, line 42: OTOH this will probably fail if we don't share/add credentials to push to the repo (they are not shared implicitly, at least not yet).

leandro-lucarella-sociomantic (changes requested):

ci/d2-convert-tag.sh
```diff
@@ -29,5 +24,5 @@ then
 	set +x
 	git -c "credential.https://github.com.username=${OCEAN_D2_USER}" \
 		-c "core.askPass=./ci/askpass.sh" \
-		push "https://github.com/sociomantic-tsunami/ocean-d2.git" "$d2tag"
+		push "https://github.com/sociomantic-tsunami/ocean.git" "$d2tag"
```
leandro-lucarella-sociomantic, line 27: Oh, this is also wrong, I think, because we need to add the `+d2` suffix so they don't oeverride the original tag.

Also, for this step I think the [deploy](https://docs.travis-ci.com/user/deployment/) section is more appropriate, so we can use the condition (`on: tag: true`) to only run when is a tag (and also to "error" the build if it fails, `after_success` exit code is ignored).

> The exit code of `after_success`, `after_failure`, `after_script` and subsequent stages do not affect the build result. However, if one of these stages times out, the build is marked as a failure.

leandro-lucarella-sociomantic: > Oh, this is also wrong, I think, because we need to add the +d2 suffix so they don't oeverride the original tag.

https://github.com/sociomantic/backend/issues/303

mihails-strasuns-sociomantic: `d2tag` is set to `$TRAVIS_TAG+d2.auto` here https://github.com/sociomantic-tsunami/ocean/pull/192/files/d484b914bb329fcc645597a2ceff953a13b78d47#diff-c4405eca91b5ba6f9ea44c38cdd19ebfR20

.travis.yml
mihails-strasuns-sociomantic: Hm, I think in this case it is better to create new beaver release first. Passing all env vars to docker container is hell of boilerplate to copy and in general beaver should provide better control over how credential are passed. I will look into that soon.

ci/d2-convert-tag.sh
leandro-lucarella-sociomantic: Do we want to keep the `.auto` though? If so, we should update the internal doc about how D2 tags should be named.

leandro-lucarella-sociomantic: I would not use the `.auto` since having `+d2` is enough indication that it was automatically generated. When we switch completely to D2 we won't add the `+d2` and we won't have a `+d1` or anything like that.

mihails-strasuns-sociomantic: I think it is nice to keep as it immediately tells that content was auto-generated. Is it documented anywhere?

.travis.yml
leandro-lucarella-sociomantic: Mmm, I think is usually more useful to start with an use case and then generalize in beaver. I don't feel we have all the info yet to have something general for beaver, and I don't want to block this on beaver.

mihails-strasuns-sociomantic:
Blocked on beaver tweaks (including https://github.com/sociomantic-tsunami/beaver/pull/14)

.travis.yml
mihails-strasuns-sociomantic: For me it is clear that at least automatic passing of credentials in all contexts in beaver was wrong approach and has to be reverted.

ci/d2-convert-tag.sh
leandro-lucarella-sociomantic: Yes, I sent you a link via PM.

leandro-lucarella-sociomantic:
I think this should be closed in favour of #209.

mihails-strasuns-sociomantic:
Done with a beaver help now.

PR #193, opened 2017-07-17
Don't decode padded chunks with less than two bytes

nemanja-boric-sociomantic:
In case the last quartet is padded, it still needs to contain at least
two bytes for the decoding to work. In case it isn't, it's simply
ignored, which is the same behaviour Base64 decoder had before
introducing the support for the optional padding.

### DIFF
--- a/src/ocean/util/encode/Base64.d
+++ b/src/ocean/util/encode/Base64.d
@@ -322,6 +322,12 @@ body
     previous one), data may or may not contain the trailing padding, and it will
     still be decoded correctly.
 
+    Note: With explicit padding, three padding bytes will cause decoder to
+    throw an Exception. However, in the case of implicit padding, the last byte
+    will be silently ignored if it contains less than two bytes. This was the
+    undocumented behaviour of the decoder, and it will be changed in the next
+    major version.
+
 
     Params:
       Table = The decode table to use. Variadic template parameter is used to
@@ -454,7 +460,7 @@ body
         }
 
         // this will try and decode whatever is left, even if it isn't terminated properly (ie: missing last one or two =)
-        if (paddedPos)
+        if (paddedPos >= 2)
         {
             auto padded = data[($ - paddedPos) .. $];
             assert (padded.length <= 4);
@@ -552,6 +558,22 @@ unittest
         test!("==")(result, decode("TWE"));
         test!("==")(result, decode("TWE="));
         test!("!=")(result, decode(""));
+    }
+
+    // in case when implicit padding is used and the last quadruple
+    // has less than two bytes needed for the decoding in which case
+    // it will be silently ignored
+    {
+        auto result = cast(Const!(ubyte)[])"Maq";
+        // B should be ignored since it's missing the second byte
+        test!("==")(result, decode("TWFxB"));
+    }
+
+    // same as previous test, just there's nothing valid to be
+    // decoded in the first place
+    {
+        ubyte[] empty;
+        test!("==")(empty, decode("T"));
     }
 }
 


nemanja-boric-sociomantic:
Though this is technically bad Base64, throwing on such things would be a breaking change, and the change itself was introduced in a patch release, where the intention was not to change the original behavior.

gavin-norman-sociomantic:
1. I think the commit message could explain the reason for this commit better. How about this:
```
Don't decode padded chunks with less than two bytes

fd813adfc91c0f56616577d6706db4899cce7917 inadvertently modified the
behaviour of the base64 decoder. Previous to that commit, trailing data of
less than 2 bytes was simply ignored. fd813adfc91c0f56616577d6706db4899cce7917
turned this into an error.

As this was a breaking change in behaviour, this commit restores the original
behaviour: ignoring a single, trailing byte.
```
2. Is this envisaged as a permanent change or would you remove this quirk in a future major release?

nemanja-boric-sociomantic:
1. Sounds all right.

2. I would throw in the future major release, similar how it throws now if it gets `===` (three padding bytes).

gavin-norman-sociomantic:
Cool. Could you update the commit message and make a PR or issue for the next major?

nemanja-boric-sociomantic:
Updated commit message.

nemanja-boric-sociomantic:
Issue for 4.0.0: https://github.com/sociomantic-tsunami/ocean/issues/196

(will submit a PR as well)

gavin-norman-sociomantic (approved):

src/ocean/util/encode/Base64.d
```diff
@@ -553,6 +553,19 @@ unittest
         test!("==")(result, decode("TWE="));
         test!("!=")(result, decode(""));
     }
+
+    // don't decode short things at the end
+    {
+        auto result = cast(Const!(ubyte)[])"Maq";
+        // B should be ignored since it's missing the second byte
+        test!("==")(result, decode("TWFxB"));
```
david-eckardt-sociomantic: This could use a bit more explanation: What is a *short thing*: Less than two extra bytes?
Maybe make it obvious in code what the valid and what the ignored invalid base 64 bytes are, like
```d
auto plain = cast(Const!(ubyte)[])"Maq";
auto base64 = "TWFx";
// Append one extra byte, say 'B', to base64, which should
// be ignored because it is {insert explanation of "short thing" here}.
test!("==")(plain, decode(encoded ~ "B"));
```

david-eckardt-sociomantic: > What is a *short thing*

https://github.com/sociomantic-tsunami/ocean/pull/193#issue-243352284 contains the answer. Suggest adding it in here as a code comment.

nemanja-boric-sociomantic: But it says just in the comment: "B should be ignored since it's missing the second byte". Anyway, I'll update, that's a fair suggestion.

david-eckardt-sociomantic:
This PR changes unspecified to defined and tested behaviour, which should be mentioned in the documentation of `decode`.

src/ocean/util/encode/Base64.d
david-eckardt-sociomantic: In the context of this PR it is obvious, but just reading the code it may be not, especially because this defines a way of handling invalid Base64 code, so what happens here isn't covered by any specification.

nemanja-boric-sociomantic:
Updated with @david-eckardt-sociomantic suggestions

david-eckardt-sociomantic (approved):

PR #200, opened 2017-07-19
Deprecate `Stopwatch.stop` in favour of `StopWatch.sec`

joseph-wakeling-sociomantic:
The old `stop` method of `StopWatch` was both misleadingly named and misleadingly documented: contrary to what it claimed, it did not stop the timer.  With this in mind it has been renamed to `sec` (by analogy to the existing `microsec` method).

In order to validate the changes, unittests have been added (for both old and new functionality), which as a side effect demonstrate the wicked wicked lies that the `stop` method and its documentation were telling us ... ;-)  These tests replace a wonderful bit of prehistoric debug checking the like of not we shall hopefully never see again.

Note, this is submitted against 3.x.x (given that it involves a deprecation), but can easily be rebased on 2.x.x if preferred.

### DIFF
--- a/src/ocean/time/StopWatch.d
+++ b/src/ocean/time/StopWatch.d
@@ -20,6 +20,12 @@ module ocean.time.StopWatch;
 // CLOCK_MONOTONIC and clock_gettime will be added to core.sys.posix.time in
 // tangort v1.7.0, see tangort issue #6.
 import core.sys.posix.time; // clockid_t, timespec
+
+version (UnitTest)
+{
+    import ocean.core.Test;
+}
+
 extern (C) private
 {
     enum: clockid_t
@@ -125,22 +131,20 @@ public struct StopWatch
         }
 }
 
-
-/*******************************************************************************
-
-*******************************************************************************/
-
-debug (StopWatch)
+unittest
 {
-        import ocean.io.Stdout_tango;
+    StopWatch t;
+    t.start;
 
-        void main()
-        {
-                StopWatch t;
-                t.start;
+    auto wait_time = timespec(0, 10_000);  // 10_000 ns = 10 microsec
 
-                for (int i=0; i < 100_000_000; ++i)
-                    {}
-                Stdout.format ("{:f9}", t.stop).newline;
-        }
+    // validate that stopwatch time counts up
+    nanosleep(&wait_time, null);
+    auto stopwatch_sec = t.stop;
+    test!(">")(stopwatch_sec, 0);
+
+    // prove that `stop` doesn't stop anything ...
+    nanosleep(&wait_time, null);
+    auto stopwatch_sec_again = t.stop;
+    test!(">")(stopwatch_sec_again, stopwatch_sec);
 }
--- a/src/ocean/time/StopWatch.d
+++ b/src/ocean/time/StopWatch.d
@@ -95,14 +95,17 @@ public struct StopWatch
 
         /***********************************************************************
 
-                Stop the timer and return elapsed duration since start()
+                Return elapsed duration (in seconds) since start()
 
         ***********************************************************************/
 
-        double stop ()
+        double sec ()
         {
                 return multiplier * this.microsec;
         }
+
+        deprecated("Use `sec` instead of `stop` (which does not stop anything)")
+        alias sec stop;
 
         /***********************************************************************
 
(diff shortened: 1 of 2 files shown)


Burgos:
Re: 

> As an amusing observation, it seems that the actual count of time that
the stopwatch records is rather greater than the 10 microseconds that
`nanosleep` requests.  It obviously has some expensive operations under
the hood somewhere...

Note that `nanosleep` doesn't tell the process to sleep for _exactly_ specified amount of time, but for _at least_ the specified amount of time. After you program goes to sleep, it may be selected for waking up and executing without any guarantees on the time (which may be as well 10ms after the specified time already elapsed), _and_ it will still take some time to actually make a context switch and get up to speed. It's also worth saying that `clock_gettime` is highly optimised function, one of the few that's utilising virtual system call facility of a linux kernel, so it shouldn't cause any additional context switch on its own. Moral of the story: Don't rely on timing of non-realtime operating system ;-).

Burgos (approved):

src/ocean/time/StopWatch.d
```diff
@@ -141,6 +144,24 @@ unittest
 
     // prove that stopwatch time at least counts up ...
     nanosleep(&wait_time, null);
+    auto stopwatch_sec = t.sec;
+    test!(">")(stopwatch_sec, 0);
+
+    // prove that calling `sec` doesn't stop anything ...
```
gavin-norman-sociomantic: Given the new behaviour, it seems odd to test this. Nothing in the doc suggests that calling `sec` _could_ stop the timer.

joseph-wakeling-sociomantic: Would it help to edit the comment?  I think it's still viable to test the expected behaviour.

gavin-norman-sociomantic: You could say the same for `microsec`.

joseph-wakeling-sociomantic: Yes, but `microsec` is called under the hood by `sec` (and by old `stop`), so its behaviour in this respect is being implicitly validated already.

gavin-norman-sociomantic: Fair enough. Yeah maybe just change the comment to `Validate that checking the elapsed time doesn't stop the timer`?

joseph-wakeling-sociomantic:
Updated the unittest comments (in both patches) and rewrote the 1st commit message slightly in response to feedback.

```diff
diff --git a/src/ocean/time/StopWatch.d b/src/ocean/time/StopWatch.d
index 82d7b96..9320d6b 100644
--- a/src/ocean/time/StopWatch.d
+++ b/src/ocean/time/StopWatch.d
@@ -142,12 +142,12 @@ unittest
 
     auto wait_time = timespec(0, 10_000);  // 10_000 ns = 10 microsec
 
-    // prove that stopwatch time at least counts up ...
+    // validate that stopwatch time counts up
     nanosleep(&wait_time, null);
     auto stopwatch_sec = t.sec;
     test!(">")(stopwatch_sec, 0);
 
-    // prove that calling `sec` doesn't stop anything ...
+    // validate that checking the elapsed time doesn't stop the timer
     nanosleep(&wait_time, null);
     auto stopwatch_sec_again = t.sec;
     test!(">")(stopwatch_sec_again, stopwatch_sec);
@@ -160,12 +160,12 @@ deprecated unittest
 
     auto wait_time = timespec(0, 10_000);  // 10_000 ns = 10 microsec
 
-    // prove that stopwatch time at least counts up ...
+    // validate that stopwatch time counts up
     nanosleep(&wait_time, null);
     auto stopwatch_sec = t.stop;
     test!(">")(stopwatch_sec, 0);
 
-    // prove that stop doesn't stop anything ...
+    // prove that `stop` doesn't stop anything ...
     nanosleep(&wait_time, null);
     auto stopwatch_sec_again = t.stop;
     test!(">")(stopwatch_sec_again, stopwatch_sec);
```

joseph-wakeling-sociomantic:
BTW shall I keep this as targeting v3.x.x or would you like me to rebase on v2.x.x ... ?  It seems reasonable either way.

gavin-norman-sociomantic:
v2.x.x is the default branch for minor releases (which include deprecations), so it should go against that branch please.

joseph-wakeling-sociomantic:
OK, give me 2 seconds ...

joseph-wakeling-sociomantic:
Rebased on v2.x.x, in 2 seconds as measured by a `nanosleep` call.

gavin-norman-sociomantic (approved):

leandro-lucarella-sociomantic:
Mmm, it seems like travis had problems to set the status, can you force-push to force travis again?

leandro-lucarella-sociomantic (approved):

leandro-lucarella-sociomantic (approved):

joseph-wakeling-sociomantic:
Rebased on latest v2.x.x branch.  @leandro-lucarella-sociomantic will this do? :-)

leandro-lucarella-sociomantic:
I don't like these codecov leaving commints in red when they are fine... :unamused:

PR #201, opened 2017-07-20
Fix compilation with 2.071.2.s1 (again)

mihails-strasuns-sociomantic:
Soon (TM) it shall be part of CI :)

### DIFF
--- a/src/ocean/core/VersionCheck.d
+++ b/src/ocean/core/VersionCheck.d
@@ -39,9 +39,19 @@ unittest
 
 private bool hasFeaturesFromFunc ( istring libname, ulong major, ulong minor ) ()
 {
-    struct Library
+    // D1 does not support plain nested imports, D2 prohibits aggregate import
+    // exposure, thus two different approaches:
+
+    version (D_Version2)
     {
-        mixin("import " ~ libname ~ ".LibFeatures;");
+        mixin("import Library = " ~ libname ~ ".LibFeatures;");
+    }
+    else
+    {
+        struct Library
+        {
+            mixin("import " ~ libname ~ ".LibFeatures;");
+        }
     }
 
     mixin("return is(typeof(Library.has_features_"
--- a/src/ocean/net/http/TaskHttpConnectionHandler.d
+++ b/src/ocean/net/http/TaskHttpConnectionHandler.d
@@ -36,7 +36,7 @@ abstract class TaskHttpConnectionHandler : TaskConnectionHandler
     import ocean.net.http.consts.HttpMethod: HttpMethod;
     import ocean.net.http.consts.HeaderFieldNames;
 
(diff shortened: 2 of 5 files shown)


src/ocean/core/VersionCheck.d
```diff
@@ -39,11 +39,23 @@ unittest
...
 {
-    struct Library
+    // D1 does not support plain nested imports, D2 prohibits aggregate import
+    // exposure, thus two different approaches:
+
+    version (D_Version2)
     {
-        mixin("import " ~ libname ~ ".LibFeatures;");
+        mixin("import Library = " ~ libname ~ ".LibFeatures;");
+        mixin("return is(typeof(Library.has_features_"
```
nemanja-boric-sociomantic: Could this mixin here and the mixing at the line 59 (they are the same) be outside this `version/else`?

mihails-strasuns-sociomantic: Yep

src/ocean/core/VersionCheck.d
```diff
@@ -39,9 +39,19 @@ unittest
 
 private bool hasFeaturesFromFunc ( istring libname, ulong major, ulong minor ) ()
 {
-    struct Library
+    // D1 does not support plain nested imports, D2 prohibits aggregate import
+    // exposure, thus two different approaches:
```
mathias-lang-sociomantic, line 43: Fixes #99 ?

mihails-strasuns-sociomantic: Yep, linked now.

mathias-lang-sociomantic (approved):
LGTM if tested

leandro-lucarella-sociomantic (approved):
LGTM

leandro-lucarella-sociomantic:
Except for the codecov things.

I think we should find a way in which we require all tests to pass before merge (at least on the normal case) or remove codecov if we are going to ignore it anyway.

mathias-lang-sociomantic:
I think as long as https://github.com/sociomantic-tsunami/ocean/issues/204 is not solved we cannot rely on codecov. Also some code will always be untestable by codecov's standard. E.g. it only includes unittests, not integration tests, and we'd need a special tool to merge reports for it to be required.

mathias-lang-sociomantic:
Ping ?

mihails-strasuns-sociomantic:
Sorry, I am on vacation. But will look into this again in a week or two ;)

mathias-lang-sociomantic:
I don't think it requires your attention, just someone to merge. @leandro-lucarella-sociomantic 's concern was about CodeCov, which is orthogonal to this P.R. IMO.

nemanja-boric-sociomantic:
I've just tested this, it sill works with v2.071.2.s1. I'm going to merge this as it LGTM.

PR #202, opened 2017-07-21
Fix missing coverage reports

mihails-strasuns-sociomantic:


mihails-strasuns-sociomantic:
Auto-merging to fix reports while @leandro-lucarella-sociomantic is away.

leandro-lucarella-sociomantic:
What is broken and how is this fixing it? There is way too little information in this PR.

PR #203, opened 2017-07-21
Stop using vararg for formatting

mihails-strasuns-sociomantic:
API affecting part of https://github.com/sociomantic-tsunami/ocean/pull/187

### DIFF
--- a/relnotes/vararg.migration.md
+++ b/relnotes/vararg.migration.md
@@ -1,0 +1,4 @@
+* `ocean.io.console.AppStatus`
+
+  `formatStaticLine` method was changed to use template-based variadic argument
+  list instead of runtime one.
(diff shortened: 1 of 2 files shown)


leandro-lucarella-sociomantic (approved):
LGTM but not to travis.

mathias-lang-sociomantic:
I think it's because it need #187 first :)

mihails-strasuns-sociomantic:
Rebased

relnotes/vararg.migration.md
```diff
@@ -0,0 +1,4 @@
+* `ocean.io.console.AppStatus`
+
+  `formatStaticLine` method was changed to use template-based variadic argument
+  list instead of runtime one.
```
nemanja-boric-sociomantic, line 4: What kind of change this actually require from the user? It should be called in the same way, shouldn't it?

mihails-strasuns-sociomantic: If user code was passing in `__va*`, calling method will need to also be rewritten to use template instead. Direct calls will keep working as-is.

nemanja-boric-sociomantic: I see :+1:

nemanja-boric-sociomantic (approved):

PR #210, opened 2017-08-11
Add BTree and operations over it to ocean

nemanja-boric-sociomantic:
This PR introduces BTree data structure and algorithms over it to the ocean. This implementation of a BTree is templated on the type it carries, on the tree's degree, and on the allocation policy that should be used for the allocating the internal nodes (it is useful if the extending the BTree is done once in a while, but when we shouldn't give chance to GC to do mark&sweep). Additionally, GC tracking of the memory can be turned of, so the GC will not (somewhat expensively) traverse the tree, which is useful if we're storing value types, or the reference types for which we're storing the reference somewhere else.  In addition to basic functionality, inorder traversal of the tree, either via opApply or InputRange is supported.

### DIFF
--- a/src/ocean/util/container/btree/BTreeMap.d
+++ b/src/ocean/util/container/btree/BTreeMap.d
@@ -1,0 +1,594 @@
+/********************************************************************************
+
+    B-Tree data structure and operations on it.
+
+    Copyright:
+        Copyright (c) 2017 sociomantic labs GmbH. All rights reserved.
+
+    License:
+        Boost Software License Version 1.0. See LICENSE_BOOST.txt for details.
+        Alternatively, this file may be distributed under the terms of the Tango
+        3-Clause BSD License (see LICENSE_BSD.txt for details).
+
+*******************************************************************************/
+
+module ocean.util.container.btree.BTreeMap;
+
+import ocean.transition;
+import ocean.util.container.mem.MemManager;
+
+/******************************************************************************
+
+    BTree structure.
+
+
+    BTree is a rooted tree, with the following properties:
+
+    - A BTree has characteristic called "degree". This indicates branching
+      factor of the tree, so that maximum number of the subtrees for a single
+      node is `2 * degree`. Having a high degree makes the tree shorter, and,
+      because the node contains large number of elements, it allows prefetching
+      many elements from the memory in one go.
+
+    - The layout of one node of the tree is as follows:
+
+      -----------------------------------------------------------------
+      | p1 | element1 | p2 | element2 | ... | pn-1 | element n-1 | pn |
+      -----------------------------------------------------------------
+
+      Every element (a key/value pair) in the node (for the non-leaf nodes) is
+      surrounded by two pointers. We call these pointers child nodes.
+
+    - Each pointer is a root of a subtree, for which the following holds true:
+      all the elements' keys in a tree pointed by p(i-1) (pointer left of the
+      elements) are smaller (in respect to the ordering defined by the key's
+      type) of the element(i), and all the elements' keys in a subtree pointed by
+      p(i) are larger than a element(i).
+
+    - Every node may contain at most 2*degree - 1 elements. Therefore, an internal
+      node may be root for at most 2*degree subtrees.
+
+    - Every node other than root must have at least (degree-1) elements.
+
+    - Every internal node other than the root has at least degree children.
+
+    - Every node has the following attributes:
+
+        node.number_of_elements = number of elements node currently contains
+            (as nodes are statically allocated block, the number of the elements
+             may not necessarily be maximal).
+        node.elements = the elements themselves
+        node.is_leaf = indicator if the node is a leaf.
+        node.child_nodes = number_of_elements+1 pointers to the subtrees whose
+            roots are this node. We refer to these nodes as child-nodes (this
+            node is a direct parent of them).
+
+    - All leaves have the same depth
+
+    Other than insert and delete operations, inorder traversal method is provided,
+    which traverses the tree in inorder order (left subtree is visited first,
+    then the element separating these subtrees, then the right subtree, then
+    the next element, etc.). This provides sequential access to the tree.
+    There's a recursive and range-based implementation of the inorder
+    traversal.
+
+    Range-based traversal is supported through `BTreeMapRange` structure, found in
+    in ocean.util.container.btree.BTreeMapRange.
+
+    CPU complexity of the operations (n is number of the elements, t is degree,
+    h is the tree height):
+
+        - Searching: O(th) = O(t * log_t(n))
+        - Inserting: O(th) = O(t * log_t(n))
+        - Deleting:  O(th) = O(t * log_t(n))
+
+    Complexity of the memory access (to read/write the node from/in memory):
+
+        - Searching O(h) = O(log_t(n))
+        - Inserting O(h) = O(log_t(n))
+        - Deleting: O(h) = O(log_t(n))
+
+    References:
+        - https://en.wikipedia.org/wiki/B-tree
+        - Thomas H. Cormen et al. "Introduction to Algorithms, 3rd edition"
+
+    Params:
+        TreeKeyType = type of the key.
+        TreeValueType = type of the element to store.
+        tree_degree = degree of the tree (refer to the documentation)
+
+******************************************************************************/
+
+struct BTreeMap(TreeKeyType, TreeValueType, int tree_degree)
+{
+    import ocean.util.container.btree.Implementation;
+
+    static assert (tree_degree > 0);
+
+    /**************************************************************************
+
+        Constructor.
+
+        Params:
+            allocator = memory manager to use to allocate/deallocate memory
+
+    ***************************************************************************/
+
+    package void initialize (IMemManager allocator = mallocMemManager)
+    {
+        this.impl.initialize(allocator);
+    }
+
+    // Disable constructor, so user always needs to use the makeBTreeMap method
+    version (D_Version2)
+    {
+        mixin("@disable this();");
+    }
+
+    /**************************************************************************
+
+        Inserts the (key, value) in the tree. This is passing the copy of the
+        value into the tree, so it's not necessary to manually create copy of
+        it.  Note that for the reference types, this will just copy the
+        reference.
+
+        Params:
+            key = key to insert
+            value = value associated to the key to insert.
+
+        Returns:
+            true if the element with the given key did not exist and it
+            was inserted, false otherwise
+
+        Complexity:
+            CPU: O(degree * log_degree(n))
+            Memory:O(log_degree(n))
+
+    ***************************************************************************/
+
+    public bool insert (KeyType key, ValueType value)
+    {
+        return this.impl.insert(key, value);
+    }
+
+    /******************************************************************************
+
+        Deletes the element from the BTreeMap.
+
+        Params:
+            key = key of the element to delete.
+
+        Returns:
+            true if the element was found and deleted, false otherwise.
+
+        Complexity:
+            CPU: O(degree * log_degree(n))
+            Memory:O(log_degree(n))
+
+     ******************************************************************************/
+
+    public bool remove (KeyType key)
+    {
+        return this.impl.remove(key);
+    }
+
+    /******************************************************************************
+
+        Searches the tree for the element with the given key and returns the
+        associated value.
+
+        Params:
+            key =  key to find in a tree.
+            found_element = indicator if the element with the given key has been
+                found in the map.
+
+        Returns:
+            copy of the value associated with the key, if found, ValueType.init
+            otherwise.
+
+        Complexity:
+            CPU: O(degree * log_degree(n))
+            Memory:O(log_degree(n))
+
+    *******************************************************************************/
+
+    public ValueType get (KeyType key, out bool found_element)
+    {
+        return this.impl.get(key, found_element);
+    }
+
+    /***********************************************************************
+
+        Recursive inorder iteration over keys and values. Note that, in case
+        the tree is changed during the iteration, iteration will halt.
+
+        Params:
+            dg = opApply's delegate
+
+        Returns:
+            return value of dg
+
+    ***********************************************************************/
+
+    public int opApply (int delegate (ref KeyType value, ref ValueType) dg)
+    {
+        return this.impl.inorder(dg);
+    }
+
+    /***********************************************************************
+
+        Recursive inorder iteration over values only. Note that, in case the
+        tree is changed during the iteration, iteration will halt.
+
+        Params:
+            dg = opApply's delegate
+
+        Returns:
+            return value of dg
+
+    ***********************************************************************/
+
+    public int opApply (int delegate (ref ValueType) dg)
+    {
+        return this.impl.inorder(dg);
+    }
+
+
+    /// Convenience alias for the implementation
+    package alias BTreeMapImplementation!(KeyType, ValueType, tree_degree)
+        Implementation;
+
+    /// Convenience alias for the key type
+    package alias TreeKeyType KeyType;
+
+    /// Convenience alias for the element type
+    package alias TreeValueType ValueType;
+
+    /// Private implementation
+    package Implementation impl;
+}
+
+/*******************************************************************************
+
+    Constructor-like method. Constructs the BTreeMap and initializes it.
+
+    Params:
+        KeyType = type of the key.
+        ValueType = type of the value to store.
+        tree_degree = degree of the tree (refer to the documentation)
+        memManager = memory management strategy to use
+
+    Returns:
+        empty BTreeMap which maps KeyType to ValueType of a given degree
+        which sues memManager allocation strategy.
+
+******************************************************************************/
+
+public BTreeMap!(KeyType, ValueType, tree_degree) makeBTreeMap
+    (KeyType, ValueType, int tree_degree)(IMemManager allocator = mallocMemManager)
+{
+    auto tree = BTreeMap!(KeyType, ValueType, tree_degree).init;
+    tree.initialize(allocator);
+    return tree;
+}
+
+version (UnitTest)
+{
+    import ocean.util.container.btree.BTreeMapRange;
+}
+
+///
+unittest
+{
+    // import ocean.util.container.btree.BTreeMapRange;
+    struct MyVal
+    {
+        int x;
+    }
+
+    auto map = makeBTreeMap!(int, MyVal, 2);
+
+    for (int i = 0; i <= 10; i++)
+    {
+        map.insert(i, MyVal(i*2));
+    }
+
+    // find the element by key
+    bool found;
+    auto val = map.get(5, found);
+    test!("==")(found, true);
+    test!("==")(val.x, 10);
+
+    // remove the element
+    map.remove(10);
+    map.get(10, found);
+    test!("==")(found, false);
+
+    // iterate over using opApply
+    foreach (key, val; map)
+    {
+        test!("==")(key * 2, val.x);
+    }
+
+    // iterate over using range
+    void[] buff;
+    for(auto range = byKeyValue(map, &buff);
+            !range.empty; range.popFront())
+    {
+        test!("==")(range.front.key * 2, range.front.value.x);
+    }
+
+    // Let's check that the memory is ready to be reused
+    test(buff.ptr !is null);
+}
+
+/*
+
+    Unittests. Compile with -debug=BTreeMapSanity to turn on
+    the invariant checks for the tree.
+
+*/
+
+version (UnitTest)
+{
+    import ocean.core.Test;
+}
+
+unittest
+{
+    BTreeMap!(int, File, 2) tree = makeBTreeMap!(int, File, 2);
+    bool found_element;
+
+    File f;
+
+    f.setName("5");
+    tree.insert(5, f);
+
+    f.setName("9");
+    tree.insert(9, f);
+
+    f.setName("3");
+    tree.insert(3, f);
+
+    f.setName("7");
+    tree.insert(7, f);
+
+    size_t index;
+    auto res = tree.get(7, found_element);
+
+    test!("==")(found_element, true);
+    test!("==")(res, f);
+
+    f.setName("1");
+    tree.insert(1, f);
+
+    f.setName("2");
+    tree.insert(2, f);
+
+    f.setName("8");
+    tree.insert(8, f);
+
+    f.setName("6");
+    tree.insert(6, f);
+
+    f.setName("0");
+    tree.insert(0, f);
+
+    f.setName("4");
+    tree.insert(4, f);
+}
+
+unittest
+{
+    bool found_element;
+    auto number_tree = makeBTreeMap!(int, int, 2);
+
+    for (int i = -1; i < 9; i++)
+    {
+        number_tree.insert(i, i);
+    }
+
+    // find "random" element
+    int i = 3;
+    auto xres = number_tree.get(i, found_element);
+    number_tree.remove(i);
+    number_tree.get(i, found_element);
+    test!("==")(found_element, false);
+
+    i = 7;
+    number_tree.remove(i);
+    number_tree.get(i, found_element);
+    test!("==")(found_element, false);
+
+    i = 9;
+    number_tree.remove(i);
+
+    i = 0;
+    number_tree.remove(i);
+
+    i = 1;
+    number_tree.remove(i);
+
+    i = 2;
+    number_tree.remove(i);
+
+    for (i = 0; i < 9; i++)
+    {
+        number_tree.remove(i);
+        number_tree.get(i, found_element);
+        test!("==")(found_element, false);
+    }
+
+    for (i = 0; i < 9; i++)
+    {
+        number_tree.insert(i, i);
+    }
+
+    for (i = 9; i > 0; i--)
+    {
+        number_tree.remove(i);
+        number_tree.get(i, found_element);
+        test!("==")(found_element, false);
+    }
+
+    i = 0;
+    number_tree.insert(i, i);
+    i = 1;
+    number_tree.insert(i, i);
+    i = 2;
+    number_tree.insert(i, i);
+
+    for (i = 0; i < 100000; i++)
+    {
+        number_tree.insert(i, i);
+    }
+
+    i = 0;
+    number_tree.remove(i);
+
+    i = 1;
+    number_tree.remove(i);
+
+    i = 2;
+    number_tree.remove(i);
+
+    for (i = 0; i < 100000; i++)
+    {
+        number_tree.remove(i);
+        number_tree.get(i, found_element);
+        test!("==")(found_element, false);
+    }
+
+    // remove reverse
+    for (i = 0; i < 100000; i++)
+    {
+        number_tree.insert(i, i);
+    }
+
+    for (i = 100000; i >= 0; i--)
+    {
+        number_tree.remove(i);
+        number_tree.get(i, found_element);
+        test!("==")(found_element, false);
+    }
+
+    // remove reverse from half
+    for (i = 0; i < 100000; i++)
+    {
+        number_tree.insert(i, i);
+    }
+
+
+    for (i = 5000; i >= 0; i--)
+    {
+        number_tree.remove(i);
+        number_tree.get(i, found_element);
+        test!("==")(found_element, false);
+    }
+
+
+    // random deletion
+    for (i = 0; i < 5000; i++)
+    {
+        number_tree.insert(i, i);
+    }
+
+    bool started;
+    int previous_value;
+    int counter;
+    foreach (value; number_tree)
+    {
+        counter++;
+
+        if (!started)
+        {
+            previous_value = value;
+            started = true;
+            continue;
+        }
+
+        test!("<")(previous_value, value);
+        previous_value = value;
+    }
+    test!("==")(counter, 100000);
+
+    foreach (key, value; number_tree)
+    {
+        test!("==")(key, value);
+    }
+
+    // Randomized tests
+
+    auto random = new Random();
+    int to_remove = 5864;
+    number_tree.remove(to_remove);
+    test(!number_tree.get(to_remove, found_element));
+
+    for (i = 0; i < 100000; i++)
+    {
+        to_remove = random.uniform!(int);
+        number_tree.remove(to_remove);
+        test(!number_tree.get(to_remove, found_element));
+    }
+}
+
+unittest
+{
+    class X
+    {
+        int x;
+
+        version (D_Version2)
+        {
+            mixin("immutable this () {}");
+        }
+    }
+
+    // Test immutable support
+    auto const_tree = makeBTreeMap!(void*, Const!(X), 2);
+
+    version (D_Version2)
+    {
+        mixin("Immut!(X) a = new immutable X;");
+    }
+    else
+    {
+        Immut!(X) a = new X;
+    }
+
+    const_tree.insert(cast(void*)&a, a);
+    bool found;
+    auto res = const_tree.get(cast(void*)&a, found);
+    test(found);
+    test(res == a);
+    test(const_tree.remove(cast(void*)&a));
+}
+
+version (UnitTest)
+{
+    import ocean.core.Test;
+    import ocean.core.Enforce;
+    import ocean.math.random.Random;
+
+    /// File structure - represents a directory or a file
+    /// For testing if the BTreeMap works with other value types
+    public struct File
+    {
+        /// Name of the directory/file
+        char[48] name_buf;
+        ubyte name_length;
+
+        cstring name () /* d1to2fix_inject: const */
+        {
+            return name_buf[0..name_length];
+        }
+
+        void setName (cstring name)
+        {
+            //logger.error("setting name: {}", name);
+            enforce (name.length <= ubyte.max);
+            this.name_length = cast(ubyte)name.length;
+            this.name_buf[0..this.name_length] = name[];
+        }
+   }
+}
--- a/src/ocean/util/container/btree/BTreeMapRange.d
+++ b/src/ocean/util/container/btree/BTreeMapRange.d
@@ -1,0 +1,409 @@
+/********************************************************************************
+
+    Range traversal support for BTreeMap.
+
+    Copyright:
+        Copyright (c) 2017 sociomantic labs GmbH. All rights reserved.
+
+    License:
+        Boost Software License Version 1.0. See LICENSE_BOOST.txt for details.
+        Alternatively, this file may be distributed under the terms of the Tango
+        3-Clause BSD License (see LICENSE_BSD.txt for details).
+
+*******************************************************************************/
+
+module ocean.util.container.btree.BTreeMapRange;
+
+import ocean.transition;
+import ocean.util.container.btree.BTreeMap;
+
+/*******************************************************************************
+
+    Exception thrown from the BTreeMapRange if a tree is modified during iteration.
+
+*******************************************************************************/
+
+class RangeInvalidatedException: Exception
+{
+    import ocean.core.Exception: DefaultExceptionCtor;
+    mixin DefaultExceptionCtor;
+}
+
+// shared between different ranges
+private RangeInvalidatedException range_exception;
+
+static this ()
+{
+    .range_exception = new RangeInvalidatedException(
+            "Container has been changed during iteration, please restart.");
+}
+
+/*******************************************************************************
+
+    Returns a range that performs an inorder iteration over the specified tree.
+
+    If tree has changed during the iteration, range will throw an
+    RangeInvalidatedException. In case user knows that the tree can be changed
+    any time, it is possible to call BTreeMapRange.isValid before performing any
+    action on the range to check if fetching the front element, or moving to
+    the next one is guaranteed to success. This is only true for the
+    single-threaded environment.
+
+    Params:
+        tree = tree to iterate over
+        buff = buffer for storing the state of the iteration. Useful
+               for the reusable-memory management strategy. The buffer size
+               is proportional to the height of the tree, so in most cases
+               it should be at most the size of the several pointers.
+               If the allocations are not a concern, null can passed for buff,
+               and range will allocate a buffer on its own.
+
+    Returns:
+        BTreeMapRange performing the full inorder iteration of the tree.
+
+*******************************************************************************/
+
+public BTreeMapRange!(BTreeMap) byKeyValue (BTreeMap) (ref BTreeMap tree, void[]* buff)
+{
+    if (buff is null)
+    {
+        // Struct wrapper is used to woraround D inability to allocate slice
+        // itself on heap via new
+        static struct Buffer
+        {
+            void[] data;
+        }
+
+        auto buf = new Buffer;
+        buff = &buf.data;
+    }
+    else
+    {
+        (*buff).length = 0;
+        enableStomping(*buff);
+    }
+
+    auto range = BTreeMapRange!(BTreeMap)(&tree.impl);
+    range.stack = cast(typeof(range.stack))buff;
+
+    range.start();
+    return range;
+}
+
+
+/*******************************************************************************
+
+    key/value pair. Akin to runtime's byKeyValue
+    NOTE: ideally, this should be optional. One might choose not to
+    care about keys at all. However, making this work without copying
+    entire code and without crashing dmd1 seems to be impossible.
+
+*******************************************************************************/
+
+private struct Pair(KeyType, ValueType)
+{
+    void* keyp;
+    void* valp;
+
+    KeyType key () { return *cast(KeyType*)keyp; }
+    ValueType value () { return *cast(ValueType*)valp; }
+}
+
+/*******************************************************************************
+
+    Structure representing an inorder range over a BTreeMap.
+
+    If tree has changed during the iteration, range will throw an
+    RangeInvalidatedException. If the user knows that the tree will not change
+    during the range traversal, it is safe to call range's methods - it will throw
+    an exception if this promise is broken. In case user knows that the tree
+    can be changed any time, it is possible to call BTreeMapRange.isValid before
+    performing any action on the range to check if the accessing is safe. This
+    is only true for the single-threaded environment.
+
+*******************************************************************************/
+
+public struct BTreeMapRange(BTreeMap)
+{
+    import ocean.core.Verify;
+    import ocean.core.array.Mutation;
+
+    /// (key, value) pair to return to the caller
+    alias Pair!(BTreeMap.KeyType, BTreeMap.ValueType) KeyValue;
+
+    /// Root node of a tree
+    private BTreeMap.Implementation* tree;
+
+
+    /// Copy of the content_version field of the tree at the time iteration began
+    private ulong tree_version;
+
+    /// Pair of the node/index for which we've paused iteration and pushed
+    /// to stack
+    private struct NodeElement
+    {
+        BTreeMap.Implementation.BTreeMapNode* node;
+        size_t index;
+
+        BTreeMap.ValueType* value ()
+        {
+            return &(node.elements[index].value);
+        }
+
+        BTreeMap.KeyType* key ()
+        {
+            return &(node.elements[index].key);
+        }
+    }
+
+    /// When IterationStep() moves from a node to iterate over its children,
+    /// the parent node is pushed to this stack, and popped upon completion
+    /// of traversal over the nodes.
+    private NodeElement[]* stack;
+
+    /// Next (node, element_index) pair we're should walk over
+    private NodeElement item;
+
+    /// Value of the current key to return with .front
+    private BTreeMap.KeyType* current_key;
+
+    /// Value of the current value to return with .front
+    private BTreeMap.ValueType* current_value;
+
+    invariant ()
+    {
+        verify(this.stack !is null);
+    }
+
+
+    /***************************************************************************
+
+        Returns:
+            the current element in the range.
+
+        Throws:
+           RangeInvalidatedException if the underlying tree has changed
+
+    ***************************************************************************/
+
+    public KeyValue front ()
+    {
+        this.enforceValid();
+        return Pair!(BTreeMap.KeyType, BTreeMap.ValueType)(this.current_key, this.current_value);
+    }
+
+    /***************************************************************************
+
+        Pops the next element from the tree.
+
+        Throws:
+           RangeInvalidatedException if the underlying tree has changed
+
+    ***************************************************************************/
+
+    public void popFront ()
+    {
+        this.enforceValid();
+        this.iterationStep();
+    }
+
+    /***************************************************************************
+
+        Returns:
+            true if there are no more elements to iterate over
+
+        Throws:
+           RangeInvalidatedException if the underlying tree has changed
+
+    ***************************************************************************/
+
+    public bool empty ()
+    {
+        this.enforceValid();
+        return this.tree.root is null ||
+            (this.item.node is null && this.stack.length == 0);
+    }
+
+    /***************************************************************************
+
+        Returns:
+            true if the underlying tree has not changed and it's safe
+            to use front/popFront/empty methods.
+
+    ***************************************************************************/
+
+    public bool isValid ()
+    {
+        return tree_version == this.tree.content_version;
+    }
+
+    /***************************************************************************
+
+        Ensures that a tree has not changed since the creation of this range.
+
+        Throws:
+           RangeInvalidatedException if the underlying tree has changed
+
+    ***************************************************************************/
+
+    private void enforceValid ()
+    {
+        if (!this.isValid())
+        {
+            throw .range_exception;
+        }
+    }
+
+    /***************************************************************************
+
+        Prepares the range over the tree.
+
+    ***************************************************************************/
+
+    private void start ()
+    {
+        if (this.tree.root !is null)
+        {
+            this.tree_version = this.tree.content_version;
+            this.item = NodeElement(this.tree.root, 0);
+            this.iterationStep();
+        }
+    }
+
+    /***************************************************************************
+
+        Goes to the next element in the tree.
+
+    ***************************************************************************/
+
+    private void iterationStep ()
+    {
+        while (true)
+        {
+            if (this.item.node.is_leaf)
+            {
+                // Leaf doesn't have subtrees, we can just iterate over it
+                if (this.item.index < this.item.node.number_of_elements)
+                {
+                    this.current_value = this.item.value;
+                    this.current_key = this.item.key;
+                    this.item.index++;
+                    return;
+                }
+            }
+            else
+            {
+                // do we have a subtree in the child_elements[index]?
+                if (this.item.index <= this.item.node.number_of_elements)
+                {
+                    *this.stack ~= NodeElement(this.item.node, this.item.index);
+                    this.item = NodeElement(item.node.child_nodes[this.item.index], 0);
+                    continue;
+                }
+            }
+
+            if ((*this.stack).pop(this.item) == false)
+            {
+                return;
+            }
+
+            // We got the item from the stack, and we should see if we
+            // have any more elements to call the delegate on it
+            if (this.item.index < this.item.node.number_of_elements)
+            {
+                this.current_value = this.item.value;
+                this.current_key = this.item.key;
+                this.item.index++;
+                return;
+            }
+            // There's no more elements, just skip to the next one
+            this.item.index++;
+        }
+    }
+}
+
+version (UnitTest)
+{
+    import ocean.math.random.Random;
+    import ocean.core.Test;
+}
+
+unittest
+{
+    // Build a random tree, and use the range over it
+    auto random = new Random();
+    bool found;
+    int res;
+
+    BTreeMap!(int, int, 3) random_tree = BTreeMap!(int, int, 3).init;
+    random_tree.initialize();
+
+    int removed_count;
+    int total_elements;
+    int to_insert = 10_000;
+    size_t my_index;
+
+    // create completely random tree and remove completely random values
+    for (int i = 0; i < to_insert; i++)
+    {
+        int element = random.uniformR(to_insert);
+        // don't count double elements (they are not inserted)
+        total_elements += random_tree.insert(element, element)? 1 : 0;
+        res = random_tree.get(element, found);
+        test(found);
+        test!("==")(res, element);
+    }
+
+    for (int i = 0; i < to_insert; i++)
+    {
+        int element = random.uniformR(to_insert);
+        removed_count += random_tree.remove(element)? 1 : 0;
+        res = random_tree.get(element, found);
+        test(!found);
+    }
+
+    void testRange(void[]* buffer_ptr)
+    {
+        bool started;
+        int previous_value;
+        int remaining_elements;
+
+        for (
+            auto range = byKeyValue(random_tree, buffer_ptr);
+            !range.empty; range.popFront())
+        {
+            if (!started)
+            {
+                previous_value = range.front.value;
+                started = true;
+                remaining_elements++;
+                continue;
+            }
+
+            // enforce that the order is preserved
+            test!(">")(range.front.value, previous_value);
+            previous_value = range.front.value;
+            test!("==")(range.front.value, range.front.key);
+            remaining_elements++;
+        }
+
+        test!("==")(total_elements - remaining_elements, removed_count);
+    }
+
+    void[] buff;
+    testRange(&buff);
+    testRange(null);
+
+    // test invalidation
+    auto range = byKeyValue(random_tree, &buff);
+    auto first = range.front;
+
+    bool insert_res;
+    do
+    {
+        auto val = random.uniformR(2 * to_insert);
+        insert_res = random_tree.insert(val, val);
+    }
+    while (!insert_res);
+    testThrown!(RangeInvalidatedException)(range.popFront());
+}
--- a/src/ocean/util/container/btree/Implementation.d
+++ b/src/ocean/util/container/btree/Implementation.d
@@ -1,0 +1,1196 @@
+/********************************************************************************
+
+    Internal (non-user facing) implementation of BTreeMap.
+
+    Copyright:
+        Copyright (c) 2017 sociomantic labs GmbH. All rights reserved.
+
+    License:
+        Boost Software License Version 1.0. See LICENSE_BOOST.txt for details.
+        Alternatively, this file may be distributed under the terms of the Tango
+        3-Clause BSD License (see LICENSE_BSD.txt for details).
+
+*******************************************************************************/
+
+module ocean.util.container.btree.Implementation;
+
+import ocean.transition;
+
+/*******************************************************************************
+
+    Internal (non-user facing) implementation of BTreeMap.
+
+    Params:
+        KeyType = type of the key
+        ValueType = type of the value
+        tree_degree = degree of the tree.
+
+*******************************************************************************/
+
+package struct BTreeMapImplementation (KeyType, ValueType, int tree_degree)
+{
+    import ocean.core.array.Mutation;
+    import ocean.core.array.Search;
+    import ocean.core.Enforce;
+    import ocean.core.Traits;
+    import ocean.core.Verify;
+    import ocean.util.container.mem.MemManager;
+
+    /**************************************************************************
+
+        Node of the tree. Contains at most (degree * 2 - 1) elements and
+        (degree * 2) subtrees.
+
+    **************************************************************************/
+
+    package struct BTreeMapNode
+    {
+        /**********************************************************************
+
+            KeyValue structure which binds key and a value to be stored in the
+            node.
+
+            NOTE: In addition of ordering notion, this is really the only thing
+            that makes this an implementation of the map, not of the set. If we're
+            going to add set support, we should probably do it via templating
+            this implementation on the actual content of the node's element
+            (replace KeyValue with just a value) and with the ordering operation
+            inserted as a policy (to compare the values, and not the nodes).
+
+        ***********************************************************************/
+
+        private struct KeyValue
+        {
+            KeyType key;
+            // Storing the unqual ValueType here, as we need to reorder
+            // the elements in the node without creating new nodes. User
+            // facing API never sees unqualed type.
+            Unqual!(ValueType) value;
+        }
+
+        /// Number of the elements currently in the node
+        int number_of_elements;
+
+        /// Indicator if the given node is a leaf
+        bool is_leaf;
+
+        /// Array of the elements
+        package KeyValue[tree_degree * 2 - 1] elements;
+
+        /// Array of the pointers to the subtrees
+        package BTreeMapNode*[tree_degree * 2] child_nodes;
+    }
+
+    /**************************************************************************
+
+        Root of the tree.
+
+    ***************************************************************************/
+
+    package BTreeMapNode* root;
+
+    /***************************************************************************
+
+        Type of the element actually stored in the node. The type stored in the
+        node is unqualed version of the type that's accessible to the user, since
+        the container needs to change the content of the actual array, without
+        a need to always allocate a new one.
+
+    ***************************************************************************/
+
+    private alias typeof(BTreeMapNode.elements[0].value) StoredValueType;
+
+    /***************************************************************************
+
+        Convenience constant describing the tree's degree.
+
+    ***************************************************************************/
+
+    private const degree = tree_degree;
+
+    /***************************************************************************
+
+        Memory manager used for allocating and deallocating memory. Refer to
+        ocean.util.container.mem.MemManager for potential options.
+
+    ***************************************************************************/
+
+    package IMemManager allocator;
+
+
+    /***************************************************************************
+
+        "version" of the tree's state. Will be incremented on any removal/adding
+        the element. On the iteration's start, range will record the current 
+        "version" of the tree, and, and before proceeding with the iteration,
+        it will check if any changes were performed on the tree, to prevent
+        iteration over an undefined region.
+
+     ***************************************************************************/
+
+    package ulong content_version;
+
+    /***************************************************************************
+
+        Constructor-like method (as a workaround of a fact that D1 doesn't provide
+        struct constructors).
+
+        Params:
+            allocator = memory manager to use to allocate/deallocate memory
+
+    ***************************************************************************/
+
+    package void initialize (IMemManager allocator = mallocMemManager)
+    {
+        verify (this.allocator is null);
+        this.allocator = allocator;
+    }
+
+
+    /**************************************************************************
+
+        Inserts the element in the tree.
+
+        Params:
+            key = key to insert
+            value = associated value to insert.
+
+        Returns:
+            true if the element with the given key did not exist and it
+            was inserted, false otherwise
+
+    ***************************************************************************/
+
+    package bool insert (KeyType key, ValueType el)
+    {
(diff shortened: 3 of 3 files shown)


src/ocean/util/container/btree/util/Allocators.d
```diff
@@ -0,0 +1,265 @@
+/*******************************************************************************
```
gavin-norman-sociomantic: This module looks similar to https://github.com/sociomantic-tsunami/ocean/blob/v2.x.x/src/ocean/util/container/mem/MemManager.d. Compare and contrast...

src/ocean/util/container/btree/Internal.d
```diff
@@ -0,0 +1,612 @@
...
+
+import ocean.transition;
+import ocean.core.array.Mutation;
+import ocean.core.array.Search;
+import ocean.core.Enforce;
+import ocean.core.Traits;
+
+template BTreeImplementation()
+{
+    /// Inserts the element inside a leaf
```
gavin-norman-sociomantic: Non-standard function doc. (Ditto for several other functions below.)

src/ocean/util/container/btree/Internal.d
```diff
@@ -0,0 +1,612 @@
+/********************************************************************************
+
+    Internal operations over BTree.
+
+    Copyright:
+        Copyright (c) 2009-2017 Sociomantic Labs GmbH.
```
gavin-norman-sociomantic: Copyright 2017. (Check all modules.)

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,695 @@
...
+import ocean.core.Traits;
+
+import ocean.util.container.btree.util.Allocators;
+import ocean.util.container.btree.Internal;
+
+public import ocean.util.container.btree.BTreeRange;
+
+/******************************************************************************
+
+    BTree structure.
```
gavin-norman-sociomantic: Maybe move this comment block to the header and put `/// ditto` here?

gavin-norman-sociomantic: Ah but then what about the `Params:` section... :(

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,695 @@
...
+    package int content_version;
+
+    /// Allocator used to allocate and deallocate memory
+    private Allocator!(BTreeNode, TrackingPolicy) allocator;
+
+    /// Private implementation
+    mixin BTreeImplementation!() impl;
+}
+
+/*
```
gavin-norman-sociomantic: The comments from here to line 294 are out-dated, or?

src/ocean/util/container/btree/Internal.d
```diff
@@ -0,0 +1,612 @@
...
+
+module ocean.util.container.btree.Internal;
+
+import ocean.transition;
+import ocean.core.array.Mutation;
+import ocean.core.array.Search;
+import ocean.core.Enforce;
+import ocean.core.Traits;
+
+template BTreeImplementation()
```
gavin-norman-sociomantic: What's the reason for splitting this into a separate module?

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,695 @@
...
+
+/******************************************************************************
+
+    BTree structure.
+
+    BTree is a rooted tree, with the following properties:
+
+    - Every node has the following attributes:
+
+        x.number_of_elements = number of elements node contains
```
gavin-norman-sociomantic: What does `x.` refer to here?

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,695 @@
...
+        x.elements = the elements themselves
+        x.is_leaf = indicator if the node is the leaf.
+        x.child_nodes = number_of_elements+1 pointers to its children
+
+    - Elements in the node split the ranges of the elements stored
+      in each subtree.
+
+    - All leaves have the same depth
+
+    - Every node other than root must have at least (degree-1) elements.
```
gavin-norman-sociomantic: Maybe explain what "degree" is before using the term?

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,695 @@
...
+    BTree structure.
+
+    BTree is a rooted tree, with the following properties:
+
+    - Every node has the following attributes:
+
+        x.number_of_elements = number of elements node contains
+        x.elements = the elements themselves
+        x.is_leaf = indicator if the node is the leaf.
+        x.child_nodes = number_of_elements+1 pointers to its children
```
gavin-norman-sociomantic: Maybe explain previously the difference between elements and children?

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,695 @@
...
+
+    - Elements in the node split the ranges of the elements stored
```
gavin-norman-sociomantic: "in a node" ? ("the node" implies there's only one, which sounds wrong.)

gavin-norman-sociomantic: Generally I don't understand this sentence.

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,695 @@
...
+      in each subtree.
+
+    - All leaves have the same depth
+
+    - Every node other than root must have at least (degree-1) elements.
+
+    - Every internal node other than the root has at least degree children.
+
+    - Every node may contain at most 2*degree - 1 keys. Therefore, an internal
```
gavin-norman-sociomantic: "keys"?

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,695 @@
...
+      node may have at most 2*degree children.
+
+    Other than insert and delete operations, inorder traversal method is
+    added, which traverses the tree in inorder order (left subtree is visited
```
gavin-norman-sociomantic: "left subtree is visited first, then the element, then the right subtree" -- a couple of questions:
1. This makes it sound like each non-leaf node has exactly two children (the left subtree and the right subtree), which contradicts what is stated above.
2. "the element" makes it sound like each node has one single element. I guess it's a typo and you mean "the elements".

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,695 @@
...
+    first, then the element, then the right subtree). This provides sequential
+    access to the tree. There's a recursive and range-based implementation
+    of the inorder traversal.
+
+    CPU complexity of the operations (n is number of the elements, t is degree,
+        h is the tree height):
```
gavin-norman-sociomantic: Extra indent.

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,695 @@
...
+        - Searching: O(th) = O(t * log_t(n))
+        - Inserting: O(th) = O(t * log_t(n))
+        - Deleting:  O(th) = O(t * log_t(n))
+
+    Complexity of the memory access (to read/write the node from/in memory):
+
+        - Searching O(h) = O(log_t(n))
+        - Inserting O(h) = O(log_t(n))
+        - Deleting: O(h) = O(log_t(n))
+
```
gavin-norman-sociomantic: Would be nice to compare against other tree implementations in ocean. (There's at least the ebtree.)

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,695 @@
...
+    References:
+        - https://en.wikipedia.org/wiki/B-tree
+        - Thomas H. Cormen et al. "Introduction to Algorithms, 3rd edition"
+
+    Params:
+        TreeElementType = type of the element to store.
+        tree_degree = degree of a tree (refer to the documentation)
```
gavin-norman-sociomantic: "the tree"

gavin-norman-sociomantic: `TrackingPolicy` not mentioned.

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,695 @@
...
+
+******************************************************************************/
+
+struct BTree(TreeElementType, int tree_degree, alias Allocator = MallocAllocator,
+        alias TrackingPolicy = TrackAllocationsPolicy)
+{
+    /***********************************************************************
+
+        Recursive inorder iteration. Note that the iteration is undefined
+        in case the tree is changed during the iteration.
```
gavin-norman-sociomantic: Would be nice the expand on that a little. Will the iteration just stop? Will it continue but maybe miss elements? Is it possible elements could be iterated twice?

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,695 @@
...
+            pointer to a copy of the element in the tree, if the element
+            didn't already exist, null otherwise.
+
+    ***************************************************************************/
+
+    public Const!(ElementType)* insertElement (ElementType el)
+    {
+        if (this.root is null)
+        {
+            this.root = insertNewNode();
```
gavin-norman-sociomantic: `this.insertNewNode()`?

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,695 @@
...
+        Params:
+            el = element to insert.
+
+        Returns:
+            pointer to a copy of the element in the tree, if the element
+            didn't already exist, null otherwise.
+
+    ***************************************************************************/
+
+    public Const!(ElementType)* insertElement (ElementType el)
```
gavin-norman-sociomantic: Should `el` be const?

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,695 @@
...
+
+        Params:
+            element = element to delete.
+
+        Returns:
+            true if the element was found and deleted, false otherwise.
+
+     ******************************************************************************/
+
+    public bool removeElement (ref ElementType element)
```
gavin-norman-sociomantic: Why are you passing the element by ref here but not to `insertElement`? Should it be const?

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,695 @@
...
+
+        if (res)
+            this.content_version++;
+
+        return res;
+    }
+
+    /******************************************************************************
+
+        Finds the highest value in the tree.
```
gavin-norman-sociomantic: Highest/largest according to what definition?

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,695 @@
...
+        Params:
+            tree = tree to search in
+            element = element to find in a tree.
+
+        Returns:
+            pointer to the element, or null if not found.
+
+    *******************************************************************************/
+
+    public Const!(ElementType)* search (ref ElementType element)
```
gavin-norman-sociomantic: const?

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,695 @@
...
+    /**************************************************************************
+
+        Node of the tree. Contains degree * 2 - 1 elements and degree * 2
+        subtrees.
+
+    **************************************************************************/
+
+    package struct BTreeNode
+    {
+        /// Number of the elements currently in the node
```
gavin-norman-sociomantic: This contradicts the statement above "Contains degree * 2 - 1 elements". Should that say "at most"?

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,695 @@
...
+        - Thomas H. Cormen et al. "Introduction to Algorithms, 3rd edition"
+
+    Params:
+        TreeElementType = type of the element to store.
+        tree_degree = degree of a tree (refer to the documentation)
+        Allocator = allocation technique to use
+
+******************************************************************************/
+
+struct BTree(TreeElementType, int tree_degree, alias Allocator = MallocAllocator,
```
gavin-norman-sociomantic: What happens if `tree_degree` is negative or 0? :trollface:

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,695 @@
...
+
+    **************************************************************************/
+
+    package struct BTreeNode
+    {
+        /// Number of the elements currently in the node
+        int number_of_elements;
+
+        /// Indicator if the given node is a leaf
+        bool is_leaf;
```
gavin-norman-sociomantic: If this is true, are any conditions about the other fields implied? (invariant?)

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,695 @@
...
+
+        /// Array of the elements
+        package TreeElementType[tree_degree * 2 - 1] elements;
```
gavin-norman-sociomantic: Nice that these are fixed-size arrays :+1:

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,695 @@
...
+        package alias TreeElementType ElementType;
+
+        /// Convenience alias of the tree degree
+        private const degree = tree_degree;
+    }
+
+    /// Convenience alias of the element type
+    package alias TreeElementType ElementType;
+
+    /// degree of a tree
```
gavin-norman-sociomantic: "the tree"

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,695 @@
...
+    package const degree = tree_degree;
+
+    /// Root directory
```
gavin-norman-sociomantic: node?

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,695 @@
...
+    package BTreeNode* root;
+
+    /// "version" of the tree's state. Will be incremented on any removal/adding
+    /// the element. Used to check if the ranges are valid while doing the iteration
```
gavin-norman-sociomantic: "doing an iteration"

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,695 @@
...
+    package int content_version;
```
gavin-norman-sociomantic: Is a signed integer large enough? If it's incremented on each modification, it wouldn't seem crazy that it could wrap under certain usage conditions.

gavin-norman-sociomantic (commented):
First high-level review focussing on doc, API, and some overall design questions.

src/ocean/util/container/btree/BTree.d
nemanja-boric-sociomantic: Yeah, it should be const. What I wanted actually is pass `ref Const!(ElementType)`, but it didn't work in D1 for some reason (I'll reevaluate this), since I don't need a copy of the element, and I don't want to change it. I'll look into normalizing passing convention through the entire module.

nemanja-boric-sociomantic: Ah, I know - it's the artifact from the time when this was a free-standing templated function, the template deduction wouldn't work.

src/ocean/util/container/btree/BTree.d
nemanja-boric-sociomantic: Yeah... I see what I did here. In dlsnode, this structure used to contain only value types. This way it was very convenient when I would create a copy of the element to insert here. This approach doesn't work for the reference types.

src/ocean/util/container/btree/BTree.d
nemanja-boric-sociomantic: I'll look to define this undefined behavior :+1:.

src/ocean/util/container/btree/BTree.d
nemanja-boric-sociomantic: excellent question - I'm not sure :-) :+1:

src/ocean/util/container/btree/util/Allocators.d
nemanja-boric-sociomantic: Ah, I wasn't aware of this one. I'll give my best to use the existing one.

src/ocean/util/container/btree/BTree.d
nemanja-boric-sociomantic: Yes, I will add the invariant for the `Node`.

src/ocean/util/container/btree/BTree.d
nemanja-boric-sociomantic: Yeah, already force-pushed them.

src/ocean/util/container/btree/BTree.d
gavin-norman-sociomantic: Though... as it doesn't have any methods, it wouldn't be run...

nemanja-boric-sociomantic: I think it should have some methods :-). I'll see about abstracting some of the elementary operations on the node to a node's methods.

gavin-norman-sociomantic: Sounds good.

src/ocean/util/container/btree/BTree.d
nemanja-boric-sociomantic: This is tricky. One one hand, we don't worry about the overflow here, since we are only interested in the difference between the current tree's version and the version which was there when the range was created. However, with this in place, it is possible that if you start a range and do 2^32 operations, while the range is still active, game over. I don't think this edge case should be allowed in a general purpose data structure. I'll see what I can do.

nemanja-boric-sociomantic: Perhaps just using long (ulong really, we want defined wraparound) would be sufficient here. I'll see if I can make it cheap and defined, anyway.

src/ocean/util/container/btree/BTree.d
nemanja-boric-sociomantic: I thought about this one, and I think it needs to stay as it is. In case of value-types, this would create a copy of the value type, which we'll put into the tree. In case of the reference type, we will do the same - this will provide the copy of the reference and it would be put into the tree as well. My feeling about this being `const` is that it's too harsh - the object itself doesn't need to be const, _except_ when it comes to ordering - ordering needs to stay the same. Templating `BTree` on `Const!(Object)` would provide the guarantee to the user, if they wish. What do you think?

gavin-norman-sociomantic:
>Move to ocean.util.container.mem.MemManager

Cool that this was possible!

nemanja-boric-sociomantic:
Squashed and ready for the next review round @gavin-norman-sociomantic .

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,507 @@
...
+public import ocean.util.container.btree.BTreeRange;
+
+/******************************************************************************
+
+    BTree structure.
+
+
+    BTree is a rooted tree, with the following properties:
+
+    - There's a characteristic of the tree, tree's degree, which indicates
```
gavin-norman-sociomantic: Clearer: "A BTree has a characteristic called "degree". This indicates how many items are stored in each node of the tree. Having a high degree..."

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,507 @@
...
+      how many items are stored in each node. Having a high degree makes the
+      tree shorter, and allows prefetching many elements from the memory in
+      one go.
+
+
+    - The layout of one node of the tree is similar to the following:
```
gavin-norman-sociomantic: "similar to" makes it sound like this is a trick and not actually the layout :D

"is as follows" ?

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,507 @@
...
+
+      -----------------------------------------------------------------
+      | p1 | element1 | p2 | element2 | ... | pn-1 | element n-1 | pn |
+      -----------------------------------------------------------------
+
+      Every element in the node (for the non-leaf nodes) is surrounded by
+      two pointers. We call these pointers child nodes. Each pointer is a root
+      of a subtree, for which the following stands: all the elements in a tree
```
gavin-norman-sociomantic: "for which the following holds true"

gavin-norman-sociomantic: I think it would be clearer to move this to another "section":
```
- The layout of one node of the tree is as follows:
  -----------------------------------------------------------------
  | p1 | element1 | p2 | element2 | ... | pn-1 | element n-1 | pn |
  -----------------------------------------------------------------

  Every element in the node (for the non-leaf nodes) is surrounded by
  two pointers. We call these pointers child nodes.

- Ordering of elements: each child node pointer is the root of a subtree,
  for which the following holds true: all the elements in the tree ...
```

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,507 @@
...
+      pointed by p(i-1) (pointer left of the elements) are smaller (in respect
+      to the ordering defined by the element's type) of the element(i), and
+      all the elements in a subtree pointed by p(i) are larger than a element(i).
+
+    - Every node has the following attributes:
+
+        node.number_of_elements = number of elements node contains
```
gavin-norman-sociomantic: Is this really stored or is it a constant? It sounds like this should == degree, which is a constant. Maybe the explanation of degree needs to make it clear that this is the _maximum_ number of elements in one node?

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,507 @@
...
+        node.elements = the elements themselves
+        node.is_leaf = indicator if the node is the leaf.
```
gavin-norman-sociomantic: "a leaf"

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,507 @@
...
+        node.child_nodes = number_of_elements+1 pointers to it the subtrees whose
```
gavin-norman-sociomantic: Typo "to it the"

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,507 @@
...
+            roots are this node. We refer to these nodes as child-nodes (this
+            node is a direct parent of them).
+
+    - All leaves have the same depth
+
+    - Every node other than root must have at least (degree-1) elements.
```
gavin-norman-sociomantic: And at most degree?

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,507 @@
...
+
+    - Every internal node other than the root has at least degree children.
```
gavin-norman-sociomantic: They can have more than degree children?

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,507 @@
...
+
+    - Every node may contain at most 2*degree - 1 elements. Therefore, an internal
```
gavin-norman-sociomantic: Ah ok, this contradicts what was written above about degree...

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,507 @@
...
+      node may have at most 2*degree children.
+
+    Other than insert and delete operations, inorder traversal method is added,
```
gavin-norman-sociomantic: "is added" is unusual wording. It implies that something has changed.

src/ocean/util/container/btree/BTree.d
gavin-norman-sociomantic: An idea: maybe a README in `util.container.tree`?

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,507 @@
...
+
+******************************************************************************/
+
+struct BTree(TreeElementType, int tree_degree)
+{
+    static assert (tree_degree > 0);
+
+    /**************************************************************************
+
+        Constructor.
```
gavin-norman-sociomantic: Params

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,507 @@
...
+    ***************************************************************************/
+
+    package void init (IMemManager allocator = mallocMemManager)
+    {
+        this.impl.init(allocator);
+    }
+
+    /**************************************************************************
+
+        Inserts the element in the tree.
```
gavin-norman-sociomantic: The element seems to be copied by value into the tree. It'd be nice to explicitly mention this, for user convenience.

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,507 @@
...
+
+        Params:
+            element = element to delete.
+
+        Returns:
+            true if the element was found and deleted, false otherwise.
+
+     ******************************************************************************/
+
+    public bool remove (ref ElementType element)
```
gavin-norman-sociomantic: Would be useful to note why this is ref.

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,507 @@
...
+    {
+        return this.impl.remove(element);
+    }
+
+    /******************************************************************************
+
+        Finds the highest value in the tree.
+
+        Returns:
+            largest element in the tree, or ElementType.init if the tree is empty.
```
gavin-norman-sociomantic: Using `ElementType.init` as the null value could be tricky with certain types. e.g. int: 0 is both a valid highest value and the null value.

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,507 @@
...
+
+        Searches the tree for the element and returns the pointer to it, or null
+        if not found.
+
+        Params:
+            element = element to find in a tree.
+            found = indicator if the element was found
+
+        Returns:
+            pointer to the element, or ElementType.init if not found.
```
gavin-norman-sociomantic: Sounds impossible.

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,507 @@
...
+    /******************************************************************************
+
+        Finds the highest value in the tree.
+
+        Returns:
+            largest element in the tree, or ElementType.init if the tree is empty.
+
+    ******************************************************************************/
+
+    public ElementType findMaximum ()
```
gavin-norman-sociomantic: Why no `findMinimum` btw?

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,507 @@
...
+
+    public ElementType search (ref ElementType element, out bool found)
+    {
+        return this.impl.search(element, found);
+    }
+
+    /***********************************************************************
+
+        Recursive inorder iteration. Note that, in case the tree is changed
+        during the iteration, iteration will not continue.
```
gavin-norman-sociomantic: "iteration will halt" is a bit clearer.

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,507 @@
...
+
+    ***********************************************************************/
+
+    public int opApply (int delegate (ref ElementType) dg)
+    {
+        return this.impl.inorder(dg);
+    }
+
+
+    alias BTreeImplementation!(ElementType, tree_degree)
```
gavin-norman-sociomantic: Doc

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,507 @@
...
+        Implementation;
+
+    alias TreeElementType ElementType;
+
+    /// Private implementation
+    package Implementation impl;
+}
+
```
gavin-norman-sociomantic: Maybe add a documented unittest / usage example here?

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,507 @@
...
+    Unittests. Compile with -debug=BTreeSanity to turn on
+    the invariant checks for the tree.
+
+*/
+
+unittest
+{
+    BTree!(File, 2) tree;
+    tree.init();
+    assert (tree.findMaximum == File.init);
```
gavin-norman-sociomantic: Any reason not to use `test`?

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,507 @@
...
+    License:
+        Boost Software License Version 1.0. See LICENSE_BOOST.txt for details.
+        Alternatively, this file may be distributed under the terms of the Tango
+        3-Clause BSD License (see LICENSE_BSD.txt for details).
+
+*******************************************************************************/
+
+module ocean.util.container.btree.BTree;
+
+import ocean.transition;
```
gavin-norman-sociomantic: Do any of these need to be imported at module scope?

src/ocean/util/container/btree/BTree.d
```diff
@@ -0,0 +1,507 @@
...
+import ocean.core.array.Mutation;
+import ocean.core.array.Search;
+import ocean.core.Enforce;
+import ocean.core.Traits;
+
+import ocean.util.container.mem.MemManager;
+import ocean.util.container.btree.Implementation;
+
+public import ocean.util.container.btree.BTreeRange;
```
gavin-norman-sociomantic: Probably would help to note the existence of this module in the doc (at a high level) somewhere. Otherwise, the user will only discover it by chance.

src/ocean/util/container/btree/BTreeRange.d
```diff
@@ -0,0 +1,355 @@
...
+
+module ocean.util.container.btree.BTreeRange;
+
+import ocean.transition;
+import ocean.core.array.Mutation;
+import ocean.util.container.btree.BTree;
+
+/*******************************************************************************
+
+    Exception thrown from the BTreeRange if the iteration over the changed
```
gavin-norman-sociomantic: "if a tree is modified during iteration" ?

src/ocean/util/container/btree/BTreeRange.d
```diff
@@ -0,0 +1,355 @@
...
+
+static this ()
+{
+    .range_exception = new RangeInvalidatedException(
+            "Container has been changed during iteration, please restart.");
+}
+
+/*******************************************************************************
+
+    Returns a range that iterates the tree via the inorder.
```
gavin-norman-sociomantic: "that performs an inorder iteration over the specified tree"

src/ocean/util/container/btree/BTreeRange.d
```diff
@@ -0,0 +1,355 @@
...
+
+    If tree has changed during the iteration, range will throw an
+    RangeInvalidatedException. If the user knows that the tree will not change
```
gavin-norman-sociomantic: I'm not sure the 2nd sentence says anything that the 1st hasn't already said.

src/ocean/util/container/btree/BTreeRange.d
```diff
@@ -0,0 +1,355 @@
...
+    during the range traversal, it is safe to call range's methods - it will throw
+    an exception if this promise is broken. In case user knows that the tree
+    can be changed any time, it is possible to call BTreeRange.isValid before
+    performing any action on the range to check if the accessing is safe. This
```
gavin-norman-sociomantic: What does "safe" mean here?

src/ocean/util/container/btree/BTreeRange.d
```diff
@@ -0,0 +1,355 @@
...
+    is only true for the single-threaded environment.
+
+    Params:
+        tree = tree to iterate over
+        buff = buffer for storing the state of the iteration. Useful
```
gavin-norman-sociomantic: Maybe specify what magnitude of data might be stored here. (A few bytes? A few Mb? A copy of the whole tree?)

src/ocean/util/container/btree/BTreeRange.d
```diff
@@ -0,0 +1,355 @@
...
+    is only true for the single-threaded environment.
+
+*******************************************************************************/
+
+public struct BTreeRange(BTree)
+{
+    /// Root node of a tree
+    private BTree.Implementation* tree;
+
+    /// "Version" of the tree at the iteration time.
```
gavin-norman-sociomantic: "Copy of the content_version field of the tree at the time iteration began" ?

src/ocean/util/container/btree/BTreeRange.d
```diff
@@ -0,0 +1,355 @@
...
+    /// Used to check if the iterator is invalid
+    private ulong tree_version;
+
+    /// Pair of the node/index for which we've paused iteration and pushed
+    /// to stack
```
gavin-norman-sociomantic: The "for which" part isn't always true, I think. (See the `stack` field.)

src/ocean/util/container/btree/BTreeRange.d
```diff
@@ -0,0 +1,355 @@
...
+    private struct IterationItem
+    {
+        BTree.Implementation.BTreeNode* node;
+        size_t index;
+    }
+
+    /// Stack of the nodes that we've postponed iteration over
```
gavin-norman-sociomantic: Maybe add "used by ..." ?

src/ocean/util/container/btree/BTreeRange.d
```diff
@@ -0,0 +1,355 @@
...
+    /// "Version" of the tree at the iteration time.
+    /// Used to check if the iterator is invalid
+    private ulong tree_version;
+
+    /// Pair of the node/index for which we've paused iteration and pushed
+    /// to stack
+    private struct IterationItem
+    {
+        BTree.Implementation.BTreeNode* node;
+        size_t index;
```
gavin-norman-sociomantic: Index of what exactly?

src/ocean/util/container/btree/BTreeRange.d
```diff
@@ -0,0 +1,355 @@
...
+    }
+
+    /// Stack of the nodes that we've postponed iteration over
+    private IterationItem[]* stack;
+
+    /// Current item we're on
```
gavin-norman-sociomantic: Is there a more specific term than "item"?

src/ocean/util/container/btree/BTreeRange.d
```diff
@@ -0,0 +1,355 @@
...
+    private IterationItem item;
+
+    /// Current element over which we're iterating
```
gavin-norman-sociomantic: Current element within the current `<something>`?

src/ocean/util/container/btree/BTreeRange.d
```diff
@@ -0,0 +1,355 @@
...
+
+    public BTree.ElementType front ()
+    {
+        this.ensureValid();
+        return *this.current_element;
+    }
+
+    /***************************************************************************
+
+        Pops the next element from the tree.
```
gavin-norman-sociomantic: "from the range"? (Not actually removing from the tree.)

src/ocean/util/container/btree/BTreeRange.d
```diff
@@ -0,0 +1,355 @@
...
+    /***************************************************************************
+
+        Ensures that a tree has not changed since the creation of this range.
+
+        Throws:
+           RangeInvalidatedException if the underlying tree has changed
+
+    ***************************************************************************/
+
+    private void ensureValid ()
```
gavin-norman-sociomantic: `enforceValid`?

src/ocean/util/container/btree/BTreeRange.d
```diff
@@ -0,0 +1,355 @@
...
+            // new item, call delegate with the element if that's ok:
+            if (item.index++ < this.item.node.number_of_elements)
+            {
+                this.current_element = &this.item.node.elements[this.item.index-1];
+                return;
+            }
+        }
+    }
+}
+
```
gavin-norman-sociomantic: Documented unittest / usage example?

src/ocean/util/container/btree/Implementation.d
```diff
@@ -0,0 +1,1121 @@
...
+    package BTreeNode* root;
+
+    /**************************************************************************
+
+        Node of the tree. Contains at most (degree * 2 - 1) elements and
+        (degree * 2) subtrees.
+
+    **************************************************************************/
+
+    package struct BTreeNode
```
gavin-norman-sociomantic: Slightly more sensible to define this struct before using it (`root`).

src/ocean/util/container/btree/Implementation.d
```diff
@@ -0,0 +1,1121 @@
...
+        /// Array of the elements
+        package Unqual!(ElementType)[tree_degree * 2 - 1] elements;
+
+        /// Array of the pointers to the subtrees
+        package BTreeNode*[tree_degree * 2] child_nodes;
+    }
+
+    /***************************************************************************
+
+        Type of the element actually stored in the node. The type stored in the
```
gavin-norman-sociomantic: I'd like D2 const masters to check this out.

@mihails-strasuns-sociomantic @mathias-lang-sociomantic -- casting away qualifiers inside a container... kosher or suspect?

src/ocean/util/container/btree/Implementation.d
```diff
@@ -0,0 +1,1121 @@
...
+
+import ocean.transition;
+import ocean.core.array.Mutation;
+import ocean.core.array.Search;
+import ocean.core.Enforce;
+import ocean.core.Traits;
+import ocean.util.container.mem.MemManager;
+
+
+/// ditto
```
gavin-norman-sociomantic: Params :(

src/ocean/util/container/btree/Implementation.d
```diff
@@ -0,0 +1,1121 @@
...
+        it will check if any changes were performed on the tree, to prevent
+        iteration over an undefined region.
+
+     ***************************************************************************/
+
+    package ulong content_version;
+
+    /***************************************************************************
+
+        Constructor's like initialization memory. Prepares the tree to be used.
```
gavin-norman-sociomantic: Comprehensibility of the first sentence is low ;)

src/ocean/util/container/btree/Implementation.d
```diff
@@ -0,0 +1,1121 @@
...
+            parent = parent of the node (needed when merging the neighbour nodes)
+            element = element to remove
+            rebalance_parent = indicator if while traversing back to the root, the
+                               parent node should be rebalanced
+
+         Returns: true if the current node needs to be rebalanced
+
+    ******************************************************************************/
+
+    private bool deleteFromNode ( BTreeNode* node, BTreeNode* parent,
```
gavin-norman-sociomantic: I notice a lot of methods that accept a `BTreeNode*` as their first argument. Would it make sense to place these inside `BTreeNode`?

gavin-norman-sociomantic: I also wonder about the pros and cons of having `BTreeNode` nested. It seems the only real reason is to share the template args of the outer struct. A pro of moving it outside would be (perhaps) increased code readability.

gavin-norman-sociomantic (commented):
2nd pass. Getting into mid-level stuff and touching on a few broad topics re: the lowest level.

src/ocean/util/container/btree/BTree.d
nemanja-boric-sociomantic: This is actually stored. Remember that the we're allocating nodes as a static arrays. This keeps track of how many elements are _currently_ in the node. (Maximum is always specified as 2*degree-1).

src/ocean/util/container/btree/BTree.d
nemanja-boric-sociomantic: Ah, yes, of course, sorry for the confusion.

src/ocean/util/container/btree/BTree.d
nemanja-boric-sociomantic: I didn't need it. Will add it for completeness.

src/ocean/util/container/btree/BTree.d
nemanja-boric-sociomantic: Ah, I think the API should be

```
bool findMaximum (out ElementType max)
```

nemanja-boric-sociomantic: (there's one method, I think `search` where I return the indicator if the element was found. I think this is better way).

src/ocean/util/container/btree/BTree.d
nemanja-boric-sociomantic: Whoops. It's a leftover. I'll change all asserts here in test.

src/ocean/util/container/btree/BTreeRange.d
nemanja-boric-sociomantic: I'll clarify more. It means that either fetching the current element, or moving to the next one is guaranteed to succeed in a single-thread environment.

src/ocean/util/container/btree/BTree.d
gavin-norman-sociomantic: Yeah, sounds more generic.

src/ocean/util/container/btree/BTreeRange.d
nemanja-boric-sociomantic: I will clarify. It just stores at most `tree_height * sizeof(pointer_to_a_node)` which should be really small, since tree height of this tree is small.

src/ocean/util/container/btree/Implementation.d
nemanja-boric-sociomantic: Yeah good idea. The only way around it I could think is that we recreate this array on every change.

src/ocean/util/container/btree/Implementation.d
nemanja-boric-sociomantic: Might be a good idea, I will evaluate this carefuly.

nemanja-boric-sociomantic:
Thanks for the great comments @gavin-norman-sociomantic !

src/ocean/util/container/btree/BTreeRange.d
gavin-norman-sociomantic: Oh I see, "iteration item". I wonder whether the struct might be better named something like `NodeElement` or something that indicates what it's actually "pointing" to.

nemanja-boric-sociomantic: Good one.

src/ocean/util/container/btree/BTreeRange.d
nemanja-boric-sociomantic: Correct, from the range, not from the tree.

src/ocean/util/container/btree/BTreeRange.d
nemanja-boric-sociomantic: This is now tricky. Because it doesn't _enforce_ anything, it just ensures that the range is valid before accessing it.

nemanja-boric-sociomantic: I mean, it's not tricky, that was just my reasoning :-)

gavin-norman-sociomantic: My reasoning was just that it does exactly what `enforce` does (checks a condition and throws, if it's false).

nemanja-boric-sociomantic: You are correct, of course.

nemanja-boric-sociomantic:
@gavin-norman-sociomantic I've squashed the changes, so it's ready again.

The most important change is that I moved from the set to the map. This makes the API more clear and akin to other containers in `ocean.util.containers.*`. I don't think adding set would be much work, but I would leave that for latter. Some renamings are done, and the iteration now iterates over the (key, value) pairs.

src/ocean/util/container/btree/Implementation.d
nemanja-boric-sociomantic: Ok, I've looked into this and I think these methods should stay outside the structure. Reason is that it looks much more natural to look into the implementation, knowing these methods are using recursion to do work, when you're explicitly passing the node on which they are going to work.

src/ocean/util/container/btree/Implementation.d
nemanja-boric-sociomantic: haha, I've just noticed this. Epic sentence.

src/ocean/util/container/btree/Implementation.d
gavin-norman-sociomantic: I've lost track of this a bit. `deleteFromNode` is now _inside_ a struct  (`BTreeMapImplementation`). I'm not sure if that was the one I was referring to originally. This seems to contradict your comment:
>I've looked into this and I think these methods should stay outside the structure

nemanja-boric-sociomantic: Sorry, I was referring to the fact that these methods are not nested inside the struct they are operating on (`BTreeNode`). This is just a style issue, and tbh, it feels more natural to define the methods that are operating recursively over the data structure as to have their node explicitly passed as the first argument. If you look into `deleteFromNode` it starts from the node that's given (usually we start with `root` and then it either finds the element in the node, or descends into one of the subtrees using recursion `return deleteFromNode(subtree_root, element)`. I feel this is much more natural (and I'm biased because this is the usually how the algorithms over the data structure are represented) than `subtree_root.deleteFromNode(element)`.

src/ocean/util/container/btree/BTreeMap.d
```diff
@@ -0,0 +1,664 @@
...
+                if (o is null)
+                    return -1;
+                if (this.x >= o.x)
+                    return this.x > o.x;
+                return -1;
+            }");
+        }
+        else
+        {
+            int opCmp(typeof(this) rhs)
```
mihails-strasuns-sociomantic: coverage check shows this method is not used

src/ocean/util/container/btree/BTreeMap.d
```diff
@@ -0,0 +1,664 @@
...
+        3-Clause BSD License (see LICENSE_BSD.txt for details).
+
+*******************************************************************************/
+
+module ocean.util.container.btree.BTreeMap;
+
+import ocean.transition;
+import ocean.util.container.mem.MemManager;
+
+public import ocean.util.container.btree.BTreeMapRange;
```
mihails-strasuns-sociomantic: I am not a fan of public imports in non-`package.d` modules. Explicitly importing range style API when you need it fits our code style better.

src/ocean/util/container/btree/BTreeMap.d
```diff
@@ -0,0 +1,664 @@
...
+      -----------------------------------------------------------------
+      | p1 | element1 | p2 | element2 | ... | pn-1 | element n-1 | pn |
+      -----------------------------------------------------------------
+
+      Every element in the node (for the non-leaf nodes) is surrounded by
+      two pointers. We call these pointers child nodes.
+
+    - Each pointer is a root of a subtree, for which the following holds true: all
+      the elements in a tree pointed by p(i-1) (pointer left of the elements) are
+      smaller (in respect to the ordering defined by the element's type) of the
```
gavin-norman-sociomantic: Does this discussion of ordering need to be updated now to account for the presence of keys?

src/ocean/util/container/btree/BTreeMap.d
```diff
@@ -0,0 +1,664 @@
...
+
+    - Every node other than root must have at least (degree-1) elements.
+
+    - Every internal node other than the root has at least degree children.
+
+    - Every node has the following attributes:
+
+        node.number_of_elements = number of elements node currently contains
+            (as nodes are statically allocated block, the number of the elements
+             may not necessarily be maximal.
```
gavin-norman-sociomantic: Closing ) missing.

src/ocean/util/container/btree/BTreeMap.d
```diff
@@ -0,0 +1,664 @@
...
+    - All leaves have the same depth
+
+    Other than insert and delete operations, inorder traversal method is provided,
+    which traverses the tree in inorder order (left subtree is visited first,
+    then the element separating these subtrees, then the right subtree, then
+    the next element, etc.). This provides sequential access to the tree.
+    There's a recursive and range-based implementation of the inorder
+    traversal.
+
+    Range-based traversal is supported through `BTreeMapRange` structure, found in
```
gavin-norman-sociomantic, line 75: "found in ocean.util.container.btree.Btree and publicly imported here."

src/ocean/util/container/btree/BTreeMap.d
```diff
@@ -0,0 +1,664 @@
...
+      because the node contains large number of elements, it allows prefetching
+      many elements from the memory in one go.
+
+    - The layout of one node of the tree is as follows:
+
+      -----------------------------------------------------------------
+      | p1 | element1 | p2 | element2 | ... | pn-1 | element n-1 | pn |
+      -----------------------------------------------------------------
+
+      Every element in the node (for the non-leaf nodes) is surrounded by
```
gavin-norman-sociomantic: Maybe mention here that each element consists of a key and a value.

src/ocean/util/container/btree/Implementation.d
```diff
@@ -0,0 +1,1259 @@
...
+    private const degree = tree_degree;
+
+    /***************************************************************************
+
+        Memory manager used for allocating and deallocating memory. Refer to
+        ocean.util.container.mem.MemManager for potential options.
+
+    ***************************************************************************/
+
+    package IMemManager allocator;
```
gavin-norman-sociomantic, line 118: Should probably `verify` that this is non-null before using it. (It's possible to use an instance without calling `init`.)

src/ocean/util/container/btree/Implementation.d
```diff
@@ -0,0 +1,1259 @@
...
+        struct constructors).
+
+        Params:
+            allocator = memory manager to use to allocate/deallocate memory
+
+    ***************************************************************************/
+
+    package void init (IMemManager allocator = mallocMemManager)
+    {
+        assert (this.allocator is null);
```
gavin-norman-sociomantic: `verify`

src/ocean/util/container/btree/Implementation.d
gavin-norman-sociomantic: An alternative would be to remove `init`, rely on this field being set when constructing an instance, and add an invariant that checks it's non-null.

src/ocean/util/container/btree/BTreeMap.d
```diff
@@ -0,0 +1,664 @@
...
+
+    /**************************************************************************
+
+        Inserts the element in the tree. This is passing the copy of the
+        element into the tree, so it's not necessary to manually create copy of
+        the element.  Note that for the reference types, this will just copy
+        the reference.
+
+        Params:
+            el = element to insert.
```
gavin-norman-sociomantic: Check params doc for all methods that deal with keys now.

src/ocean/util/container/btree/BTreeMap.d
```diff
@@ -0,0 +1,664 @@
...
+
+    public bool insert (KeyType key, ElementType el)
+    {
+        return this.impl.insert(key, el);
+    }
+
+    /******************************************************************************
+
+        Deletes the element from the BTreeMap. The element is passed by reference,
+        since we'll just use it for the comparison with the value stored in the tree.
```
gavin-norman-sociomantic: No longer true.

src/ocean/util/container/btree/BTreeMap.d
```diff
@@ -0,0 +1,664 @@
...
+    {
+        return this.impl.remove(key);
+    }
+
+    /******************************************************************************
+
+        Finds the highest value in the tree.
+
+        Params:
+            el = out parameter to be filled with the copy of the highest element
```
gavin-norman-sociomantic: Params/returns backwards. Ditto for `findMinimum`. What was the reasoning for switching these?

src/ocean/util/container/btree/BTreeMap.d
```diff
@@ -0,0 +1,664 @@
...
+    *******************************************************************************/
+
+    public ElementType search (KeyType key, out bool found_element)
+    {
+        return this.impl.search(key, found_element);
+    }
+
+    /***********************************************************************
+
+        Recursive inorder iteration. Note that, in case the tree is changed
```
gavin-norman-sociomantic: Please add "over keys and values".

src/ocean/util/container/btree/BTreeMap.d
```diff
@@ -0,0 +1,664 @@
...
+    ***********************************************************************/
+
+    public int opApply (int delegate (ref KeyType value, ref ElementType) dg)
+    {
+        return this.impl.inorder(dg);
+    }
+
+    /***********************************************************************
+
+        Recursive inorder iteration. Note that, in case the tree is changed
```
gavin-norman-sociomantic: Please add "over values only".

src/ocean/util/container/btree/BTreeMap.d
```diff
@@ -0,0 +1,664 @@
...
+/*******************************************************************************
+
+    Constructor-like method. Constructs the BTreeMap and initializes it.
+
+    Params:
+        KeyType = type of the key.
+        ElementType = type of the element to store.
+        tree_degree = degree of the tree (refer to the documentation)
+        memManager = memory management strategy to use
+
```
gavin-norman-sociomantic: Returns

src/ocean/util/container/btree/BTreeMap.d
```diff
@@ -0,0 +1,664 @@
...
+public BTreeMap!(KeyType, ElementType, tree_degree) createBTreeMap
+    (KeyType, ElementType, int tree_degree)(IMemManager allocator = mallocMemManager)
+{
+    BTreeMap!(KeyType, ElementType, tree_degree) tree;
+    tree.init(allocator);
+    return tree;
+}
+
+
+///
```
gavin-norman-sociomantic: I wonder if this should be a usage example for the struct, not for the constructor function.

src/ocean/util/container/btree/BTreeMap.d
gavin-norman-sociomantic: Interesting. This searches for the largest value.
1. Is this useful? The key determines the ordering, so I would have thought the largest key would be the expected function.
2. Whatever the search method, the user will require the key & value returned.

src/ocean/util/container/btree/BTreeMap.d
```diff
@@ -0,0 +1,664 @@
...
+            needle = element to find in a tree.
+            found_element = out parameter to be filled with the copy of the found
+                element.
+
+        Returns:
+            true if the element is found, false otherwise
+
+    *******************************************************************************/
+
+    public ElementType search (KeyType key, out bool found_element)
```
gavin-norman-sociomantic: Maybe `get` would be a better name. `search` has a slight implication of just checking whether something exists, not necessarily returning its value.

src/ocean/util/container/btree/BTreeMapRange.d
```diff
@@ -0,0 +1,388 @@
...
+*******************************************************************************/
+
+class RangeInvalidatedException: Exception
+{
+    import ocean.core.Exception: DefaultExceptionCtor;
+    mixin DefaultExceptionCtor;
+}
+
+// shared between different ranges
+private static RangeInvalidatedException range_exception;
```
gavin-norman-sociomantic: `static` is meaningless here.

src/ocean/util/container/btree/BTreeMapRange.d
```diff
@@ -0,0 +1,388 @@
...
+    If tree has changed during the iteration, range will throw an
+    RangeInvalidatedException. In case user knows that the tree can be changed
+    any time, it is possible to call BTreeMapRange.isValid before performing any
+    action on the range to check if fetching the front element, or moving to
+    the next one is guaranteed to success. This is only true for the
+    single-threaded environment.
+
+    Params:
+        tree = tree to iterate over
+        buff = buffer for storing the state of the iteration. Useful
```
gavin-norman-sociomantic, line 54: Mention behaviour if null.

src/ocean/util/container/btree/BTreeMapRange.d
```diff
@@ -0,0 +1,388 @@
...
+    /// Pair of the node/index for which we've paused iteration and pushed
+    /// to stack
+    private struct NodeElement
+    {
+        BTreeMap.Implementation.BTreeMapNode* node;
+        size_t index;
+    }
+
+    /// Stack of the nodes that we've postponed iteration over
+    private NodeElement[]* stack;
```
gavin-norman-sociomantic: Could an invariant check that this is never null?

gavin-norman-sociomantic: Invariant checking that this is never null?

src/ocean/util/container/btree/BTreeMapRange.d
```diff
@@ -0,0 +1,388 @@
...
+
+    /// Current (node, element_index) pair we're on
+    private NodeElement item;
+
+    /// Current Element (item.node[item.element_index].value)
```
gavin-norman-sociomantic: "Current value"

gavin-norman-sociomantic: Would it make sense to make `current_element` and `current_key` methods of `NodeElement`? (i.e. `key` and `value`.)

src/ocean/util/container/btree/BTreeMap.d
```diff
@@ -0,0 +1,664 @@
...
+        - Inserting O(h) = O(log_t(n))
+        - Deleting: O(h) = O(log_t(n))
+
+    References:
+        - https://en.wikipedia.org/wiki/B-tree
+        - Thomas H. Cormen et al. "Introduction to Algorithms, 3rd edition"
+
+    Params:
+        KeyType = type of the key.
+        TreeElementType = type of the element to store.
```
gavin-norman-sociomantic: The terminology around values / elements could be tidied up. I'd suggest that an "element" is a key/value pair. Otherwise always refer to values (not elements).

src/ocean/util/container/btree/BTreeMapRange.d
```diff
@@ -0,0 +1,388 @@
...
+
+        Returns:
+            the current element in the range.
+
+        Throws:
+           RangeInvalidatedException if the underlying tree has changed
+
+    ***************************************************************************/
+
+    public Pair!(BTreeMap.KeyType, BTreeMap.ElementType) front ()
```
gavin-norman-sociomantic: `alias Pair!(BTreeMap.KeyType, BTreeMap.ElementType) KeyValue` (or something)?

src/ocean/util/container/btree/BTreeMap.d
```diff
@@ -0,0 +1,664 @@
...
+        - Searching O(h) = O(log_t(n))
+        - Inserting O(h) = O(log_t(n))
+        - Deleting: O(h) = O(log_t(n))
+
+    References:
+        - https://en.wikipedia.org/wiki/B-tree
+        - Thomas H. Cormen et al. "Introduction to Algorithms, 3rd edition"
+
+    Params:
+        KeyType = type of the key.
```
gavin-norman-sociomantic: `TreeKeyType`

src/ocean/util/container/btree/BTreeMapRange.d
gavin-norman-sociomantic: Maybe expand this a bit. Something like: "When interationStep() moves from a node to iterate over its children, the parent node is pushed to the stack.".

src/ocean/util/container/btree/BTreeMapRange.d
```diff
@@ -0,0 +1,388 @@
...
+        }
+    }
+
+    /***************************************************************************
+
+        Goes to the next element in the tree.
+
+    ***************************************************************************/
+
+    private void iterationStep ()
```
gavin-norman-sociomantic, line 279: It occurs to me that the logic in this method -- moving from one node-element to the next, in iteration order -- must be duplicated somewhere inside the foreach functions of the tree implementation. Is there any sensible way this could be abstracted?

gavin-norman-sociomantic (commented):
Next pass. Looking into the range.

src/ocean/util/container/btree/BTreeMap.d
nemanja-boric-sociomantic: Yes, you're right. I should mention this now.

src/ocean/util/container/btree/BTreeMap.d
nemanja-boric-sociomantic: Oh, I'm going to do a documentation scan for the key/value.

src/ocean/util/container/btree/BTreeMap.d
nemanja-boric-sociomantic: Yes, this is more useful for the set, it's not very useful now. I'll remove it, since the ordering is done by keys and this would return the item with the highest key.

src/ocean/util/container/btree/BTreeMap.d
nemanja-boric-sociomantic: Very nice point, I'll do it.

src/ocean/util/container/btree/BTreeMap.d
nemanja-boric-sociomantic: Yes, but I would highly discourage using this struct without constructor function. I think that can be doable with removing `init` function and adding invariant.

src/ocean/util/container/btree/Implementation.d
nemanja-boric-sociomantic: Ha! :-)

src/ocean/util/container/btree/BTreeMapRange.d
nemanja-boric-sociomantic: Sorry, C habits die hard.

src/ocean/util/container/btree/BTreeMapRange.d
nemanja-boric-sociomantic: I'll look into this. The logic inside the tree uses the call-stack for the recursion, while this allows us to suspend the iteration and resume it later (unless the map has changed in the meantime).

src/ocean/util/container/btree/BTreeMap.d
```diff
@@ -0,0 +1,664 @@
...
+    which traverses the tree in inorder order (left subtree is visited first,
+    then the element separating these subtrees, then the right subtree, then
+    the next element, etc.). This provides sequential access to the tree.
+    There's a recursive and range-based implementation of the inorder
+    traversal.
+
+    Range-based traversal is supported through `BTreeMapRange` structure, found in
+    (and here publicly imported) in ocean.util.container.btree.Btree.
+
+    CPU complexity of the operations (n is number of the elements, t is degree,
```
mihails-strasuns-sociomantic: More useful is to state complexity for reach public method in its own DDOC

src/ocean/util/container/btree/BTreeMap.d
```diff
@@ -0,0 +1,664 @@
...
+    /**************************************************************************
+
+        Constructor.
+
+        Params:
+            allocator = memory manager to use to allocate/deallocate memory
+
+    ***************************************************************************/
+
+    package void init (IMemManager allocator = mallocMemManager)
```
mihails-strasuns-sociomantic: It is not a good idea to name method `init` because of paren-less calls and built-in `T.init` defined for all types.

src/ocean/util/container/btree/BTreeMap.d
```diff
@@ -0,0 +1,664 @@
...
+
+    Params:
+        KeyType = type of the key.
+        ElementType = type of the element to store.
+        tree_degree = degree of the tree (refer to the documentation)
+        memManager = memory management strategy to use
+
+******************************************************************************/
+
+public BTreeMap!(KeyType, ElementType, tree_degree) createBTreeMap
```
mihails-strasuns-sociomantic: Phobos uses `makeX` notation (but we don't need to care ;))

src/ocean/util/container/btree/BTreeMap.d
mihails-strasuns-sociomantic: You can `@disable this` for actual struct and prepare it in constructor function from `T.init` state.

mihails-strasuns-sociomantic (commented):
Some very brief comments without reading into implementation/tests

src/ocean/util/container/btree/BTreeMap.d
```diff
@@ -0,0 +1,570 @@
+/********************************************************************************
+
+    B-Tree data structure and operations on it.
+
+    Copyright:
+        Copyright (c) 2017 Sociomantic Labs GmbH.
```
leandro-lucarella-sociomantic: I would go with the updated lowercase copyright notice in new files :)

nemanja-boric-sociomantic:
Updated with feedback from all of you.

mihails-strasuns-sociomantic (approved):
@gavin-norman-sociomantic do you want to give another review pass? To me API looks reasonable enough to merge and everything else can be fixed later if needed.

src/ocean/util/container/btree/BTreeMap.d
```diff
@@ -0,0 +1,590 @@
...
+    tree.insert(4, f);
+
+    auto number_tree = makeBTreeMap!(int, int, 2);
+
+    for (int i = -1; i < 9; i++)
+    {
+        number_tree.insert(i, i);
+    }
+
+    // let's find 7_001
```
gavin-norman-sociomantic: What's 7,001?

src/ocean/util/container/btree/BTreeMap.d
```diff
@@ -0,0 +1,590 @@
...
+
+*/
+
+version (UnitTest)
+{
+    import ocean.core.Test;
+}
+
+unittest
+{
```
gavin-norman-sociomantic, line 338: Might be nice to break this down into separate unittests, to make it easier to follow. There's at very least the test for the file map and the tests for the integer map.

src/ocean/util/container/btree/BTreeMapRange.d
```diff
@@ -0,0 +1,412 @@
...
+    {
+        int element = random.uniformR(to_insert);
+        // don't count double elements (they are not inserted)
+        total_elements += random_tree.insert(element, element)? 1 : 0;
+        res = random_tree.get(element, found);
+        test(found);
+        test!("==")(res, element);
+    }
+
+    // reseed, so that the difference two random number generated sets
```
gavin-norman-sociomantic: Where does it reseed?

src/ocean/util/container/btree/Implementation.d
```diff
@@ -0,0 +1,1200 @@
...
+            this implementation on the actual content of the node's element
+            (replace KeyValue with just a value) and with the ordering operation
+            inserted as a policy (to compare the values, and not the nodes).
+
+        ***********************************************************************/
+
+        private struct KeyValue
+        {
+            KeyType key;
+            Unqual!(ValueType) value;
```
gavin-norman-sociomantic: Is the use of `Unqual` here described anywhere? This seems like an important implementation detail to mention at a high level.

src/ocean/util/container/btree/Implementation.d
```diff
@@ -0,0 +1,1200 @@
...
+            tree = tree to visit
+            dg = delegate to call for each visited element. Delegate should return
+                 non-zero value if the visiting should be aborted.
+
+        Returns:
+            return value of the last dg call.
+
+    ******************************************************************************/
+
+    package int inorder (int delegate(ref KeyType key, ref ValueType value) dg)
```
gavin-norman-sociomantic, line 929: Might be nice to combine the two `inorder` overloads. A template with a `static if` to call the different delegate types?

nemanja-boric-sociomantic:
Updated to address @gavin-norman-sociomantic 's comments (mostly outdated comments, separate unittests, de-duplication of `inorder`, commend on unqualed value type).

mihails-strasuns-sociomantic:
Ok, let's do it :P

PR #211, opened 2017-08-13
Fix some D2 deprecations

Geod24:
Doesn't fix all of them since some are in soon-to-be-deprecated modules.
Also found a new DMD bug in the JSON module: https://issues.dlang.org/show_bug.cgi?id=17752

### DIFF
--- a/src/ocean/sys/stats/linux/ProcVFS.d
+++ b/src/ocean/sys/stats/linux/ProcVFS.d
@@ -585,7 +585,7 @@ private ProcMemInfo parseProcMemInfoData (cstring delegate() read_next_line)
 
         switch (field_name)
         {
-            foreach (i, ref field; meminfo.tupleof)
+            foreach (i, FieldT; typeof(meminfo.tupleof))
             {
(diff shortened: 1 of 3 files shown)


mathias-lang-sociomantic:
Rebased on v3.x.x as one of the commit was redundant with https://github.com/sociomantic-tsunami/ocean/pull/201

src/ocean/sys/stats/linux/ProcVFS.d
```diff
@@ -585,7 +585,7 @@ private ProcMemInfo parseProcMemInfoData (cstring delegate() read_next_line)
 
         switch (field_name)
         {
-            foreach (i, ref field; meminfo.tupleof)
```
mihails-strasuns-sociomantic, line 588: What was deprecated here?

mihails-strasuns-sociomantic: Nevermind, missed commit body

mihails-strasuns-sociomantic (approved):

PR #215, opened 2017-08-16
Remove lst files in make clean, no matter if they are hidden

nemanja-boric-sociomantic:
*.lst files generated by the compiler are hidden and simple `rm *.lst`
will not remove them. `rm .*.lst` will.


nemanja-boric-sociomantic:
I don't think this warrants patch release on its own.

mihails-strasuns-sociomantic (approved):

PR #216, opened 2017-08-16
Remove redundant scope around enforce

JensKMueller:
<s>This change adds the else branch which seems missing in the code.
Supposedly an else branch was intended here because of the introduced
scope for the enforce statement. The original code is not buggy. It only
executed the enforce statement when the if branch was taken. In this
case the enforce statement is redundant since its condition is always
true. With the added else branch the code looks as it was originally
intended especially with respect to similar code a few lines above.
It also seems cleaner.</s>
Right after an if statement an enforce is placed inside its own scope.
This is a leftover from old code. The intention here is to always
enforce. For clarity the scope is removed since otherwise one might
think an else branch is missing. The changed code should be easier to
understand raising less questions about its intentions.

### DIFF
--- a/src/ocean/task/util/StreamProcessor.d
+++ b/src/ocean/task/util/StreamProcessor.d
@@ -169,32 +169,27 @@ class StreamProcessor ( TaskT : Task )
 
         if (throttler_config.suspend_point == size_t.max)
             throttler_config.suspend_point = total / 3 * 2;
-        else
-        {
-            enforce(
-                this.throttler_failure_e,
-                throttler_config.suspend_point < total,
-                format(
-                    "Trying to configure StreamProcessor with suspend point ({}) " ~
-                        "larger or equal to task queue size {}",
-                    throttler_config.suspend_point, total
-                )
-            );
-        }
+        enforce(
+            this.throttler_failure_e,
+            throttler_config.suspend_point < total,
+            format(
+                "Trying to configure StreamProcessor with suspend point ({}) " ~
+                    "larger or equal to task queue size {}",
+                throttler_config.suspend_point, total
+            )
+        );
 
         if (throttler_config.resume_point == size_t.max)
             throttler_config.resume_point = total / 5;
-        {
-            enforce(
-                this.throttler_failure_e,
-                throttler_config.resume_point < total,
-                format(
-                    "Trying to configure StreamProcessor with resume point ({}) " ~
-                        "larger or equal to task queue size {}",
-                    throttler_config.resume_point, total
-                )
-            );
-        }
+        enforce(
+            this.throttler_failure_e,
+            throttler_config.resume_point < total,
+            format(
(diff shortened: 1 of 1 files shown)


src/ocean/task/util/StreamProcessor.d
```diff
@@ -184,6 +184,7 @@ class StreamProcessor ( TaskT : Task )
 
         if (throttler_config.resume_point == size_t.max)
             throttler_config.resume_point = total / 5;
+		else
```
gkotian: Wouldn't it better instead to remove the `else` from the previous if-block (i.e. always execute both the `enforce`s)? If you go this way, the additional scope surrounding both the enforces can also be removed.

nemanja-boric-sociomantic: also tabs ;-)

nemanja-boric-sociomantic: @gkotian I'm not sure this is how you want to do this. Because you want to actually execute first `enforce` only if the `suspend_point != size_t` max. If you remove `else`, both of the enforces are becoming redundant, but not equivalent the original code (or code that was supposed to be here) - only test if the `suspend point`/`resume_point` is not `size_t`.

nemanja-boric-sociomantic:
The target branch should be v2.8.x, as this is a bugfix.

src/ocean/task/util/StreamProcessor.d
mihails-strasuns-sociomantic: My original intention was exactly like @gkotian describes - to always do enforce, even if it is a bit redundant (because it is more refactoring-friendly). There is no bug, just broken formatting leftover from old code structure.

mihails-strasuns-sociomantic (changes requested):

src/ocean/task/util/StreamProcessor.d
JensKMueller: What about the `else` branch a few lines above? Was that also intended to be always enforce?

JensKMueller:
@nemanja-boric-sociomantic How do I change the target branch? I can't change it as explained [here](https://github.com/blog/2224-change-the-base-branch-of-a-pull-request). There is no drop down menu for me.

mathias-lang-sociomantic:
@jenskmueller you need to clic 'Edit' first, as if you were editing the title. However as Mihails mentioned it is unnecessary as there is no bug.

src/ocean/task/util/StreamProcessor.d
mihails-strasuns-sociomantic: Yep, for same reason. I must have been distracted with something else before fully changing the code accordingly which is why it looks so weird now.

JensKMueller:
Alright. Pushed changed code. Any more feedback?

mihails-strasuns-sociomantic (approved):

mihails-strasuns-sociomantic:
LGTM, thanks!

PR #219, opened 2017-08-17
beaver/travis after script tweaks

mihails-strasuns-sociomantic:


leandro-lucarella-sociomantic (approved):
I want to change this soonish in beaver itself, but for now I don't see anything against this either.

PR #221, opened 2017-08-21
UnitTestRunner: improve -p option

leandro-lucarella-sociomantic:
The --package / -p option behaved a bit weird. Instead of really meaning
"test only modules in package PKG" it meant "test only modules with
fully qualified names starting with PKG". So if you used `-p ocean`,
a top-level module `ocean` would be tested (which is not probably what
you want) and even an entire package `oceanic`would be tested
too (anything matching `^ocean.*` would be tested)..

This commit makes `-p` only test modules strictly in the PKG package
(i.e. the module fully qualified name must match `^PKG\..*`) and a new
options `--module MOD` / `-m MOD` was added to be able to specify single
modules.

Related to https://github.com/sociomantic-tsunami/makd/pull/86.

### DIFF
--- a/relnotes/utrunner.migration.md
+++ b/relnotes/utrunner.migration.md
@@ -1,0 +1,7 @@
+* `core.UnitTestRunner`
+
+  There is a minor breaking bug fix in the `--package` / `-p` option; now it can't be used as some sort of wildcard.
+
+  The `PKG` argument will only match fully qualified names that start with `PKG.` or the exact module `PKG`. Before using `PKG` as argument would have matched `PKG*`, so `PKGfoo` was also a match.
+
+  As a transitional step, `PKG.` will also be accepted as a package specification, although it will match both `PKG` (exact match) and `PKG.*`.
--- a/src/ocean/core/UnitTestRunner.d
+++ b/src/ocean/core/UnitTestRunner.d
@@ -90,6 +90,7 @@ import ocean.text.xml.Document: Document;
 import ocean.text.xml.DocPrinter: DocPrinter;
 import ocean.text.convert.Formatter;
 import ocean.core.Test: TestException, test;
+import ocean.core.array.Mutation: uniq;
 import core.memory;
 
 
@@ -212,8 +213,8 @@ private scope class UnitTestRunner
             {
                 no_match++;
                 if (this.verbose > 1)
-                    Stdout.formatln("{}: {}: skipped (not in packages to test)",
-                            this.prog, m.name);
+                    Stdout.formatln("{}: {}: skipped (not in " ~
+                            "modules to test)", this.prog, m.name);
                 continue;
             }
 
@@ -312,7 +313,7 @@ private scope class UnitTestRunner
             if (!this.keep_going && failed)
                 Stdout.format(", {} skipped", skipped);
             if (this.verbose > 1)
-                Stdout.format(", {} didn't match --package", no_match);
+                Stdout.format(", {} didn't match -p/--package", no_match);
             Stdout.formatln(" [{}]", this.toHumanTime(total_time));
         }
 
@@ -736,8 +737,13 @@ private scope class UnitTestRunner
 
         foreach (pkg; this.packages)
         {
-            if (name.length >= pkg.length &&
-                    strncmp(pkg.ptr, name.ptr, pkg.length) == 0)
+            // It matches as a module
+            if (name == pkg)
+                return true;
+            // If name is part of the package, it must start with "pkg."
+            if (name.length > pkg.length &&
+                    strncmp(pkg.ptr, name.ptr, pkg.length) == 0 &&
+                    name[pkg.length] == '.')
                 return true;
         }
 
@@ -827,8 +833,18 @@ private scope class UnitTestRunner
             case "-p":
             case "--package":
                 auto opt_arg = getOptArg(i);
-                if (opt_arg is null)
+                if (opt_arg.length == 0)
                     return false;
+                // FIXME: This is just a compatibility-mode because at some
+                //        point this worked implicitly as some sort of pattern
+                //        matching ARG*, so to specify a strict package it was
+                //        common to use "pkg.". Now we just remove the trailing
+                //        "." if we find one, it's not a valid package name
+                //        anyway. This was introduced in v3, so it should be
+                //        safe to remove it in v4 or maybe v5 go make sure all
+                //        dependencies were updated.
+                if (opt_arg[$-1] == '.')
+                    opt_arg = opt_arg[0..$-1];
                 this.packages ~= opt_arg;
                 break;
 
@@ -863,6 +879,9 @@ private scope class UnitTestRunner
             Stderr.newline();
             return false;
         }
+
+        // Remove any package duplicates
+        this.packages = uniq(this.packages);
 
         return true;
     }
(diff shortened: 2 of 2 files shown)


leandro-lucarella-sociomantic:
> Merging #221 into v3.x.x will decrease coverage by 69.57%.

Mmm, this module has no unit tests. I think sort of on purpose, since it is running the unit tests, but maybe it was just a reflex I had coming from GC development, where I definitely couldn't use the GC :D.

leandro-lucarella-sociomantic:
Even when this change is strictly a breaking change, I think it is worth including in a minor release, since is not really affecting the API and it is very unlikely to be disruptive.

leandro-lucarella-sociomantic:
I did some simple tests and it seems to work as expected.

mihails-strasuns-sociomantic:
Wouldn't be it more simple to just define `-p` to mean "package" OR "exact module name" match? We only need to support individual modules because of `ocean/package.d` and this is more intuitively covered by plain `-p ocean`. This will also allow to make makd change forward compatible without any complicated special cases.

mihails-strasuns-sociomantic:
(and I would totally categorize this as a disruptive bug-fix)

mathias-lang-sociomantic:
Agree with @mihails-strasuns-sociomantic here, let's just fix `-p`

leandro-lucarella-sociomantic:
> Wouldn't be it more simple to just define -p to mean "package" OR "exact module name" match? We only need to support individual modules because of ocean/package.d and this is more intuitively covered by plain -p ocean. This will also allow to make makd change forward compatible without any complicated special cases.

No, because I don't actually want to match a `ocean` module name when using `-p ocean`, I want to explicitly only accept modules inside a package (i.e. `ocean.*` but not `ocean`). I actually started implementing this as an extension of `-p` but then realized it doesn't work as intended if I do that.

And if we are going to fix this because the current behaviour is weird, I prefer to fix it right once and for all instead of having a `-p` that is better than before but still weird.

mathias-lang-sociomantic:
Why don't you want to run the package test in this case ?

leandro-lucarella-sociomantic:
> Why don't you want to run the package test in this case ?

What do you mean by "run the package test"?

mihails-strasuns-sociomantic:
All tests contained in package "X". Idea is that there are just two possible cases when you may get module name with exact match to "X":

- `X/package.d`, which is part of `X` package already and would be expected to run if you supply `-p X`
- `X.d` in one of import root paths. If that happens to be the case (very rare!), it is quite likely that supplied `-p X` is meant to refer to it specifically.

So it seems that distinction between `-m` and `-p` only makes sense from purist "what D spec calls package and what it calls module" point of view. From a practical "what tests should be run" PoV it only adds unnecessary complexity.

leandro-lucarella-sociomantic:
> If that happens to be the case (very rare!), it is quite likely that supplied -p X is meant to refer to it specifically.

And if it doesn't? I really don't see the complexity. For MakD I have the exact information I need to decide if it is `-m` or `-p` to make it precise, and if we follow the holly `-p` flag, then if you hit a false positive, you have no way to do it proper, and we are back to having to fix this.

I really don't understand why you just want to adopt an okaish but flawed solution when the right solution already here (OK, I see one reason, it will require a transition period, but is just temporary).

mihails-strasuns-sociomantic:
> I really don't understand why you just want to adopt an okaish but flawed solution when the right solution already here

KISS. Right solution creates unnecessary cognitive load by adding two flags when you almost always think in terms of only one.

leandro-lucarella-sociomantic:
Still not convinced. You are not addressing my points. Warning: I'm willing to fight for this one.

mathias-lang-sociomantic:
> For MakD I have the exact information I need to decide if it is -m or -p to make it precise, and if we follow the holly -p flag, then if you hit a false positive, you have no way to do it proper, and we are back to having to fix this.

But you never wants to do that. The distinction itself *is* counter-intuitive. The package module is meant as either an interface over the package, a simplification (when you don't care about dependency) or a transitional mean. There is really no reason to not run the package module's unittest with the rest of the package.
I could understand if you had a reason to run the package module unittest and *not* the rest (high level test vs low level test), but if you reach that point were you cannot afford to run extra UT, you're better off going for an integration test anyway.

leandro-lucarella-sociomantic:
> But you never wants to do that. The distinction itself is counter-intuitive. The package module is meant as either an interface over the package, a simplification (when you don't care about dependency) or a transitional mean.

In the kind of scenario I am thinking of, the distinction is important. Now I'm thinking if, because of D import rules, this scenario is impossible, maybe it is... I'm thinking of this kind of scenario where you are using some external library that provides, say, package `tools` (with no `tools/package.d` and then you, in your code, have a module called `tools`. I want to run unittests in my `tools` module, but not in the rest of the `tools` package.

In any case, I'm not talking about the case where the whole package (having a `package.d`) is under your control, I'm thinking of cases where there are weird interactions between the project and its dependencies. In those cases, unless you can prove all corner cases works with an unified `-p`, I will fight for a separate `-m`.

If you prove there are no corner cases where `-m` would be useful, I would happily go with a unified `-p`. Since I can't, I want to have the future-proof, safe option.

mathias-lang-sociomantic:
```
mkdir -p lol/tools/
echo "module tools.foo;" > lol/tools/foo.d
echo "module tools;" > tools.d
echo "import tools; import tools.foo; void main () {}" > main.d

rdmd --compiler=dmd1 -Ilol main.d                                                               :(
main.d(2): Error: can only import from a module, not from a member of module tool. Did you mean `import tool : foo`?
Failed: ["dmd1", "-Ilol", "-v", "-o-", "main.d", "-I."]

rdmd -Ilol main.d
lol/tools/foo.d(1): Error: package name 'tool' conflicts with usage as a module name in file tools.d
Failed: ["dmd", "-Ilol", "-v", "-o-", "main.d", "-I."]
```

From my experience, anything remotely close to the case you describe would wreak havoc on most, if not all, D tooling out there.

leandro-lucarella-sociomantic:
Mmmm, maybe I overestimate D in this respect. OK, I will change it back, once again. If we find problems with this in the future at least I get to tell you "I told you so".

OTOH, this is being "fixed" because there is a duplicate `-p` in Makd (sociomantic-tsunami/makd#86). This duplicate will still exist in this scheme (removing it in Makd is complicated because it can't be done by a simple `find` as it is now), so I don't see it happening. We could remove duplicates in the `UnitTestRunner`, I can try that which should be simpler.

leandro-lucarella-sociomantic:
Updated. Thinking of the migration notes, I am more and more convinced this should go to the next major. Even when the breaking change shouldn't affect most people, this is not really urgent at all, so I don't see the point on forcing a breaking change in a non-major update.

I wonder if we should print any warnings if we find a module that would have matched the old scheme and now it doesn't to warn the user that the module/package will not be tested now unless it's included explicitly. The only problem with this is I don't see any good way to silence the warning if you don't actually want to test the module (if you want to test it we can silence the warning if the module matches with other `-p` arguments, i.e. never printing warnings for stuff that matches).

mihails-strasuns-sociomantic:
> I am more and more convinced this should go to the next major. 

Sounds fine to me.

leandro-lucarella-sociomantic:
Rebased on top of v4 and changed the milestone.

mihails-strasuns-sociomantic:
Please rebase again

leandro-lucarella-sociomantic:
Updated.

mihails-strasuns-sociomantic (approved):

PR #224, opened 2017-08-23
Make DeserializationException a ReusableException.

scott-gibson-sociomantic:
Previously the formatting call would allocate a new buffer every time the enforce triggered.

### DIFF
--- a/src/ocean/core/Exception.d
+++ b/src/ocean/core/Exception.d
@@ -100,9 +100,10 @@ unittest
 public template ReusableExceptionImplementation()
 {
     import ocean.transition;
-    static import ocean.text.convert.Integer_tango;
     import ocean.core.Buffer;
     static import ocean.core.array.Mutation;
+    static import ocean.text.convert.Formatter;
+    static import ocean.text.convert.Integer_tango;
 
     static assert (is(typeof(this) : Exception));
 
@@ -247,6 +248,25 @@ public template ReusableExceptionImplementation()
                 ocean.text.convert.Integer_tango.format(buff, num));
         return this;
     }
+
+    /**************************************************************************
+
+        Appends formatted string
+
+        Params:
+            fmt = string format to use
+            args = variadic args list to format
+
+        Returns:
+            this instance
+
+    **************************************************************************/
+
+    public typeof (this) fmtAppend ( Args... ) ( cstring fmt, Args args )
+    {
+        ocean.text.convert.Formatter.sformat(this.reused_msg, fmt, args);
+        return this;
+    }
 }
 
 ///
@@ -281,6 +301,13 @@ unittest
     catch (SomeReusableException) { }
     assert (getMsg(e) == "Wrong name (NAME) 0x2A 42");
     assert (old_ptr is e.reused_msg[].ptr);
+}
+
+unittest
+{
+    auto e = new SomeReusableException;
+    e.set("aaa").fmtAppend("{0}{0}", 42);
+    test!("==")(getMsg(e), "aaa4242");
 }
 
 version (UnitTest)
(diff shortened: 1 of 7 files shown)


scott-gibson-sociomantic:
Stress testing an application revealed that even though the exception is static in `Deserializer`, the formatting calls are creating a new buffer each time.

nemanja-boric-sociomantic:
Alternative (and more expected IMHO) approach would be to make `DeserializationException` reusable exception: https://github.com/sociomantic-tsunami/ocean/blob/bcd88397f484ca4821ceed942b518f05c2e2274f/src/ocean/core/Exception.d#L100

mihails-strasuns-sociomantic:
Indeed, let's avoid repeating boilerplate.

scott-gibson-sociomantic:
Alright i'll update it to a ReusableException now.

scott-gibson-sociomantic:
Updated to use ReusableException and also added some coverage tests.

mihails-strasuns-sociomantic:
@scott-gibson-sociomantic it never came to my mind that now we can do nice formatting of reusable messages with new formatter sink!

I took the opportunity to generalize it and move into `ReusableExceptionImpl` itself and force pushes result into your branch :P FYI @mathias-lang-sociomantic too

src/ocean/core/Exception.d
```diff
@@ -198,53 +197,40 @@ public template ReusableExceptionImplementation()
...
+        static if (is(T : long) && args.length <= 1)
+        {
+            istring fmt;
+            static if (args.length == 1 && is(Args[0] == bool))
+                fmt = args[0] ? "0x{:X}" : "{}";
+            else
+                fmt = "{}";
+            ocean.text.convert.Formatter.sformat(this.reused_msg, fmt, arg);
+        }
+        else static if (is(T : cstring))
```
mathias-lang-sociomantic: Not so fan of mixing semantics though...
`append!(cstring, cstring)` => format
`append!(uint, cstring)` => append

That is a weird API to me.

mihails-strasuns-sociomantic: It was not intended as a feature - can't overload two different templated `append` methods in D1 but must preserve backwards compatibility. It is one of stupid legacy quirks that will have to wait for full D2 :(

mathias-lang-sociomantic: I would much prefer to introduce a new symbol then... Maybe `format` ?

mihails-strasuns-sociomantic: Sure, renamed to `fmtAppend` and restored old methods as-is. Ugly but I see your point.

mihails-strasuns-sociomantic:

nemanja-boric-sociomantic (approved):
good stuff!

src/ocean/core/Exception.d
```diff
@@ -100,9 +100,9 @@ unittest
 public template ReusableExceptionImplementation()
 {
     import ocean.transition;
-    static import ocean.text.convert.Integer_tango;
```
nemanja-boric-sociomantic, line 103: is this really gone?

mihails-strasuns-sociomantic: not anymore, good catch

nemanja-boric-sociomantic: Good thing we don't have any unittest so this doesn't go red ;-)

mihails-strasuns-sociomantic: Updated

mihails-strasuns-sociomantic: > Good thing we don't have any unittest so this doesn't go red ;-)

We have but there are other imports in same module that provide it so it kept being green :(

scott-gibson-sociomantic:
Thanks @mihails-strasuns-sociomantic for your work. I'm unsure about the codecov/changes check but i'm happy with the new method.

PR #223, opened 2017-08-23
Produce coverage report only if asked

nemanja-boric-sociomantic:
By default build should not generate coverage report. This allows
one to produce them by passing COVERAGE=1 to make.

PR #225, opened 2017-08-23
Properly escape OCEAN_D2_USER OCEAN_D2_PASS

nemanja-boric-sociomantic:
Noticed in and based on #223

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,6 @@ script: BEAVER_DOCKER_VARS=AFTER_SCRIPT beaver dlang make
 
 after_success:
     - test "$AFTER_SCRIPT" = "1" && beaver run ci/codecov.sh
-    - test "$AFTER_SCRIPT" = "1" && \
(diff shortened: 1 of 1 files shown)


mihails-strasuns-sociomantic:
It doesn't really need to be based on the other one ;)

nemanja-boric-sociomantic:
I'm curious why it fails: https://travis-ci.org/sociomantic-tsunami/ocean/jobs/267559688#L1409

Should it be maybe:

```
BEAVER_DOCKER_VARS=... test "$AFTER_SCRIPT" = "1" && beaver run ci/d2tags.sh
```

?

mihails-strasuns-sociomantic:
Yeah could be the case. I thought one can define env var at start of any expression but it doesn't seem to be the case!

nemanja-boric-sociomantic:
OK, I got it. Turns out there's no need for any escaping in travis.yml.

mihails-strasuns-sociomantic (approved):
Even better!

PR #226, opened 2017-08-24
[WIP] travis: Use stages for post-build actions

leandro-lucarella-sociomantic:


leandro-lucarella-sociomantic:
It will be slower, but it should be much cleaner.

leandro-lucarella-sociomantic:
And also more correct, as now tags could be pushed even if not all matrix jobs succeed. As for coverage, I'm not sure what to do, now is only sent if all tests pass.

Oh, and I realized I missed passing the generated reports via the cache. Do you know where are the report files?

mihails-strasuns-sociomantic:
> Do you know where are the report files?

Right now they are written straight in the root folder, bunch of separate `.lst` files.

.travis.yml
```diff
@@ -19,15 +19,46 @@ env:
...
         - DMD=1.081.0 DIST=xenial F=devel
         - DMD=2.070.2.s12 DIST=xenial F=production
         - DMD=2.070.2.s12 DIST=xenial F=devel
         - DMD=2.071.2.s1 DIST=xenial F=production
         - DMD=2.071.2.s1 DIST=xenial F=devel
 
-install: beaver dlang install
+install: |
+    zcat "build/last/tmp/docker.img.gz" |
+        /usr/bin/time -f 'Load took %es\n' docker load -q || true
```
mihails-strasuns-sociomantic: Travis UI shows stage timings, what is this for?

.travis.yml
```diff
@@ -19,15 +19,46 @@ env:
...
+    beaver dlang install
 
-script: beaver dlang make
+script: /usr/bin/time -f 'Build took %es\n' beaver dlang make
 
-after_success: ci/travis-after-success.sh
+after_success: |
+    # Write the image cache
+    docker images --all --format '{{.Repository}}:{{.Tag}} {{.ID}}' |
```
mihails-strasuns-sociomantic: Won't it write new image for each matrix row?

.travis.yml
```diff
@@ -19,15 +19,46 @@ env:
...
+        sed 's|<none>:<none>||g' |
+        /usr/bin/time -f 'Save took %es\n' xargs docker save |
+        gzip --fast > "build/last/tmp/docker.img.gz"
+
+jobs:
+    include:
+        - stage: Coverage
```
mihails-strasuns-sociomantic: What will happen if you also define `stage: Test` and move env matrix there? Will it work as I would hope or fail miserably? :)

.travis.yml
```diff
@@ -19,15 +19,46 @@ env:
...
+          env: DMD=1.081.0 DIST=xenial F=production
+          install: |
+            zcat "build/last/tmp/docker.img.gz" |
+                /usr/bin/time -f 'Load took %es\n' docker load -q
+            beaver dlang install
+          script: echo Sending coverage reports
+          after_success: ci/cov
+        - stage: Release D2
+          env: DMD=1.081.0 DIST=xenial F=production
```
mihails-strasuns-sociomantic: Is it needed for d2conv?

ci/cov
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
```
mihails-strasuns-sociomantic: I'd much prefer to having `.sh` suffixes in shell scripts, it saves quite some time navigating in projects ;)

mihails-strasuns-sociomantic (commented):
Had no idea travis supports this, looks promising!

Added time is no big deal because matrix take much longer anyway.

.travis.yml
leandro-lucarella-sociomantic: Just copy&paste from other parts where there is not the only command. I don't think I will keep up after debugging the options.

.travis.yml
leandro-lucarella-sociomantic: I don't follow this question.

.travis.yml
leandro-lucarella-sociomantic: I think it should work, but then we'll have to redefine the `install:` step for all the stages, which is just annoying. Everything you don't put in a stage goes to the default `Test` stage anyway.

.travis.yml
leandro-lucarella-sociomantic: A docker container, yes. The `env` here is to say "I want to run this stage only for the matrix cell X", otherwise it will run all the global matrix for the stage too.

ci/cov
leandro-lucarella-sociomantic: This is an experiment of something I will move to beaver, so becoming a *command* where the suffix makes no sense.

leandro-lucarella-sociomantic:
BTW, this PR is more a way for me to test approaches rather than something really ready for review.

I think I will discard the stages approaches, even when visually is nicer, it adds quite some time, because is a synchronization point too. At least for coverage, it means it will need to first wait for **ALL** matrix job to finish, then acquire a new VM, create a docker image, and run the stuff there. So is not like just adding one more cell because a stage waits for the previous to finish.

I think for D2 tags conversion a stage might make sense because it only runs on tags, it doesn't slow down every job (although I think because we can't use a condition on an entire stage, the job will still have to wait for travis to allocate a new VM just to figure out there is nothing to be done for this stage, so even for this it might be overkill.

Travis really needs a way to make stages available in the same VM for this to make sense, or at least to be able to pass around artefacts between stages. Otherwise it really sucks...

leandro-lucarella-sociomantic:
And using a cached docker image is still slower than building the image (and sometimes it even doesn't use the cached one and rebuilds anyway).

mihails-strasuns-sociomantic:
Should this still be opened?

leandro-lucarella-sociomantic:
Yes, I want to use it as a base for the D2 tag conversion in beaver, I just didn't have the time to finally get into it :(

mihails-strasuns-sociomantic:
I think it makes most sense to start using this for D2 conversion, but keep code cov as part of regular script. Latter now runs for each matrix combination (and reuses artifacts from build), while former is independent and should only be done if everything else succeeds. This feels like a natural separation.

leandro-lucarella-sociomantic:
Yes, that's the plan. I'm about to close this when I do the test PR for the new `beaver dlang d2-release`.

leandro-lucarella-sociomantic:
I will close this now that I finally got what I wanted with D2 releases, codecov should still be done as a `after_success` step for all matrix builds.

PR #227, opened 2017-08-28
Don't unregister TimerExt if exception is thrown from DaemonApp.onSta…

nemanja-boric-sociomantic:
…tsTimer

In case the exception is thrown from DaemonApp.onStatsTimer (for
example, if there's no disk space left on device), we don't want to
unregister timer and never to write stats again. Instead, we should keep
trying.

### DIFF
--- a/src/ocean/util/app/DaemonApp.d
+++ b/src/ocean/util/app/DaemonApp.d
@@ -60,6 +60,7 @@ public abstract class DaemonApp : Application,
         ISignalExtExtension
 {
     import ocean.io.select.EpollSelectDispatcher;
+    import ocean.io.Stdout;
     import ocean.sys.Stats;
     import ocean.task.Scheduler;
     import ocean.text.Arguments : Arguments;
@@ -563,7 +564,23 @@ public abstract class DaemonApp : Application,
 
     private bool statsTimer ( )
     {
-        this.onStatsTimer();
+        try
+        {
+            this.onStatsTimer();
+        }
+        catch (Exception e)
+        {
+            try
+            {
+                Stderr.formatln(
+                        "Exception caught in DaemonApp.statsTimer: {}", e);
+            }
+            catch (Exception)
+            {
+                // ignore the failed stderr output, keep the timer registered
+            }
+        }
+
         return true;
     }
 


src/ocean/util/app/DaemonApp.d
```diff
@@ -563,7 +564,16 @@ public abstract class DaemonApp : Application,
...
     {
-        this.onStatsTimer();
+        try
+        {
+            this.onStatsTimer();
+        }
+        catch (Exception e)
+        {
+            Stderr.formatln("Exception caught in DaemonApp.statsTimer: {}",
+                    getMsg(e));
```
mathias-lang-sociomantic: I would really like a `toString` here (so just `e`) because this way we get a stack trace for free in D2.

src/ocean/util/app/DaemonApp.d
```diff
@@ -563,7 +564,16 @@ public abstract class DaemonApp : Application,
...
     private bool statsTimer ( )
     {
-        this.onStatsTimer();
+        try
+        {
+            this.onStatsTimer();
+        }
+        catch (Exception e)
+        {
+            Stderr.formatln("Exception caught in DaemonApp.statsTimer: {}",
```
gavin-norman-sociomantic: Why stderr and not a log, for example?

gavin-norman-sociomantic: Ah, I just read the commit message that this is about out-of-disk errors.

src/ocean/util/app/DaemonApp.d
nemanja-boric-sociomantic: Wasn't the point that getting the stack trace is very slow and should be avoided?

src/ocean/util/app/DaemonApp.d
nemanja-boric-sociomantic: Yeah, I'm not even sure about `Stderr`. My assumption is that most of the daemon apps will have this either redirected to `/dev/null`, or to some circular buffer mounted on `/`, so it is unlikely that this will fail as well. Of course, I could do a try/empty catch here, just to be safe.

src/ocean/util/app/DaemonApp.d
mathias-lang-sociomantic: For swarm yes, because it can happen under normal condition.
If this is hit, it's arguably a bug.

mathias-lang-sociomantic: Also it's once every 30s, not hundreds of occurences / second

src/ocean/util/app/DaemonApp.d
mathias-lang-sociomantic: I would still use the logger, as someone could have configured `Stderr` as an appender, or maybe use a network appender...

nemanja-boric-sociomantic: I'm not sure that's realistic thing to assume (that one configured logger just to use `Stderr` but not a file appender), and this fix was made precisely because `log.error` was throwing.

src/ocean/util/app/DaemonApp.d
nemanja-boric-sociomantic: > If this is hit, it's arguably a bug.
  
Not sure about that. If you're out of disk quota, for example, this will happen.

nemanja-boric-sociomantic: > Also it's once every 30s, not hundreds of occurences / second
  
That's a fair point.

nemanja-boric-sociomantic:
Updated with using `Exception.toString()` and ignoring any errors from `Stderr`. There's still an open question about `Stderr` vs `log`.

src/ocean/util/app/DaemonApp.d
```diff
@@ -563,7 +564,23 @@ public abstract class DaemonApp : Application,
...
+        try
+        {
+            this.onStatsTimer();
+        }
+        catch (Exception e)
+        {
+            try
+            {
+                Stderr.formatln(
+                        "Exception caught in DaemonApp.statsTimer: {}", e);
```
gavin-norman-sociomantic, line 576: Non-standard indentation!!! :D ;)

nemanja-boric-sociomantic: Please suggest what the standard one is, I had problems fitting this :-)

gavin-norman-sociomantic: Only joking. It doesn't matter.

mihails-strasuns-sociomantic:
Ping @gavin-norman-sociomantic - is it good to go?

nemanja-boric-sociomantic:
I'll close this and make a new one where the entire `TimerExt` would have a stronger guarantee that it will keep working by default if the exception is thrown.

gavin-norman-sociomantic:
Two discussion points:
1. @nemanja-boric-sociomantic is going to investigate whether we can put this catch at a deeper location (to cover more cases), in `TimerExt`, for example.
2. I had a feeling that there used to be a low-level try/catch around timers. Turns out it was only a log/rethrow (rather than logging and swallowing the exception, preventing the timer from being unregistered), but was removed in ocean v2.6.0.

nemanja-boric-sociomantic:
Targeting v4.0.0: #229

PR #229, opened 2017-08-29
Don't unregister the TimerSet and timers if exception is thrown in Ti…

nemanja-boric-sociomantic:
…merExt

`TimerExt` is now swallowing all the unhandled exceptions, keeping the timers
registered by default. Previously, the uncaught exception would
propagate to TimerSet, unregistering it and making all TimerExt
registration not work again. That behaviour is now fixed, with the
default that, if exception is thrown, the timer will stay registered.


nemanja-boric-sociomantic:
Marked as a breaking change, because users may use `TimerExt` for purposes not covered by `DaemonApp`, so that could change the application's behavior.

gavin-norman-sociomantic (approved):

gavin-norman-sociomantic:
LGTM, but will leave a bit in case anyone else has any comments.

mathias-lang-sociomantic:
Test ?

src/ocean/util/app/ext/TimerExt.d
```diff
@@ -27,6 +27,19 @@
...
 
+import ocean.util.log.Log;
+
+/*******************************************************************************
+
+    Static module logger
+
+*******************************************************************************/
+
+static private Logger log;
```
mathias-lang-sociomantic: Redundant static. Also last time a module-level logger was added it didn't go so well. Could you make it class-local?

src/ocean/util/app/ext/TimerExt.d
```diff
@@ -27,6 +27,19 @@
 
 module ocean.util.app.ext.TimerExt;
 
+import ocean.util.log.Log;
```
mathias-lang-sociomantic: `Logger`, `Log` will be gone in v4 and is deprecated in v3 but v4.x.x is too much behind

src/ocean/util/app/ext/TimerExt.d
nemanja-boric-sociomantic: bah, I've copy/pasted it from somewhere else.

mathias-lang-sociomantic:
I suggest merging v3 into v4 (fast-forwarding?) before merging anything into it.

nemanja-boric-sociomantic:
> I suggest merging v3 into v4 (fast-forwarding?) before merging anything into it.

@mihails-strasuns-sociomantic said he's going to do this.

nemanja-boric-sociomantic:
Updated with the fix for the unittests, test for the catching behavior and `Log` -> `Logger` import.

src/ocean/util/app/ext/TimerExt.d
```diff
@@ -87,7 +87,7 @@ unittest
         }
     }
 
-    new App;
+    (new App).run(null);
```
mathias-lang-sociomantic: Not an unittest please, an integration test. `unittest` shouldn't do IO or affect global state, and `epoll` does.

nemanja-boric-sociomantic: Well, there's no IO with this particular unittest. The `epoll_create` and `timerfd_create` calls already affect the global state, with one `epoll_wait` cycle or not.

mathias-lang-sociomantic: Well, I think they shouldn't be called then :)

nemanja-boric-sociomantic: Ok, I'll move the test into an integration test and remove `new App` from here.

nemanja-boric-sociomantic:
Updated with removing instantantion of the applications inside unittests and moving the `TimerExt` test into `test/timerext/main.d`

nemanja-boric-sociomantic:
Replaced `e.toString()` with `getMsg(e)`, since the timers may be called multiple times per second, unlike in #227.

test/timerext/main.d
```diff
@@ -0,0 +1,89 @@
...
+        }
+    }
+
+    private bool third ( )
+    {
+        return false;
+    }
+}
+
+void main(istring[] args)
```
mathias-lang-sociomantic: Missing space

test/timerext/main.d
```diff
@@ -0,0 +1,89 @@
...
+{
+    auto sandbox = DirectorySandbox.create(["etc", "log"]);
+
+    File.set("etc/config.ini", "[LOG.Root]\n" ~
+               "console = false\n");
+
+    auto app = new App;
+    app.main(args);
```
mathias-lang-sociomantic: Shouldn't you check the return code also ?

test/timerext/main.d
nemanja-boric-sociomantic: :grumpy-cat:

nemanja-boric-sociomantic:
updated: check the code of the `app.main` in test.

nemanja-boric-sociomantic:
I forgot the space >_>.

mihails-strasuns-sociomantic:
Please rebase

nemanja-boric-sociomantic:
Rebased.

mihails-strasuns-sociomantic (approved):

mihails-strasuns-sociomantic:
Checked coverage stats manually and it seems correct - don't think we can do any better until reports from int tests are sent.

PR #230, opened 2017-08-30
Fix travis build matrix

nemanja-boric-sociomantic:
We're using F env variable to set the flavour of a build. This needs
to be explicitly exported into the running environment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ env:
     global:
         # Make sure beaver is in the PATH
         - PATH="$(git config -f .gitmodules submodule.beaver.path)/bin:$PATH"
+        - COVERAGE_REPORT="1"
     matrix:
         - DMD=1.081.1 DIST=xenial F=production AFTER_SCRIPT=1
         - DMD=1.081.1 DIST=xenial F=devel
@@ -26,7 +27,7 @@ env:
 
 install: beaver dlang install
 
-script: BEAVER_DOCKER_VARS="AFTER_SCRIPT F" beaver dlang make
+script: BEAVER_DOCKER_VARS="AFTER_SCRIPT COVERAGE_REPORT F" beaver dlang make
 
 after_success:
     - beaver run ci/codecov.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ install: beaver dlang install
 script: BEAVER_DOCKER_VARS="AFTER_SCRIPT F" beaver dlang make
 
 after_success:
-    - test "$AFTER_SCRIPT" = "1" && beaver run ci/codecov.sh
+    - beaver run ci/codecov.sh
     - test "$AFTER_SCRIPT" = "1" &&
         BEAVER_DOCKER_VARS="OCEAN_D2_USER OCEAN_D2_PASS"
         beaver run ci/d2tags.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ env:
 
 install: beaver dlang install
 
-script: BEAVER_DOCKER_VARS=AFTER_SCRIPT beaver dlang make
+script: BEAVER_DOCKER_VARS="AFTER_SCRIPT F" beaver dlang make
 
 after_success:
     - test "$AFTER_SCRIPT" = "1" && beaver run ci/codecov.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,8 +22,8 @@ env:
     matrix:
         - DMD=1.081.1 DIST=xenial F=production AFTER_SCRIPT=1
         - DMD=1.081.1 DIST=xenial F=devel
-        - DMD=2.070.2.s12 DIST=xenial F=production
-        - DMD=2.070.2.s12 DIST=xenial F=devel
+        - DMD=2.070.2.s13 DIST=xenial F=production
+        - DMD=2.070.2.s13 DIST=xenial F=devel
 
 install: beaver dlang install
 
--- a/Build.mak
+++ b/Build.mak
@@ -8,7 +8,7 @@ override DFLAGS += -w -version=GLIBC
 TEST_RUNNER_MODULE := ocean.core.UnitTestRunner
 
 # Do we want coverage report?
-ifeq ($(AFTER_SCRIPT),1)
+ifeq ($(COVERAGE_REPORT),1)
 COVFLAG:=-cov
 endif
 
--- a/src/ocean/core/array/Search.d
+++ b/src/ocean/core/array/Search.d
@@ -1271,7 +1271,9 @@ unittest
 
 bool startsWith ( T ) ( in T[] arr, in T[] prefix )
 {
-    return (arr.length >= prefix.length) && (arr[0..prefix.length] == prefix[]);
+    if (arr.length < prefix.length)
+        return false;
+    return arr[0..prefix.length] == prefix[];
 }
 
 unittest
@@ -1307,7 +1309,9 @@ unittest
 
 bool endsWith ( T ) ( in T[] arr, in T[] suffix )
 {
-    return (arr.length >= suffix.length) && (arr[$ - suffix.length .. $] == suffix[]);
+    if (arr.length < suffix.length)
+        return false;
+    return arr[$ - suffix.length .. $] == suffix[];
 }
 
 unittest


joseph-wakeling-sociomantic (approved):
LGTM.  You might want to add a `V=1` just for a test run to validate it does what you want.

nemanja-boric-sociomantic:
I can see that from where the makd is putting the build artifacts (they should be in `build/production` for the production flavor).

joseph-wakeling-sociomantic:
Oh, interesting alternative choice -- why the switch to doing things this way?

joseph-wakeling-sociomantic:
> I can see that from where the makd is putting the build artifacts (they should be in build/production for the production flavor).

OK, fair enough.  I guess I'm being paranoid about flags ;-)

nemanja-boric-sociomantic:
> Oh, interesting alternative choice -- why the switch to doing things this way?

Well, because we've already exporting the `AFTER_SCRIPT` in the `beaver dlang make`, and I don't know if `F` is needed/usable anywhere else except in the make step.

nemanja-boric-sociomantic:
OK, it looks good: https://travis-ci.org/sociomantic-tsunami/ocean/jobs/270054098 https://travis-ci.org/sociomantic-tsunami/ocean/jobs/270054102

joseph-wakeling-sociomantic:
> Well, because we've already exporting the `AFTER_SCRIPT` in the beaver dlang make, and I don't know if `F` is needed/usable anywhere else except in the make step.

Makes sense, and your `after_success` entry makes a different use of `BEAVER_DOCKER_VARS`, so presumably it's better that this isn't global.

joseph-wakeling-sociomantic (approved):
Looks even better to me now than the original version ;-)

mihails-strasuns-sociomantic (approved):

leandro-lucarella-sociomantic (approved):
Arrgh, damn! Good catch. I think this should be done by beaver. I wonder if somehow all env vars should be exported to beaver, at least the ones present in the `.travis.yml` file. Otherwise this will keep biting us.

leandro-lucarella-sociomantic:
https://github.com/sociomantic-tsunami/beaver/issues/25

leandro-lucarella-sociomantic:
This fucked up coverage though, I presume because production builds might skip some checks (not contracts, right? We are still compiling with contracts for production right?).

I think adding codecov as an "AFTER_SCRIPT" was a mistake, coverage reports should be uploaded for all matrix cells, codecov will merged them, and we care about the global coverage, if we have versions that only run with some DMD version for example, the coverage will be different for different builds.

nemanja-boric-sociomantic:
> This fucked up coverage though, I presume because production builds might skip some checks (not contracts, right? We are still compiling with contracts for production right?).

This is correct, we're still building with contracts, but `debug` blocks are not compiled in anymore.

> I think adding codecov as an "AFTER_SCRIPT" was a mistake, coverage reports should be uploaded for all matrix cells, codecov will merged them, and we care about the global coverage, if we have versions that only run with some DMD version for example, the coverage will be different for different builds.

If codecov does merging, that makes sense to me. Alternatively, as a quick fix, we can move `AFTER_SCRIPT` to a `devel` matrix item.

mihails-strasuns-sociomantic:
> coverage reports should be uploaded for all matrix cells

Yes, I think this what we should do now. When originally adding codecov support I didn't notice it can merge reports from same PR, otherwise I would probably do that from the start.

nemanja-boric-sociomantic:
I agree, but would you agree that I should do that in the separate PR, as it's not related to the omitted `F`?

mihails-strasuns-sociomantic:
Rename PR to "fix build matrix"? :P

nemanja-boric-sociomantic:
OK, let's try it.

nemanja-boric-sociomantic:
Updated with uploading the coverage for all matrix items.

mihails-strasuns-sociomantic:
> 63.39% (-4.82%) compared to ed68eae

This is extremely weird.

nemanja-boric-sociomantic:
I wonder if I should export matrix vars while running `ci/codecov.sh`

mihails-strasuns-sociomantic:
Oh, I get it now - it is because of your earlier change https://github.com/sociomantic-tsunami/ocean/blob/v2.x.x/Build.mak#L10-L13

nemanja-boric-sociomantic:
Ahh yes.

nemanja-boric-sociomantic:
damn:

```
src/rt/lifetime.d:265 nothrow void rt.lifetime.stomping_prevention_trigger_nonpure() [0x11a8083]
src/rt/lifetime.d:236 pure nothrow bool rt.lifetime.__setArrayAllocLength(ref core.memory.BlkInfo_, ulong, bool, const(TypeInfo), ulong) [0x11a29ba]
src/rt/lifetime.d:1989 _d_arrayappendcTX [0x118a7dc]
src/rt/cover.d:465 void rt.cover.splitLines(char[], ref char[][]) [0x11a1dfc]
src/rt/cover.d:181 void rt.cover._sharedStaticDtor390() [0x11a144b]
??:? void rt.cover.__modshareddtor() [0x11a17b4]
src/rt/minfo.d:379 void rt.minfo.__T17runModuleFuncsRevS442rt5minfo11ModuleGroup8runDtorsMFZ9__lambda1Z.runModuleFuncsRev(const(immutable(object.ModuleInfo)*)[]) [0x11a3b46]
src/rt/minfo.d:255 void rt.minfo.ModuleGroup.runDtors() [0x11a3784]
src/rt/minfo.d:332 int rt.minfo.rt_moduleDtor().__foreachbody1(ref rt.sections_elf_shared.DSO) [0x118b9a3]
src/rt/sections_elf_shared.d:63 int rt.sections_elf_shared.DSO.opApplyReverse(scope int delegate(ref rt.sections_elf_shared.DSO)) [0x118bdc2]
src/rt/minfo.d:330 rt_moduleDtor [0x118b982]
src/rt/dmain2.d:208 rt_term [0x1186b42]
src/rt/dmain2.d:476 void rt.dmain2._d_run_main(int, char**, extern (C) int function(char[][])*).runAll() [0x1186f1c]
src/rt/dmain2.d:447 void rt.dmain2._d_run_main(int, char**, extern (C) int function(char[][])*).tryExec(scope void delegate()) [0x1186e90]
src/rt/dmain2.d:480 _d_run_main [0x1186ded]
??:? main [0xf60c77]
??:? __libc_start_main [0x70782f]
submodules/makd/Makd.mak:513: recipe for target '/home/travis/build/sociomantic-tsunami/ocean/build/devel/tmp/allunittests.stamp' failed
Aborted (core dumped)
make: *** [/home/travis/build/sociomantic-tsunami/ocean/build/devel/tmp/allunittests.stamp] Error 134
```

nemanja-boric-sociomantic:
So, coverage reporting is broken in D2 and stomping prevention assert.

mihails-strasuns-sociomantic:
Ah, crap, I totally forgot about the issue. I will have a quick look if patching dmd-transitional is feasible.

mihails-strasuns-sociomantic:
https://github.com/sociomantic-tsunami/dmd-transitional/pull/8

mihails-strasuns-sociomantic:
Fails because of weird `-O -cov` compiler codegen bug I know. This patch workarounds it:

```diff
@@ -1271,7 +1271,9 @@ unittest
 
 bool startsWith ( T ) ( in T[] arr, in T[] prefix )
 {
-    return (arr.length >= prefix.length) && (arr[0..prefix.length] == prefix[]);
+    if (arr.length < prefix.length)
+        return false;
+    return arr[0..prefix.length] == prefix[];
 }
 
 unittest
@@ -1307,7 +1309,9 @@ unittest
 
 bool endsWith ( T ) ( in T[] arr, in T[] suffix )
 {
-    return (arr.length >= suffix.length) && (arr[$ - suffix.length .. $] == suffix[]);
+    if (arr.length >= suffix.length)
+        return false;
+    return arr[$ - suffix.length .. $] == suffix[];
 }
```

nemanja-boric-sociomantic:
Hm, D2 production build fails in a scary way (in a place that's pretty obvious it shouldn't fail), and on my machine I can't even compile `make DVER=2 F=production` (dmd hags, both v2.070.2.s12 and v2.070.2.s13.).

nemanja-boric-sociomantic:
OK, all green now. Let's see what codecov says.

mihails-strasuns-sociomantic:
Hm, better, but it still shows some lines that are not covered despite being covered before (if I read report correctly). This is strange.

nemanja-boric-sociomantic:
To me it doesn't look like it says that. I think it says for most of the files here: https://codecov.io/gh/sociomantic-tsunami/ocean/compare/ed68eae659b6c9b93e13d1463d11c84cce7229fa...a2849583c58575cb38a612444f97404de65fddd6/changes

that the files had no coverage (left side of `->` is empty) and now they have some (either positive or negative)

nemanja-boric-sociomantic:
For example, see `ocean.text.xml.Document`: https://codecov.io/gh/sociomantic-tsunami/ocean/src/v2.x.x/src/ocean/text/xml/Document.d

and https://codecov.io/gh/sociomantic-tsunami/ocean/src/a2849583c58575cb38a612444f97404de65fddd6/src/ocean/text/xml/Document.d

(compare can be seen here: https://codecov.io/gh/sociomantic-tsunami/ocean/compare/v2.x.x...a2849583c58575cb38a612444f97404de65fddd6/changes)

You'll see that the lines that are now not covered, were not actually included in the report before.

leandro-lucarella-sociomantic:
> Rename PR to "fix build matrix"? :P

Oh, I'm already working on a beaver-based solution for this, I would leave this to beaver. See #226. I also fixed the original problem addressed by this PR in beaver.

mihails-strasuns-sociomantic:
I would still prefer to merge this as is right now because it provides a useful baseline of expected coverage stats when later switching to beaver. Fine with you?

mihails-strasuns-sociomantic:
Abandoning to unblock 2.10 as @leandro-lucarella-sociomantic won't be available and this is not important enough. Will be done via https://github.com/sociomantic-tsunami/ocean/pull/226

mihails-strasuns-sociomantic:
This PR will remain to shame me for eternity to come

PR #231, opened 2017-08-30
[WIP] Use beaver to send codecov reports

leandro-lucarella-sociomantic:
```
* submodules/beaver v0.2.0(19f5888)...v0.2.0+9(e7121e1) (8 commits)
  > dlang: Add command to send reports to codecov
  > Merge tag v0.2.1 into v0.x.x v0.2.1
  > Add an make wrapper for dlang projects
  > docker: Allow using a custom image name
  > dlang: Don't try to convert D2 only projects
  (...)
```


.gitmodules
```diff
@@ -3,4 +3,4 @@
 	url = https://github.com/sociomantic-tsunami/makd.git
 [submodule "beaver"]
 	path = submodules/beaver
-	url = https://github.com/sociomantic-tsunami/beaver.git
+	url = https://github.com/leandro-lucarella-sociomantic/beaver.git
```
leandro-lucarella-sociomantic, line 6: Remember to fix this before merging :)

leandro-lucarella-sociomantic:
```
./submodules/beaver/bin/dlang/codecov: 25: ./submodules/beaver/bin/dlang/codecov: BEAVER_DOCKER_VARS: parameter not set
```

I wonder why the build isn't marked as errored :thinking:

leandro-lucarella-sociomantic:
> The exit code of after_success, after_failure, after_script and subsequent stages do not affect the build result. However, if one of these stages times out, the build is marked as a failure.

https://docs.travis-ci.com/user/customizing-the-build

I guess this is what we wanted, if the coverage report sending fails for some reason we don't want to fail the build, right?

leandro-lucarella-sociomantic:
Ah! Finally got it working!

Labels are interesting. We have waaaaay better coverage in `devel` builds compared to `produciton` builds. Which I guess is sort of expected since some stuff might be only hit by asserts or stuff like that, but I think we might try to aim to increase the `production` coverage by calling those debug functions in unittest or integration tests. Another reason for this might be some old tango tests might be still using `assert()` instead of `test()`?

leandro-lucarella-sociomantic:
Also, no coverage for D2, I think that is expected too, right?

Burgos:
Not sure, check my earlier PR #230, we were able to fix the druntime and produce coverage files for D2 as well.

Burgos:
> Another reason for this might be some old tango tests might be still using assert() instead of test()?

I think `F=production` does not include `-release `, so asserts should still be there.

leandro-lucarella-sociomantic:
With this we can enable coverage in integration tests without DMD support to merge them :)

mihails-strasuns-sociomantic:
> With this we can enable coverage in integration tests without DMD support to merge them :)

Only if you run them in separate docker spawn, otherwise it will overwrite previous results from plain tests.

mihails-strasuns-sociomantic:
Ping?

mihails-strasuns-sociomantic:
(should probably be closed now?)

leandro-lucarella-sociomantic:
This is like ancient, right?

leandro-lucarella-sociomantic:
Duplicate of #320

PR #232, opened 2017-09-01
Attempt shutdown of scheduler on SIGTERM by default

mihails-strasuns-sociomantic:
If not handled, signal will just kill the application, which will circumvent
runtime finalization step and thus make impossible to see things like GC stats
which are calculated during that step.

### DIFF
--- a/relnotes/is-scheduler.feature.md
+++ b/relnotes/is-scheduler.feature.md
@@ -1,0 +1,4 @@
+* `ocean.task.Scheduler`
+
+  New `isSchedulerUsed` global function allows to check if scheduler was
+  initialized previously in the application.
--- a/src/ocean/task/Scheduler.d
+++ b/src/ocean/task/Scheduler.d
@@ -933,6 +933,18 @@ body
 
 /*******************************************************************************
 
+    Returns:
+        'true' if scheduler system was initialized, 'false' otherwise
+
+*******************************************************************************/
+
+public bool isSchedulerUsed ( )
+{
+    return _scheduler !is null;
+}
+
+/*******************************************************************************
+
     Creates or re-creates the scheduler instance.
 
     Re-creating of scheduler is only allowed if previous one doesn't have
--- a/src/ocean/util/app/DaemonApp.d
+++ b/src/ocean/util/app/DaemonApp.d
@@ -79,6 +79,7 @@ public abstract class DaemonApp : Application,
     import NewLog = ocean.util.log.Logger;
     import ocean.util.log.Stats;
 
+    static import core.sys.posix.signal;
 
     /***************************************************************************
 
@@ -324,6 +325,9 @@ public abstract class DaemonApp : Application,
     {
         super(name, desc);
 
+        // DaemonApp always handles SIGTERM:
+        settings.signals ~= core.sys.posix.signal.SIGTERM;
+
         // Create and register arguments extension
         this.args_ext = new ArgumentsExt(name, desc, settings.usage,
             settings.help);
@@ -550,19 +554,32 @@ public abstract class DaemonApp : Application,
 
     /***************************************************************************
 
-        ISignalExtExtension methods dummy implementation.
+        ISignalExtExtension method default implementation.
 
-        This method is implemented with an "empty" implementation to ease
-        deriving from this class.
+        This method is implemented with behaviour most commonly desired in apps
+        that don't do any custom signal handling - attempt cleaner shutdown on
+        SIGTERM signal.
 
         See ISignalExtExtension documentation for more information on how to
-        override this method.
+        override this method with own behaviour. `super.onSignal` is not needed
+        to be called when doing so.
 
     ***************************************************************************/
 
     override public void onSignal ( int signum )
     {
-        // Dummy implementation of the interface
+        switch ( signum )
+        {
+            case core.sys.posix.signal.SIGTERM:
+                // Default implementation to shut down cleanly
+                if (isSchedulerUsed())
+                    theScheduler.shutdown();
+                else
+                    this.epoll.shutdown();
+                break;
+            default:
+                assert(false);
+        }
     }
 
     /***************************************************************************
--- a/relnotes/cleanup.feature.md
+++ b/relnotes/cleanup.feature.md
@@ -1,0 +1,7 @@
+* `ocean.util.app.ext.SignalExt`
+
+  Now instead of empty default implementation, the one that catches `SIGTERM`
+  and tries shutting down scheduler/epoll system is used. It should result in
+  same observable behaviour but with runtime finalizers being run. Applications
+  that already implement their own `onSignal` don't need to do anything either
+  (not even call `super.onSignal`).


leandro-lucarella-sociomantic (commented):
Shouldn't this be done too at least for `SIGINT` and `SIGQUIT`?

mihails-strasuns-sociomantic:
I don't have hard opinion but neither `SIGINT` ("interrupt") nor `SIGQUIT` ("exit and core dump") seem to strictly match to "cleanly quit as you please" behaviour I add here for `SIGTERM`. As this is subtle change in behaviour and I don't want to target `v4.x.x` I'd rather pick the minimal useful option.

As `SIGTERM` is the first of signals sent by turtle on termination is is sufficient to fix stat problem for me.

nemanja-boric-sociomantic (commented):
I have a mixed feelings about handling this signal on this level, and not letting application handle it (mostly because I work with the applications which must do a non-trivial amount of work before actually exiting). All of them anyway override `onSignal`, though.

On the other hand, for the applications that don't use Scheduler, this is a breaking change. Application will simply do nothing with the `SIGTERM` if there's no scheduler involved. Perhaps if there's no scheduler application should `kill(0, SIGTERM)` again?

Last but not least - I'm not sure how this worked for you, because by default `SIGTERM` is not caught by `SignalExt` at all. You want to OR `SIGTERM` here for to make this reliably to work: https://github.com/sociomantic-tsunami/ocean/blob/v2.x.x/src/ocean/util/app/DaemonApp.d#L384

mihails-strasuns-sociomantic:
> All of them anyway override onSignal, though.

That was exactly my logic - if apps already override `onSignal`, nothing will change for them. If they don't, they simply rely on default signal handling and thus implementing anything that works in the spirit of default handling is fine.

> Perhaps if there's no scheduler application should kill(0, SIGTERM) again?

Wouldn't that result in infinite signal loop? Maybe it should simply call `exit(1)`?

> Last but not least - I'm not sure how this worked for you, because by default SIGTERM is not caught by SignalExt at all.

Good catch - I was adding it in apps but forgot in this PR.

nemanja-boric-sociomantic:
> Wouldn't that result in infinite signal loop? Maybe it should simply call exit(1)?

Yes, of course, didn't had my coffee yet.

I hope nobody installs signal handler on their own. (If we want to be pedantic, removing its registration from signalext here, if possible, and then raising it would work).

mihails-strasuns-sociomantic:
Rebased and updated with `exit` and signal registration.

leandro-lucarella-sociomantic:
If this is a hack for turtle then, shouldn't turtle inject this behaviour in some way instead?

mihails-strasuns-sociomantic:
1) It is not specific to turtle, you will hit same issue trying to get GC stats by running app manually (but I use it with turtle a lot as source of more realistic work)
2) Turtle can't modify tested apps in any way by design :) Well, other than via unix control socket commands.

leandro-lucarella-sociomantic:
> I don't have hard opinion but neither SIGINT ("interrupt") nor SIGQUIT ("exit and core dump") seem to strictly match to "cleanly quit as you please"

With `SIGQUIT`, ok, but with `SIGINT` it is super common to expect applications to quit cleanly. Almost all programs handle `SIGINT` these days.

> It is not specific to turtle, you will hit same issue trying to get GC stats by running app manually (but I use it with turtle a lot as source of more realistic work)

But in those cases you should implement the `onSignal()` yourself as far as I understood from the thread. But anyway, I guess I'm too distant from this to have a meaningful input. I declare myself neutral about this.

src/ocean/util/app/DaemonApp.d
```diff
@@ -561,7 +565,18 @@ public abstract class DaemonApp : Application,
 
     override public void onSignal ( int signum )
     {
-        // Dummy implementation of the interface
```
gavin-norman-sociomantic, line 565: The method doc should be updated.

src/ocean/util/app/DaemonApp.d
```diff
@@ -323,6 +324,9 @@ public abstract class DaemonApp : Application,
     {
         super(name, desc);
 
+        // DaemonApp always handles SIGTERM:
+        settings.signals ~= core.sys.posix.signal.SIGTERM;
```
gavin-norman-sociomantic, line 329: What if SIGTERM is already in the list?

src/ocean/util/app/DaemonApp.d
```diff
@@ -561,7 +565,18 @@ public abstract class DaemonApp : Application,
 
     override public void onSignal ( int signum )
     {
-        // Dummy implementation of the interface
+        switch ( signum )
+        {
+            case core.sys.posix.signal.SIGTERM:
```
gavin-norman-sociomantic, line 573: How will this change affect apps which already have their own handling for SIGTERM?

nemanja-boric-sociomantic: It should not change, the subclass override should prevail.

src/ocean/util/app/DaemonApp.d
nemanja-boric-sociomantic: Good point, I think either `SignalExt` or `DaemonApp` when instantiating should remove duplicates.

src/ocean/util/app/DaemonApp.d
gavin-norman-sociomantic: Is that desired, though? The special scheduler shutdown behaviour would not occur then.

src/ocean/util/app/DaemonApp.d
mihails-strasuns-sociomantic: I checked that, duplicates are ignored far down the call stack when registering in `SignalFD`.

src/ocean/util/app/DaemonApp.d
mihails-strasuns-sociomantic: Can't really do anything for that case without risking breakage. Maybe I should just mention that in release notes, advising to do the same.

gavin-norman-sociomantic: This is kind of a special case, in that the method is already implemented (i.e. all user code _must_ override it) but has a dummy implementation, so almost all user code will not be calling `super.onSignal(signum)`.

I'm wondering whether this is a breaking change.

nemanja-boric-sociomantic: I would assume that whoever is handling the `SIGTERM` already does `theScheduler.shutdown()` or `epoll.shutdown()` anyway (in which case it will be replaced by `theScheduler.shutdown()` when migrating to the Scheduler). If they are handling SIGTERM without shutting down event loop, they are not exiting cleanly anyway, and I don't see anything is broken.

nemanja-boric-sociomantic: In another words: user code is not _required_ to call `super.onSignal`.

gavin-norman-sociomantic: I see. I think this should definitely be mentioned in the release notes and probably in the method doc as well.

src/ocean/task/Scheduler.d
```diff
@@ -945,6 +945,16 @@ body
 
 /*******************************************************************************
 
+    Returns: 'true' if scheduler system was initialized, 'false' otherwise
+
+*******************************************************************************/
```
nemanja-boric-sociomantic: nitpick: missing newline

relnotes/is-scheduler.feature.md
```diff
@@ -0,0 +1,4 @@
+* `ocean.task.Scheduler`
+
+  New `isSchedulerActive` global function allows to check if scheduler was
```
nemanja-boric-sociomantic: `s/isSchedulerActive/isSchedulerUse/`

src/ocean/util/app/DaemonApp.d
leandro-lucarella-sociomantic: Maybe this should be a set instead (not for now, but for a future PR)

mihails-strasuns-sociomantic:
Updated: documentation adjustments and more detailed release notes

src/ocean/util/app/DaemonApp.d
```diff
@@ -549,19 +553,32 @@ public abstract class DaemonApp : Application,
...
+                // Default implementation to shut down cleanly
+                if (isSchedulerUsed())
+                    theScheduler.shutdown();
+                else
+                    exit(signum);
```
nemanja-boric-sociomantic: I think this is the only loose end in this PR. This is not semantically same thing as removing the installed signal handler for this and killing the program with the same signal. For example, in `systemd`, with `Restart=on-failure` this will cause service to restart (exit with the non-zero code), while in reality it shouldn't (as `SIGTERM` is considered a "clean" signal by default). I would extend `SignalExt` by making portion of this code `public` (https://github.com/sociomantic-tsunami/ocean/blob/v3.x.x/src/ocean/util/app/ext/SignalExt.d#L435) where it should just accept a signal number to restore to the previous handler (akin to `ignoreSignal`) and then call that and call `kill` again.

It's a bit more work, but I don't think it's too hard to get this working.

mihails-strasuns-sociomantic: > SIGTERM is considered a "clean" signal by default)

This doesn't match my observations at all. Simple test with `killall -SIGTERM yes` shows exit status of 143.

nemanja-boric-sociomantic: But that's just shell convention, and the shell is getting this value from the `wait (2)`, inspecting:

```
WIFSIGNALED(stat_val)
Evaluates to a non-zero value if status was returned for a child process that terminated due to the receipt of a signal that was not caught (see <signal.h>).
WTERMSIG(stat_val)
If the value of WIFSIGNALED(stat_val) is non-zero, this macro evaluates to the number of the signal that caused the termination of the child process.
```

so it converts this to an exit return code. In case of the init system, there's no shell to do such thing, and instead of not restarting daemon (on `restart=on-failure` setting of systemd), it will figure out non-zero exit code (inspecting `WEXITED`/`WEXITSTATUS`) and it will restart the program, which is wrong.

nemanja-boric-sociomantic: Here's how bash calculates the exit code: https://github.com/tpruzina/bash/blob/35751626eebdec08c6ac8600b33477e6db31988a/nojobs.c#L216-L224

and here's how it's important to systemd if the service exited via `exit` or via unhandled signal:

https://github.com/systemd/systemd-stable/blob/dbd3dedfaf9165c5b9a0f159ac26c2a2bec2dae1/src/shared/exit-status.c#L165-L180

(using `waitid`:

```
       si_code     Set to one of: CLD_EXITED (child called _exit(2)); CLD_KILLED (child killed by signal); CLD_DUMPED (child killed by signal, and dumped core); CLD_STOPPED (child stopped  by  sig‐
                   nal); CLD_TRAPPED (traced child has trapped); or CLD_CONTINUED (child continued by SIGCONT).
```

mihails-strasuns-sociomantic: Thinking about this again - is there any reason why we can't simply do `epoll.shutdown()` for the second branch? It is not as reliable as scheduler version but worst case SIGKILL will just arrive next :)

nemanja-boric-sociomantic: Huh, I assumed there's no `EpollSelectDispatcher` available in that context. I might be wrong.

nemanja-boric-sociomantic: Aha, there is, it gets assigned to the private `this.epoll` instance here: https://github.com/sociomantic-tsunami/ocean/blob/v3.x.x/src/ocean/util/app/DaemonApp.d#L406

In that case, you don't even need a branch for the scheduler: `DaemonApp.epoll` will point to a scheduler's one.

mihails-strasuns-sociomantic: `theScheduler.shutdown` does a bit more than just calling epoll one - it also kills all tasks and prevents new ones from being spawned

nemanja-boric-sociomantic: Ah, fair enough. I think `this.epoll.shutdown()` in another branch is a fair compromise.

mihails-strasuns-sociomantic:
Updated with `epoll.shutdown` and matching relnotes tweak

mihails-strasuns-sociomantic:
Ping @nemanja-boric-sociomantic when you are back :)

nemanja-boric-sociomantic (approved):

src/ocean/util/app/DaemonApp.d
nemanja-boric-sociomantic: Oh, wait, I just remembered nothing in ocean is using `SignalFD` anymore!

nemanja-boric-sociomantic: Yeah, no deduplication is done: https://github.com/sociomantic-tsunami/ocean/blob/v3.x.x/src/ocean/util/app/ext/SignalExt.d#L409 Could you just use `uniq` either here or down in `SignalExt`?

mihails-strasuns-sociomantic: Updated:

```diff
diff --git a/src/ocean/util/app/DaemonApp.d b/src/ocean/util/app/DaemonApp.d
index 5db601e..4aa84d2 100644
--- a/src/ocean/util/app/DaemonApp.d
+++ b/src/ocean/util/app/DaemonApp.d
@@ -325,7 +325,9 @@ public abstract class DaemonApp : Application,
         super(name, desc);
 
         // DaemonApp always handles SIGTERM:
-        settings.signals ~= core.sys.posix.signal.SIGTERM;
+        auto term = core.sys.posix.signal.SIGTERM;
+        if (find(settings.signals, term) == settings.signals.length)
+            settings.signals ~= term;
```

gavin-norman-sociomantic: >Yeah, no deduplication is done: https://github.com/sociomantic-tsunami/ocean/blob/v3.x.x/src/ocean/util/app/ext/SignalExt.d#L409 Could you just use uniq either here or down in SignalExt?

Mihails said he checked and it's done further down:

>I checked that, duplicates are ignored far down the call stack when registering in SignalFD.

mihails-strasuns-sociomantic: I have assumed @nemanja-boric-sociomantic comment was that we no longer use `SignalFD` for it in most recent ocean. Was I wrong?

nemanja-boric-sociomantic: Correct, we're no longer using `signalfd`, neither in `SignalExt`, nor in `EpollProcess`, since ocean v3.x.x.

nemanja-boric-sociomantic: (we're using self-piping trick).

nemanja-boric-sociomantic:
Travis is red because of the flaky test it seems.

nemanja-boric-sociomantic:
Ok, let's give this a push! LGTM.

PR #235, opened 2017-09-04
Add SmartUnion.handleInvalidCases

nemanja-boric-sociomantic:
Helper const `handleInvalidCases` in `SmartUnion` is added as the support
for the `final switch`. It covers `case none` and `default:` so the actual
switch body is as clean as possible.

### DIFF
--- a/relnotes/smartunion.feature.md
+++ b/relnotes/smartunion.feature.md
@@ -1,0 +1,5 @@
+* `ocean.core.SmartUnion`
+
+  Helper mixin string `handleInvalidCases` in `SmartUnion` is added as the support
+  for the `final switch`. It covers `case none` and `default:` so the actual
+  switch body is as clean as possible.
--- a/src/ocean/core/SmartUnion.d
+++ b/src/ocean/core/SmartUnion.d
@@ -81,6 +81,15 @@ struct SmartUnion ( U )
         return this._.active_names[this._.active];
     }
 
+    /***************************************************************************
+
+        Helper mixin to support final switch over a smart union: it handles the
+        `case none:` in D2-build and `default:` (including `case none:`) in D1.
+
+    ***************************************************************************/
+
+    public alias .handleInvalidCases handleInvalidCases;
+
     /**************************************************************************
 
         Member getter/setter method definitions string mixin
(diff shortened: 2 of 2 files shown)


relnotes/smartunion.feature.md
```diff
@@ -0,0 +1,5 @@
+* `ocean.core.SmartUnion`
+
+  Helper function `handleInvalidCases` in `SmartUnion` is added as the support
```
david-eckardt-sociomantic: It's a plain mixin string now (not a function any more).

nemanja-boric-sociomantic: Ah, I forgot to update it here (already updated commit message).

src/ocean/core/SmartUnion.d
```diff
@@ -289,6 +289,43 @@ version ( UnitTest )
...
+    u.a = 1;
+
+    with (TestSmartUnion.Active) final switch (u.active)
+    {
+        case a:
+            break;
+        case b:
+            break;
+
+        mixin(handleInvalidCases);
```
gavin-norman-sociomantic: This usage makes me wonder if it might make sense to add an alias to `handleInvalidCases` inside `SmartUnion!`. So you could do `mixin(u.handleInvalidCases);` or `mixin(TestSmartUnion.handleInvalidCases);`.

nemanja-boric-sociomantic: Oh, that's clever!

nemanja-boric-sociomantic:
Updated with making `handleInvalidCases` part of `SmartUnion`

src/ocean/core/SmartUnion.d
```diff
@@ -81,6 +81,15 @@ struct SmartUnion ( U )
         return this._.active_names[this._.active];
     }
 
+    /***************************************************************************
+
+        Helper mixin to support final switch over a smart union: it handles the
+        `case none:` in D2-build and `default:` in D1.
```
gavin-norman-sociomantic: "and `default:` (including `case none:`) in D1."

nemanja-boric-sociomantic: Updated.

nemanja-boric-sociomantic:
Oh, it looks red.

nemanja-boric-sociomantic:
```
./src/ocean/core/SmartUnion.d(334): Error: variable u cannot be read at compile time
./src/ocean/core/SmartUnion.d(334): Error: argument to mixin must be a string, not (u , handleInvalidCases) of type immutable(string)
```

nemanja-boric-sociomantic:
Fixed the compile error and the coverage.

gavin-norman-sociomantic (approved):

PR #236, opened 2017-09-05
Don't write the log output to stderr in unittest

nemanja-boric-sociomantic:
By default, the logger used for testing is additive, so it will use
its ancestors Appenders as well. We don't want that, we just want to use
the test Appender.


mihails-strasuns-sociomantic (approved):

mathias-lang-sociomantic:
Oops :blush:

PR #237, opened 2017-09-05
Move the SmartUnion's active member enforcement into method body

### LINKED ISSUE #96: Ocean must pass tests when run with `-release`
Essentially another form of the goal to clean up `assert` usage in ocean to make `-release` flag feasible for our applications. Right now it doesn't even pass tests because some failed assertions are checked with `testThrown`.

nemanja-boric-sociomantic:
The enforcement which was used to confirm that the active member
has been accessed was omitted in the `-release` build. Now it's moved
in the method's body and it still active, no matter if compiled with
`-release`.

Fixes #96

### DIFF
--- a/src/ocean/core/SmartUnion.d
+++ b/src/ocean/core/SmartUnion.d
@@ -20,6 +20,7 @@ import ocean.transition;
 import ocean.core.ExceptionDefinitions;
 import ocean.core.Test;
 import ocean.core.Traits;
+import ocean.core.Verify;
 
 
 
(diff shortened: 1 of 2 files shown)


mihails-strasuns-sociomantic:
`enforce` or `verify`? Is the intent of smart union that one shall never call `get` with wrong type?

nemanja-boric-sociomantic:
I think so, I'll use verify.

nemanja-boric-sociomantic:
Updated. I've also added the matrix item with `-release`, let's see if that work.

nemanja-boric-sociomantic:
Ok, passing `DFLAGS=-release` to travis works:

https://travis-ci.org/sociomantic-tsunami/ocean/jobs/272091389

src/ocean/core/SmartUnion.d
```diff
@@ -503,9 +503,9 @@ private template Methods ( U, uint i )
     const type = "typeof(" ~ member_access ~ ")";
 
     const get = type ~ ' ' ~  member ~ "() "
-        ~ "in { enforce(_.active == _.active." ~ member ~ ", "
```
gavin-norman-sociomantic, line 506: Oof sneaky!

gavin-norman-sociomantic (approved):

mihails-strasuns-sociomantic (approved):

PR #239, opened 2017-09-07
Deprecate timer* methods in ocean.stdc.posix.sys.time

nemanja-boric-sociomantic:
They are available in runtime bindings inside core.sys.posix.sys.time.

### DIFF
--- a/src/ocean/stdc/posix/sys/time.d
+++ b/src/ocean/stdc/posix/sys/time.d
@@ -16,6 +16,7 @@ module ocean.stdc.posix.sys.time;
 
 public import core.sys.posix.sys.time;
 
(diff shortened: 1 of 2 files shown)


src/ocean/stdc/posix/sys/time.d
```diff
@@ -16,6 +16,7 @@ module ocean.stdc.posix.sys.time;
 
 public import core.sys.posix.sys.time;
 
+deprecated ("ocean.stdc.posix.sys.time.timeradd is deprecated. Use core.sys.posix.sys.time.timeradd instead.")
```
nemanja-boric-sociomantic, line 19: Long lines because D1 can't cope with `deprecated ("ocean.xxx deprecated " ~ "use blah instead")`

mihails-strasuns-sociomantic (approved):

PR #240, opened 2017-09-07
Explain rationale behind `Test` module

### LINKED ISSUE #238: Enhance doc of ocean.core.Test
The module header just says "These helpers are supposed to be used in unittest blocks instead of asserts", but doesn't explain _why_. This needs to be explained.

mihails-strasuns-sociomantic:
Fixes #238


gavin-norman-sociomantic:
Thanks!

gavin-norman-sociomantic (approved):

PR #241, opened 2017-09-07
Replace most asserts with `test` in test blocks

mihails-strasuns-sociomantic:
Done via automatic conversion + manual post-processing. Few module
(`ocean.meta.*`, `ocean.core.Enforce`) keep using assert intentionally
to minimize import cycles.


gavin-norman-sociomantic:
>Few module (ocean.meta.*, ocean.core.Enforce) keep using assert intentionally to minimize import cycles.

Is there a comment noting this in these modules? Would be nice to explicitly note why the exception.

mihails-strasuns-sociomantic:
Sure, added module header comments.

gavin-norman-sociomantic (approved):

gavin-norman-sociomantic:
Ok let's do this.

PR #243, opened 2017-09-08
Add opIn and pointer-returning `BTreeMap.insert`.

nemanja-boric-sociomantic:
This adds a missing part of the API which allows accessing the stored values by pointer. Any change to these items will not affect the integrity of the map, since the map is ordered by keys, not by values.

### DIFF
--- a/src/ocean/util/container/btree/BTreeMap.d
+++ b/src/ocean/util/container/btree/BTreeMap.d
@@ -197,6 +197,28 @@ struct BTreeMap(TreeKeyType, TreeValueType, int tree_degree)
         return this.impl.get(key, found_element);
     }
 
+    /**************************************************************************
+
+        Searches the tree for the element with the given key and returns the
+        pointer to the associated value.
+
+        Params:
+            key =  key to find in a tree.
+
+        Returns:
+            pointer to the value associated with the key, if found, null otherwise.
+
+        Complexity:
+            CPU: O(degree * log_degree(n))
+            Memory:O(log_degree(n))
+
+    ***************************************************************************/
+
+    public ValueType* opIn_r (KeyType key)
+    {
+        return this.impl.get(key);
+    }
+
     /***********************************************************************
 
         Recursive inorder iteration over keys and values. Note that, in case
@@ -320,6 +342,12 @@ unittest
 
     // Let's check that the memory is ready to be reused
     test(buff.ptr !is null);
+
+
+    // Use pointer-returning methods
+    auto x = 5 in map;
+    test(x !is null);
+    test!("==")(*x, MyVal(10));
 }
 
 /*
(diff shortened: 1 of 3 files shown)


src/ocean/util/container/btree/BTreeMap.d
```diff
@@ -148,9 +148,40 @@ struct BTreeMap(TreeKeyType, TreeValueType, int tree_degree)
...
+            Pointer to the element associated with the given key (either existing
+            or new value).
+
+        Complexity:
+            CPU: O(degree * log_degree(n))
+            Memory:O(log_degree(n))
+
+    ***************************************************************************/
+
+    public ValueType* insert (KeyType key, ValueType value, out bool added)
```
mihails-strasuns-sociomantic, line 179: Meh, seeing more of the `out bool something` pattern I ask rest of core team to reconsider addition of `Optional` type into ocean :)

nemanja-boric-sociomantic: I think that's not a bad idea at all.

nemanja-boric-sociomantic: If you make a PR to get it included in ocean, I'll go and base this PR on top of that ;-)

src/ocean/util/container/btree/Implementation.d
```diff
@@ -279,6 +278,28 @@ package struct BTreeMapImplementation (KeyType, ValueType, int tree_degree)
...
+
+    *******************************************************************************/
+
+    package ValueType* get (KeyType key)
+    {
+        size_t index;
+        auto node = this.get(key, index);
+        if (!node) return null;
+
+        return &node.elements[index].value;
```
david-eckardt-sociomantic:
```d
if (auto node = this.get(key, index))
    return &node.elements[index].value;
return null;
```

gavin-norman-sociomantic (commented):
LGTM, pending the changes for `Optional`.

src/ocean/util/container/btree/BTreeMap.d
nemanja-boric-sociomantic: I'm now looking again into this and I think there's no need for `Optional` here :-(. Trick is that this method always returns a pointer, either to the old value or the new inserted value, passing the `bool` to the user to say if it added or removed the new element (it's akin to `getOrCreate` in Ocean's caches, really). Do you have proposal for the better API?

gavin-norman-sociomantic: Ha. Yeah, this function's API was misread. So you could add `out (p) { assert(p !is null); }`.

nemanja-boric-sociomantic: Yes, will add, along with the unittests.

nemanja-boric-sociomantic:
Updated with an `out` contract and with unittests for the new features.

src/ocean/util/container/btree/BTreeMap.d
mihails-strasuns-sociomantic: :(

mihails-strasuns-sociomantic (approved):

PR #245, opened 2017-09-11
UnitTestRunner: Fix C string handling in D2

leandro-lucarella-sociomantic:
In D2 apparently command-line arguments don't come with a null-terminator as in D1, so we need to make sure to add one before sending the string to C functions.

### DIFF
--- a/src/ocean/core/UnitTestRunner.d
+++ b/src/ocean/core/UnitTestRunner.d
@@ -764,8 +764,9 @@ private scope class UnitTestRunner
     {
         // we don't care about freeing anything, is just a few bytes and the program
         // will quite after we are done using these variables
-        char* bin_c = strdup(args[0].ptr);
-        char* prog_c = basename(bin_c);
+        auto prog = args[0].dup;
+        prog ~= '\0'; // Make sure is null-terminated
+        char* prog_c = basename(prog.ptr);
         this.prog = prog_c[0..strlen(prog_c)];
 
         cstring[] unknown;


mathias-lang-sociomantic:
Yeah you'd do that with `core.runtime: Runtime.cArgs` normally, but Tango doesn't have it

joseph-wakeling-sociomantic:
> Yeah you'd do that with core.runtime: Runtime.cArgs normally, but Tango doesn't have it

Why not use a `version` block to differentiate?  Can't we rely on `core.*` for D2 builds?

src/ocean/core/UnitTestRunner.d
```diff
@@ -764,8 +764,9 @@ private scope class UnitTestRunner
     {
         // we don't care about freeing anything, is just a few bytes and the program
         // will quite after we are done using these variables
-        char* bin_c = strdup(args[0].ptr);
-        char* prog_c = basename(bin_c);
+        auto prog = args[0].dup;
```
joseph-wakeling-sociomantic, line 767: Minor note: it predates this patch, but the comment should read `quit`, not `quite` ;-)

mathias-lang-sociomantic:
Our normal policy is to avoid `version` block unless strictly necessary.

joseph-wakeling-sociomantic (approved):
LGTM.  I can confirm it fixes the observed issue (see: https://github.com/sociomantic-tsunami/makd/pull/97#discussion_r138079767)

mihails-strasuns-sociomantic (approved):

joseph-wakeling-sociomantic:
Is a point release with this fix planned for any time soon (i.e. today/tomorrow)?  Or are there other things blocking it?

mihails-strasuns-sociomantic:
This is only fixed issue so far and it doesn't affect anything critical so I'd wait a bit more if new problems arise.

joseph-wakeling-sociomantic:
OK, no worries.  I'm just a bit in two minds about whether or not to update projects to makd v2.0.x without it.

leandro-lucarella-sociomantic:
MakD v2 and this problem are totally orthogonal. This was (and still is) present in all MakD versions (because it has nothing to do with MakD, you just happened to notice it when movint to makd2 :D).

PR #246, opened 2017-09-12
Add `ocean.core.Option`

mihails-strasuns-sociomantic:


relnotes/option.feature.md
```diff
@@ -0,0 +1,24 @@
...
+
+      foo(true).visit(
+          ()              { test(false); },
+          (ref int value) { test(value == 42); }
+      );
+
+      foo(false).visit(
+          ()              { },
+          (ref int value) { test(false); }
+      );
```
nemanja-boric-sociomantic: Should also document `.get`

src/ocean/core/Option.d
```diff
@@ -0,0 +1,236 @@
+/******************************************************************************
+
+    Wraps any type in a struct that also contains boolean field indicating if
+    value is in defined state.
+
+    If T is a value type, `Option!(T)` is value type too.
+
+    Copyright:
+        Copyright (c) 2009-2016 Sociomantic Labs GmbH.
```
gavin-norman-sociomantic: 2017

src/ocean/core/Option.d
```diff
@@ -0,0 +1,236 @@
...
+
+module ocean.core.Option;
+
+version(UnitTest)
+{
+    import ocean.core.Test;
+}
+
+/// ditto
+struct Option ( T )
```
gavin-norman-sociomantic: Why `Option` not `Optional`? The word "option" makes me think of a choice between an arbitrary number of things (A, B, C, etc...), whereas "optional" implies a single thing that may or may not be present.

relnotes/option.feature.md
```diff
@@ -0,0 +1,24 @@
...
+      Option!(int) foo ( bool x )
+      {
+          if (x)
+              return option(42);
+          else
+              return undefined!(int);
+      }
+
+      foo(true).visit(
+          ()              { test(false); },
```
gavin-norman-sociomantic: Just for the sake of example simplicity: these `()` aren't necessary, are they?

src/ocean/core/Option.d
```diff
@@ -0,0 +1,236 @@
...
+        ()              { },
+        (ref int value) { test(false); }
+    );
+}
+
+/******************************************************************************
+
+    Shortcut on top of Option to created undefined value.
+
+    Exists only for syntax consistentcy with `option` helper function.
```
nemanja-boric-sociomantic: `consistency` typo

src/ocean/core/Option.d
```diff
@@ -0,0 +1,236 @@
...
+
+    /**************************************************************************
+
+        Puts `this` into undefined state
+
+    **************************************************************************/
+
+    public void reset()
+    {
+        this.defined = false;
```
david-eckardt-sociomantic: Maybe reset `this = this.init` so that `this.value` is reset as well? Otherwise, if `T` is or contains reference types, there could be dangling references (which shouldn't be possible to use, but IMHO it's a good idea to avoid them if possible), or garbage collection could be prevented by keeping a pointer to an object that is in fact not used any more.

src/ocean/core/Option.d
```diff
@@ -0,0 +1,236 @@
...
+
+    foo(false).visit(
+        ()              { },
+        (ref int value) { test(false); }
+    );
+}
+
+/******************************************************************************
+
+    Shortcut on top of Option to created undefined value.
```
nemanja-boric-sociomantic: `to create` and I feel you're missing some articles here, but I can't place them myself :-)

gavin-norman-sociomantic (commented):
Very nice!

src/ocean/core/Option.d
```diff
@@ -0,0 +1,236 @@
...
+        else
+            cb_undefined();
+    }
+
+    /**************************************************************************
+
+        A more "old-school" version of `visit`. Provided both to conform style
+        of existing code and avoid delegate bugs in dmd1.
+
+        Discouraged by default as more error-prone than `visit`.
```
nemanja-boric-sociomantic: ok, this might be a reason why `get` is not inside release notes.

src/ocean/core/Option.d
```diff
@@ -0,0 +1,236 @@
...
+        Boost Software License Version 1.0. See LICENSE_BOOST.txt for details.
+        Alternatively, this file may be distributed under the terms of the Tango
+        3-Clause BSD License (see LICENSE_BSD.txt for details).
+
+
+*******************************************************************************/
+
+module ocean.core.Option;
+
+version(UnitTest)
```
mathias-lang-sociomantic: Style: Lack of spacing before parenthesis everywhere

src/ocean/core/Option.d
```diff
@@ -0,0 +1,236 @@
...
+
+        This will become more convenient with D2 when lambda type inference
+        will happen, as well as shorter notation:
+
+        ```
+            value.visit(
+                ()  => { },
+                (x) => { }
+            );
+        ```
```
mathias-lang-sociomantic: `---` not backtips

src/ocean/core/Option.d
gavin-norman-sociomantic: "to create an"

src/ocean/core/Option.d
```diff
@@ -0,0 +1,236 @@
...
+    /**************************************************************************
+
+        A more "old-school" version of `visit`. Provided both to conform style
+        of existing code and avoid delegate bugs in dmd1.
+
+        Discouraged by default as more error-prone than `visit`.
+
+        Parameters:
+            value = out parameter, will be written to with content of `this` if
+            defined
```
david-eckardt-sociomantic: "out parameter" is a confusing wording here, it is not `out T value`. Suggest "Will be set to the content of `this` if defined or remain unchanged if undefined".

src/ocean/core/Option.d
nemanja-boric-sociomantic: Is there some overload for `cast(bool)` conversion? Maybe it should default to `defined`? Or be disabled, and force users to use `isDefined`?

src/ocean/core/Option.d
```diff
@@ -0,0 +1,236 @@
...
+
+    Parameters:
+        T = type to pass to Option
+
+    Returns:
+        `Option!(T).undefined`
+
+******************************************************************************/
+
+public Option!(T) undefined ( T ) ( )
```
mathias-lang-sociomantic: Since there is no type inference (unlike `option`) I don't think this adds much value

src/ocean/core/Option.d
nemanja-boric-sociomantic: also lowercase

src/ocean/core/Option.d
```diff
@@ -0,0 +1,236 @@
...
+            defined
+
+        Return:
+            `true` if this is defined and `value` was updated.
+
+    **************************************************************************/
+
+    public bool get ( ref T value )
+    {
+        if (this.defined) value = this.value;
```
nemanja-boric-sociomantic: No test coverage for `get` and for `isDefined`.

src/ocean/core/Option.d
nemanja-boric-sociomantic: I wonder if it _should_ be `out T value`, i.e. if it ever sense to leave the parameter unchanged.

src/ocean/core/Option.d
nemanja-boric-sociomantic: Talking with @david-eckardt-sociomantic seems that there's no implicit cast to bool for the struct anyway, and that `opCast` covers _explicit_ casts, so this is not very useful to have.

src/ocean/core/Option.d
david-eckardt-sociomantic: `ref` allows the caller to use a default value other than `T.init`.

src/ocean/core/Option.d
mihails-strasuns-sociomantic: I am in doubts. `Option` is how this concept is usually named in various functional languages (say, Haskell). But from language PoV indeed `Optional` makes more sense.

relnotes/option.feature.md
mihails-strasuns-sociomantic: For empty arguments they are not, correct. I think in such situations it is better to still put them to simplify formatting and make "undefined" case more obvious.

src/ocean/core/Option.d
mihails-strasuns-sociomantic: Correct. I added docs and tests for it though.

src/ocean/core/Option.d
mihails-strasuns-sociomantic: It is not required by style guide. Ocean has both styles with more recent code leaning towards lack of space (though I am probably to blame for this, being author of some added packages).

src/ocean/core/Option.d
mihails-strasuns-sociomantic: Yeah, I think `ref` is more practical. I will adjust docs.

src/ocean/core/Option.d
mihails-strasuns-sociomantic: I was going for plain visual syntax consistency with `option` but fter thinking more about it I agree it is not worth new function.

mihails-strasuns-sociomantic:
Updated with all required changes other than `Option` vs `Optional` naming thing (want second-thought confirmation for that).

src/ocean/core/Option.d
gavin-norman-sociomantic: @nemanja-boric-sociomantic and @mathias-lang-sociomantic :+1:ed my original comment. Does that mean they also prefer `Optional`?

mihails-strasuns-sociomantic: Probably, but I want explicit confirmation that we prefer common sense over widespread naming :)

gavin-norman-sociomantic: Right, I'm asking them for confirmation.

nemanja-boric-sociomantic: Yeah, for me `Option` also implies "take one from the set". IIRC, Haskel calls it `Maybe`, Rust calls it `Option`, and C++ calls it `std::optional`. Common sense tells me that the C++ got it right :-)

nemanja-boric-sociomantic: I say screw it all and let's go with `Maybe` :smile:

src/ocean/core/Option.d
nemanja-boric-sociomantic: I agree.

nemanja-boric-sociomantic:
D2 build is failing.

mihails-strasuns-sociomantic:
Renamed & green

src/ocean/core/Optional.d
```diff
@@ -0,0 +1,210 @@
+/******************************************************************************
+
+    Wraps any type in a struct that also contains boolean field indicating if
+    value is in defined state.
+
+    If T is a value type, `Optional!(T)` is value type too.
```
gavin-norman-sociomantic, line 6: Missing `Params:`. This is a topic that I've been unsure about before, with a module header comment block being `/// ditto`ed onto an aggregate. Seems weird to have `Params:` in the header block, but it seems weird to _not_ have it on the aggregate's comment block.

gavin-norman-sociomantic (approved):

src/ocean/core/Optional.d
mihails-strasuns-sociomantic: I considered same choice and decided `Params` in the module is much weirder.

gavin-norman-sociomantic: Yeah, for a simple case like this with just one template arg, I think this seems fine.

nemanja-boric-sociomantic (approved):

PR #247, opened 2017-09-13
VersionArgExt: Print unknown version items

leandro-lucarella-sociomantic:
Unknown version items present in the `version_info` are not printed, but
if they exist, is probably because the app author thinks is a valuable
piece of information to the user.

This commit prints any unknown version items found.

Related to https://github.com/sociomantic-tsunami/makd/pull/98.

### DIFF
--- a/relnotes/version-arg.feature.md
+++ b/relnotes/version-arg.feature.md
@@ -1,0 +1,7 @@
+* `ocean.util.app.ext.VersionArgsExt`
+
+  The `getLibsVersionsString()` method learned a new optional argument `pop`
+  that will pop any library version found while generating the string.
+
+  Also the version string now will include unknown items present in the
+  `version_info`. See the migration notes for details on the format.
--- a/relnotes/version-arg.migration.md
+++ b/relnotes/version-arg.migration.md
@@ -1,0 +1,9 @@
+* `ocean.util.app.ext.VersionArgsExt`
+
+  The version string generated will now also print any unknown item found in the
+  `version_info` with the following format: ` [key='val', ...]` after the
+  library versions. If there are no unknown items, nothing changes compared to
+  the previous behaviour.
+
+  Even when this is not strictly an API change, you should be careful if you
+  have any scripts that parse the version string or stuff like that.
--- a/src/ocean/util/app/ext/VersionArgsExt.d
+++ b/src/ocean/util/app/ext/VersionArgsExt.d
@@ -453,3 +453,25 @@ unittest
     auto app = new MyApp;
     app.main(["app_name", "--version"]);
 }
+
+/*******************************************************************************
+
+    Test the built version string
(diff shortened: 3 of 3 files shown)


leandro-lucarella-sociomantic:
After this is merged, I will like to rebump the version thingy for v4.

My idea is to:

* Change `--version` to just print the version, as most programs do.
* Add a `--version-info` to print the whole 2MiB long version information (maybe changing the format too, probably only treating the `version` key specially, and then just print all the rest of the items sorted by key to simplify things, and using a more multi-line `key=value` format).

One question is if we should leave full or short version for logging. Maybe the full version but in a one-liner version (multi-lines in the log might be weird).

Opinions?

leandro-lucarella-sociomantic:
Updated to add some tests.

mihails-strasuns-sociomantic:
Wrong base (next minor release will happen in v3.x.x only)

mihails-strasuns-sociomantic (changes requested):
LGTM apart from the base thing

mihails-strasuns-sociomantic:
As for change in major - makes sense, but I'd call it `--build-info` as it is not really about just version anymore.

llucax:
> Wrong base (next minor release will happen in v3.x.x only)

Shouldn't be v2.x.x removed then?

mihails-strasuns-sociomantic:
> Shouldn't be v2.x.x removed then?

I will remove after doing last point release for 2.9.x - don't think neptune-release will work if major branch is missing.

Note though that v3.x.x is already default branch on github.

leandro-lucarella-sociomantic:
Rebased on top of v3.x.x

leandro-lucarella-sociomantic:
https://travis-ci.org/sociomantic-tsunami/ocean/jobs/275419898#L782

```
  run build/devel/tmp/allunittests

core.exception.AssertError@./src/ocean/task/Task.d(326): Assertion failure
```

Ugh? This really doesn't look related to what I did. Do we have flakey tests? It only failed on `DMD=2.071.2.s1 DIST=xenial F=devel`.

mihails-strasuns-sociomantic:
Hm, it fires from tests for task timer code so some flakey-ness may be the cause. However this is the first time ever I see it failing in ocean PR so not sure..

leandro-lucarella-sociomantic:
I could try a re-run and see what happens.

leandro-lucarella-sociomantic:
FLAKEY

mihails-strasuns-sociomantic:
Damn :/

leandro-lucarella-sociomantic:
Ah, release notes are outdated, I'll make a tiny adjustment before merging. Hold on!

leandro-lucarella-sociomantic:
Updated.

mihails-strasuns-sociomantic (approved):

PR #252, opened 2017-09-14
Fix bug in ICache.remove_

gavin-norman-sociomantic:


gavin-norman-sociomantic:
This bug is harmless in practice. Will include these commits in a PR against the current minor branch.

PR #253, opened 2017-09-15
Convert assert -> verify in util.container

gavin-norman-sociomantic:


src/ocean/util/container/AppendBuffer.d
```diff
@@ -549,12 +551,9 @@ public class AppendBuffer ( T, Base: AppendBufferImpl ): Base, IAppendBufferRead
          **********************************************************************/
 
         T cut ( )
-        in
-        {
-            assert (this.length, "cannot cut last element: content is empty");
-        }
-        body
         {
+            verify (this.length > 0, "cannot cut last element: content is empty");
```
nemanja-boric-sociomantic, line 555: Only thing that was scaring me while I was doing the same is that, if the argument is signed (granted, it's not in this case), _and_ negative value is expected (think returning negative value on error, positive on success), this will be a change in the semantics. Again, I don't think it matters in this (or most (all?) cases), but we should be careful.

nemanja-boric-sociomantic: I was going to start doing `verify(!!this.length)`, but I opted out :-)

gavin-norman-sociomantic: Yeah, right. I tried to check them, but it's worth double-checking.

src/ocean/util/container/Container.d
```diff
@@ -320,7 +322,7 @@ struct Container
 
                 void collect (T* p)
                 {
-                        assert (p);
+                        verify(p);
```
nemanja-boric-sociomantic: You may want `p !is null` here, like in line 165. (and it matches the preference of @david-eckardt-sociomantic )

gavin-norman-sociomantic: Huh, not sure how this slipped through.

src/ocean/util/container/ebtree/EBTree128.d
```diff
@@ -353,16 +353,14 @@ class EBTree128 ( bool signed = false ) : IEBTree
...
-    {
-        assert (node !is null, "attempted to add null node (node pool returned null?)");
-    }
     out (node_out)
     {
         assert (node_out !is null);
     }
     body
     {
+        verify (node !is null, "attempted to add null node (node pool returned null?)");
```
nemanja-boric-sociomantic, line 364: Missing import?

gavin-norman-sociomantic: Yeah, curiously it compiles, though. (From the super class?)

nemanja-boric-sociomantic: Yeah, I don't think it should compile with v2.071.x though. Ah, Travis is red anyway.

nemanja-boric-sociomantic: Yeah, lot of import errors with new rules: https://travis-ci.org/sociomantic-tsunami/ocean/jobs/275789554#L763

gavin-norman-sociomantic: Ahh damn. I guess I should do whatever magic is required to update my local compiler.

nemanja-boric-sociomantic: Though, right now only swarm and ocean will compile with dmd v2.071.x compilers, but you can use it for testing/developing.

src/ocean/util/container/map/Map.d
```diff
@@ -1009,12 +1009,9 @@ public abstract class Map ( size_t V, K ) : BucketSet!(V, K)
     ***************************************************************************/
 
     public void[] opIndexAssign ( void[] val, K key )
-    in
-    {
-        assert (val.length == V, "expected a value of length " ~ V.stringof);
-    }
-    body
     {
+        verify (val.length == V, "expected a value of length " ~ V.stringof);
```
nemanja-boric-sociomantic, line 1015: Missing import.

gavin-norman-sociomantic:
This pull request was [updated](https://gist.github.com/b8d2244a805d59cd4e2545f7249eb0e1). Previous HEAD was 6e4db7dbc62f7c7b6c63b3e3dec8d72f3d25d4f9

mihails-strasuns-sociomantic:
For me big question here is about container checks inside `protected` and `private` methods. Ideally all those should use `assert` and public method that call them should sanitize input with `verify`. But for `protected` it also means trusting derived classes to do the same - can/should we d so?

gavin-norman-sociomantic:
This pull request was [updated](https://gist.github.com/65ab705d5f22ef8f57d83b1de0ca0b8d). Previous HEAD was 4a64f95639b1ed9610f290934e144a1aa00c0c43

gavin-norman-sociomantic:
>For me big question here is about container checks inside protected and private methods. Ideally all those should use assert and public method that call them should sanitize input with verify. But for protected it also means trusting derived classes to do the same - can/should we d so?

We agreed to do a simple assert -> verify conversion initially, so that these checks all remain in release builds. In the future, we can give the error handling approach of different components a closer look.

mihails-strasuns-sociomantic (approved):

PR #259, opened 2017-09-15
Convert assert to verify/test in math

david-eckardt-sociomantic:
I grouped them by topic:
- Function argument validation and unittests are straight-forward.
- Internal sanity checking are cases where `assert` might actually be correct but I'm not perfectly sure.
- Conversion of `assert(0)` to `static assert(0)` is for functions that support x86-64 but had an `assert(0)` for some other unsupported architecture.
- Conversion of `assert(0)` to `throw new SanityException` for disabled functions is for functions that don't support x86-64. `static assert(0)` cannot be used because the module wouldn't compile then. These functions are useless for us and should be removed (or fixed/extended to support x86-64 if it's worth it).
- Extremely unlikely RNG conditions are obscure, take a look yourself.

### DIFF
--- a/src/ocean/math/Distribution.d
+++ b/src/ocean/math/Distribution.d
@@ -295,13 +295,6 @@ public class Distribution ( T )
 
         auto index = cast(size_t)(fraction * (this.values.length - 1));
 
-        assert(index < this.values.length, "index greater than or equal to length of value list");
-
-        if ( index >= this.values.length )
-        {
-            index = this.values.length - 1;
-        }
-
         return this.values[index];
     }
 
--- a/src/ocean/math/Distribution.d
+++ b/src/ocean/math/Distribution.d
@@ -110,6 +110,7 @@ import ocean.core.Array : bsearch, sort;
 
 import ocean.util.container.AppendBuffer;
 
+version(UnitTest) import ocean.core.Test;
 
 
 /*******************************************************************************
@@ -474,7 +475,7 @@ private void appendDist ( T ) ( Distribution!(T) dist, T[] values )
     Params:
         dg = a delegate that calls an error throwing function
         error = whether an error should have been thrown or not
-        msg = the error message should the assert fail
+        msg = the error message should the test fail
 
 *******************************************************************************/
 
@@ -492,7 +493,7 @@ private void testForError ( bool dummy = false )
         caught = true;
     }
 
-    assert(error == caught, "Error " ~ (!error ? "not " : "")  ~ "expected: " ~ msg);
+    test(error == caught, "Error " ~ (!error ? "not " : "")  ~ "expected: " ~ msg);
 }
 
 
@@ -515,7 +516,7 @@ private void percentValueTests ( T ) ( T[] values, T middle_value )
     auto dist = new Distribution!(T);
 
     // test that percentValue always returns 0 for empty distributions regardless of type
-    assert(dist.percentValue(0.25) == 0, "percentValue should always return 0 for an empty distribution");
+    test!("==")(dist.percentValue(0.25), 0, "percentValue should always return 0 for an empty distribution");
 
     appendDist!(T)(dist, values);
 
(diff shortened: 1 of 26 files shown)


src/ocean/math/Distribution.d
```diff
@@ -295,13 +295,6 @@ public class Distribution ( T )
 
         auto index = cast(size_t)(fraction * (this.values.length - 1));
 
-        assert(index < this.values.length, "index greater than or equal to length of value list");
```
gavin-norman-sociomantic, line 298: Weird code!

gavin-norman-sociomantic (approved):
LGTM

gavin-norman-sociomantic:
Travis doesn't like it, though.

david-eckardt-sociomantic:
`immutable` strikes again.

> ./src/ocean/math/random/engines/URandom.d(95): Error: function ocean.core.Verify.verify (bool ok, lazy string msg = "", string file = __FILE__, int line = __LINE__) is not callable using argument types (bool, char[])

david-eckardt-sociomantic:
Updated.

src/ocean/math/IEEE.d
```diff
@@ -321,7 +322,7 @@ private:
               tmpval &=0xFFFF_FC00;
               asm { ld tmpval, %fsr; }
             */
-           assert(0, "Not yet supported");
+           throw new SanityException("Not yet supported");
```
mihails-strasuns-sociomantic, line 323: Shall we simply deprecate these functions? It is weird to keep code that does nothing but throwing.

david-eckardt-sociomantic: Sure. We could even just remove them because they are literally dead code, no user code can possibly call them.

mihails-strasuns-sociomantic: Let's just follow usual procedure, no benefit in making exceptions :) New major should be soon enough for this to not matter.

david-eckardt-sociomantic:
Updated, deprecating functions that don't support x86-64.

src/ocean/math/IEEE.d
```diff
@@ -307,6 +307,7 @@ private:
            static assert(0, "Not yet supported");
        }
     }
+    deprecated
```
david-eckardt-sociomantic: Could someone with x86 assembly language knowledge take a look at `resetIeeeFlags()`, please ( @don-clugston-sociomantic pr @nemanja-boric-sociomantic )? It seems to me it would work with x86-64 but uses an orphan `version` identifier.

nemanja-boric-sociomantic: It seems that that's still applicable even for x86-64 (as part of x87 FPU).

david-eckardt-sociomantic:
Updated, trivially extending `resetIeeeFlags` to support x86-64.

src/ocean/math/IEEE.d
don-clugston-sociomantic: That's right, there's no difference between x86 and x86-64 in this respect

nemanja-boric-sociomantic (approved):

don-clugston-sociomantic:
Just an observation: there are two very different things that are replaced with `verify` here. The ones which are in the `BigIntInternalXXX` functions are true asserts in the sense that Walter originally intended. They are only supposed to exist while unit tests are running. The ones which were in `in` contracts are `enforce` things which are actually expected to happen, and they need to stay in.

It's a nice example of the naivety of D's assert system.

Apart from that, LGTM.

mihails-strasuns-sociomantic:
@don-clugston-sociomantic yeah we talked about it some time before in core chat - all ocean will really need another pass checking which `verify` can really be an `assert` (and some may even be `enforce` in disguise). Right now agreement is to convert everything blindly to unblock D2 deployment of dhtnode and revisit things later with more educated approach :)

david-eckardt-sociomantic:
@don-clugston-sociomantic, in addition to what @mihails-strasuns-sociomantic says, to address your concern I put the replacement of the different types of assertions in separate commits. See also my first PR comment.

don-clugston-sociomantic:
Cool, that sounds completely reasonable to me. In fact most of the floating point functions should probably return NaN on invalid input parameters, which would sidestep the issue in a lot of cases.

PR #261, opened 2017-09-18
Fix SmartUnion unittest to compile with a modern d2 compiler

stefan-koch-sociomantic:
This pr merely uses string mixins to allow different versions of the switch.
To be used in D1/D2.
`with (...) final switch` seems to rejected by the parser.

### DIFF
--- a/src/ocean/core/SmartUnion.d
+++ b/src/ocean/core/SmartUnion.d
@@ -325,13 +325,25 @@ unittest
     TestSmartUnion u;
     u.a = 1;
 
-    with (TestSmartUnion.Active) final switch (u.active)
+    const switch_string_head = "with (TestSmartUnion.Active)";
+    const switch_string_tail = "switch (u.active)
     {
         case a:
         case b:
             break;
 
         mixin(TestSmartUnion.handleInvalidCases);
+    }";
+
+
+    version (D_Version2)
+    {
+         mixin (switch_string_head ~ "final "
+          ~ switch_string_tail);
+    }
+    else
+    {
+         mixin(switch_string_head ~ switch_string_tail);
     }
 }
 


mihails-strasuns-sociomantic:
> `with (...) final switch` seems to rejected by the parser

It is a legal syntax per https://dlang.org/spec/grammar.html#WithStatement
What compiler version do you use?

gavin-norman-sociomantic:
>with (...) final switch seems to rejected by the parser.

The D2 parser rejects this? It shouldn't. This is a very common pattern.

nemanja-boric-sociomantic:
> This pr merely uses string mixins to allow different versions of the switch.

Problem with this is that this is documented unittest, showing how this (soon) very common pattern
should be used. If this doesn't work with the modern D2 compiler, we should consider removing
support for this.

stefan-koch-sociomantic:
I used `v2.076.0-dirty`

gavin-norman-sociomantic:
Oh, right, yeah, good point about this being an actual usage example!

nemanja-boric-sociomantic:
This compiles with the v2.076.0-dirty:

```
enum E
{
    A,
    B,
    C
}

void main()
{
    E instance;
    with (E) final switch (instance)
    {
        case A:
            break;
        case B:
            break;
        case C:
            break;
    }
}
```

stefan-koch-sociomantic:
@nemanja-boric-sociomantic
Seems about right ... maybe it's another thing which is rejected then ?

stefan-koch-sociomantic:
I've got it wrong :) 
The actual bug is final switch getting rejected by D1 :)

gavin-norman-sociomantic:
Hahaha, I wondered if that's what this was about but then saw you mentioning D2 and assumed it was something else. You need the newer D1 compiler (I forget which number).

nemanja-boric-sociomantic:
v1.081.2 is your lucky number :-)

PR #262, opened 2017-09-19
Fix array of typedefs formatting with StringStructSerializer

stefan-koch-sociomantic:
Just style fixes as the original fix for the problem proofed goofy


stefan-koch-sociomantic:
Before the fix (only in D2) : 
```
Hash[] x (length 3):
         struct Hash:
            ulong value : 571933887645496831
         struct Hash:
            ulong value : 2583082765769247665
         struct Hash:
            ulong value : 6225354332464533606
```

after the fix: 
```
Hash[] x (length 3): [571933887645496831, 2583082765769247665, 6225354332464533606]
```

mihails-strasuns-sociomantic:
Why is this a bug? Is there any project that expects specific string format for typedefed field?

stefan-koch-sociomantic:
The D1 behavior is : 
Hash[] x (length 3): [571933887645496831, 2583082765769247665, 6225354332464533606]

mihails-strasuns-sociomantic:
Yes, I know, but we don't have D1 typedef anymore. There are several places where it had to be emulated to preserve ABI compatibility but in general it is expected that D2 replacement behaves differently. Thus I am asking if there is an app case behind it.

src/ocean/io/serialize/StructSerializer.d
```diff
@@ -32,6 +32,8 @@ import ocean.core.Exception;
 
 import ocean.io.model.IConduit: IOStream, InputStream, OutputStream;
 
+import ocean.io.serialize.TypeId : DropTypedef;
```
mihails-strasuns-sociomantic: https://github.com/sociomantic-tsunami/ocean/blob/v3.x.x/src/ocean/meta/types/Typedef.d#L143

src/ocean/io/serialize/StructSerializer.d
```diff
@@ -555,9 +558,20 @@ struct StructSerializer ( bool AllowUnions = false )
                 // slice array (passing a static array as ref is not allowed)
                 U[] array = field;
 
-                static if ( is(BaseTypeOfArrays!(U) == struct) )
+                alias BaseTypeOfArrays!(U) BT;
+
+                static if ( is(BT == struct) )
                 {
-                    serializeStructArray(array, field_name, serializer, data);
+                    static if ( is(BT.IsTypedef) )
```
mihails-strasuns-sociomantic: https://github.com/sociomantic-tsunami/ocean/blob/v3.x.x/src/ocean/meta/traits/Basic.d#L609

mihails-strasuns-sociomantic (changes requested):

stefan-koch-sociomantic:
@mihails-strasuns-sociomantic the usecase is much more concise output in one of our apps.

src/ocean/io/serialize/StringStructSerializer.d
```diff
@@ -345,18 +345,19 @@ public class StringStructSerializer ( Char )
...
-            array = array to append
             name = name of array item
+            array = array to append
+            typename = name of the baset type (defaults to T.stringof)
 
     ***************************************************************************/
 
     public void serializeArray ( T ) ( ref Char[] output, cstring name,
-        T[] array )
+        T[] array, cstring typename = T.stringof )
```
mihails-strasuns-sociomantic: I'd also suggest small refactoring to make `serializeStructArray` sub-routine of `serializeArray` so that this awkward extra argument is not needed.

mihails-strasuns-sociomantic (dismissed):

src/ocean/io/serialize/StringStructSerializer.d
stefan-koch-sociomantic: I am not sure I understand what you are suggesting.
so you suggest putting typedef-handling logic into serializeStructArray ?

mihails-strasuns-sociomantic:
Ok, I have checked original use case, it looks reasonable. But please make suggested changes so that formatting code remains more or less generic with minimal amount of special cases.

src/ocean/io/serialize/StructSerializer.d
stefan-koch-sociomantic: I'd like to avoid an extra template instantiation when a cheap is-Expression does the trick.
also further up in this file the same is used.

src/ocean/io/serialize/StructSerializer.d
stefan-koch-sociomantic: I'd like to avoid an extra template instantiation when a cheap is-Expression does the trick.
also further up in this file the same is used.

src/ocean/io/serialize/StringStructSerializer.d
mihails-strasuns-sociomantic: Both array and struct array start with generating the very same string - `Hash[] x (length 3):` (the only one which needs "real" name). Branching for structs only happens after.

So it should be possible to just call `serializeArray` for both and do `static if (T == struct) serializeStructArrayContent()` inside it, where `serializeStructArrayContent` is new imaginary method to replace `serializeStructArray`.

src/ocean/io/serialize/StructSerializer.d
mihails-strasuns-sociomantic: And I don't care how much memory compiler consumes if code becomes more readable.

We generate thousands of template instances with versioned serializer, don't do micro-optimizations in code that doesn't matter.

mihails-strasuns-sociomantic: > also further up in this file the same is used

Please adjust it too in that case.

stefan-koch-sociomantic: As you wish :)

src/ocean/io/serialize/StringStructSerializer.d
```diff
@@ -845,4 +846,30 @@ unittest
...
+    {
+        sta.ids[idx] = cast(AdskilletId)(64 + idx);
+    }
+
+    buffer.length = 0;
+    enableStomping(buffer);
+    serializer.serialize(buffer, sta);
+
+    t.test!("==")(buffer.length, 81);
+    t.test(buffer == "struct StructWithTypedefArray:\n" ~
```
mathias-lang-sociomantic: `t.test!("==")`?

src/ocean/io/serialize/StringStructSerializer.d
```diff
@@ -845,4 +846,30 @@ unittest
...
+    struct StructWithTypedefArray
+    {
+        AdskilletId[] ids;
+    }
+
+    StructWithTypedefArray sta;
+
+    sta.ids = new AdskilletId[](4);
+
+    for (int idx = 0; idx < 4; idx++)
```
mathias-lang-sociomantic: Foreach ref?

src/ocean/io/serialize/StructSerializer.d
```diff
@@ -555,9 +558,20 @@ struct StructSerializer ( bool AllowUnions = false )
...
+                    {
+                        alias TypedefBaseType!(BT)[] DT;
+                        serializer.serializeArray(data, field_name, cast(DT) array,
```
mathias-lang-sociomantic: I would just instantiate it with the right template parameters and rely on implicit conversion instead of casting

mihails-strasuns-sociomantic: I don't think D allows implicit casting of `struct { T field; } []` to `T[]` , does it?

stefan-koch-sociomantic: I don't quite get what you mean ;)

src/ocean/io/serialize/StringStructSerializer.d
stefan-koch-sociomantic: All the other tests in this module do == on buffer as well.

src/ocean/io/serialize/StructSerializer.d
mathias-lang-sociomantic: Ah damn no. Ignore then :)

src/ocean/io/serialize/StringStructSerializer.d
mathias-lang-sociomantic: Feel free to add a commit which switch them then. The advantage of having the operator as template argument is that you get a much more useful error message when it fails.

src/ocean/io/serialize/StructSerializer.d
stefan-koch-sociomantic: ./src/ocean/io/serialize/StructSerializer.d(590): Error: ocean.core.Traits.isTypedef(T) at ./src/ocean/core/Traits.d(873) conflicts with ocean.meta.traits.Basic.isTypedef(T) at ./src/ocean/meta/traits/Basic.d(609)

src/ocean/io/serialize/StringStructSerializer.d
stefan-koch-sociomantic: isn't the for loop is clearer ?

mihails-strasuns-sociomantic: I can a hard time imagining situation when it could be clearer. It is also more error-prone.

stefan-koch-sociomantic:
@mihails-strasuns-sociomantic I just fixed the merge conflicts and addressed the comments
... after which it failed agian; looks like it's going to take a little longer.

mihails-strasuns-sociomantic:
Taking over the PR & updated. Now contains bunch of refactorings and cleanup, actual fix boils down to 03bbda3

relnotes/meta-strip-arrays.feature.md
```diff
@@ -0,0 +1,5 @@
+* `ocean.meta.types.Arrays`
+
+  `ocean.core.Traits.BaseTypeOfArrays` was reimplemented as
```
gavin-norman-sociomantic, line 3: "reimplemented" implies that it has the same functionality. The release note should probably explain why one would use `StripAllArrays` over `BaseTypeOfArrays` (or are both still useful in different situations?).

relnotes/typedefBase.feature.md
```diff
@@ -0,0 +1,4 @@
+* `ocean.meta.types.Typedef`
+
+  New helper function `typedefBase` casts value of type created
+  by `mixin(Typedef!())` to its base type, in both D1 an D2.
```
gavin-norman-sociomantic: Typo "D1 and D2"

relnotes/typedefBase.feature.md
```diff
@@ -0,0 +1,4 @@
+* `ocean.meta.types.Typedef`
+
+  New helper function `typedefBase` casts value of type created
```
gavin-norman-sociomantic: Some articles would help here: "The new helper function `typedefBase` casts a value of a type created..."

src/ocean/meta/types/Typedef.d
```diff
@@ -162,3 +162,25 @@ unittest
...
     static assert (is(TypedefBaseType!(MyInt) == int));
 }
+
+/*******************************************************************************
+
+    Helper function to reduce typedef to its base value
+
+    Params:
+        T = type to reduce to base, must be generated by `Typedef`
+
```
gavin-norman-sociomantic, line 172: Returns:

src/ocean/meta/types/Typedef.d
```diff
@@ -162,3 +162,25 @@ unittest
...
+*******************************************************************************/
+
+TypedefBaseType!(T) typedefBase ( T ) ( T value )
```
gavin-norman-sociomantic: `castTypedefToBase` would be clearer (but a bit longer).

src/ocean/io/serialize/StructSerializer.d
```diff
@@ -554,8 +554,9 @@ struct StructSerializer ( bool AllowUnions = false )
             {
                 // slice array (passing a static array as ref is not allowed)
                 U[] array = field;
+                alias StripAllArrays!(U) Element;
 
-                static if ( is(StripAllArrays!(U) == struct) )
+                static if ( is(Element == struct) && !isTypedef!(Element) )
```
gavin-norman-sociomantic: I'm not quite clear why we're excluding arrays of typedefs here now. Maybe a comment would help.

relnotes/meta-strip-arrays.feature.md
mihails-strasuns-sociomantic: It has same functionality and old one is one day going to be deprecated with all `ocean.core.Traits`. That is true for all utils reimplemented from `ocean.core.Traits` - any specific wording you would like to see?

src/ocean/meta/types/Typedef.d
mihails-strasuns-sociomantic: `castToBase`?

gavin-norman-sociomantic: Given that it's in a module named `Typedef`, yep :+1:

mihails-strasuns-sociomantic:
Updated with wording and docs tweaks

relnotes/meta-strip-arrays.feature.md
gavin-norman-sociomantic: What you added answers my question :+1:

gavin-norman-sociomantic (approved):

gavin-norman-sociomantic:
Travis is sad.

mihails-strasuns-sociomantic:
Greenified

PR #263, opened 2017-09-19
Deprecate ocean.text.Text

mathias-lang-sociomantic:
> Another Layout_tango user which is not used anywhere in our project.

Well actually it was used in one single place, but the replacement was trivial. There were some lingering imports for which I've submitted P.R. to cleanup.


gavin-norman-sociomantic (approved):
Ha! Don't think I've ever even seen that module :laughing:

LGTM but will leave for a while in case anyone else has any comments.

mihails-strasuns-sociomantic (approved):

PR #266, opened 2017-09-20
Remove all "imports" comment blocks

mihails-strasuns-sociomantic:


nemanja-boric-sociomantic:
Maybe we should wait with other large chunks of code changes (asserts -> verify) being merged first, to avoid conflicts?

mihails-strasuns-sociomantic:
I did it exactly so that I don't need to split it into separate commits in my verify PRs :P

gavin-norman-sociomantic (approved):
Yeah why not. The time has probably come.

gavin-norman-sociomantic:
Er... red...?

mihails-strasuns-sociomantic:
Overly eager pattern matching :) Fixed

PR #268, opened 2017-09-21
Fix malformed scheduler test case

mihails-strasuns-sociomantic:
- Fix of test case which was presumably causing problems for @ben-palmer-sociomantic 
- Unrelated fix of `task_pending_cleanup` which I noticed while investigating

### DIFF
--- a/src/ocean/task/Task.d
+++ b/src/ocean/task/Task.d
@@ -462,7 +462,16 @@ public abstract class Task : ISuspendable
             Task.task_pending_cleanup = this;
             this.fiber.yieldAndThrow(e);
             Task.task_pending_cleanup = null;
-            return true;
+
+            // if task was resumed with `kill` from 'pending cleanup' state
+            // just quit the method as if it was originally terminated by kill
+            if (this.to_kill)
+            {
+                debug_trace("<{}> termination (killed)", cast(void*) this);
+                return false;
+            }
+            else
+                return true;
         }
 
         debug_trace("<{}> termination (end of main function)", cast(void*) this);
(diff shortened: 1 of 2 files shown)


mihails-strasuns-sociomantic:
@ben-palmer-sociomantic can you check this please?

ben-palmer-sociomantic:
> @ben-palmer-sociomantic can you check this please?

Yes! fixes my issue, thanks!

mihails-strasuns-sociomantic:

src/ocean/task/Task.d
```diff
@@ -462,7 +462,16 @@ public abstract class Task : ISuspendable
             Task.task_pending_cleanup = this;
             this.fiber.yieldAndThrow(e);
             Task.task_pending_cleanup = null;
-            return true;
+
+            // if task was resumed with `kill` from 'pending cleanup' state
+            // just quit the method as if it was originally terminated by kill
+            if (this.to_kill)
+            {
+                debug_trace("<{}> termination (killed)", cast(void*) this);
```
gavin-norman-sociomantic, line 471: Just to check: the `to_kill` flag should remain set to true, in this case?

mihails-strasuns-sociomantic: Yes, but looks like my other fix to reset flags in `assignTo` method was only merged into later branch. I will add it here too.

mihails-strasuns-sociomantic: Updated

gavin-norman-sociomantic (approved):

PR #269, opened 2017-09-22
Fixes and unit tests for the XML generator

ben-palmer-sociomantic:
Previously the XML generator would slice the values that were passed through for each node. This meant that if you added multiple nodes with different strings from the same buffer the values would be over written.

Additionally the module was missing any unit tests that were actually run.

### DIFF
--- a/src/ocean/text/xml/Document.d
+++ b/src/ocean/text/xml/Document.d
@@ -17,6 +17,7 @@
 
 module ocean.text.xml.Document;
 
+import Array = ocean.core.Array;
 import ocean.transition;
 
 package import ocean.text.xml.PullParser;
@@ -409,9 +410,12 @@ else
                 p.lastChild =
                 p.firstAttr =
                 p.lastAttr = null;
-                p.rawValue =
-                p.localName =
-                p.prefixed = null;
+                p.rawValue.length = 0;
+                enableStomping(p.rawValue);
+                p.localName.length = 0;
+                enableStomping(p.localName);
+                p.prefixed.length = 0;
+                enableStomping(p.prefixed);
 }
                 return p;
         }
@@ -588,9 +592,9 @@ version (Filter)
                 public void*            user;           /// open for usage
                 package Document        doc;            // owning document
                 package XmlNodeType     id;             // node type
-                package T[]             prefixed;       // namespace
-                package T[]             localName;      // name
-                package T[]             rawValue;       // data value
+                package MutT[]          prefixed;       // namespace
+                package MutT[]          localName;      // name
+                package MutT[]          rawValue;       // data value
 
                 package Node            host,           // parent node
                                         prevSibling,    // prior
@@ -688,7 +692,7 @@ version (Filter)
 
                 Node prefix (T[] replace)
                 {
-                        prefixed = replace;
+                        Array.copy(this.prefixed, replace);
                         return this;
                 }
 
@@ -711,7 +715,7 @@ version (Filter)
 
                 Node name (T[] replace)
                 {
-                        localName = replace;
+                        Array.copy(this.localName, replace);
                         return this;
                 }
 
@@ -749,7 +753,7 @@ version(discrete)
                                      if (child.id is XmlNodeType.Data)
                                          return child.value (val);
 }
-                        rawValue = val;
+                        Array.copy(this.rawValue, val);
                         mutate;
                 }
 
@@ -992,7 +996,7 @@ version(discrete)
 }
 else
 {
-                        node.rawValue = value;
+                        Array.copy(node.rawValue, value);
 }
                         return node;
                 }
@@ -1141,8 +1145,8 @@ else
 
                 private Node set (T[] prefix, T[] local)
                 {
-                        this.localName = local;
-                        this.prefixed = prefix;
+                        Array.copy(this.localName, local);
+                        Array.copy(this.prefixed, prefix);
                         return this;
                 }
 
@@ -1155,7 +1159,7 @@ else
                 private Node create (XmlNodeType type, T[] value)
                 {
                         auto node = document.allocate;
-                        node.rawValue = value;
+                        Array.copy(node.rawValue, value);
                         node.id = type;
                         return node;
                 }
--- a/src/ocean/text/xml/Document.d
+++ b/src/ocean/text/xml/Document.d
@@ -152,7 +152,7 @@ version=discrete;
 class Document(T_) : PullParser!(T_)
 {
         public alias NodeImpl*  Node;
-        
+
         private alias Const!(T_)     T;
         private alias T_             MutT;
         private alias XmlPath!(MutT) XmlPathT;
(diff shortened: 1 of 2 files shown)


mathias-lang-sociomantic:
D2 needs some love (red)

src/ocean/text/xml/Document.d
```diff
@@ -409,9 +410,9 @@ else
                 p.lastChild =
                 p.firstAttr =
                 p.lastAttr = null;
-                p.rawValue =
-                p.localName =
-                p.prefixed = null;
+                p.rawValue.length = 0;
```
mihails-strasuns-sociomantic, line 430: `enableStomping`

mihails-strasuns-sociomantic (changes requested):

ben-palmer-sociomantic:
Once the tests are passing I would like to block this until I have done some additional testing in krill for generating the video adgroup XML.

ben-palmer-sociomantic:
Fixed additional problem with the method for adding a header to the xml file always allocating.

src/ocean/text/xml/Document.d
```diff
@@ -151,7 +152,7 @@ version=discrete;
 class Document(T_) : PullParser!(T_)
```
mihails-strasuns-sociomantic, line 152: Do we still want to support non-char `T` here?

mihails-strasuns-sociomantic (changes requested):
Needs release notes. Actual changes look OK but I am really tempted to deprecate wchar/dchar support while we are touching it.

mathias-lang-sociomantic:
> mihails-strasuns-sociomantic added this to the v3.5.0 milestone 12 minutes ago

This is targeting v3.3.x though (I guess because an internal library?). So deprecation can't happen now.

mihails-strasuns-sociomantic:
> This is targeting v3.3.x though

Thanks for noticing! @ben-palmer-sociomantic any reason you target this at point release? These are some non-trivial changes, I wouldn't want to have it in point release until it is really necessary.

ben-palmer-sociomantic:
> @ben-palmer-sociomantic any reason you target this at point release? These are some non-trivial changes, I wouldn't want to have it in point release until it is really necessary.

Well I was viewing these as bugs. Basically extra memory allocations that will cause a memory leak if we try and use this in krill (so yes it would make adding it in to krill a lot easier).

gavin-norman-sociomantic:
Ok so if this is to be included in a patch release, the minor code tidy ups should be removed.

ben-palmer-sociomantic:
> Ok so if this is to be included in a patch release, the minor code tidy ups should be removed.

You mean f4c47375ec6841db7840d97e43add593f0e75e89 and 1911a158305b8baef8a3d531795a91089dd1c106 right?

gavin-norman-sociomantic:
Yeah.

ben-palmer-sociomantic:
Removed minor code tidy ups.

src/ocean/text/xml/Document.d
```diff
@@ -2082,6 +2082,95 @@ public class XmlPath(T)
...
 
+version ( UnitTest )
+{
+    import ocean.core.Test;
+    import ocean.text.xml.DocPrinter;
+}
+
+unittest
+{
+    auto t = new NamedTest("ocean.text.xml.Document");
```
gavin-norman-sociomantic: Why using a named test?

src/ocean/text/xml/Document.d
```diff
@@ -2082,6 +2082,95 @@ public class XmlPath(T)
...
+    // check for no memory allocations.
+    testNoAlloc({
+        doc.reset();
+        generateBasicXML("12345", "one", "one", "two", "two");
+    }());
+
+    auto result2 = printer.print(doc);
+    t.test!("==")(result2, doc2);
+
+    // Generate a more complex XML document.
```
gavin-norman-sociomantic: For readability, maybe split this (and `complex_doc`) into a separate unittest? It's a completely separate test to what comes before, IIUC.

src/ocean/text/xml/Document.d
```diff
@@ -272,7 +272,24 @@ version(d)
 
         ***********************************************************************/
 
-        final Document header (Immut!(T)[] encoding = null)
+        final Document header ()
```
gavin-norman-sociomantic, line 275: Hmm this looks like a new feature, not a bug fix.

ben-palmer-sociomantic: Yeah this is an interesting one. It fixes the bug that you can't use the `header` function without allocation, but I chose to do it by adding a new method.

ben-palmer-sociomantic: Otherwise I would have introduced a breaking change by changing the API.

src/ocean/text/xml/Document.d
ben-palmer-sociomantic: No reason really, I tend to always used named test to get slightly more information on the test. I can remove it though.

src/ocean/text/xml/Document.d
ben-palmer-sociomantic: Yep, will do.

ben-palmer-sociomantic:
Fixed comments on unit tests.

mihails-strasuns-sociomantic:
> Well I was viewing these as bugs.

This is not sufficient to put it into point release per se - the fix itself also has to be more or less straightforward and without any breaking changes in semantics. I don't think this is applicable for this PR. So if it is only matter of convenience (as opposed to something like fixing already released krill version that leaks memory), I'd prefer to move it to minor branch.

ben-palmer-sociomantic:
> This is not sufficient to put it into point release per se - the fix itself also has to be more or less straightforward and without any breaking changes in semantics. I don't think this is applicable for this PR. 

For the sake of argument and so I understand the system better, I fail to see how not being able to add a header to an XML file without allocating is not a bug. Is this only to do with the changes being more complex?

ben-palmer-sociomantic:
Ah, I did just notice that I changed the signature of the old method from:
`final Document header (Immut!(T)[] encoding = null)` to `final Document header (Immut!(T)[] encoding)` which was a mistake.

mihails-strasuns-sociomantic:
> I fail to see how not being able to add a header to an XML file without allocating is not a bug

We have more than one app that doesn't have strict memory allocation policy. So being inefficient does not make something broken on its own. On the other hand every non-trivial change risks introducing new subtle changes and bugs which are currently not reliably catchable with ocean level of testing.

Essentially we try to get to situation where developer can trust any point release to be able to switch to it without any serious testing (unless it is annotated with `+breaking` but that is reserved for truly dire cases). Because of that any doubt is interpreted in favor of moving more controversial fixes to minors or sometimes even majors.

ben-palmer-sociomantic:
> We have more than one app that don't have strict memory allocation policy. So being inefficient does not make something broken on its own. On the other hand every non-trivial change risks introducing new subtle changes and bugs which are currently not reliably catchable with ocean level of testing.

Ok I can understand the logic here. I am just used to having alarms bells go off for unneeded memory allocations :) I'll rebase it on v3.x.x.

mihails-strasuns-sociomantic:
Thanks!

ben-palmer-sociomantic:
Also just added some release notes for the new `header` method.

gavin-norman-sociomantic:
The change for copying values internally rather than slicing is a valid bug fix, I feel. (Albeit a rather wide-ranging one.)

gavin-norman-sociomantic:
>The change for copying values internally rather than slicing is a valid bug fix, I feel. (Albeit a rather wide-ranging one.)

Well, actually, no. The old code just puts the responsibility for handling this onto the user. Ok, new features all around, I guess.

mathias-lang-sociomantic (approved):
Hum this module could really use some cleanup, but I guess that's orthogonal to the aim of this P.R.
I feel like the first commit could really use a release note, as it's a change of semantic.

gavin-norman-sociomantic:
>I feel like the first commit could really use a release note, as it's a change of semantic.

I agree.

ben-palmer-sociomantic:
Added release notes for both commits. I used the same file for both and just appended to it.

mihails-strasuns-sociomantic (approved):

mihails-strasuns-sociomantic:
Thanks!

PR #270, opened 2017-09-22
Verification of `ocean.sys`

mihails-strasuns-sociomantic:


mihails-strasuns-sociomantic:
I think this should be the last one, right?

mihails-strasuns-sociomantic:

src/ocean/sys/Pipe.d
```diff
@@ -116,7 +116,7 @@ class Pipe
         }
         else
         {
-            assert(false, "Unknown platform");
```
gavin-norman-sociomantic, line 119: Haha. Wow.

gavin-norman-sociomantic (approved):

PR #271, opened 2017-09-22
Rework logger test to not rely on state

mihails-strasuns-sociomantic:
Previosuly test case would fail if some other module configures log
system in module constructor in some incompatible way.

### DIFF
--- a/src/ocean/util/log/Logger.d
+++ b/src/ocean/util/log/Logger.d
@@ -988,7 +988,7 @@ unittest
     const TestStr = "Ce qui se conçoit bien s'énonce clairement - Et les mots pour le dire arrivent aisément";
     scope appender = new Buffer();
     char[32] log_buffer;
-    Logger log = Log.lookup("ocean.util.log.Logger.TestLogTrim")
+    Logger log = (new Logger(Log.hierarchy(), "dummy"))
         .add(appender).buffer(log_buffer);
     log.info("{}", TestStr);
     log.error(TestStr);


mathias-lang-sociomantic:
What problem did you see ?

src/ocean/util/log/Logger.d
```diff
@@ -988,7 +988,7 @@ unittest
     const TestStr = "Ce qui se conçoit bien s'énonce clairement - Et les mots pour le dire arrivent aisément";
     scope appender = new Buffer();
     char[32] log_buffer;
-    Logger log = Log.lookup("ocean.util.log.Logger.TestLogTrim")
+    Logger log = (new Logger(Log.hierarchy(), "dummy"))
```
mathias-lang-sociomantic, line 991: We could make it `scope` then

mathias-lang-sociomantic (approved):

mihails-strasuns-sociomantic:
> What problem did you see?

One of turtle integration tests would fail when running this ocean unit tests (this was with earlier makd version which runs integration tests with `-unittest` flag). My understanding is that this turtle module https://github.com/sociomantic-tsunami/turtle/blob/v8.x.x/src/turtle/runner/Logging.d modification of the root logger is the problem.

gavin-norman-sociomantic (approved):

src/ocean/util/log/Logger.d
mihails-strasuns-sociomantic: AFAICS there is not destructor for `Logger` so it shouldn't matter

PR #275, opened 2017-10-01
Version out TypeInfo_Typedef references in D2

Geod24:
It was removed from D2 in 2.076.1.
Since TypeInfo_Enum inherited from it, the fix in RuntimeTraits is just to keep things working.
RuntimeTraits is only used by Layout_Tango and will be killed with it.


mihails-strasuns-sociomantic:
Can we really do that in v3.x.x though? (while `Layout_Tango` is still present)

mihails-strasuns-sociomantic:
Ah, of course, it is not used in our D2 builds anyway.

mihails-strasuns-sociomantic (approved):

PR #276, opened 2017-10-02
Verification of `ocean.text`

mihails-strasuns-sociomantic:
Includes some reverse switching of tests to `assert` as it was originally intended to prevent circular imports.


src/ocean/text/convert/Float.d
```diff
@@ -36,8 +36,6 @@ import ocean.core.ExceptionDefinitions;
 static import tsm = core.stdc.math;
 static import Integer = ocean.text.convert.Integer_tango;
 
-version(UnitTest) import ocean.core.Test;
-
```
mathias-lang-sociomantic, line 41: Two points:
- `test` should be killed with fire, as the compiler should really be more clever about this
- If the first point is not addressed, another approach would be to move those tests to a separate module.

src/ocean/text/xml/Document.d
```diff
@@ -1204,7 +1205,7 @@ else
                         else
                            if (nextSibling)
                               {
-                              debug assert(host.firstChild == this);
```
mathias-lang-sociomantic, line 1207: Those were `debug` tho

src/ocean/text/xml/PullParser.d
```diff
@@ -27,6 +27,11 @@ import Integer = ocean.text.convert.Integer_tango;
 
 import Utf = ocean.text.convert.Utf : toString;
 
+version (UnitTest)
+{
+    import ocean.core.Test;
+}
+
```
mathias-lang-sociomantic, line 34: Commit message looks weird: `Switch tests from assert to`

src/ocean/text/xml/PullParser.d
```diff
@@ -704,60 +709,53 @@ version (UnitTest)
...
+	        test(itr.next);
+	        test(itr.localName == "el3");
+	        test(itr.depth == 2);
+	        test(itr.next);
+	        test(itr.type == XmlTokenType.EndEmptyElement);
+	        test(itr.next);
+	        test(itr.value == "data");
+	        test(itr.next);
+	      //  test(itr.qvalue == "pi", itr.qvalue);
+	      //  test(itr.value == "test");
```
mathias-lang-sociomantic: Since you removed the previous ones, remove this as well ?

src/ocean/text/convert/Float.d
mihails-strasuns-sociomantic: I assume first statement has typo and is supposed to either mean "should not" or `assert`?

src/ocean/text/xml/Document.d
mihails-strasuns-sociomantic: Yep, and thus almost never really run. I think they are simple and useful enough to keep them as verifies.

src/ocean/text/convert/Float.d
mathias-lang-sociomantic: I mean that the compiler should provide an helpful error message on assert error by default.
It could very well turn `assert(a == b)` into `druntimeAssert!("==")(a, b)` or anything of the like. But that's a bit OT.

mihails-strasuns-sociomantic: Yes, but won't make `assert` usable for testing. The fact that it throws `AssertError` is much bigger problem.

mihails-strasuns-sociomantic: (good suggestion about moving all tests to separate module though)

mathias-lang-sociomantic:
D2 is not happy

mihails-strasuns-sociomantic:
Updated:
- moved formatting tests to `_test` module instead of switching back to `assert`
- removed more commented out tests
- fix D2 build

nemanja-boric-sociomantic (approved):

src/ocean/text/convert/Float.d
gavin-norman-sociomantic: Yes, good idea :+1:

src/ocean/text/csv/CSV.d
```diff
@@ -228,7 +228,7 @@ public class CSV
                     // FIXME: if we need to be able to successfully parse
                     // fields like `"hello"world,` then we'll need to come
                     // up with something clever here instead of this assert.
-                    assert(i == row.length - 1 || row[i + 1] == this.separator,
+                    enforce(i == row.length - 1 || row[i + 1] == this.separator,
```
gavin-norman-sociomantic: Could you add `Throws:`?

gavin-norman-sociomantic: Also the doc mentioned "this assert".

gavin-norman-sociomantic: (Suggest moving the whole FIXME up to the Throws section.)

mihails-strasuns-sociomantic:
@gavin-norman-sociomantic done

PR #277, opened 2017-10-02
Support Buffer!void in (de)serializer

mihails-strasuns-sociomantic:
Makes it possible to use `Buffer` in apps working with serialized
structs.


mihails-strasuns-sociomantic:
There is also a very minor breakage in case anyone was using specific way to call `store` (see release notes) but I think we can get away with `3.5.0+breaking` for it.

src/ocean/util/serialize/contiguous/Serializer.d
```diff
@@ -25,6 +25,10 @@ import ocean.core.Traits : ContainsDynamicArray;
 
 import ocean.core.Test;
 
+import ocean.core.Buffer;
+
+import core.memory;
```
gavin-norman-sociomantic, line 28: Looks unused.

relnotes/decorator.feature.md
```diff
@@ -0,0 +1,12 @@
+* `ocean.util.serialize.contiguous.Serializer`
```
gavin-norman-sociomantic, line 1: Isn't this release note for the previous commit?

gavin-norman-sociomantic (commented):
I'm not sure whether the breaking change is necessary for the rest of the PR. Even if it's minor, it'd be nice to stay strict about breaking changes only in major releases.

src/ocean/util/serialize/contiguous/Serializer.d
mihails-strasuns-sociomantic: `GC.setAttr`

mihails-strasuns-sociomantic:
> I'm not sure whether the breaking change is necessary for the rest of the PR

I simply can't introduce new overload of `store` without it. What would be the "rest" in that case?

src/ocean/util/serialize/contiguous/Serializer.d
gavin-norman-sociomantic: Ah yes, I see.

gavin-norman-sociomantic:
I mean the other two commits could be merged alone and the breaking change put into v4.x.x. Is there some urgency? Going for `v3.5.0+breaking` should be a last resort, I'd say.

mihails-strasuns-sociomantic:
I don't see any benefit in adding support to serializer but not version decorator. It will only spread confusion - if we have to move it to v4.x.x, I'd rather move the whole PR.

There doesn't seem to be real urgency - it came from investigating @joseph-wakeling-sociomantic question about using `FreeList` of serialization buffers. When doing so I have realized that `FreeList!(void[])` can't be made to work and `FreeList!(Buffer!(void))` is not applicable because of lack of support in serializer. Which means workarounds and casts with `FreeList!(something_else[])` will have to be used.

So I have evaluated benefit of not adding new hacks to app code against chance of purely syntactical breakage and it felt like worth getting earlier. But I won't argue and move it to v4.x.x if you feel it is not worthy enough :)

gavin-norman-sociomantic:
My feeling is that it's better to stick to the semver norms, given that there's a (reasonably non-nasty) workaround.

The code style commit can stay here.

joseph-wakeling-sociomantic:
> I don't see any benefit in adding support to serializer but not version decorator.

We'd want to use this when serializing versioned structs, so it would be a bit annoying if we couldn't rely on `{Multi}VersionDecorator` support.

The specific use-case is using `FreeList!(Buffer!(void))` to create a pool of recyclable buffers that can be used to serialize data for use with individual DHT `Put` requests.  If this is controversial then we might look into some other solution (e.g. a `StructPool` of `struct S { void[] buffer; }`), but that feels a bit boilerplate-y.

FWIW I think `FreeList!(char[])` should work for our use-case (though @mihails-strasuns-sociomantic has some doubts IIRC), so if there's controversy about this PR, it's fine to take some time to really think through the use-cases and the possible solutions.

mihails-strasuns-sociomantic:
Rebased on latest v4.x.x HEAD.
Is anything else needed?

relnotes/decorator.breaking.md
```diff
@@ -0,0 +1,10 @@
+* `ocean.util.serialize.contiguous.MultiVersionDecorator`
```
nemanja-boric-sociomantic: Is the `breaking` or `migration` correct name for these notes?

relnotes/decorator.feature.md
nemanja-boric-sociomantic: Ping. Not too important, but nice to have.

src/ocean/util/serialize/contiguous/MultiVersionDecorator.d
```diff
@@ -181,6 +314,136 @@ class VersionDecorator
...
+            S = needed struct type
+            Source = struct type seralized into buffer
+
+        Params:
+            buffer = contains serialized Source instance, will be modified to
+                store deserialized S instance instead.
+
+    ***************************************************************************/
+
+    public Contiguous!(S) convert (S, Source) (ref Buffer!(void) buffer)
```
nemanja-boric-sociomantic: Is this intended to be used by the user? If so, does it missing feature notes? If not, can we make `/* package */` comment for the future use?

nemanja-boric-sociomantic: Also, perhaps add some unittests for this method.

mihails-strasuns-sociomantic: Right now it is used from `MapSerializer` - https://github.com/sociomantic-tsunami/ocean/blob/5c01dc0adee97b833054a8cd1efd0ca29c8d8cea/src/ocean/util/container/map/utils/MapSerializer.d#L1181

It is tested as part of generic version decorator tests.

mihails-strasuns-sociomantic: In case that was not clear - it is not a new method, and it was already public. Only difference is that it used to come from mixin and now defined directly in the class.

mihails-strasuns-sociomantic:
Fixed release notes.

mihails-strasuns-sociomantic:

leandro-lucarella-sociomantic (commented):
I have nothing against this PR but I prefer someone else to merge it, as I just gave it a super quick look.

mihails-strasuns-sociomantic:
Rebased, ping

david-eckardt-sociomantic (approved):

PR #283, opened 2017-10-04
Start adding v3 based versions to LibFeatures

mihails-strasuns-sociomantic:
Now that v2.x.x is not supported anymore, having few versio
overlap should simplify things for downstream.


mihails-strasuns-sociomantic:
https://github.com/sociomantic-tsunami/ocean/pull/286

PR #284, opened 2017-10-04
Fix missing v2.10 in LibFeatures and add v3.4

mihails-strasuns-sociomantic:


mihails-strasuns-sociomantic:
FYI @mathias-lang-sociomantic

mathias-lang-sociomantic:
I don't think we should be adding this to v3.4.x. Updating to a patch release should not change functionality, but this adds functionality transitively (because other libraries might be affected)

mihails-strasuns-sociomantic:
You think of case when app upgrades from v3.4.2 to 3.4.3 and some feature in downstream library gets silently enabled as a side effect?

mathias-lang-sociomantic:
Yes. Of course nobody uses it atm so we can get away with it this time I guess, but in general I think that's not something we should allow to be done in patch releases. For my use case it can wait for v3.5.0

mihails-strasuns-sociomantic:
Sadly, your concerns seem totally legit :) Damn, this makes https://github.com/sociomantic-tsunami/neptune/issues/48 even more important as we can't even fix missing flag in point release.

mihails-strasuns-sociomantic:
https://github.com/sociomantic-tsunami/ocean/pull/286

PR #286, opened 2017-10-05
Fix and augment LibFeatures.d versions

mihails-strasuns-sociomantic:
Now that v2.x.x is not supported anymore, having few version
overlap should simplify things for downstream.

Also fixes missing 2_10 version.

### DIFF
--- a/src/ocean/LibFeatures.d
+++ b/src/ocean/LibFeatures.d
@@ -17,13 +17,19 @@
 
 module ocean.LibFeatures;
 
-const has_features_3_5 = true;
-const has_features_2_9 = true;
-const has_features_2_8 = true;
-const has_features_2_7 = true;
-const has_features_2_6 = true;
-const has_features_2_5 = true;
-const has_features_2_4 = true;
-const has_features_2_3 = true;
-const has_features_2_1 = true;
-const has_features_2_0 = true;
+const has_features_3_5  = true;
+const has_features_3_4  = true;
+const has_features_3_3  = true;
+const has_features_3_2  = true;
+const has_features_3_1  = true;
+const has_features_3_0  = true;
+const has_features_2_10 = true;
+const has_features_2_9  = true;
+const has_features_2_8  = true;
+const has_features_2_7  = true;
+const has_features_2_6  = true;
+const has_features_2_5  = true;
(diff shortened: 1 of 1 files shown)


src/ocean/LibFeatures.d
```diff
@@ -17,13 +17,16 @@
...
-const has_features_2_7 = true;
-const has_features_2_6 = true;
-const has_features_2_5 = true;
-const has_features_2_4 = true;
-const has_features_2_3 = true;
-const has_features_2_1 = true;
-const has_features_2_0 = true;
+const has_features_3_5  = true;
+const has_features_3_4  = true;
+const has_features_3_3  = true;
```
mathias-lang-sociomantic, line 22: Not including v3.2.0 / v3.1.0 / v3.0.0 ?
Also, now that we start using it, a question came about the design: Why not just write the library version here, instead of those `has_feature` ? Is it assumed we'll have other usages than those based on the version number ?

gavin-norman-sociomantic (commented):
I don't really understand the commit message.

src/ocean/LibFeatures.d
mihails-strasuns-sociomantic: This design is driven by the fact that we merge features from oldest major upwards. Which means just knowing that your library is currently at version `3.4.6` doesn't tell a thing about if it includes some feature first introduced in 2.10.0, not unless you examine git history. Thus in practice you don't want to check what is the actual library version, you only want to know if it is recent enough to include specific feature set.

So the idea is that at some point of transitioning to new major you have to go through all `static if(hasVersion` cases and change them to use feature flag applicable to new minimal ocean version. And by the time ocean v5.0.0 is out we will be removing all `has_features_2_*` to avoid the list growing indefinitely.

With that all being said I indeed should also add `3_0`, `3_1` and `3_2` now.

mathias-lang-sociomantic: Ah right I completely forgot about multiple major branch. Sounds about right then.

mihails-strasuns-sociomantic:
Updated:
- add other older v3 versions
- reword commit message

gavin-norman-sociomantic (approved):

PR #291, opened 2017-10-06
Fix another log test relying on state

mihails-strasuns-sociomantic:
Similar to the same case in previous unittest block which
was fixed earlier. New test case was introduced in 3.5.0


mathias-lang-sociomantic (approved):
Sorry :/

PR #293, opened 2017-10-08
Implement specialized pool support in the scheduler

mihails-strasuns-sociomantic:
Shows planned design and passes most simple test. PR right now only exists as form of preview and code coverage report abuse.

TODO:

- [x] More comprehensive tests
- [x] Review / fix documentation
- [ ] Test using one of existing applications extensively using tasks

### DIFF
--- a/relnotes/fiber-pool.feature.md
+++ b/relnotes/fiber-pool.feature.md
@@ -1,0 +1,30 @@
+* `ocean.task.Scheduler`
+
+  It is now possible to define dedicated worker fiber pools to process tasks of
+  specific type:
+
+  ```D
+  SchedulerConfiguration config;
+
+  with (config)
+  {
+    specialized_pools = [
+      PoolDescription(MyTask.classinfo, 10240),
+      PoolDescription(MyOtherTask.classinfo, 2048),
+    ];
+  }
+
+  initScheduler(config);
+
+  // will be processed immediately in own worker fiber pool, skipping the queue:
+  theScheduler.schedule(new MyTask);
+  ```
+
+  Such functionality is intended to be used when there are some task types that
+  need real-time handling and thus can't afford waiting in the queue.  It can
+  also be used for the case where some very long-lived tasks would occupy worker
+  fibers from the shared pool forever, if processed normally
+
+  Note that this functionality can be used side by side with regular scheduling
+  queue with no issues - each task will be handled by own sub-system depending
+  on its `ClassInfo`.
--- a/src/ocean/task/Scheduler.d
+++ b/src/ocean/task/Scheduler.d
@@ -27,7 +27,6 @@ import ocean.transition;
 import ocean.core.Enforce;
 import ocean.core.Verify;
 import ocean.core.Traits;
-import ocean.core.array.Mutation : reverse;
 import ocean.io.select.EpollSelectDispatcher;
 import ocean.util.container.queue.FixedRingQueue;
 
@@ -639,31 +638,6 @@ final class Scheduler
         task.assignTo(fiber);
         // execute the task
         bool had_exception = task.entryPoint();
-        assert (task.finished());
-
-        if (task.termination_hooks.length)
-        {
-            debug_trace("Calling {} termination_hooks for task <{}>",
-                task.termination_hooks.length, cast(void*) task);
-
-            auto hooks = reverse(task.termination_hooks[]);
-            task.termination_hooks.reset();
-
-            foreach (hook; hooks)
-            {
-                hook();
-                assert(
-                    task.termination_hooks.length == 0,
-                    "Adding new hooks while running existing ones is not" ~
-                        "supported"
-                );
-            }
-        }
-
-        // allow task to recycle any shared resources it may have (or recycle
-        // task instance itself)
-        debug_trace("Recycling task <{}>", cast(void*) task);
-        task.recycle();
 
         // in case task was resumed after unhandled exception, delay further
         // execution for one cycle to avoid situation where exception handler
--- a/src/ocean/task/Scheduler.d
+++ b/src/ocean/task/Scheduler.d
@@ -31,7 +31,7 @@ import ocean.io.select.EpollSelectDispatcher;
 import ocean.util.container.queue.FixedRingQueue;
 
 import ocean.task.Task;
-import ocean.task.internal.FiberPool;
+import ocean.task.internal.FiberPoolWithQueue;
 
 version (UnitTest)
 {
@@ -127,15 +127,6 @@ final class Scheduler
 
     /***************************************************************************
 
-        Thrown when all fibers are busy, task queue is full and no custom
-        delegate to handle queue overflow is supplied.
-
-    ***************************************************************************/
-
-    private TaskQueueFullException queue_full_e;
-
-    /***************************************************************************
-
         Thrown when task is being added for later resuming via
         Scheduler.processEvents but matching queue is full
 
@@ -149,7 +140,7 @@ final class Scheduler
 
     ***************************************************************************/
 
-    private FiberPool fiber_pool;
+    private FiberPoolWithQueue fiber_pool;
 
     /***************************************************************************
 
@@ -177,11 +168,27 @@ final class Scheduler
 
     /***************************************************************************
 
-        Queue of tasks awaiting idle worker fiber to be executed in
+        Set delegate to call each time task is attempted to be queue but size
+        limit is reached. Both the queue and the task will be supplied as
+        arguments.
 
     ***************************************************************************/
 
-    private FixedRingQueue!(Task) queued_tasks;
+    public void task_queue_full_cb ( TaskQueueFullCB dg )
+    {
+        this.fiber_pool.task_queue_full_cb = dg;
+    }
+
+    /***************************************************************************
+
+        Gets current callback for case of full queue
+
+    ***************************************************************************/
+
+    public TaskQueueFullCB task_queue_full_cb ( )
+    {
+        return this.fiber_pool.task_queue_full_cb;
+    }
 
     /***************************************************************************
 
@@ -190,15 +197,6 @@ final class Scheduler
     ***************************************************************************/
 
     private FixedRingQueue!(Task) suspended_tasks;
-
-    /***************************************************************************
-
-        Called each time task is attempted to be queue but size limit is
-        reached. Both the queue and the task will be supplied as arguments.
-
-    ***************************************************************************/
-
-    public void delegate ( Task, FixedRingQueue!(Task) ) task_queue_full_cb;
 
     /***************************************************************************
 
@@ -263,7 +261,6 @@ final class Scheduler
             config.suspended_task_limit
         );
 
-        this.queue_full_e = new TaskQueueFullException;
         this.suspend_queue_full_e = new SuspendQueueFullException;
 
         verify(
@@ -277,12 +274,13 @@ final class Scheduler
         else
             this._epoll = epoll;
 
-        this.fiber_pool = new FiberPool(
+        this.fiber_pool = new FiberPoolWithQueue(
+            config.task_queue_limit,
             config.worker_fiber_stack_size,
-            config.worker_fiber_limit
+            config.worker_fiber_limit,
+            &this.processEvents
         );
 
-        this.queued_tasks = new FixedRingQueue!(Task)(config.task_queue_limit);
         this.suspended_tasks = new FixedRingQueue!(Task)(config.suspended_task_limit);
     }
 
@@ -298,12 +296,15 @@ final class Scheduler
 
     public void shutdown ( )
     {
-        debug_trace("Shutting down initiated. {} queued tasks will be " ~
-            " discardead, {} suspended tasks will be killed",
-            this.queued_tasks.length(), this.suspended_tasks.length());
+        debug_trace(
+            "Shutting down initiated. {} queued tasks will be " ~
+                " discardead, {} suspended tasks will be killed",
+            this.fiber_pool.queued_tasks.length(),
+            this.suspended_tasks.length()
+        );
 
         this.state = State.Shutdown;
-        this.queued_tasks.clear();
+        this.fiber_pool.queued_tasks.clear();
 
         Task task;
         while (this.suspended_tasks.pop(task))
@@ -329,8 +330,8 @@ final class Scheduler
     public SchedulerStats getStats ( )
     {
         SchedulerStats stats =  {
-            task_queue_busy : this.queued_tasks.length(),
-            task_queue_total : this.queued_tasks.maxItems(),
+            task_queue_busy : this.fiber_pool.queued_tasks.length(),
+            task_queue_total : this.fiber_pool.queued_tasks.maxItems(),
             suspended_queue_busy : this.suspended_tasks.length(),
             suspended_queue_total : this.suspended_tasks.maxItems(),
             worker_fiber_busy : this.fiber_pool.num_busy(),
@@ -367,18 +368,16 @@ final class Scheduler
                 caller_task.kill();
         }
 
-        if (this.fiber_pool.num_busy() >= this.fiber_pool.limit())
+        try
         {
-            this.queue(task);
+            this.fiber_pool.runOrQueue(task);
         }
-        else
+        catch (Exception e)
         {
-            auto fiber = this.fiber_pool.get();
-            debug_trace("running task <{}> via worker fiber <{}>",
-                cast(void*) task, cast(void*) fiber);
-            fiber.reset(&this.worker_fiber_method);
-            task.assignTo(fiber);
-            this.resumeTask(task);
+            if (this.exception_handler !is null)
+                this.exception_handler(task, e);
+            else
+                throw e;
         }
     }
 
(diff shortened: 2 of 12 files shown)


src/ocean/task/internal/FiberPoolEager.d
```diff
@@ -0,0 +1,82 @@
...
+
+        Params:
+            task = derivative from `ocean.task.Task` defining some application
+                task to execute
+
+    ***************************************************************************/
+
+    public void run ( Task task )
+    {
+        auto fiber = this.get();
```
gavin-norman-sociomantic, line 62: So this is an unlimited pool of fibers, right?

src/ocean/task/internal/SpecializedPools.d
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    /***************************************************************************
+
+        Runs task in one of dedicated worker fiber pools if it is registered in
+        the mapping. Returns 'false' otherwise.
+
+    ***************************************************************************/
+
+    bool run ( Task task )
```
gavin-norman-sociomantic: Do you anticipate use cases where the caller isn't sure whether the task is handled by this class or not? I would have expected it to throw if the task isn't of a supported type.

gavin-norman-sociomantic (commented):
Interesting that all specialized task pools are unlimited. What's the concept behind that?

src/ocean/task/internal/FiberPoolEager.d
mihails-strasuns-sociomantic: Yes. My line of though went like that - we are only expected to use those pools for real-time kind of handling where waiting in queue is not appropriate. Like HTTP connection handler. If we allow the limit, what would be a reasonable fallback for case when it is reached? Close the connection, bring down the app? Nothing really made sense. It feels like having unbound limit and just monitoring graphs is the only practical approach.

src/ocean/task/internal/SpecializedPools.d
mihails-strasuns-sociomantic: Right now it is used to simplify code in scheduler. It does something like `if (!custom_pools.run(task)) generic_pools.run(task);`.

gavin-norman-sociomantic: Fair enough.

src/ocean/task/internal/FiberPoolEager.d
gavin-norman-sociomantic: I see. This works fine for normal usage where the number of connections (say) gradually grows over time. We also need to consider erroneous or malicious behaviour, though, where the number of connections could grow extremely quickly. OOM is not an acceptable solution ;)

mihails-strasuns-sociomantic:
I think this is ready for review. Changes from last update:

- new higher level test which I used to manually trace scheduling logic
- bunch of small fixes everywhere

Testing with some real application would be nice but I don't think it is important enough right now to distract app maintainers (as the effort is non-trivial).

src/ocean/task/Scheduler.d
```diff
@@ -27,7 +27,6 @@ import ocean.transition;
 import ocean.core.Enforce;
 import ocean.core.Verify;
 import ocean.core.Traits;
-import ocean.core.array.Mutation : reverse;
```
gavin-norman-sociomantic, line 30: Commit message comment:
"This is the inherent part of running task, not matter the context it is being run from."
->
"This is an inherent part of running a task, not a matter for the context it is being run from."
(Assuming I understand what you want to say.)

src/ocean/task/Task.d
```diff
@@ -456,8 +456,39 @@ public abstract class Task : ISuspendable
         debug_trace("<{}> start of main function", cast(void*) this);
 
         scope(exit)
+        {
             this.state_bitmask |= TaskState.Finished;
 
+            if (this.termination_hooks.length)
```
gavin-norman-sociomantic, line 473: This seems like a potentially dangerous change: running the termination hooks inside the task, rather than from the calling context.

src/ocean/task/internal/FiberPoolWithQueue.d
```diff
@@ -0,0 +1,241 @@
...
+    ***************************************************************************/
+
+    public TaskQueueFullCB task_queue_full_cb;
+
+    /**************************************************************************
+
+        Constructor
+
+        Params:
+            stack_size = fiber stack size to use in this poll
```
gavin-norman-sociomantic: Incomplete params doc.

src/ocean/task/internal/FiberPoolWithQueue.d
```diff
@@ -0,0 +1,241 @@
...
+        {
+            this.queue(task);
+            return;
+        }
+
+        auto fiber = this.get();
+        debug_trace("running task <{}> via worker fiber <{}>",
+            cast(void*) task, cast(void*) fiber);
+        task.assignTo(fiber);
+        // trick task by using different method for real entry point:
```
gavin-norman-sociomantic: This comment is a little mysterious...

src/ocean/task/internal/FiberPoolEager.d
gavin-norman-sociomantic: As far as I remember from our discussion on this before, the conclusion was that this is fine/irrelevant, as the select listener already implements a limit on the number of handled connections. (And rejects any incoming connections beyond the limit.)

src/ocean/task/Scheduler.d
```diff
@@ -69,6 +70,16 @@ struct SchedulerConfiguration
...
     /// `theScheduler.processEvents` in between scheduler dispatch cycles
     size_t suspended_task_limit = 16;
+
+    /// optional array that defines specialized worker fiber pools to be
+    /// used for handling specific task kinds. Scheduled task is checked
+    /// against this array every time thus it is not recommended to configure
+    /// it to more than a few dedicated extra pools
+    PoolDescription[] specialized_pools;
+
+    /// specialized worker fiber pool description. Tasks handled by such / pools
```
gavin-norman-sociomantic: The "/" at the end of this line seems erroneous.

src/ocean/task/Task.d
```diff
@@ -83,6 +83,17 @@ public class WorkerFiber : core.thread.Fiber
 
     /***************************************************************************
 
+        Terminates itself by killing underlying task
```
gavin-norman-sociomantic, line 86: This seems like it should be a separate commit.

src/ocean/task/internal/SpecializedPools.d
```diff
@@ -0,0 +1,124 @@
+/*******************************************************************************
+
```
gavin-norman-sociomantic, line 2: Module description missing.

src/ocean/task/internal/SpecializedPools.d
```diff
@@ -0,0 +1,124 @@
...
+    Defines all mappings between task kinds and dedicated worker fiber pools
+    processing those.
+
+*******************************************************************************/
+
+public class SpecializedPools
+{
+    /// mapping as a simple array (expected to contain few elements)
+    private SpecializedPool[] mapping;
```
gavin-norman-sociomantic, line 53: What's the reason you're not using an AA?

src/ocean/task/internal/SpecializedPools.d
```diff
@@ -0,0 +1,124 @@
...
+    private static struct SpecializedPool
+    {
+        ClassInfo task_kind;
+        FiberPoolEager pool;
+    }
+
+    /***************************************************************************
+
+        Constructor
+
```
gavin-norman-sociomantic, line 65: Params

src/ocean/task/internal/SpecializedPools.d
```diff
@@ -0,0 +1,124 @@
...
+    {
+        foreach (description; config)
+        {
+            verify(description.task_kind !is null);
+            verify(
+                description.stack_size >= 1024,
+                "Configured stack size is suspiciously low"
+            );
+
+            this.mapping ~= SpecializedPool(
```
gavin-norman-sociomantic: Using an AA would also allow us to detect multiple mappings for the same task type.

src/ocean/task/internal/SpecializedPools.d
```diff
@@ -0,0 +1,124 @@
...
+                )
+            );
+        }
+    }
+
+    /***************************************************************************
+
+        Runs task in one of dedicated worker fiber pools if it is registered in
+        the mapping. Returns 'false' otherwise.
+
```
gavin-norman-sociomantic: Params / returns.

src/ocean/task/internal/SpecializedPools.d
```diff
@@ -0,0 +1,124 @@
...
+        return false;
+    }
+
+    /***************************************************************************
+
+        Kills all worker fibers in all pools
+
+    ***************************************************************************/
+
+    void kill ( )
```
gavin-norman-sociomantic: Protection attribute missing on all methods. (All public, I suppose.)

src/ocean/task/internal/SpecializedPools.d
```diff
@@ -0,0 +1,124 @@
...
+        foreach (meta; this.mapping)
+        {
+            scope iterator = meta.pool.new BusyItemsIterator;
+            foreach (ref fiber; iterator)
+                fiber.kill();
+        }
+    }
+}
+
+private void debug_trace ( T... ) ( cstring format, T args )
```
gavin-norman-sociomantic, line 171: Unused?

gavin-norman-sociomantic (commented):
Looking good!

gavin-norman-sociomantic:
Oh, needs release notes, too.

src/ocean/task/Task.d
mihails-strasuns-sociomantic: There is no change, it already works like that (the fiber is the same).

gavin-norman-sociomantic: Ah I see, `runTask` is called from inside the fiber. :+1:

src/ocean/task/internal/FiberPoolEager.d
mihails-strasuns-sociomantic: Correct. Consensus was to think about implementing a limit once we have real app use case that may require it.

src/ocean/task/internal/SpecializedPools.d
mihails-strasuns-sociomantic: Cons: slower, inherent heap fragmentation, have to keep in mind all const-related AA bugs
Pros: nice syntax

I think for purely private entity in stdlib choice is obvious :P

mihails-strasuns-sociomantic:
Updated: https://gist.github.com/mihails-strasuns-sociomantic/ce7257baa7584449dce0c21be29d0a7e

mihails-strasuns-sociomantic:
I am also bumping priority because investigation on https://github.com/sociomantic-tsunami/ocean/pull/335 shows there are pretty big changes that need to be done on scheduler - and I'd prefer to do those on top of refactoring from this PR.

mihails-strasuns-sociomantic:
Added release notes

mihails-strasuns-sociomantic:
Green

src/ocean/task/internal/FiberPoolWithQueue.d
gavin-norman-sociomantic: Still incomplete.

relnotes/fiber-pool.feature.md
```diff
@@ -0,0 +1,30 @@
...
+
+  initScheduler(config);
+
+  // will be processed immediately in own worker fiber pool, skipping the queue:
+  theScheduler.schedule(new MyTask);
+  ```
+
+  Such functionality is intended to be used when there are some task types that
+  need real-time handling and thus can't afford waiting in the queue. It can
+  also be used for case when some tasks are very long-living would hold worker
```
gavin-norman-sociomantic: Grammar / clarity: "It can also be used for the case where some very long-lived tasks would occupy worker fibers from the shared pool forever, if processed normally."

src/ocean/task/internal/FiberPoolWithQueue.d
mihails-strasuns-sociomantic: https://github.com/sociomantic-tsunami/ocean/pull/293/files#diff-2c0e70ae37423c24132186053ded1353R68

gavin-norman-sociomantic: Ah, the comment fix was in the wrong commit.

gavin-norman-sociomantic (commented):
A couple of minor comments.

mihails-strasuns-sociomantic:
Updated:
- relnotes wording
- move doc fix to proper commit

gavin-norman-sociomantic (approved):

gavin-norman-sociomantic:
Feel free to auto-merge when green.

src/ocean/task/internal/FiberPoolWithQueue.d
```diff
@@ -0,0 +1,249 @@
...
+            stack_size = fiber stack size to use in this poll
+            limit = limit to pool size. If set to 0 (default), there is no
+                app limit and pool growth will be limited only by OS
+                resources
+            process_events = reference to scheduler method which does temporary
+                suspend, passed to constructor to avoid circular dependency
+
+    **************************************************************************/
+
+    this ( size_t queue_limit, size_t stack_size, size_t limit,
```
daniel-zullo, line 70: missing protection attribute

src/ocean/task/internal/FiberPoolEager.d
```diff
@@ -0,0 +1,83 @@
...
+    /**************************************************************************
+
+        Constructor
+
+        Params:
+            stack_size = fiber stack size to use in this poll
+
+    **************************************************************************/
+
+    this ( size_t stack_size )
```
daniel-zullo, line 42: missing protection attribute

src/ocean/task/internal/FiberPoolEager.d
```diff
@@ -0,0 +1,83 @@
...
+import ocean.meta.types.Qualifiers;
+import ocean.core.Enforce;
+import ocean.task.internal.FiberPool;
+import ocean.task.Task;
+
+debug (TaskScheduler)
+    import ocean.io.Stdout;
+
+/// Ditto
+class FiberPoolEager : FiberPool
```
daniel-zullo, line 31: missing protection attribute

src/ocean/task/Scheduler.d
```diff
@@ -299,12 +318,15 @@ final class Scheduler
 
     public void shutdown ( )
     {
-        debug_trace("Shutting down initiated. {} queued tasks will be " ~
-            " discardead, {} suspended tasks will be killed",
-            this.queued_tasks.length(), this.suspended_tasks.length());
+        debug_trace(
+            "Shutting down initiated. {} queued tasks will be " ~
+                " discardead, {} suspended tasks will be killed",
```
daniel-zullo, line 323: typo *discardead

src/ocean/task/Task.d
```diff
@@ -299,8 +310,8 @@ public abstract class Task : ISuspendable
     public void suspend ( )
     {
         assert (this.fiber !is null);
```
daniel-zullo, line 312: shouldn't it be replaced to `verify()` too?

src/ocean/task/internal/FiberPoolEager.d
```diff
@@ -0,0 +1,83 @@
...
+        3-Clause BSD License (see LICENSE_BSD.txt for details).
+
+*******************************************************************************/
+
+module ocean.task.internal.FiberPoolEager;
+
+import core.thread : Fiber;
+
+import ocean.meta.types.Qualifiers;
+import ocean.core.Enforce;
```
daniel-zullo, line 23: this import should go on the top of ocean imports

src/ocean/task/internal/FiberPoolEager.d
```diff
@@ -0,0 +1,83 @@
...
+
+/// Ditto
+class FiberPoolEager : FiberPool
+{
+    /**************************************************************************
+
+        Constructor
+
+        Params:
+            stack_size = fiber stack size to use in this poll
```
daniel-zullo, line 38: s/poll/pool ?

src/ocean/task/internal/FiberPoolWithQueue.d
```diff
@@ -0,0 +1,249 @@
...
+    public TaskQueueFullCB task_queue_full_cb;
+
+    /**************************************************************************
+
+        Constructor
+
+        Params:
+            queue_limit = max size of task queue, recommened to be at least as
+                large as pool limit
+            stack_size = fiber stack size to use in this poll
```
daniel-zullo, line 61: ditto *poll

src/ocean/task/internal/SpecializedPools.d
```diff
@@ -0,0 +1,178 @@
...
+*******************************************************************************/
+
+module ocean.task.internal.SpecializedPools;
+
+import ocean.meta.types.Qualifiers;
+import ocean.task.internal.FiberPoolEager;
+import ocean.task.Task;
+import ocean.core.Verify;
+import ocean.core.Enforce;
+import ocean.core.Optional;
```
daniel-zullo, line 22: Should the imports be sorted?

daniel-zullo (commented):
A bit late for my minor comments but I'd submit them anyway

src/ocean/task/Task.d
mihails-strasuns-sociomantic: No, `task` package is currently the only one which uses both as intended (depending on what kind of condition is protected against), contrary to rest of ocean which was converted 1-to-1.

I will likely have a full re-read of all scheduler source code soon and will also re-evaluate all assert/verify usage cases.

src/ocean/task/internal/SpecializedPools.d
mihails-strasuns-sociomantic: No, there is no particular import order requirement for our libs/apps.

mihails-strasuns-sociomantic:
@daniel-zullo-sociomantic thanks, I will submit separate style cleanup PR

src/ocean/task/Task.d
daniel-zullo: ok, good to know that

src/ocean/task/internal/SpecializedPools.d
daniel-zullo: fair enough

PR #294, opened 2017-10-09
Exit the client process after write/read, don't run the next tests in it

nemanja-boric-sociomantic:
Intended return was to stop the client process after it would send/read
data from the socket. However, this would just exit the frame,
continuing running the test next test, as the main would not exit
immediately.


mihails-strasuns-sociomantic:
Needs rebase

nemanja-boric-sociomantic:
Whoops. Done.

mihails-strasuns-sociomantic (approved):

PR #299, opened 2017-10-10
Use custom exception type for out of bound in Stack

### LINKED ISSUE #297: Fix JsonParser throwing inconsistent exceptions or errors
Depending on the D version used, JsonParser will throw a different exception or error. In order to make it consistent it should throw always a `JsonParserException` so a program using the module can catch the exception whatever is the D version currently in use.
 
```D
unittest
{
    static immutable istring json =
    `{
        "products": {
        }}
    }`;

    auto p = new JsonParser!(char)(json);
    assert(p);
    assert(p.type == p.Token.BeginObject);
    assert(p.next);
    assert(p.type == p.Token.Name);
    assert(p.value == "products", p.value);
    assert(p.next);
    assert(p.value == "", p.value);
    assert(p.next);
    assert(p.type == p.Token.EndObject);
    assert(p.next);
    assert(p.type == p.Token.EndObject);

    try
    {
        assert(p.next);
    }
    catch (Exception e)
    {
        // catch in D1
    }
    catch
    {
        // catch in D2
    }
}
```

mihails-strasuns-sociomantic:
Fixes #297

### DIFF
--- a/src/ocean/util/container/more/Stack.d
+++ b/src/ocean/util/container/more/Stack.d
@@ -17,7 +17,9 @@
 
 module ocean.util.container.more.Stack;
 
-import ocean.core.ExceptionDefinitions : ArrayBoundsException;
+import ocean.core.Enforce;
+version (UnitTest)
+    import ocean.core.Test;
 
 /******************************************************************************
 
@@ -133,7 +135,7 @@ struct Stack (V, int Size = 0)
                           if (depth < stack.length)
                               stack[depth++] = value;
                           else
-                             error (__LINE__);
+                              enforce(.e_bounds, false);
                           }
                 return this;
         }
@@ -166,7 +168,8 @@ struct Stack (V, int Size = 0)
                 if (depth)
                     return stack[--depth];
 
-                return error (__LINE__);
+                enforce(.e_bounds, false);
+                assert(false);
         }
 
         /**********************************************************************
@@ -182,7 +185,8 @@ struct Stack (V, int Size = 0)
                 if (depth)
                     return stack[depth-1];
 
-                return error (__LINE__);
+                enforce(.e_bounds, false);
+                assert(false);
         }
 
         /**********************************************************************
@@ -203,7 +207,8 @@ struct Stack (V, int Size = 0)
                    return p[1] = v;
                    }
 
-                return error (__LINE__);
+                enforce(.e_bounds, false);
+                assert(false);
         }
 
         /**********************************************************************
@@ -220,7 +225,8 @@ struct Stack (V, int Size = 0)
                 if (i < depth)
                     return stack [depth-i-1];
 
-                return error (__LINE__);
+                enforce(.e_bounds, false);
+                assert(false);
         }
 
         /**********************************************************************
@@ -242,7 +248,7 @@ struct Stack (V, int Size = 0)
                    *p = t;
                    }
                 else
-                   error (__LINE__);
+                   enforce(.e_bounds, false);
                 return this;
         }
 
@@ -265,7 +271,7 @@ struct Stack (V, int Size = 0)
                    *p = t;
                    }
                 else
-                   error (__LINE__);
+                   enforce(.e_bounds, false);
                 return this;
         }
 
@@ -284,17 +290,6 @@ struct Stack (V, int Size = 0)
                 return stack [0 .. depth];
         }
 
-        /**********************************************************************
-
-                Throw an exception
-
-        **********************************************************************/
-
-        private V error (size_t line)
-        {
-                throw new ArrayBoundsException (__FILE__, line);
-        }
-
         /***********************************************************************
 
                 Iterate from the most recent to the oldest stack entries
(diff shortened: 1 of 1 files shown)


nemanja-boric-sociomantic:
I'm scared to say anything about breaking the code cause the different exception is now thrown on this one :no_mouth:

src/ocean/util/container/more/Stack.d
```diff
@@ -133,7 +133,7 @@ struct Stack (V, int Size = 0)
                           if (depth < stack.length)
```
nemanja-boric-sociomantic, line 135: These all `if (...)` can be converted to enforce for a nicer code.

mihails-strasuns-sociomantic:
Oh, you mean our D1-only apps doing `catch (RangeException)` with previous code?

src/ocean/util/container/more/Stack.d
mihails-strasuns-sociomantic: Full rewrite of module code style is my next step :)

nemanja-boric-sociomantic:
Yeah.

mihails-strasuns-sociomantic:
This is actually both totally legit and damn scary concern :/

nemanja-boric-sociomantic:
I wonder if we can keep throwing `ArrayBoundsException` (`StackException: ArrayBoundsException`) in D1 and in D2 do `StackException: Exception`. Given that the users couldn't (yeah, right) catch `RangeError` in the past, we could get away with this approach.

mihails-strasuns-sociomantic:
Sounds on the edge of acceptable :)

nemanja-boric-sociomantic:
What's happening with the codecov? Is it not attached, is it #298 a cause for it? I doubt it, but IIRC we had some issues with it in #230, that's why it got too complicated and forgotten in the end.

nemanja-boric-sociomantic:
It looks like it's just not reported to the PR: https://codecov.io/github/sociomantic-tsunami/ocean/commit/77b6c23bc60e9b3c23dafe5c68a48989e8bb2c8e

mihails-strasuns-sociomantic:
Not entirely sure. If I go their website and open ocean PR list in the reports, I can find it there (and log shows it gets submitted without problems). But it doesn't set any status.

mihails-strasuns-sociomantic:
Updated with coding style and tests

mihails-strasuns-sociomantic:
Updated: fix another undefined behaviour bug found by added tests :)

src/ocean/util/container/more/Stack.d
```diff
@@ -311,6 +306,48 @@ struct Stack (V, int Size = 0)
...
+    this ( )
+    {
+        version (D_Version2)
+            super("Out of bounds access attempt to stack struct");
+        else
+            super("", 0);
+    }
+}
+
+// HACK: some D1 code may be trying to catch ArrayBoundsException specifically
```
gavin-norman-sociomantic, line 413: Would be worth mentioning what the plan is for removing this hack.

src/ocean/util/container/more/Stack.d
```diff
@@ -311,6 +306,48 @@ struct Stack (V, int Size = 0)
...
+{
+    Stack!(int) stack;
+    stack.push(42);
+    test!("==")(stack.pop(), 42);
+    testThrown!(StackBoundsException)(stack.pop());
+}
+
+/*******************************************************************************
+
+    Exception used instead of uncatchable `RangeError`.
```
gavin-norman-sociomantic: This comment only seems relevant from the perspective of this moment in time (i.e. this commit). It doesn't seem relevant doc for this module, in general.

src/ocean/util/container/more/Stack.d
```diff
@@ -284,17 +290,6 @@ struct Stack (V, int Size = 0)
                 return stack [0 .. depth];
         }
 
-        /**********************************************************************
-
-                Throw an exception
-
-        **********************************************************************/
-
-        private V error (size_t line)
```
gavin-norman-sociomantic, line 293: What? Good riddance.

src/ocean/util/container/more/Stack.d
```diff
@@ -203,7 +207,8 @@ struct Stack (V, int Size = 0)
                    return p[1] = v;
                    }
 
-                return error (__LINE__);
```
gavin-norman-sociomantic, line 206: `return`? This code was super weird!

src/ocean/util/container/more/Stack.d
```diff
@@ -1,309 +1,316 @@
 /*******************************************************************************
```
gavin-norman-sociomantic, line 1: >Fix Stack coding style and cover it with tests

In a patch release?

src/ocean/util/container/more/Stack.d
```diff
@@ -351,43 +429,43 @@ static this ( )
 
 /*******************************************************************************
 
-*******************************************************************************/
+ *******************************************************************************/
 
 debug (Stack)
```
mathias-lang-sociomantic: Remove ?

mihails-strasuns-sociomantic: Already removed in 3.5.x, don't want conflicts

src/ocean/util/container/more/Stack.d
mihails-strasuns-sociomantic: This module had exactly 0 tests so I had to do something to actually check that things work. Now that they do, we could possibly move all the part adding tests & style to minor but it would be a bit scary to merge fix without tests coming in a pair.

gavin-norman-sociomantic: The massive coding style changes are more what I was referring to.

src/ocean/util/container/more/Stack.d
mihails-strasuns-sociomantic: On this topic - any idea how we could possibly handle it? I can only think of deprecating `ArrayBoundsException` alias in `ocean.core` to ensure no one refers to it directly.

src/ocean/util/container/more/Stack.d
mathias-lang-sociomantic: But you changed the indent, so you'll get them anyway

src/ocean/util/container/more/Stack.d
nemanja-boric-sociomantic: I think that's fair enough.

mihails-strasuns-sociomantic:
Modified PR to only contains fixes and added tests. Will make another one against v3.x.x after point release with the rest.

mihails-strasuns-sociomantic:
Pushed one more UB fix. This module is about to overload my crap detector.

mihails-strasuns-sociomantic:
Green

nemanja-boric-sociomantic (approved):

PR #301, opened 2017-10-10
Temporarily disable `final` in only `final switch`

mihails-strasuns-sociomantic:
As development infrastructure was not yet ready for switching to 1.081,
the usage of this feature was premature.

### DIFF
--- a/src/ocean/core/SmartUnion.d
+++ b/src/ocean/core/SmartUnion.d
@@ -312,6 +312,8 @@ private const handleInvalidCases = "case none: assert(false);" ~
     // has no default so we add a default here just to avoid the warning.
     "version (D_Version2){} else {default: assert(false);}";
 
+// Disabled until DMD 1.081 becomes widely available
+/*
 ///
 unittest
 {
@@ -334,6 +336,7 @@ unittest
         mixin(TestSmartUnion.handleInvalidCases);
     }
 }
+*/
 
 /******************************************************************************
 


src/ocean/core/SmartUnion.d
```diff
@@ -325,7 +325,7 @@ unittest
     TestSmartUnion u;
     u.a = 1;
 
-    with (TestSmartUnion.Active) final switch (u.active)
+    with (TestSmartUnion.Active) /* final */ switch (u.active)
```
gavin-norman-sociomantic: Could you create an issue describing when we should reenable this?

gavin-norman-sociomantic (approved):

src/ocean/core/SmartUnion.d
mihails-strasuns-sociomantic: Done

mathias-lang-sociomantic: I would comment the whole unittest, since the point is to test `handleInvalidCases`, isn't it ?

nemanja-boric-sociomantic: Yeah, I would just comment entire unittest.

mihails-strasuns-sociomantic:
Updated to comment out whole unit test

nemanja-boric-sociomantic (approved):

mihails-strasuns-sociomantic:
Green

PR #303, opened 2017-10-10
HttpException: Use catch (Exception) instead of pokemon catch

Geod24:
Pokemon catch is deprecated.

### DIFF
--- a/src/ocean/net/http/HttpException.d
+++ b/src/ocean/net/http/HttpException.d
@@ -128,7 +128,7 @@ class HttpException : HttpServerException
             auto path = "/path/with/errors";
             e.enforce(false, HttpResponseCode.NotFound, "Unable to locate URI path:", path);
         }
-        catch
+        catch (Exception)
         {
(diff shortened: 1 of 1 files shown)


gavin-norman-sociomantic (approved):
This is on a minor release branch as it doesn't affect currently supported compiler versions, right?

mihails-strasuns-sociomantic:
It affects any D2 compiler by not trying to catch `Error` anymore. I still think it is acceptable.

mathias-lang-sociomantic:
We don't intend (and and couldn't without some effort) support upstream compiler in v3.x.x. I think that should be a reasonable goal for v4.x.x though.

mihails-strasuns-sociomantic:
It is actually inside unittest, so doesn't matter at all :)

mathias-lang-sociomantic:
Oh I think this part was obvious :P

mihails-strasuns-sociomantic:
You have overestimated me :)

PR #306, opened 2017-10-11
Annotate callback passed to C function to match upstream

mathias-lang-sociomantic:
Since delegate parameter contravariance is not implemented, this error out when compiled with upstream compiler,
which annotate the parameter of the callback with scope.
With our version of dmd(-transitional), it doesn't make a difference though.
Moreover, thanks to the prototype change, some implicit convertion is not done anymore.

Part of #9

### DIFF
--- a/src/ocean/io/select/selector/TimeoutSelectedKeysHandler.d
+++ b/src/ocean/io/select/selector/TimeoutSelectedKeysHandler.d
@@ -28,6 +28,7 @@ import ocean.io.select.client.model.ISelectClient;
 
 import ocean.time.timeout.model.ITimeoutManager,
                ocean.time.timeout.model.ITimeoutClient;
+import ocean.transition;
 
 import ocean.util.container.AppendBuffer: AppendBuffer;
 
@@ -143,7 +144,8 @@ class TimeoutSelectedKeysHandler: SelectedKeysHandler
 
                     verify(client !is null);
 
-                    if (!bsearch(cast (void*) client, timed_out_clients.ptr,
+                    Const!(void)* typed_ptr = timed_out_clients.ptr;
+                    if (!bsearch(cast (void*) client, typed_ptr,
                                  timed_out_clients.length, timed_out_clients[0].sizeof, &cmpPtr!(true)))
                     {
                         this.handleSelectedKey(key, unhandled_exception_hook);
@@ -195,8 +197,9 @@ class TimeoutSelectedKeysHandler: SelectedKeysHandler
 
     ***************************************************************************/
 
-    extern (C) private static int cmpPtr ( bool searching ) ( in void* a_,
-        in void* b_ )
+    extern (C) private static int cmpPtr ( bool searching ) (
+        /* d1to2fix_inject: scope */ Const!(void*) a_,
+        /* d1to2fix_inject: scope */ Const!(void*) b_ )
     {
         static if (searching)
         {


mihails-strasuns-sociomantic:
Is there any chance we can fix `scope` delegates to not be checked in `@system` code instead, same as rest of `scope`? I am so not happy about dealing with any of dip1000 consequences >_<

mathias-lang-sociomantic:
> Is there any chance we can fix scope delegates to not be checked in @system code instead, same as rest of scope? I am so not happy about dealing with any of dip1000 consequences >_<

Yeah me neither, but the issue is not `scope` related. Since delegate parameter contravariance is not implemented, there's not much we can do. And implementing it is quite a project.

IOW, since this code doesn't compile:
```D
void foo (void function(mstring) dg);
void bar (cstring);
foo(&bar); // Error !?!
````

This code cannot possibly compile without hacks:
```D
void foo (void function(scope void*) dg);
void bar (void*);
foo(&bar); // Ditto error
```

mihails-strasuns-sociomantic:
Ah, right, so the problem boils down to extra `scope` added in `druntime` to C functions args?

mihails-strasuns-sociomantic:
I wonder if redefining the binding as private symbol in this module would be less of a mess in that case :(

mathias-lang-sociomantic:
> Ah, right, so the problem boils down to extra `scope` added in druntime to C functions args?

Yes

> I wonder if redefining the binding as private symbol in this module would be less of a mess in that case :(

Hum no ? We'll eventually have to comply with some of those things, so it just sounds like extra work. This change is not *wrong* per se. Things would be smoother if the type system was better, but I don't see a problem in adding this annotation *for this case*.

src/ocean/io/select/selector/TimeoutSelectedKeysHandler.d
```diff
@@ -195,8 +197,9 @@ class TimeoutSelectedKeysHandler: SelectedKeysHandler
 
     ***************************************************************************/
 
-    extern (C) private static int cmpPtr ( bool searching ) ( in void* a_,
-        in void* b_ )
+    extern (C) private static int cmpPtr ( bool searching ) (
+        /* d1to2fix_inject: scope */ Const!(void*) a_,
```
david-eckardt-sociomantic, line 201: `in void* a_`?

mihails-strasuns-sociomantic: Can't mix `in` and `scope` because former was supposed to include the latter (but not anymore). Hm, speaking about that, @mathias-lang-sociomantic do you think this restriction can/should be removed now?

mathias-lang-sociomantic: Which restriction ? That `in` doesn't include `scope` ? I would love that, but don't think Walter will agree. Also, [they don't play along very well](https://issues.dlang.org/show_bug.cgi?id=17408) :(

mihails-strasuns-sociomantic: Other way around, prohibiting `in scope` combo.

mathias-lang-sociomantic: Should the bug be fixed ? Definitely. Just didn't have  the time (nor considered it important enough, given how DIP1000 is broken) to get around it.

mihails-strasuns-sociomantic: I wasn't even sure it is viewed as bug now. Was definitely intentional when `in` was still supposed to mean `const scope`.

src/ocean/io/select/selector/TimeoutSelectedKeysHandler.d
```diff
@@ -143,7 +144,8 @@ class TimeoutSelectedKeysHandler: SelectedKeysHandler
 
                     verify(client !is null);
 
-                    if (!bsearch(cast (void*) client, timed_out_clients.ptr,
```
mihails-strasuns-sociomantic, line 146: Why is this change necessary btw?

mathias-lang-sociomantic: Because the second parameter is `inout(void*)` but we pass a typed parameter (`Timeout*` IIRC). Apparently DMD cannot figure out the conversion.

mihails-strasuns-sociomantic (approved):

PR #312, opened 2017-10-18
Submit code coverage for all builds

mihails-strasuns-sociomantic:


leandro-lucarella-sociomantic (approved):

mihails-strasuns-sociomantic:
Right I remember now that there was issue with D2 builds because -cov triggers stomping prevention assert

leandro-lucarella-sociomantic:
How do you see this issue in the long term? Is there a way to make `-cov` not stomp? Why is the adding of `-cov` making any changes in stomping?

mihails-strasuns-sociomantic:
> Is there a way to make -cov not stomp?

I am working on dmd-transitional patch right now

> Why is the adding of -cov making any changes in stomping?

Because it is `-cov` implementation code inside druntime which triggers the assertion (and it is not being used at all unless you use `-cov`)

nemanja-boric-sociomantic:
https://github.com/sociomantic-tsunami/ocean/pull/230#issuecomment-326316241
https://github.com/sociomantic-tsunami/ocean/pull/230#issuecomment-326255956
https://github.com/sociomantic-tsunami/ocean/pull/230#issuecomment-326272862

:grumpy:

mihails-strasuns-sociomantic:
Yeah, so it is supposed to be in s3

mihails-strasuns-sociomantic:
Picked required changes from old @nemanja-boric-sociomantic PR :D

mihails-strasuns-sociomantic:
Green

src/ocean/core/array/Search.d
```diff
@@ -958,13 +958,13 @@ unittest
 
    Performs a linear scan of haystack from [0 .. haystack.length$(RP), returning
    a count of the number of elements where pred returns true.
-  
+
```
nemanja-boric-sociomantic, line 961: Not really important, but this should be in a separate commit (or not committed at all).

src/ocean/core/array/Search.d
```diff
@@ -1278,7 +1278,9 @@ unittest
 
 bool startsWith ( T ) ( in T[] arr, in T[] prefix )
 {
-    return (arr.length >= prefix.length) && (arr[0..prefix.length] == prefix[]);
+    if (arr.length < prefix.length)
+        return false;
+    return arr[0..prefix.length] == prefix[];
```
leandro-lucarella-sociomantic: I would like to see a better explanation for this change in the commit message, at simple sight it looks like rearranging the code. So what's the real problem and what's the intended change should be stated in the commit message.

mihails-strasuns-sociomantic: It is simply re-arranging code indeed. Problem is major code generation bug in compiler (can't find bugzilla number right now). Looks like it doesn't do early return from `&&` expression here when both `-O` and `-cov` are supplied.

mihails-strasuns-sociomantic:
Updated:
- separate whitespace commit
- more info about DMD bug

src/ocean/core/array/Search.d
leandro-lucarella-sociomantic: I strongly suggest adding a comment here too, otherwise someone could try to rearrange the code again in the future. Ideally a comment with a link to the upstream bug.

mihails-strasuns-sociomantic: Done

leandro-lucarella-sociomantic (approved):

Issue #190, opened 2017-07-13
Allow reloading specific files through UnixSocket


mathias-lang-sociomantic:
For files which are registered with `ReopenableFilesExt`, we need to be able to reopen a specific file via an `UnixSocket` command.
This should speed up the process for applications with many opened files.

For example it would be used as `echo "App reopen /srv/httpd/log/root.log /srv/httpd/log/important.log" | tee /srv/httpd/httpd.socket`.
@andrew-stevenson-sociomantic would that cover your needs ?

PR #314, opened 2017-10-19
 Support reopening single files via unix domain socket in DaemonApp

nemanja-boric-sociomantic:
`DaemonApp`'s `ReopenableFilesExt` now supports reopening commands from the
UnixSocketExt. If you have the UnixSocketExt bound to the socket, sending command
`reopen_logfile file_path` will reopen this file, if registered, and acknowledge
the action with `ACK` or `ERROR` strings.

This is different than reopening files with the reopen signal, which
is not capable passing the file path to it, so it forces reopening all
files.

### DIFF
--- a/relnotes/reopensocket.feature.md
+++ b/relnotes/reopensocket.feature.md
@@ -1,0 +1,40 @@
+* `ocean.util.app.DaemonApp`
+
+  `DaemonApp`'s `ReopenableFilesExt` now supports reopening commands from the
+  UnixSocketExt. If you have the UnixSocketExt bound to the socket, sending
+  command `reopen_logfile file_path` will reopen the specified file, if
+  registered, and acknowledge the action with `ACK` or `ERROR` strings (followed
+  by the file name which failed to reopen in case of `ERROR`).
+
+  With this in place, reopening only affected files (vs all open registered files)
+  with logrotate can be done as follows:
+
+    1) use `nosharedscripts` directive, so the post-rotate script is called
+       once per file, with the absolute file path passed to it.
+    2) From the absolute file path, get the log directory and deduce socket
+       location from it (application could place the socket in its `run/`
+       directory, so the socket path would be
+       `SOCKETPATH=$(dirname $LOGPATH)/../application.socket`.
+    3) Finally, in post-rotate script, simply executing
+       `test "$(echo reopen_files $LOGPATH | nc -U $SOCKETPATH)" = ACK` will
+       instruct the application to reopen the given logfile and report status
+       to logrotate.
+
+  Example logrotate file:
+
+  ```
+  /srv/dlsnode/log/*.log
+  {
+      rotate 10
+      missingok
+      notifempty
+      delaycompress
+      compress
+      size 1M
+      nosharedscripts
+      postrotate
+        export SOCKETPATH="$(dirname $1)/../dlsnode.socket"
+        test "$(echo reopen_files $1 | nc -U $SOCKETPATH)" = ACK
+      endscript
+    }
+    ```
--- a/src/ocean/io/device/File.d
+++ b/src/ocean/io/device/File.d
@@ -412,6 +412,18 @@ class File : Device, Device.Seek, Device.Truncate
                 return idup(path_);
         }
 
+        /**********************************************************************
+
+            Returns:
+                path used by this file.
+
+        **********************************************************************/
+
+        public cstring path ()
+        {
+            return this.path_;
+        }
+
         /***********************************************************************
 
                 Convenience function to return the content of a file.
--- a/src/ocean/util/app/DaemonApp.d
+++ b/src/ocean/util/app/DaemonApp.d
@@ -290,6 +290,16 @@ public abstract class DaemonApp : Application,
 
         /***********************************************************************
 
+            Unix domain socket command to trigger reopening of files which are
+            registered with the ReopenableFilesExt. (Typically used for log
+            rotation).
+
+        ***********************************************************************/
+
+        istring reopen_command = "reopen_files";
+
+        /***********************************************************************
+
             Set of signals to ignore. Delivery of the signals specified in this
             set will have no effect on the application -- they are not passed
             to the default signal handler.
@@ -367,11 +377,6 @@ public abstract class DaemonApp : Application,
         this.signal_ext.registerExtension(this);
         this.registerExtension(this.signal_ext);
 
-        // Create and register repoenable files extension
-        this.reopenable_files_ext = new ReopenableFilesExt(this.signal_ext,
-            settings.reopen_signal);
-        this.registerExtension(this.reopenable_files_ext);
-
         this.pidlock_ext = new PidLockExt();
         this.config_ext.registerExtension(this.pidlock_ext);
         this.registerExtension(this.pidlock_ext);
@@ -379,6 +384,23 @@ public abstract class DaemonApp : Application,
         this.unix_socket_ext = new UnixSocketExt();
         this.config_ext.registerExtension(this.unix_socket_ext);
         this.registerExtension(this.unix_socket_ext);
+
+        // Create and register repoenable files extension
+        this.reopenable_files_ext = new ReopenableFilesExt();
+
+        if (settings.reopen_signal)
+        {
+            this.reopenable_files_ext.setupSignalHandler(this.signal_ext,
+                    settings.reopen_signal);
+        }
+
+        if (settings.reopen_command.length)
+        {
+            this.reopenable_files_ext.setupUnixSocketHandler(
+                    this.unix_socket_ext, settings.reopen_command);
+        }
+
+        this.registerExtension(this.reopenable_files_ext);
 
         this.system_stats = new CpuMemoryStats();
     }
--- a/src/ocean/util/app/ext/ReopenableFilesExt.d
+++ b/src/ocean/util/app/ext/ReopenableFilesExt.d
@@ -23,17 +23,24 @@ module ocean.util.app.ext.ReopenableFilesExt;
 
 import ocean.transition;
 
+import ocean.core.Buffer;
+import ocean.core.Verify;
+
+import ocean.sys.Environment;
+
 import ocean.util.app.model.IApplication;
 import ocean.util.app.model.IApplicationExtension;
 
 import ocean.util.app.ext.SignalExt;
+import ocean.util.app.ext.UnixSocketExt;
 
 import ocean.util.app.ext.model.ISignalExtExtension;
+
+import ocean.text.convert.Formatter;
 
 import core.sys.posix.signal : SIGHUP;
 
 import ocean.io.device.File;
-
 
 
 public class ReopenableFilesExt : IApplicationExtension, ISignalExtExtension
@@ -56,6 +63,23 @@ public class ReopenableFilesExt : IApplicationExtension, ISignalExtExtension
 
     private int reopen_signal;
 
+    /***************************************************************************
+
+        Current working directory. Used for building absolute paths for the
+        registered files.
+
+    ***************************************************************************/
+
+    private istring cwd;
+
+    /***************************************************************************
+
+        Buffer for rendering the absolute paths.
+
+    ***************************************************************************/
+
+    private Buffer!(char) path_buffer;
+
 
     /***************************************************************************
 
@@ -72,12 +96,62 @@ public class ReopenableFilesExt : IApplicationExtension, ISignalExtExtension
 
     public this ( SignalExt signal_ext = null, int reopen_signal = SIGHUP )
     {
-        if ( signal_ext )
+        if (signal_ext)
         {
-            this.reopen_signal = reopen_signal;
-            signal_ext.register(this.reopen_signal);
-            signal_ext.registerExtension(this);
+            this.setupSignalHandler(signal_ext, reopen_signal);
         }
+    }
+
+
+    /***************************************************************************
+
+        Registers this extension with the signal extension and activates the
+        handling of the specified signal, which will cause the registered files
+        to be reopened.
+
+        Params:
+            signal_ext = SignalExt instance
+            reopen_signal = signal to trigger reopening of registered files
+
+    ***************************************************************************/
+
+    public void setupSignalHandler ( SignalExt signal_ext,
+            int reopen_signal = SIGHUP )
+    {
+        verify(signal_ext !is null);
+        verify(this.reopen_signal == this.reopen_signal.init,
+            "Either pass SignalExt to constructor or to setupSignalHandler, " ~
+            "not to both.");
+
+        this.reopen_signal = reopen_signal;
+        signal_ext.register(this.reopen_signal);
+        signal_ext.registerExtension(this);
+    }
+
+
+    /***************************************************************************
+
+        Registers this extension with the unix socket extension and activates the
+        handling of the specified unix socket command, which will cause reopening
+        files specified as arguments to the command.
+
+        Params:
+            unix_socket_ext = UnixSocketExt instance to register with
+            reopen_command = command to trigger reopening of the registered
+                file (passed as arguments to the command).
+
+    ***************************************************************************/
+
+    public void setupUnixSocketHandler ( UnixSocketExt unix_socket_ext,
+            istring reopen_command = "reopen_files" )
+    {
+        verify(unix_socket_ext !is null);
+
+        this.cwd = Environment.directory();
+        verify (this.cwd[$-1] == '/');
+
+        unix_socket_ext.addHandler(reopen_command,
+            &this.socketReloadCommand);
     }
 
 
@@ -114,6 +188,42 @@ public class ReopenableFilesExt : IApplicationExtension, ISignalExtExtension
 
     /***************************************************************************
 
+        Reopens all registered files with the given path
+
+        Params:
+            file_path = path of the file to open
+
+        Returns:
+            true if the file was registered and reopen, false otherwise.
+
+    ***************************************************************************/
+
+    public bool reopenFile ( cstring file_path )
+    {
+        // It might be the case that we have
+        // several files registered with the same path.
+        // We need to reopen all of them
+        bool reopened;
+
+        foreach (file; this.open_files)
+        {
+            path_buffer.reset();
+            sformat(path_buffer, "{}{}", this.cwd, file.path());
+
+            // Check both relative and absolute paths
+            if (file.path() == file_path || path_buffer[] == file_path)
+            {
+                file.close();
+                file.open(file.toString(), file.style);
+                reopened = true;
+            }
+        }
+
+        return reopened;
+    }
+
+    /***************************************************************************
+
         Signal handler. Called by SignalExt when a signal occurs. Reopens all
         log files.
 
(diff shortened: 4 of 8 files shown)


nemanja-boric-sociomantic:
```
echo reopen_logfile /srv/dlsnode/log/root.log | sudo nc -U /srv/dlsnode/dlsnode.socket
ACK
```

relnotes/reopensocket.feature.md
```diff
@@ -0,0 +1,6 @@
+* `ocean.util.app.DaemonApp`
+
+  `DaemonApp`'s `ReopenableFilesExt` now supports reopening commands from the
+  UnixSocketExt. If you have the UnixSocketExt bound to the socket, sending command
+  `reopen_logfile file_path` will reopen this file, if registered, and acknowledge
```
gavin-norman-sociomantic: "will reopen the specified file"

nemanja-boric-sociomantic:
With this in place, reopening only affected files with the logrotate can be done as follows:

1. use `nosharedscripts` directive, so the post-rotate script is called once per file, with the absolute file path passed to it.
2. From the absolute file path, get the log directory and deduce socket placement from it (we could have convention that the socket is in the application's `run/` directory, so the socket path would be `
SOCKETPATH=$(dirname $LOGPATH)/../run/application.socket`
3. Simply execute:

```
test $(echo reopen_logfile $LOGPATH | sudo nc -U $SOCKETPATH) = ACK
```

and that should rotate your single logfile.

gavin-norman-sociomantic:
https://github.com/sociomantic-tsunami/ocean/pull/314#issuecomment-337876590 would be super useful in the release notes. We should try to make this feature as easy as possible for people to start using.

nemanja-boric-sociomantic:
Updated with the extended release notes.

src/ocean/util/app/ext/ReopenableFilesExt.d
```diff
@@ -70,14 +93,27 @@ public class ReopenableFilesExt : IApplicationExtension, ISignalExtExtension
 
     ***************************************************************************/
 
-    public this ( SignalExt signal_ext = null, int reopen_signal = SIGHUP )
+    public this ( SignalExt signal_ext = null,
```
gavin-norman-sociomantic: Oh wait... this is a breaking change, or?

gavin-norman-sociomantic (commented):
Submitting an early review because of the breaking change question.

src/ocean/util/app/ext/ReopenableFilesExt.d
nemanja-boric-sociomantic: Puh, yeah, I've missed that - I originally targeted v4.x.x. Just having new constructor with suffice, though. Will update.

nemanja-boric-sociomantic:
Updated with the dodged breaking change.

src/ocean/io/device/File.d
```diff
@@ -412,6 +412,18 @@ class File : Device, Device.Seek, Device.Truncate
...
+
+            Returns:
+                path used by this file.
+
+        **********************************************************************/
+
+        public cstring path ()
+        {
+            return this.path_;
+        }
```
leandro-lucarella-sociomantic, line 425: Why isn't this done in `toString()` too?

relnotes/reopensocket.feature.md
```diff
@@ -0,0 +1,20 @@
+* `ocean.util.app.DaemonApp`
+
+  `DaemonApp`'s `ReopenableFilesExt` now supports reopening commands from the
+  UnixSocketExt. If you have the UnixSocketExt bound to the socket, sending
+  command `reopen_logfile file_path` will reopen the specified file, if
+  registered, and acknowledge the action with `ACK` or `ERROR` strings.
```
leandro-lucarella-sociomantic: I wonder if, for debugging purposes, a message should be returned with the `ERROR` too (like `ERROR <free text>`).

relnotes/reopensocket.feature.md
```diff
@@ -0,0 +1,20 @@
...
+  with logrotate can be done as follows:
+
+    1) use `nosharedscripts` directive, so the post-rotate script is called
+       once per file, with the absolute file path passed to it.
+    2) From the absolute file path, get the log directory and deduce socket
+       location from it (application could place the socket in its `run/`
+       directory, so the socket path would be
+       `SOCKETPATH=$(dirname $LOGPATH)/../application.socket`.
+    3) Finally, in post-rotate script, simply executing
+       `test $(echo reopen_logfile $LOGPATH | nc -q 2 -U $SOCKETPATH) = ACK` will
```
leandro-lucarella-sociomantic: You probably should add quotes to `$(echo ...)`, otherwise it could be an empty string and make `test` give an error, right?

src/ocean/util/app/DaemonApp.d
```diff
@@ -380,6 +375,12 @@ public abstract class DaemonApp : Application,
         this.config_ext.registerExtension(this.unix_socket_ext);
         this.registerExtension(this.unix_socket_ext);
 
+        // Create and register repoenable files extension
+        this.reopenable_files_ext = new ReopenableFilesExt(this.signal_ext,
+            this.unix_socket_ext,
+            settings.reopen_signal);
+        this.registerExtension(this.reopenable_files_ext);
+
```
leandro-lucarella-sociomantic: Why is this now registered twice?

src/ocean/util/app/ext/ReopenableFilesExt.d
leandro-lucarella-sociomantic: Ah, I guess it's registered twice because of this... Maybe add an issue and a comment to clean this up in a following major?

src/ocean/util/app/ext/ReopenableFilesExt.d
```diff
@@ -70,14 +94,54 @@ public class ReopenableFilesExt : IApplicationExtension, ISignalExtExtension
...
+            reopen_command = if unix socket ext is non-null, command to trigger
+                reopenning of the registered file (passed as the first
+                argument to the command).
+
+    ***************************************************************************/
+
+    public this ( SignalExt signal_ext,
+        UnixSocketExt unix_socket_ext,
+        int reopen_signal = SIGHUP,
+        cstring reopen_command = "reopen_logfile")
```
leandro-lucarella-sociomantic: Do we really want to make this configurable? It seems strange to be able to introduce inconsistencies in the socket interface...

src/ocean/util/app/ext/ReopenableFilesExt.d
```diff
@@ -130,6 +224,27 @@ public class ReopenableFilesExt : IApplicationExtension, ISignalExtExtension
...
+            args = list of arguments received from the socket - should contain
+                the filename of the log to rotate.
+            send_response = delegate to send the response to the client
+
+    *****************************************************************************/
+
+    private void socketReloadCommand ( cstring[] args,
+            void delegate ( cstring response ) send_response )
+    {
+        // Todo: reopen the file here.
```
leandro-lucarella-sociomantic: I guess this `TODO` is outdated?

src/ocean/io/device/File.d
nemanja-boric-sociomantic: Because `toString()` must return `istring` (from `Object`).

src/ocean/util/app/DaemonApp.d
nemanja-boric-sociomantic: Probably accident - we need to register it only once. Will fix.

src/ocean/util/app/ext/ReopenableFilesExt.d
nemanja-boric-sociomantic: Well, we do the same thing for the SIGHUP. Note that people don't actually use this class directly, but it's used withing DaemonApp.

src/ocean/util/app/ext/ReopenableFilesExt.d
nemanja-boric-sociomantic: Oh, right.

relnotes/reopensocket.feature.md
nemanja-boric-sociomantic: Yeah.

nemanja-boric-sociomantic: Good catch!

relnotes/reopensocket.feature.md
nemanja-boric-sociomantic: I could return ERROR + the error text, no problem.

src/ocean/util/app/DaemonApp.d
nemanja-boric-sociomantic: Wait, where do you see it's registered twice?

nemanja-boric-sociomantic: I've just moved it to be registered after UnixSocketExt, since we need it for registering the handler.

nemanja-boric-sociomantic:
Updated.

src/ocean/util/app/DaemonApp.d
leandro-lucarella-sociomantic: I guess I was colorblind and didn't see the red above... All good.

leandro-lucarella-sociomantic (approved):
All the pending stuff is minor, so approving, feel free to merge and then work on any improvements.

mathias-lang-sociomantic:
Any chance we could get this tested by the right people ( :wave: @andrew-stevenson-sociomantic ) so we are sure it fulfills their use-case ? :)
Thanks a lot for taking care of this @nemanja-boric-sociomantic 
Fixes https://github.com/sociomantic-tsunami/ocean/issues/190

src/ocean/util/app/ext/ReopenableFilesExt.d
mathias-lang-sociomantic: Travis says this should be an `istring`
Also this works with *any* file opened with `ReusableFilesExt` right ? In this case `logfile` might be a bit too restrictive of a name.

src/ocean/util/app/ext/ReopenableFilesExt.d
```diff
@@ -130,6 +224,37 @@ public class ReopenableFilesExt : IApplicationExtension, ISignalExtExtension
...
+    private void socketReloadCommand ( cstring[] args,
+            void delegate ( cstring response ) send_response )
+    {
+        if (args.length == 0)
+        {
+            send_response("ERROR: missing name of the file to rotate.\n");
+            return;
+        }
+
+        auto filename = args[0];
```
mathias-lang-sociomantic: Could we use a `while` loop here, so one can pass multiple files ?

src/ocean/util/app/ext/ReopenableFilesExt.d
```diff
@@ -70,14 +94,54 @@ public class ReopenableFilesExt : IApplicationExtension, ISignalExtExtension
...
+
+    public this ( SignalExt signal_ext,
+        UnixSocketExt unix_socket_ext,
+        int reopen_signal = SIGHUP,
+        cstring reopen_command = "reopen_logfile")
     {
+        this.cwd = Environment.directory();
+
+        // Environment.directory makes sure for the trailing /, confirm it
+        verify (this.cwd[$-1] == '/');
```
mathias-lang-sociomantic: Extra space

mathias-lang-sociomantic: Also the comment sounds weird to me

relnotes/reopensocket.feature.md
```diff
@@ -0,0 +1,20 @@
...
+    1) use `nosharedscripts` directive, so the post-rotate script is called
+       once per file, with the absolute file path passed to it.
+    2) From the absolute file path, get the log directory and deduce socket
+       location from it (application could place the socket in its `run/`
+       directory, so the socket path would be
+       `SOCKETPATH=$(dirname $LOGPATH)/../application.socket`.
+    3) Finally, in post-rotate script, simply executing
+       `test "$(echo reopen_logfile $LOGPATH | nc -U $SOCKETPATH)" = ACK` will
+       instruct the application to reopen the given logfile and report status
+       to logrotate.
```
mathias-lang-sociomantic: Maybe give a very simple sample file ?

src/ocean/util/app/ext/ReopenableFilesExt.d
```diff
@@ -70,14 +94,54 @@ public class ReopenableFilesExt : IApplicationExtension, ISignalExtExtension
 
     ***************************************************************************/
 
-    public this ( SignalExt signal_ext = null, int reopen_signal = SIGHUP )
+    public this ( SignalExt signal_ext = null,
+        int reopen_signal = SIGHUP )
```
mathias-lang-sociomantic: Why the change indent ?

mathias-lang-sociomantic: Since this is internal-only, should it be deprecated ?

src/ocean/util/app/ext/ReopenableFilesExt.d
mathias-lang-sociomantic: Another comment about this: It looks weird to have signal-related thing, unixsocket, signal-related, unix-socket.

I understand why you went with this, and the alternative is not too appealing as well, so I'm going to propose an alternate solution: since both `SignalExt` and `UnixSocketExt` are optional and independent of each other, why don't we move the logic to `setup` methods, e.g. `setupUnixSocket(UnixSocketExt, istring = "reopen_file")` and `setupSignalExt(SignalExt, int = SIGHUP)` (naming up for discussion).

src/ocean/util/app/ext/ReopenableFilesExt.d
nemanja-boric-sociomantic: I didn't anticipated that one can do that, since logrotate doesn't do it, but it makes sense, I'll update it.

src/ocean/util/app/ext/ReopenableFilesExt.d
nemanja-boric-sociomantic: Interesting thing that this is not internal only - once could create their completely separate application hierarchy and use this. However, I will deprecate this, good catch :+1:

Indentation was changed via copy/pasting from other constructor. Will fix.

src/ocean/util/app/ext/ReopenableFilesExt.d
nemanja-boric-sociomantic: Yes, should be `Environment.directory should always return trailing /, confirm that that's the case`

src/ocean/util/app/ext/ReopenableFilesExt.d
nemanja-boric-sociomantic: Right, deprecating constructor and having setup methods replace it would work, good idea :+1:

leandro-lucarella-sociomantic:
BTW, adding some test would be nice. It is a bit annoying but it should be possible doing something like: create extension, register file X, write, move file externally to X and create a new file X, write again, check it's written to Y (and not to X), reopen files, write again and check it's written to X (and not to Y).

nemanja-boric-sociomantic:
Yeah, the test could/should be added.

src/ocean/util/app/ext/ReopenableFilesExt.d
```diff
@@ -114,6 +178,36 @@ public class ReopenableFilesExt : IApplicationExtension, ISignalExtExtension
...
+        Returns:
+            true if the file was registered and reopen, false otherwise.
+
+    ***************************************************************************/
+
+    public bool reopenFile ( cstring file_path )
+    {
+        foreach (file; this.open_files)
+        {
+            path_buffer.length = 0;
```
mihails-strasuns-sociomantic: I'd recommend `path_buffer.reset()`, it makes it more obvious you don't need `enableStomping` :)

mathias-lang-sociomantic: Yeah I started typing `enableStomping` and had the "oh..." moment :P

nemanja-boric-sociomantic: The sole purpose of Buffer :-P

src/ocean/util/app/ext/ReopenableFilesExt.d
nemanja-boric-sociomantic: Damn, I can't deprecate the existing constructor as all the methods are optional, so simply `auto ext = new ReopenableFilesExt` will trigger the deprecation :(.

nemanja-boric-sociomantic:
Updated with all points except the example logrotate file and test.

nemanja-boric-sociomantic:
Updated with the test.

nemanja-boric-sociomantic:
Updated with the example logrotate config file and the fact that there may be multiple files registered with the same filename, and they all needs to be reloaded.

nemanja-boric-sociomantic:
I think this includes all requested changes on this PR.

nemanja-boric-sociomantic:
Updated with the last bit: `ReopenableFilesExt` was getting the file's path via `File.toString` which would always allocate, given it needs to return `istring`. It's now tweaked to use newly added `File.path`.

mihails-strasuns-sociomantic (approved):

mathias-lang-sociomantic (approved):
LGTM, pending testing I assume ?

nemanja-boric-sociomantic:
Yes, I've tested it with the DLS node and it works. I wonder if @gautam-kotian-sociomantic could help us test this as well, if @andrew-stevenson-sociomantic doesn't have much time.

If nobody has time for a test, I think it's fair to merge this as is, given that I did a bit of testing, and fix issues in the patch releases, than to leave this PR rotting.

andrew-stevenson-sociomantic:
> if @andrew-stevenson-sociomantic doesn't have much time.

I suspect it will take a lot of time for me to test as I would need to update ocean, add command socket support to an app, switch from our logging stuff to ocean and then test :-(

nemanja-boric-sociomantic:
Just a note there: adding command socket support is done automatically with the recent oceans (just add `[UNIX_SOCKET]path=app.path` in the config file). But yeah, removing the custom stuff is more problematic.

leandro-lucarella-sociomantic:
If we have unit/integration tests and simple test seems to work, I would just merge.

nemanja-boric-sociomantic:
Yes, the intergration test has been added.

leandro-lucarella-sociomantic:
I don't get why the coverage is so crappy now. Lately I couldn't make sense of the coverage reports :angry: 

Also, needs rebase now.

nemanja-boric-sociomantic:
The biggest problem we have IMHO is that the coverage is reported file per file only. So, having something tested in the different file (like we're doing in the integration tests), doesn't help coverage at all :(. The second, separate problem, which is way more annoying is that these unexpected coverage changes. I wonder if that's related to the fact that we started uploading coverage for all matrix items (and actually building stuff with F=production), so codecov is going crazy because of that.

nemanja-boric-sociomantic:
Rebased.

leandro-lucarella-sociomantic:
> I wonder if that's related to the fact that we started uploading coverage for all matrix items (and actually building stuff with F=production), so codecov is going crazy because of that.

Mmm, maybe it is worth trying to disable codecov for F=production builds then. It's really annoying right now.

PR #316, opened 2017-10-19
Fix DaemonApp.onSignal crash on unkown signal

mihails-strasuns-sociomantic:
New code added in v3.5.0 was using `assert(false)` instead of a `break`.
Any application which was overriding `onSignal` was not affected.

### DIFF
--- a/src/ocean/util/app/DaemonApp.d
+++ b/src/ocean/util/app/DaemonApp.d
@@ -578,7 +578,7 @@ public abstract class DaemonApp : Application,
                     this.epoll.shutdown();
                 break;
             default:
-                assert(false);
+                break;
         }
     }
 


nemanja-boric-sociomantic (approved):

joseph-wakeling-sociomantic:
Typo in PR title and commit message: should be `unknown`, not `unkown`

src/ocean/util/app/DaemonApp.d
```diff
@@ -578,7 +578,7 @@ public abstract class DaemonApp : Application,
                     this.epoll.shutdown();
                 break;
             default:
-                assert(false);
```
gavin-norman-sociomantic, line 581: Oof.

gavin-norman-sociomantic: Given the SIGTERM logic, is the comment above still correct?

>See ISignalExtExtension documentation for more information on how to override this method with own behaviour. `super.onSignal` is not needed to be called when doing so.

nemanja-boric-sociomantic: Yes, it's still correct - if you're aren't `SIGTERM` in your app (like dhtnode) you still don't need to call this method. If you are not, this will still do what you probably want - terminate application (by shutting epoll) on SIGTERM. It's just the assert that was wrong, as the SignalExt calls all registered clients with all signals (no matter if you don't ask for that signal).

gavin-norman-sociomantic: There's at least one negative too many in your comment ;)

My point is that the derived class would have to reimplement the SIGTERM handling behaviour relating to the scheduler. Is that desirable?

gavin-norman-sociomantic: (If it doesn't call `super.onSignal`, that is.)

mihails-strasuns-sociomantic: > Is that desirable?

Yes. Assumption is that app knows better how to cleanly terminate itself - this is only default implementation that is expected to be not worse than crashing on unhandled signal.

gavin-norman-sociomantic: I see. Fair enough.

PR #281, opened 2017-10-20
Fix flakey `ocean.task` test (major `TimeSet` bug)

mihails-strasuns-sociomantic:
We have observed two times that some CI runs fail with totally unrelated test failures coming from task code. Previous case was ~3 weeks ago in one of PRs and it has happened again on v3.x.x HEAD after merging green PR today. It looks like one of task tests has some very low chance of failure that is not foreseen.

Will need further investigation.

### DIFF
--- a/relnotes/expiry.feature.md
+++ b/relnotes/expiry.feature.md
@@ -1,0 +1,6 @@
+* `ocean.time.timeout.model.ExpiryRegistrationBase`
+
+  New method, `drop`, provides same functionality as `unregister`, but also
+  removes the expiry from current list of fired events if it was present there.
+  This makes it possible to disable timer events from the timeout callbacks of
+  other timer events.
--- a/src/ocean/io/select/client/TimerSet.d
+++ b/src/ocean/io/select/client/TimerSet.d
@@ -202,6 +202,7 @@ public class TimerSet ( EventData ) : TimerEventTimeoutManager
         public void unregister ( )
         {
             this.expiry_registration.drop();
+            this.outer.events.recycle(this);
         }
 
 
--- a/src/ocean/io/select/client/TimerSet.d
+++ b/src/ocean/io/select/client/TimerSet.d
@@ -201,7 +201,7 @@ public class TimerSet ( EventData ) : TimerEventTimeoutManager
 
         public void unregister ( )
         {
-            this.expiry_registration.unregister();
+            this.expiry_registration.drop();
         }
 
 
--- a/src/ocean/time/timeout/TimeoutManager.d
+++ b/src/ocean/time/timeout/TimeoutManager.d
@@ -60,7 +60,6 @@ import ocean.transition;
 
 debug
 {
-    import ocean.io.Stdout;
     import core.stdc.time: time_t, ctime;
     import core.stdc.string: strlen;
 }
@@ -217,13 +216,51 @@ abstract class TimeoutManagerBase : ITimeoutManager
                          the IExpiryRegistration instance to unregister
 
             In:
-                Must not be called from within timeout().
+                Must not be called from within timeout(). Doing so would still
+                leave already fired timer events in the TimeoutManager internal
+                list and their respective `timeout` will be called despite
+                unregistration. be called from within timeout().
 
         ***********************************************************************/
 
         void unregister ( ref Expiry expiry )
         {
             this.outer.unregister(expiry);
+        }
+
+        /***********************************************************************
+
+            Unregisters the specified expiry. If the expiry is present in the
+            list of expired registrations being currently iterated over by
+            checkTimeouts, then it will be removed (its timeout method will not
+            be called). (This means that drop can be called from timeout
+            callbacks, unlike unregister.)
+
+            Params:
+                expiry = expiry token returned by register() when registering
+                         the IExpiryRegistration instance to unregister
+
+        ***********************************************************************/
+
+        protected void drop ( ref Expiry expiry )
+        {
+            // If this method is called while checkTimeouts is iterating over
+            // the list of expired registrations, then we need to check whether
+            // the expiry to be dropped is present in the list and remove it, if
+            // it is.
+            // This makes it possible to disable timer events from the
+            // timeout callbacks of other timer events.
+
+            IExpiryRegistration registration =
+                *this.outer.expiry_to_client.get(&expiry);
+            foreach (ref pending; this.outer.expired_registrations[])
+            {
+                if (pending is registration)
+                    pending = null;
+            }
+
+            // do everything else as for regular unregistration
+            this.unregister(expiry);
         }
 
         /***********************************************************************
@@ -296,6 +333,9 @@ abstract class TimeoutManagerBase : ITimeoutManager
     /***************************************************************************
 
         List of expired registrations. Used by the checkTimeouts() method.
+
+        Elements can be set to `null` by `drop` method, in which case they
+        are ignored.
 
     ***************************************************************************/
 
@@ -513,6 +553,11 @@ abstract class TimeoutManagerBase : ITimeoutManager
         // the optional delegate returns false.
         foreach (ref registration; this.expired_registrations[])
         {
+            // registration can be disabled by `drop` method by being set to
+            // `null`
+            if (registration is null)
+                continue;
+
             ITimeoutClient client = registration.timeout();
 
             if (dg !is null) if (!dg(client)) break;
--- a/src/ocean/time/timeout/model/ExpiryRegistrationBase.d
+++ b/src/ocean/time/timeout/model/ExpiryRegistrationBase.d
@@ -88,11 +88,30 @@ abstract class ExpiryRegistrationBase : IExpiryRegistration
                          the IExpiryRegistration instance to unregister
 
             In:
-                Must not be called from within timeout().
+                Must not be called from within timeout(). Doing so would still
+                leave already fired timer events in the TimeoutManager internal
+                list and their respective `timeout` will be called despite
+                unregistration.
 
         ***********************************************************************/
 
         void unregister ( ref Expiry expiry );
+
+        /***********************************************************************
+
+            Unregisters the specified expiry. If the expiry is present in the
+            list of expired registrations being currently iterated over by
+            checkTimeouts, then it will be removed (its timeout method will not
+            be called). (This means that drop can be called from timeout
+            callbacks, unlike unregister.)
+
+            Params:
+                expiry = expiry token returned by register() when registering
+                         the IExpiryRegistration instance to unregister
+
+        ***********************************************************************/
+
+        void drop ( ref Expiry expiry );
 
         /***********************************************************************
 
@@ -190,22 +209,23 @@ abstract class ExpiryRegistrationBase : IExpiryRegistration
 
     public bool unregister ( )
     {
-        if (this.expiry) try
-        {
-            debug ( TimeoutManager ) Stderr("*** unregister ")(this.id)('\n').flush();
+        return this.unregisterWith({ this.mgr.unregister(*this.expiry); });
+    }
 
-            this.mgr.unregister(*this.expiry);
+    /***************************************************************************
 
-            return true;
-        }
-        finally
-        {
-            this.expiry = null;
-        }
-        else
-        {
-            return false;
-        }
+        Same as `unregister` but also removes expirty from list of currently
+        expired registrations. That means it can be called from `timeout`
+        callbacks, contrary to `unregister`.
+
+        Returns:
+            true on success or false if no client was registered.
+
+    ***************************************************************************/
+
+    public bool drop ( )
+    {
+        return this.unregisterWith({ this.mgr.drop(*this.expiry); });
     }
 
     /***************************************************************************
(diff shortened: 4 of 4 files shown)


mihails-strasuns-sociomantic:
It looks like origin of the problem comes from here https://github.com/sociomantic-tsunami/ocean/commit/43e881d96f5a7e760bf8427266e3d9e0de371672#diff-cdfe73f7a32941d6ef2be1802a4a7819R680

mihails-strasuns-sociomantic:
But it is only manifestation of real problem - `terminationHook` system is broken when several delegates with same pointer are registered. In this specific case same hooks get added from both `wait` and `awaitOrTimeout` for the same task.

It will probably be more clear once I come up with fixup PR :)

mihails-strasuns-sociomantic:
On a related topic, @gavin-norman-sociomantic could you explain me something about `TimerSet` internals please? It looks like normally event is recycled from this method - https://github.com/sociomantic-tsunami/ocean/blob/v3.x.x/src/ocean/io/select/client/TimerSet.d#L212-L216 . But what happens when `unregister` is called instead (https://github.com/sociomantic-tsunami/ocean/blob/v3.x.x/src/ocean/io/select/client/TimerSet.d#L199-L202)?

mihails-strasuns-sociomantic:
Thanks to some very productive brain storming session with @nemanja-boric-sociomantic , root bug is identified as the fix is attached here :)

relnotes/expiry.feature.md
```diff
@@ -0,0 +1,6 @@
+* `ocean.time.timeout.model.ExpiryRegistrationBase`
+
+  New method, `drop`, provides same functionality as `unregister`, but also
+  removes the expiry from current list of fired events if it was present there.
+  This method makes possible to disable one expiry from timeout callback of
+  another expiry, even if they fire in the same time slot.
```
mathias-lang-sociomantic: This is a patch release though

mihails-strasuns-sociomantic: Yes, this is one of rare exception cases "have to add feature to fix critical bug" in my opinion. I would probably even make the method `package(ocean)` protected if I could, to stick with formalities. But I can't, so better at least document it.

src/ocean/time/timeout/TimeoutManager.d
```diff
@@ -228,6 +227,37 @@ abstract class TimeoutManagerBase : ITimeoutManager
...
+            expired registrations. That means it can be called from `timeout`
+            callbacks, contrary to `unregister`.
+
+            Params:
+                expiry = expiry token returned by register() when registering
+                         the IExpiryRegistration instance to unregister
+
+        ***********************************************************************/
+
+        protected void drop ( ref Expiry expiry )
```
nemanja-boric-sociomantic, line 245: Only thing I wonder about is how much is `unregister` useful now. Should anything call `unregister` now?

mihails-strasuns-sociomantic: Yes, it is the very same reason why I couldn't just change `unregister` to do the same thing : https://github.com/mihails-strasuns-sociomantic/ocean/blob/a3b792b1dc6d1b0a87897fa9c5647467815f9b79/src/ocean/time/timeout/TimeoutManager.d#L524-L531

nemanja-boric-sociomantic: Yes, now I remember. Feels like `unregister` should be private then, but I don't know how could we pull that off.

mihails-strasuns-sociomantic: I don't think so as it has to be used across ocean modules. `package(ocean)` would fix it but it is not available in D1.

nemanja-boric-sociomantic: Yeah. Oh well, such is life.

src/ocean/time/timeout/TimeoutManager.d
```diff
@@ -228,6 +227,37 @@ abstract class TimeoutManagerBase : ITimeoutManager
...
+            // list that is processed right now, remove it. This makes possible
+            // to disable timer events from timeout callbacks of other timer
+            // events, even if they compete for the same time slot.
+
+            IExpiryRegistration registration =
+                *this.outer.expiry_to_client.get(&expiry);
+            foreach (ref pending; this.outer.expired_registrations[])
+            {
+                if (pending is registration)
+                    pending = null;
```
nemanja-boric-sociomantic, line 259: According to the test coverage, this line is red :-). I wonder if it we merge it and suddenly becomes green :-).

Also, I wonder what about "files not in diff unexpectedly changed coverage". I hope codecov has some threshold for this.

nemanja-boric-sociomantic (approved):

src/ocean/time/timeout/TimeoutManager.d
mihails-strasuns-sociomantic: The task test which uncovered this was only added in 3.5.0 and this PR targets 3.4.x - I decided that adding yet another one just for the purpose of more coverage in old branch is not rational :)

src/ocean/task/Scheduler.d
```diff
@@ -568,6 +568,8 @@ final class Scheduler
         assert (this.state != State.Shutdown);
         this.state = State.Running;
 
+        debug_trace("Starting `theScheduler.eventLoop`");
```
gavin-norman-sociomantic: Bug fix?

src/ocean/time/timeout/TimeoutManager.d
```diff
@@ -228,6 +227,37 @@ abstract class TimeoutManagerBase : ITimeoutManager
 
         /***********************************************************************
 
+            Same as `unregister` but also removes expirty from list of currently
```
gavin-norman-sociomantic: expirty

src/ocean/time/timeout/TimeoutManager.d
```diff
@@ -228,6 +227,37 @@ abstract class TimeoutManagerBase : ITimeoutManager
...
+
+            Params:
+                expiry = expiry token returned by register() when registering
+                         the IExpiryRegistration instance to unregister
+
+        ***********************************************************************/
+
+        protected void drop ( ref Expiry expiry )
+        {
+            // If this expiry is already present in the expired registration
```
gavin-norman-sociomantic: This is a bit clearer and more specific:

"If this method is called while `checkTimeouts` is iterating over the list of expired registrations, then we need to check whether the expiry to be dropped is present in the list and remove it, if it is. This makes it possible to disable timer events from the timeout callbacks of other timer events, even if they compete for the same time slot."

>even if they compete for the same time slot

What does this mean, by the way?

mihails-strasuns-sociomantic:
Rebased

src/ocean/time/timeout/TimeoutManager.d
gavin-norman-sociomantic: I'd make this more specific:

"
Unregisters the specified expiry. If the expiry is present in the list of expired registrations being currently iterated over by `checkTimeouts`, then it will be removed (its `timeout` method will not be called). (This means that `drop` can be called from `timeout` callbacks, unlike `unregister`.)

If the specified expiry is currently not registered, nothing is done.
"

I'm not sure if this part of the doc for `unregister` is also relevant for `drop`:
>The subclass may reset the client instance to null after it has called this method (even if it throws an exception).

src/ocean/time/timeout/TimeoutManager.d
```diff
@@ -228,6 +227,37 @@ abstract class TimeoutManagerBase : ITimeoutManager
...
+            // list that is processed right now, remove it. This makes possible
+            // to disable timer events from timeout callbacks of other timer
+            // events, even if they compete for the same time slot.
+
+            IExpiryRegistration registration =
+                *this.outer.expiry_to_client.get(&expiry);
+            foreach (ref pending; this.outer.expired_registrations[])
+            {
+                if (pending is registration)
```
gavin-norman-sociomantic, line 258: Could you add a comment on the `this.outer.expired_registrations` field stating that the elements may be set to null by this method?

src/ocean/time/timeout/TimeoutManager.d
```diff
@@ -513,6 +543,9 @@ abstract class TimeoutManagerBase : ITimeoutManager
         // the optional delegate returns false.
         foreach (ref registration; this.expired_registrations[])
         {
+            if (registration is null)
```
gavin-norman-sociomantic: Maybe add a comment: `// May be null if removed by drop`.

src/ocean/time/timeout/model/ExpiryRegistrationBase.d
```diff
@@ -96,6 +96,20 @@ abstract class ExpiryRegistrationBase : IExpiryRegistration
 
         /***********************************************************************
 
+            Same as `unregister` but also removes expirty from list of currently
```
gavin-norman-sociomantic: Same comment suggestion as above.

src/ocean/time/timeout/model/ExpiryRegistrationBase.d
```diff
@@ -96,6 +96,20 @@ abstract class ExpiryRegistrationBase : IExpiryRegistration
...
+            expired registrations. That means it can be called from `timeout`
+            callbacks, contrary to `unregister`.
```
gavin-norman-sociomantic: Could you extend the doc of `update` to explain why it should not be called from a `timeout` callback?

Also: should `unregister` be called at all by any non-library code any more? (If not, please document this as well.)

src/ocean/time/timeout/model/ExpiryRegistrationBase.d
```diff
@@ -190,22 +204,23 @@ abstract class ExpiryRegistrationBase : IExpiryRegistration
...
-        }
-        finally
-        {
-            this.expiry = null;
-        }
-        else
-        {
-            return false;
-        }
+        Same as `unregister` but also removes expirty from list of currently
```
gavin-norman-sociomantic, line 217: Same comment suggestion.

src/ocean/time/timeout/model/ExpiryRegistrationBase.d
```diff
@@ -347,6 +362,44 @@ abstract class ExpiryRegistrationBase : IExpiryRegistration
 
     /***************************************************************************
 
+        Extracted common code of `drop` and `unregister`.
+
```
gavin-norman-sociomantic: Please document what the extracted logic does.

src/ocean/time/timeout/TimeoutManager.d
gavin-norman-sociomantic: Should we make `unregister` private, in a future release? (Either as a breaking change or via a deprecation path.)

src/ocean/io/select/client/TimerSet.d
```diff
@@ -202,6 +202,7 @@ public class TimerSet ( EventData ) : TimerEventTimeoutManager
         public void unregister ( )
         {
             this.expiry_registration.drop();
+            this.outer.events.recycle(this);
```
gavin-norman-sociomantic: Good catch!

src/ocean/time/timeout/TimeoutManager.d
mihails-strasuns-sociomantic: > What does this mean, by the way?

It was a shortest intuitive explanation of how `TimeoutManager` works I could think of :) i.e. that it fires once for actual epoll timer but multiplexes to several timer events for that one "slot".

src/ocean/time/timeout/model/ExpiryRegistrationBase.d
mihails-strasuns-sociomantic: > should unregister be called at all by any non-library code any more?

I was hoping you can answer that :) To me it seems 100% wrong to call it with old semantics but I don't have a good overview of how it is used across apps.

src/ocean/time/timeout/TimeoutManager.d
nemanja-boric-sociomantic: I think this answers your last question @mihails-strasuns-sociomantic :-)

src/ocean/time/timeout/TimeoutManager.d
mihails-strasuns-sociomantic: > I'm not sure if this part of the doc

I have no idea what it does even mean

src/ocean/time/timeout/TimeoutManager.d
nemanja-boric-sociomantic: Oh, yes, I asked the same question @gavin-norman-sociomantic , but unregister can't be private, since it needs to be accessible from other ocean internal modules - `package` would be appropriate, but it doesn't work.

mihails-strasuns-sociomantic:
Updated:
- documentation changes as suggested by @gavin-norman-sociomantic 
- removed unnecessary commit with extra scheduler traces

src/ocean/time/timeout/TimeoutManager.d
mihails-strasuns-sociomantic: Maybe we can do some trick with `protected`, but I'd delay that for v3.x.x

src/ocean/time/timeout/TimeoutManager.d
gavin-norman-sociomantic: Me neither. Maybe leave it off ;)

src/ocean/time/timeout/TimeoutManager.d
gavin-norman-sociomantic: Ah I see. It seems pretty redundant to me, as timeout callbacks are always called from `checkTimeouts`. I'd just leave it as:
"This makes it possible to disable timer events from the timeout callbacks of other timer events."

src/ocean/time/timeout/TimeoutManager.d
```diff
@@ -228,6 +230,41 @@ abstract class TimeoutManagerBase : ITimeoutManager
...
+        ***********************************************************************/
+
+        protected void drop ( ref Expiry expiry )
+        {
+            // If this method is called while checkTimeouts is iterating over
+            // the list of expired registrations, then we need to check whether
+            // the expiry to be dropped is present in the list and remove it, if
+            // it is. This makes it possible to disable timer events from the
+            // timeout callbacks of other timer
+            // events, even if they compete for the same time slot.
```
gavin-norman-sociomantic: Ditto about leaving off the bit about competing for a time slot.

src/ocean/time/timeout/TimeoutManager.d
mihails-strasuns-sociomantic: Updated

relnotes/expiry.feature.md
gavin-norman-sociomantic: "even if they fire in the same time slot" -- here too. (Sorry to be pedantic, but I think this phrase adds unnecessary confusion.)

mihails-strasuns-sociomantic:
Fixed relnotes too

gavin-norman-sociomantic (approved):

PR #320, opened 2017-10-20
Update beaver to add D2 auto-conversion for tags and upload codecov reports

leandro-lucarella-sociomantic:
Now when a tag is released, a new stage will also convert the tag to D2, then it will commit the changes, create a new tag with a +d2 suffix, push the changes and create a GitHub release for it.

This is done via the new beaver dlang d2-release command.

* submodules/beaver v0.2.2(480e6f5)...v0.2.2+10(55e9471) (10 commits)
  > dlang: Add a command to release converted D2 tags
  > make-wrapper-dlang: Fix matrix row selection
  > Drop trusty support
  > Change default matrix row for make-wrapper-dlang
  > Add an make wrapper for dlang projects
  (...)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,12 +26,25 @@ env:
         - DMD=2.071.2.s1 DIST=xenial F=production
         - DMD=2.071.2.s1 DIST=xenial F=devel
 
+# Don't build tags already converted to D2
+if: NOT tag =~ \+d2$
+
 install: beaver dlang install
 
 script: BEAVER_DOCKER_VARS="F AFTER_SCRIPT" beaver dlang make
 
-after_success:
-    - test "$AFTER_SCRIPT" = "1" && beaver run ci/codecov.sh
-    - test "$AFTER_SCRIPT" = "1" &&
-        BEAVER_DOCKER_VARS="OCEAN_D2_USER OCEAN_D2_PASS"
-        beaver run ci/d2tags.sh
+after_success: test "$AFTER_SCRIPT" = "1" && beaver run ci/codecov.sh
+
+jobs:
+    include:
+        - stage: D2 Release
+          # We need to include the exclusion of D2 tags because this "if"
+          # replaces the global one
+          if: tag IS present AND NOT tag =~ \+d2$
+          # We override it, otherwise is inherited from the first matrix row,
+          # we need DIST and DMD for `beaver dlang install`
+          env: DIST=xenial DMD=2.070.2.s12
+          # before_install and install are inherited
+          script: beaver dlang d2-release
+          # Overridden to NOP, otherwise is inherited
+          after_success: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ env:
         # Make sure beaver is in the PATH
         - PATH="$(git config -f .gitmodules submodule.beaver.path)/bin:$PATH"
     matrix:
-        - DMD=1.081.1 DIST=xenial F=production AFTER_SCRIPT=1
+        - DMD=1.081.1 DIST=xenial F=production
         - DMD=1.081.1 DIST=xenial F=devel
         - DMD=2.070.2.s12 DIST=xenial F=production
         - DMD=2.070.2.s12 DIST=xenial F=devel
@@ -31,9 +31,9 @@ if: NOT tag =~ \+d2$
 
 install: beaver dlang install
 
-script: BEAVER_DOCKER_VARS="F AFTER_SCRIPT" beaver dlang make
+script: beaver dlang make
 
-after_success: test "$AFTER_SCRIPT" = "1" && beaver run ci/codecov.sh
+after_success: beaver dlang codecov
 
 jobs:
     include:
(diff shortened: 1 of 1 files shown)


leandro-lucarella-sociomantic:
FYI @joseph-wakeling-sociomantic @mathias-lang-sociomantic

leandro-lucarella-sociomantic:
Now that I think of, doing a simple test in a fork shouldn't be too hard...

.travis.yml
```diff
@@ -26,12 +26,22 @@ env:
         - DMD=2.071.2.s1 DIST=xenial F=production
         - DMD=2.071.2.s1 DIST=xenial F=devel
 
+# Don't build tags already converted to D2
+if: NOT tag =~ ^.*\+d2^
```
mihails-strasuns-sociomantic: To continue my question from beaver PR - I really think this check is much better in actual convert script, less chance to forget this line in travis script and get things broken.

.travis.yml
```diff
@@ -26,12 +26,22 @@ env:
...
+jobs:
+    include:
+        - stage: D2 Release
+          if: tag IS present
+          # We override it, otherwise is inherited from the first matrix row
+          env: DIST=xenial
+          # before_install and install are inherited
+          script: beaver dlang d2-release
+          # Overridden to NOP, otherwise is inherited
+          after_success: true
```
mihails-strasuns-sociomantic: Maybe we should just move codecov script to `script` phase to make things simpler here?

.travis.yml
leandro-lucarella-sociomantic: Mmm, for me it's better in here, as we don't even want to trigger a travis build at all for a tag like this (like the status shouldn't be set, if we run the job anyway and make it a NOP it will get a bogus status).

Maybe we can include the check in both parts, but then we might forget to keep them in sync.

.travis.yml
leandro-lucarella-sociomantic: I think even if we move codecov to script, this is an accident waiting to happen, so I would be inclined to keep the comments. To be honest travis stages seem like an ugly hack for now.

And having codecov as a `after_success` script is slightly more correct because if it fails it won't make the build fail.

.travis.yml
leandro-lucarella-sociomantic: OTOH, I'm failing and getting these conditions to work. In my previous tests, using `if: tag IS present` in the stage works fine (you can see in this PR job that there is no `D2 Release` stage), but in here the job is still triggered for the `+d2` tag. I don't know if I'm getting the regex wrong or what, I will contact travis support.

gavin-norman-sociomantic:
Moved to v3.4.7, as it's still marked WIP.

leandro-lucarella-sociomantic:
Alright! The current version seems to work as expected, I just tested it in my fork. You can see in my travis that no builds were triggered by Travis for the +d2 tag: https://travis-ci.org/leandro-lucarella-sociomantic/ocean/branches

Also, the automatically converted tag is there: https://github.com/leandro-lucarella-sociomantic/ocean/releases/tag/v2.3.5-alpha4%2Bd2

And also, the build for the branch (as for this PR) doesn't even include the conversion stage: https://travis-ci.org/sociomantic-tsunami/ocean/builds/296352146, while for the original tag is there: https://travis-ci.org/leandro-lucarella-sociomantic/ocean/builds/296351091.

leandro-lucarella-sociomantic:
So even when we first need a beaver release, this is ready for a final review.

.travis.yml
leandro-lucarella-sociomantic: Fixed, I was just mixing up `^` and `$` :roll_eyes:

mihails-strasuns-sociomantic (approved):
I don't like how much magic lines it does require but we won't be able to do any better with YML/Travis AFAICS.

mathias-lang-sociomantic:
Shouldn't that target v3.x.x ?

leandro-lucarella-sociomantic:
The earliest we get D2 releases the better IMHO, this is an infra change only, so no risk for end users. But I'm fine with whatever.

leandro-lucarella-sociomantic:
Updated to use the codecov reporting from beaver too.

leandro-lucarella-sociomantic:
Now using beaver v0.3.0. Ready for the final review / merge.

.travis.yml
```diff
@@ -19,19 +19,32 @@ env:
...
-        beaver run ci/d2tags.sh
+jobs:
+    include:
+        - stage: D2 Release
+          # We need to include the exclusion of D2 tags because this "if"
+          # replaces the global one
+          if: tag IS present AND NOT tag =~ \+d2$
+          # We override it, otherwise is inherited from the first matrix row,
+          # we need DIST and DMD for `beaver dlang install`
+          env: DIST=xenial DMD=2.070.2.s12
```
mihails-strasuns-sociomantic: Remind me, why did it matter for `DIST` and `DMD` to be different from first matrix row? `make d2conv` doesn't depend on D compiler in any way.

leandro-lucarella-sociomantic: IIRC if you don't "fix" the matrix, the job will be run for all the matrix rows. I just used D2 because it seemed to make more sense.

mihails-strasuns-sociomantic: Right, you should fix your comment in that case - it says that first row would be inherited, now whole matrix

leandro-lucarella-sociomantic: I'll do another test to find out, it was some time ago and better not to put misleading comments...

leandro-lucarella-sociomantic: It works without the `env:` override: https://travis-ci.org/leandro-lucarella-sociomantic/ocean/builds/299741606

But I think now I remember why it was like that. I started based on a branch that has `DIST` moved as a global. Maybe that works too though, I'll check.

leandro-lucarella-sociomantic: It seems to work too: https://travis-ci.org/leandro-lucarella-sociomantic/ocean/builds/299743144

I'll update.

leandro-lucarella-sociomantic:
Updated.

mihails-strasuns-sociomantic (approved):
LGTM!

PR #326, opened 2017-10-25
Unregister connection socket before closing it

gavin-norman-sociomantic:
As noted in
https://github.com/sociomantic-tsunami/dhtnode/issues/25#issuecomment-339293270https://github.com/sociomantic-tsunami/dhtnode/issues/25#issuecomment-339293270,
it is unsafe (at least in D1 builds with the CDGC) to rely on the OS to
unregister a socket from epoll when it's closed. When the GC kicks in, it
forks, causing all fds of the process to be duplicated. While the fork is
alive, this means that closing a socket that existed when the GC fork began
will not cause it to be unregistered from epoll, as they fork's copy
remains. This can result in events firing for a socket that has been
closed.

This commit makes sure that a closed connection socket cannot fire in epoll
by unregistering it before closing it.

Note that TaskConnectionHandler will also need to be adapted to override
unregisterSocket(), but this will likely require new features and is not
regarded as an urgent bugfix (as TaskConnectionHandler is not yet in active
use).

### DIFF
--- a/src/ocean/net/server/connection/IConnectionHandler.d
+++ b/src/ocean/net/server/connection/IConnectionHandler.d
@@ -321,6 +321,14 @@ abstract class IConnectionHandler : IConnectionHandlerInfo,
                 this.error(this.socket_error.setSock("error closing connection", __FILE__, __LINE__));
             }
 
+            // The socket is closed below. However, this does not always mean
+            // that it is unregistered from epoll -- in situations where the
+            // process has forked, the fork's reference to the underlying kernel
+            // file description will prevent it from being unregistered until
+            // the fork exits. Therefore, to be certain that the socket will not
+            // fire again in epoll, we need to explicitly unregister it.
+            this.unregisterSocket();
+
             this.socket.close();
         }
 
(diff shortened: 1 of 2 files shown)


src/ocean/net/server/connection/IConnectionHandler.d
```diff
@@ -321,6 +321,14 @@ abstract class IConnectionHandler : IConnectionHandlerInfo,
                 this.error(this.socket_error.setSock("error closing connection", __FILE__, __LINE__));
             }
 
+            // The socket is closed below. However, this does not always mean
+            // that it is unregistered from epoll -- in situations where the
+            // process has forked, the fork's reference to the fd will prevent
```
nemanja-boric-sociomantic: I suggest: "the fork's reference to the underlying kernel's file description" to emphasize that there are two file descriptors pointing to the same underlying file description.

src/ocean/net/server/connection/IFiberConnectionHandler.d
```diff
@@ -221,6 +221,25 @@ abstract class IFiberConnectionHandlerBase : IConnectionHandler
...
+        being unregistered until the fork exits. Therefore, to be certain that
+        the socket will not fire again in epoll, we need to explicitly
+        unregister it.
+
+    ***************************************************************************/
+
+    override protected void unregisterSocket ( )
+    {
+        if ( auto registered_client = this.fiber.registered_client )
+            if ( this.socket.fd == registered_client.fileHandle() )
```
nemanja-boric-sociomantic, line 237: What would be the reason when this is not true?

gavin-norman-sociomantic: It would mean that the connection fiber is suspended waiting for another event, not I/O on the socket. (Thus the socket is not registered with epoll and we don't need to worry about it.)

src/ocean/net/server/connection/IConnectionHandler.d
gavin-norman-sociomantic: Updated.

nemanja-boric-sociomantic (approved):

nemanja-boric-sociomantic:
@david-eckardt-sociomantic your turn.

david-eckardt-sociomantic (approved):

mihails-strasuns-sociomantic (approved):

PR #330, opened 2017-10-27
Deprecate `Time.toString` methods

mihails-strasuns-sociomantic:
First step before removing them in ocean v4.0.0 and breaking one of import cycles vanilla druntime complains about.

### DIFF
--- a/src/ocean/text/convert/DateTime_tango.d
+++ b/src/ocean/text/convert/DateTime_tango.d
@@ -36,6 +36,7 @@ import ocean.core.ExceptionDefinitions;
 import ocean.stdc.posix.langinfo;
 import Integer = ocean.text.convert.Integer_tango;
 import Utf = ocean.text.convert.Utf;
+import ocean.text.convert.Formatter;
 import ocean.text.util.StringC;
 import ocean.time.chrono.Calendar;
 import ocean.time.chrono.Gregorian;
@@ -967,6 +968,41 @@ private struct Result
     private char[] get ()
     {
         return target_[0 .. index];
+    }
+}
+
+/*******************************************************************************
+
+    Params:
+        value = time value to wrap in pretty-printing struct
+
+    Returns:
+        wrapper struct which, when supplied to `Formatter`, prints `value`
+        using current default locale settings.
+
+*******************************************************************************/
+
+public AsPrettyStr asPrettyStr ( Time value )
+{
+    return AsPrettyStr(value);
+}
+
+/*******************************************************************************
+
+    Wrapper struct which, when supplied to `Formatter`, prints `value`
+    using current default locale settings.
+
+*******************************************************************************/
+
+public struct AsPrettyStr
+{
+    private Time value;
+
+    public void toString (FormatterSink sink)
+    {
+        // Layout defaults to 'G'
+        scope dg = (cstring s) { sink(s); return s.length; };
+        DateTimeDefault.format(dg, this.value, "");
     }
 }
 
--- a/src/ocean/time/Time_test.d
+++ b/src/ocean/time/Time_test.d
@@ -66,3 +66,19 @@ unittest
         "01/06/70 00:00:00"
     );
 }
+
+// raw time
+unittest
+{
+    // NB: ocean is compiled with deprecations-as-errors, so deprecated
+    // toString method is not picked even before it is completely removed:
+
+    test!("==")(
+        format("{}", Time.epoch1970),
+        "{ ticks_: 621355968000000000 }"
+    );
+    test!("==")(
+        format("{}", Time.epoch1970 + TimeSpan.fromDays(5)),
+        "{ ticks_: 621360288000000000 }"
+    );
+}
--- a/src/ocean/time/Time_test.d
+++ b/src/ocean/time/Time_test.d
@@ -1,0 +1,53 @@
+/*******************************************************************************
+
+    Copyright:
+        Copyright (c) 2009-2016 Sociomantic Labs GmbH.
+        All rights reserved.
+
+    License:
+        Boost Software License Version 1.0. See LICENSE_BOOST.txt for details.
+        Alternatively, this file may be distributed under the terms of the Tango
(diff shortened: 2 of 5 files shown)


mihails-strasuns-sociomantic:
Competing proposal to https://github.com/sociomantic-tsunami/ocean/pull/93

mihails-strasuns-sociomantic:
Reason why I would prefer such more intrusive approach should be preferred is that I view every single circular import as a technical debt. You do it because it is often fast and convenient, but in the long run it is always a problem that needs to be solved. Circular imports break modularization, making impossible to import only subset of library functionality you really use and often harming build times as a side effect.

While the matter is not important enough to hide and kill every circular import case intentionally, it should not be hidden with workarounds if such case is shown by compiler/runtime.

In this specific case I also don't think it was a good idea in the first place to do any kind of non-trivial time formatting inside default `toString` method, especially considering it relies on global state.

mathias-lang-sociomantic:
Noticed, and I think the concept is sane. I would like to give it a look on a day I'm a bit more functional (feel free to ping me from Wednesday on ;) ).
Ultimately I think we could/should bring both P.R. in, because those static ctors are also nasty. But we can keep that discussion on the other P.R.

mihails-strasuns-sociomantic:
Wednesday ping @mathias-lang-sociomantic :)

src/ocean/text/convert/DateTime_tango.d
```diff
@@ -970,6 +970,41 @@ private struct Result
...
+    Wrapper struct which, when supplied to `Formatter`, prints `value`
+    using current default locale settings.
+
+*******************************************************************************/
+
+public struct AsPrettyStr
+{
+    private Time value;
+
+    public void toString (void delegate(cstring) sink)
```
mathias-lang-sociomantic: `FormatterSink` ?

src/ocean/text/convert/DateTime_tango.d
```diff
@@ -970,6 +970,41 @@ private struct Result
...
+    {
+        // Layout defaults to 'G'
+        scope dg = (cstring s) { sink(s); return s.length; };
+        DateTimeDefault.format(dg, this.value, "");
```
mathias-lang-sociomantic: Hum since those two differ I think we can just change `format` here... Will look into it

src/ocean/text/convert/Formatter_test.d
```diff
@@ -546,6 +546,22 @@ unittest
...
+    test!("==")(
+        format("{}", Time.epoch1970),
+        "{ ticks_: 621355968000000000 }"
+    );
+    test!("==")(
+        format("{}", Time.epoch1970 + TimeSpan.fromDays(5)),
+        "{ ticks_: 621360288000000000 }"
+    );
+}
+
```
mathias-lang-sociomantic: I think this belong to `Time_test` more than formatter's

src/ocean/time/Time_test.d
```diff
@@ -0,0 +1,57 @@
...
+    test (TimeSpan.fromHours(5000).hours is 5000);
+    test (TimeSpan.fromDays(6).days is 6);
+    test (TimeSpan.fromDays(5000).days is 5000);
+
+    test (TimeSpan.fromSeconds(50).time.seconds is 50);
+    test (TimeSpan.fromSeconds(5000).time.seconds is 5000 % 60);
+    test (TimeSpan.fromMinutes(50).time.minutes is 50);
+    test (TimeSpan.fromMinutes(5000).time.minutes is 5000 % 60);
+    test (TimeSpan.fromHours(23).time.hours is 23);
+    test (TimeSpan.fromHours(5000).time.hours is 5000 % 24);
```
mathias-lang-sociomantic: Nitpick: extra space before parent

src/ocean/text/convert/DateTime_tango.d
mihails-strasuns-sociomantic: What do you mean?

mihails-strasuns-sociomantic:
Updated:
- moved all tests to `Time_test.d`
- whitespaces
- `FormatterSink`

src/ocean/text/convert/DateTime_tango.d
mathias-lang-sociomantic: This uses a `size_t`-returning delegate and shouldn't. But we can take care of that later.

src/ocean/time/Time_test.d
```diff
@@ -0,0 +1,84 @@
...
+    test(TimeSpan.fromMinutes(5000).time.minutes is 5000 % 60);
+    test(TimeSpan.fromHours(23).time.hours is 23);
+    test(TimeSpan.fromHours(5000).time.hours is 5000 % 24);
+
+    auto tod = TimeOfDay (25, 2, 3, 4);
+    tod = tod.span.time;
+    test(tod.hours is 1);
+    test(tod.minutes is 2);
+    test(tod.seconds is 3);
+    test(tod.millis is 4);
```
mathias-lang-sociomantic, line 54: All those tests could be converted to `test!("...")` but it's a minor point.

mathias-lang-sociomantic (approved):
LGTM

src/ocean/time/Time_test.d
```diff
@@ -51,3 +53,16 @@ unittest
     test(tod.seconds is 3);
     test(tod.millis is 4);
 }
+
+// pretty time formatting
+unittest
+{
+    test!("==")(
+        format("{}", asPrettyStr(Time.epoch1970)),
+        "01/01/70 00:00:00"
```
nemanja-boric-sociomantic, line 62: I sure hope the locale on the test machines will not change :-)

mihails-strasuns-sociomantic: A _lot_ of tests will break when that happens :)

nemanja-boric-sociomantic (approved):

mihails-strasuns-sociomantic:
Thanks!

PR #331, opened 2017-10-27
Move import outside of unittest block

mathias-lang-sociomantic:
IODevice is used as a template constraint by connect, so it should not be imported only in the unittest block.


mihails-strasuns-sociomantic:
Bug fix?

mathias-lang-sociomantic:
It's only detected by newer compiler, I haven't seen it while compiling with older release. But sure can do.

mathias-lang-sociomantic:
Rebased. Will conflict on merge, as another import (`verify`) was added in v3.5.x

mihails-strasuns-sociomantic (approved):

PR #332, opened 2017-10-27
Add -debug=EpollFdSanity debug flag

nemanja-boric-sociomantic:
In case `EpollFdSanity` debug flag is set `EpollSelectDispatcher` and
related utilities will exploit the limitation in the user address space
and stuff the least significant byte of the fd into the highest bits
of the epoll registration data. This allows checking if the epoll firing
the event on the ISelectClient is actually valid: it can happen that the
ISelectClient gets reused, possibly changing its fd, but still leaving
the registration for the old file descriptor in the epoll. This should
help us debugging these issues more easily.

I've tested this with the dhtnode and #326 and it indeed catches the problems
immediately, and that problem disappears with  #326 applied!

### DIFF
--- a/src/ocean/io/select/EpollSelectDispatcher.d
+++ b/src/ocean/io/select/EpollSelectDispatcher.d
@@ -62,6 +62,16 @@ import core.stdc.errno: errno, EINTR, ENOENT, EEXIST, ENOMEM, EINVAL;
 
 debug ( ISelectClient ) import ocean.io.Stdout;
 
+version (UnitTest)
+{
+    debug = EpollFdSanity;
+}
+
+debug (EpollFdSanity)
+{
+    import ocean.io.select.selector.EpollFdSanity;
+}
+
 /*******************************************************************************
 
     EpollSelectDispatcher
@@ -320,7 +330,7 @@ public class EpollSelectDispatcher : IEpollSelectDispatcherInfo
         else
         {
             this.e.enforce(
-                this.epoll.ctl(epoll.CtlOp.EPOLL_CTL_ADD, client.fileHandle,
+                this.epollCtl(epoll.CtlOp.EPOLL_CTL_ADD, client.fileHandle,
                     client.events, client) == 0,
                 "error adding epoll registration",
                 "epoll_ctl"
@@ -372,7 +382,7 @@ public class EpollSelectDispatcher : IEpollSelectDispatcherInfo
                 this.registered_clients -= client;
             }
 
-            if (!this.epoll.ctl(epoll.CtlOp.EPOLL_CTL_DEL, client.fileHandle,
+            if (!this.epollCtl(epoll.CtlOp.EPOLL_CTL_DEL, client.fileHandle,
                 client.events, client))
             {
                 return 0;
@@ -585,7 +595,7 @@ public class EpollSelectDispatcher : IEpollSelectDispatcherInfo
 
     public bool modify ( ISelectClient client )
     {
-        if (!this.epoll.ctl(epoll.CtlOp.EPOLL_CTL_MOD, client.fileHandle,
+        if (!this.epollCtl(epoll.CtlOp.EPOLL_CTL_MOD, client.fileHandle,
             client.events, client))
         {
             return false;
@@ -596,7 +606,7 @@ public class EpollSelectDispatcher : IEpollSelectDispatcherInfo
 
             if (errnum == ENOENT)
             {
-                if (!this.epoll.ctl(epoll.CtlOp.EPOLL_CTL_ADD, client.fileHandle,
+                if (!this.epollCtl(epoll.CtlOp.EPOLL_CTL_ADD, client.fileHandle,
                     client.events, client))
                 {
                     return true;
@@ -796,6 +806,37 @@ public class EpollSelectDispatcher : IEpollSelectDispatcherInfo
                         "error waiting for epoll events");
                 }
             }
+        }
+    }
+
+    /***************************************************************************
+
+        Creates/deletes/modifies registration inside the epoll. Wrapped to
+        allow for the sanity checks on the EpollSelectDispatcher's level.
+
+        Params:
+            op     = epoll_ctl opcode
+            fd     = file descriptor to register for events
+            events = events to register fd for
+            client = user ISelectClient to set data.obj of the created epoll_data_t
+                     instance to
+
+        Returns:
+            0 on success or -1 on error. On error errno is set appropriately.
+
+    ***************************************************************************/
+
+    private int epollCtl ( Epoll.CtlOp op, int fd, Epoll.Event events,
+            ISelectClient client )
+    {
+        debug (EpollFdSanity)
+        {
+            return this.epoll.ctl(op, fd, client.events,
+                    FdObjEpollData.encode(client, client.fileHandle));
+        }
+        else
+        {
+            return this.epoll.ctl(op, fd, client.events, client);
         }
     }
 
--- a/src/ocean/io/select/selector/EpollFdSanity.d
+++ b/src/ocean/io/select/selector/EpollFdSanity.d
@@ -1,0 +1,135 @@
+/*******************************************************************************
+
+    As linux on 64 machines is using only up to 56 bits for the user address
+    space (see https://www.kernel.org/doc/Documentation/x86/x86_64/mm.txt) we
+    can use bits from at least 56 to 63 for stuffing in the least significant 8
+    bits of the fd into the epoll registration. This can be used to check if
+    the ISelectClient's registration triggering in epoll wait is made for the
+    previous usage of the ISelectClient (for the wrong fd). This is not 100%
+    accurate, as the file descriptors could be reused but it can be in general
+    this is the best we can to find out expired registrations that are not
+    unregistered from epoll.
+
+    This module provides FdObjEpollData structure that binds fd
+    and the object used at the registration time, so that they both can
+    be registered with epoll and returned with epoll_wait to the user.
+
+    Copyright:
+        Copyright (c) 2017 sociomantic labs GmbH.
+        All rights reserved.
+
+    License:
+        Boost Software License Version 1.0. See LICENSE_BOOST.txt for details.
+        Alternatively, this file may be distributed under the terms of the Tango
+        3-Clause BSD License (see LICENSE_BSD.txt for details).
+
+*******************************************************************************/
+
+module ocean.io.select.selector.EpollFdSanity;
+
+version (UnitTest)
+{
+    import ocean.core.Test;
+}
+
+/*******************************************************************************
+
+    Struct binding fd and a Object.
+
+*******************************************************************************/
+
+public struct FdObjEpollData
+{
+    /// Number of bits to store address in
+    private const address_bits = 56;
+
+    /// Mask for the user-space address
+    private const address_mask = (1UL << address_bits) - 1;
+
+    /// Mask for the storing part of the fd
+    private const fd_mask = ~address_mask;
+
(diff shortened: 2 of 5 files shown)


mihails-strasuns-sociomantic:
Could you just enable it always in `version(UnitTest)`, same as it is currently done with `-debug=ContiguousIntegrity`?

nemanja-boric-sociomantic:
Good point. Then I'll drop the second commit.

nemanja-boric-sociomantic:
Updated.

nemanja-boric-sociomantic:
Ah, it's segfaulting because this flag is per-module only. Is there any easy way to define it only once and use it in all affected modules, or should I just add it in all three affected modules @mihails-strasuns-sociomantic ?

nemanja-boric-sociomantic:
As for the unexpected coverage changes not visible in diff, I think it's just about the diff/coverage report being screwed a bit. The accused line (`event.data.obj = obj;`) is really in the diff.

mihails-strasuns-sociomantic:
For <5 modules I would just copy-paste it.

mihails-strasuns-sociomantic (approved):

src/ocean/io/select/EpollSelectDispatcher.d
```diff
@@ -482,12 +484,27 @@ public class EpollSelectDispatcher : IEpollSelectDispatcherInfo
...
+            debug (EpollFdSanity)
+            {
+                // In this case, we've already registered this fd with the
+                // client, and we need to make sure that we unregister it
+                // first, as the sanity check would fail while trying
+                // to handle the new select client
+                this.unregister(current);
+                return this.register(next);
             }
+            else
```
gavin-norman-sociomantic: Why `else`?

src/ocean/io/select/EpollSelectDispatcher.d
```diff
@@ -482,12 +484,27 @@ public class EpollSelectDispatcher : IEpollSelectDispatcherInfo
                     Stderr.formatln("  Replaced {}", current).flush();
                     Stderr.formatln("  with     {}", next).flush();
                 }
+            }
 
-                this.registered_clients -= current;
-                this.registered_clients += next;
+            debug (EpollFdSanity)
+            {
+                // In this case, we've already registered this fd with the
```
gavin-norman-sociomantic: I don't understand why this different logic here is required. (We didn't make this change in our rough testing, did we?)

src/ocean/sys/Epoll.d
```diff
@@ -722,3 +733,92 @@ struct Epoll
...
+        in the least significant 8 bits of the fd into the epoll
+        registration. This is used so we can check if the ISelectClient's
+        registration triggering in epoll wait is made for the previous
+        usage of the ISelectClient (for the wrong fd). This is not 100%
+        accurate, as the file descriptors could be reused but it can
+        be in general this is the best we can to find out expired
+        registrations that are not unregistered from epoll.
+
+        Params:
+            client = ISelectClient to create the registration value
```
gavin-norman-sociomantic: Wrong param.

src/ocean/sys/Epoll.d
```diff
@@ -722,3 +733,92 @@ struct Epoll
...
+                by stuffing it's fd into it's address
+
+        Returns:
+            combination of the ISelectClient's current fd and it's address
+            to register with the epoll.
+
+    ***********************************************************************/
+
+
+    public ulong createRegistationMask (Object obj, int fd)
```
gavin-norman-sociomantic: The word "mask" is misleading, I think. (It implies that this function returns a bit-mask which will then be used to stuff the fd into the ulong.)

How about this:
```D
struct FdObjEpollData
{
  Object obj;
  ubyte fd;

  static ulong encode ( Object obj, int fd ) { ... }

  static FdObjEpollData decode ( ulong encoded ) { ...}
}
```
?

src/ocean/sys/Epoll.d
```diff
@@ -722,3 +733,92 @@ struct Epoll
...
+        Struct binding fd and the ISelectClient.
+
+    ***********************************************************************/
+
+    public struct Registration
+    {
+        /// Object instance
+        Object obj;
+        /// client's fd at the time of the registration (least significant byte)
+        int fd;
```
gavin-norman-sociomantic: Hm any reason to expose this as a full `int`? It's deceptive, as only a single byte is stored.

src/ocean/sys/Epoll.d
```diff
@@ -669,7 +671,16 @@ struct Epoll
         epoll_event_t event;
 
         event.events   = events;
-        event.data.obj = obj;
+
+        debug (EpollFdSanity)
+        {
+            ulong registration = createRegistationMask(obj ,fd);
```
gavin-norman-sociomantic: May as well be written on one line, without the intermediary variable.

test/signalfd/main.d
```diff
@@ -436,7 +436,19 @@ private class SignalFDTest
         // FLAKY: call to epoll_wait() may fail or return wrong number of events
         test!("!=")(epoll_res, -1); // epoll_wait() error
         test!("==")(epoll_res, 1); // one event fired
-        assert(fired_events[0].data.obj is signal_fd, "unexpected event data");
+
+        // We need to handle if the test is built with EpollFdSanity
+        // as we're using epoll event directly, which may have registered
+        // fd stuffed in into the highest byte.
+        debug (EpollFdSanity)
```
gavin-norman-sociomantic: Huh, it's annoying to have to bear this in mind for any test which might use epoll directly. The only alternative would be to not run this test at all, in `-debug=EpollFdSanity`, which I don't think is an especially better option. What do you think? (I'm especially considering future tests that may be added, that would also need to bear this in mind.)

src/ocean/io/select/EpollSelectDispatcher.d
nemanja-boric-sociomantic: Ah, this is a good catch. I assumed I'll perform the lookup with the new client's address in the `epoll_ctl(EPOLL_CTL_MOD)` - this is not right - I'm doing it only by `fd`, which is same for the current and next. I think this is not needed.

src/ocean/io/select/EpollSelectDispatcher.d
nemanja-boric-sociomantic: Not needed, as this entire thing will just be `this.modify`

src/ocean/sys/Epoll.d
nemanja-boric-sociomantic: Yes, good suggestion.

src/ocean/sys/Epoll.d
nemanja-boric-sociomantic: Ah, good point. I'll just store it here as `ubyte`.

test/signalfd/main.d
nemanja-boric-sociomantic: I'll have to think a bit more about this. Somehow this doesn't taste good - there's asymmetry in the `Epoll`'s API now - registration doesn't match return values. I think what I'll need to do is to not embed registration protection this inside `epoll.ctl`, but to add new method into `Epoll` (maybe `epoll.safeCtl`), which will be used by `EpollSelectDispatcher` (and others, but then they'll know they need to handle the event parsing afterwards.

nemanja-boric-sociomantic: Hmm, other thing to do is to handle this safety check inside `EpollSelectDispatcher` and on the other side `SelectedKeysHandler`, and not touch `Epoll` at all. I think that might be the best approach.

gavin-norman-sociomantic: Yes, that would be good to try.

nemanja-boric-sociomantic:
Ok, reworked patch so that `EpollFdSanity` it's only `EpollSelectDispatcher/SelectedKeysHandler`'s business. Applications using `Epoll` directly will be oblivious about this flag.

src/ocean/sys/Epoll.d
```diff
@@ -722,3 +758,108 @@ struct Epoll
         return .close(this.fd);
     }
 }
+
+debug (EpollFdSanity)
+{
+    /***********************************************************************
+
+        Struct binding fd and the ISelectClient.
```
gavin-norman-sociomantic: "an fd and an Object"

src/ocean/sys/Epoll.d
```diff
@@ -722,3 +758,108 @@ struct Epoll
         return .close(this.fd);
     }
 }
+
+debug (EpollFdSanity)
```
gavin-norman-sociomantic: It's slightly odd that this struct is defined here but isn't used in this module. Maybe put this in a separate module?

gavin-norman-sociomantic: Maybe `ocean.io.select.selector.EpollFdSanity`?

nemanja-boric-sociomantic: Yeah, I'll move it to a separate module.

src/ocean/sys/Epoll.d
nemanja-boric-sociomantic: Thanks!

src/ocean/sys/Epoll.d
nemanja-boric-sociomantic: Done.

src/ocean/io/select/selector/EpollFdSanity.d
```diff
@@ -0,0 +1,127 @@
...
+        Returns:
+            FdObjEpollData struct containing Object and accompanying fd.
+
+    ***************************************************************************/
+
+    public static FdObjEpollData decode (ulong registration)
+    {
+        FdObjEpollData data;
+        data.obj = cast(Object)cast(void*)(registration & mask);
+        data.fd = cast(ubyte)(((registration & ~mask) >> 56) & 0xFF);
```
gavin-norman-sociomantic: The last `& 0xFF` looks redundant.

gavin-norman-sociomantic (commented):
Yep, that looks better in a separate module.

src/ocean/io/select/selector/EpollFdSanity.d
nemanja-boric-sociomantic: Updated, you're right.

nemanja-boric-sociomantic:
Please wait for the merging, something might be off.

src/ocean/io/select/selector/EpollFdSanity.d
```diff
@@ -0,0 +1,127 @@
...
+/*******************************************************************************
+
+    Struct binding fd and a Object.
+
+*******************************************************************************/
+
+public struct FdObjEpollData
+{
+    /// Mask for the user-space address bits.
+    private const mask = 0x00FF_FFFF_FFFF_FFFF;
```
gavin-norman-sociomantic: Small suggestion:
```D
const address_bits = 56;
const address_mask = (2 ^^ (address_bits + 1)) - 1;
const fd_mask = ~address_mask;
```
Then you could remove the two instances of `56` and the `~mask` below.

gavin-norman-sociomantic: `static assert(address_mask + fd_mask == ulong.max);` :)

nemanja-boric-sociomantic:
OK, I've tested this with the DHT node and if the d9525cf11bee7bd5d325eb7f00a015b3e057d2ac is not present in the ocean's tree, setting this debug flag catches the issue during the testing, so that's good. I'm going to apply your last suggestion, but other than that I'm quite happy how this worked out.

nemanja-boric-sociomantic:
Updated with the mask constants.

gavin-norman-sociomantic (approved):

PR #333, opened 2017-10-27
Change assert(false) to verify(false) in SmartUnion.handleInvalidCases

nemanja-boric-sociomantic:
This should be verify(false), but we missed it while converting.

### DIFF
--- a/src/ocean/core/SmartUnion.d
+++ b/src/ocean/core/SmartUnion.d
@@ -306,7 +306,7 @@ version ( UnitTest )
 
 *******************************************************************************/
 
-private const handleInvalidCases = "case none: assert(false);" ~
+private const handleInvalidCases = "case none: verify(false); break;" ~
     // The D2 build makes sure all enum cases are covered so in fact we don't
     // need a default. However, D1 emits a compiler warning if a switch
     // has no default so we add a default here just to avoid the warning.


src/ocean/core/SmartUnion.d
```diff
@@ -306,7 +307,7 @@ version ( UnitTest )
 
 *******************************************************************************/
 
-private const handleInvalidCases = "case none: assert(false);" ~
+private const handleInvalidCases = "case none: verify(false);" ~
```
mathias-lang-sociomantic: Then you need a break, because DMD cannot detect that `verify` never returns

nemanja-boric-sociomantic: Ah right.

src/ocean/core/SmartUnion.d
```diff
@@ -306,11 +307,11 @@ version ( UnitTest )
 
 *******************************************************************************/
 
-private const handleInvalidCases = "case none: assert(false);" ~
+private const handleInvalidCases = "case none: verify(false); break;" ~
     // The D2 build makes sure all enum cases are covered so in fact we don't
     // need a default. However, D1 emits a compiler warning if a switch
     // has no default so we add a default here just to avoid the warning.
-    "version (D_Version2){} else {default: assert(false);}";
+    "version (D_Version2){} else {default: verify(false); break; }";
```
mathias-lang-sociomantic: Well I'd be inclined to keep the `assert` here, as this case is *truly* unreachable. And it's never compiled in D2 anyway.

nemanja-boric-sociomantic: I did think about it and this case makes me sad. While it's true it's truly unreachable, it might not be for the D1-only compiled application. In which case, it doesn't matter if it's `AssertException` or `SanityException`, but I've changed it only for the consistency (as it's not compiled in D2 anyway). I don't know, let's see what other think, I don't feel right about this change, I'll tell you that.

david-eckardt-sociomantic: > While it's true it's truly unreachable, it might not be for the D1-only compiled application.

This is a contradiction, and I think it is truly unreachable, the D2 build makes sure of that. Or when do you think it could be reachable in a D1 build?
Here "truly unreachable" does not mean there is a formal proof, instead it means there is a guarantee -- the D2 build -- that each enum member has a `case`. Even D2 `final switch` doesn't prevent `u.active` from returning a value outside the range of the enum members.

nemanja-boric-sociomantic: Yes, I was thinking about cases when there's no D2 build for the project. But then, they shouldn't use `handleInvalidCases` anyway. I'll just revert this change :+1:

mathias-lang-sociomantic (approved):

src/ocean/core/SmartUnion.d
```diff
@@ -19,6 +19,7 @@ module ocean.core.SmartUnion;
 import ocean.transition;
 import ocean.core.ExceptionDefinitions;
 import ocean.core.Test;
+import ocean.core.Verify;
```
mihails-strasuns-sociomantic: Already imported few lines below

mihails-strasuns-sociomantic (commented):
I don't think either way matters in this case, so whatever suits you

src/ocean/core/SmartUnion.d
nemanja-boric-sociomantic: Updated, thanks!

mihails-strasuns-sociomantic (approved):

joseph-wakeling-sociomantic:
Unfortunately this creates a breaking change: code that could previously use `handleInvalidCases` just fine now fails to compile because the module using it does not import `ocean.core.Verify`. :-(

mihails-strasuns-sociomantic:
https://github.com/sociomantic-tsunami/ocean/pull/358

PR #337, opened 2017-11-02
Update documentation of release notes procedure

gavin-norman-sociomantic:


gavin-norman-sociomantic:
Requires https://github.com/sociomantic-tsunami/neptune/pull/117 to be merged.

nemanja-boric-sociomantic (approved):
I think this is fine. I would leave decision to people that are likely to have more/different opinion on this. cc @leandro-lucarella-sociomantic @mihails-strasuns-sociomantic

README.rst
```diff
@@ -81,6 +84,18 @@ submodule update --init``. This will fetch the `Makd
 <https://github.com/sociomantic-tsunami/makd>`_ project in ``submodules/makd``.
 
 
+Submodules
+**********
+
+============================= ===========
+Dependency                     Version
+============================= ===========
+makd                           v1.3.x
```
gavin-norman-sociomantic, line 93: Note: it's possible these dependencies are horribly outdated. (Need automation.)

joseph-wakeling-sociomantic: Checking out the branch and typing `git submodule` should give a clear picture of what's required.

gavin-norman-sociomantic: The currently used submodules is different to the minimum required, though, or?

README.rst
```diff
@@ -143,3 +158,10 @@ which are the mandatory steps that users have to do to update to a new version,
 use but will not break any old code, and the **New Features** which are optional
 new features available in the new version that users might find interesting.
 Using them is optional, but encouraged.
+
+Contributing
+============
+
+See the guide for `contributing to Neptune-versioned libraries
+<https://github.com/sociomantic-tsunami/neptune/blob/master/doc/library-contributor.rst>`_.
```
joseph-wakeling-sociomantic, line 166: While this link covers details, I wonder if we should mention the `relnotes` folder at all in the README.

Perhaps add a paragraph to the end of the `Releases` section along the lines of:

```
Release notes are maintained in the `relnotes` folder of the source tree,
with individual files for each individual change.
```

gavin-norman-sociomantic: My feeling is that that's too much detail, at this level. The `Releases` section is about the final releases in github (i.e. https://github.com/sociomantic-tsunami/ocean/releases). I don't think how those release notes are written is relevant here.

gavin-norman-sociomantic: (Happy to hear arguments to the contrary, though, if others feel differently :)

README.rst
joseph-wakeling-sociomantic: @gavin-norman-sociomantic I'm not sure how we can reasonably enforce minimum-required constraints outside of having the submodules at that level.  How is this done?

README.rst
```diff
@@ -81,6 +84,18 @@ submodule update --init``. This will fetch the `Makd
...
 
+Submodules
+**********
+
+============================= ===========
+Dependency                     Version
+============================= ===========
+makd                           v1.3.x
+libtangort-dmd-dev (for D1)    v1.6.0
+dmd-transitional   (for D2)    2.070.2.s9
```
joseph-wakeling-sociomantic, line 95: `libtango` and `dmd-transitional` are not submodule dependencies.

This should probably be in the `Packages` section ... ?  I would suggest explicitly noting the repository from where these packages can be installed.  We should probably also reference the `dmd1` compiler.

README.rst
gavin-norman-sociomantic: Currently there is no nice way. There's an ongoing discussion about it in an internal repo. I'll send you the link.

README.rst
gavin-norman-sociomantic: Aha, good point. At this juncture, I would defer to the knowledge of someone else ( @mihails-strasuns-sociomantic ? @leandro-lucarella-sociomantic ?) about what exactly would be the best thing to note here.

joseph-wakeling-sociomantic: BTW, do I recall right that `Makd` relies on `rdmd` being present?  We might want to mention that as well.

All things considered, we might want a separate `Compiler/Runtime` dependencies section to cover compilers and runtime/standard libraries.

joseph-wakeling-sociomantic: The other option might be to have a standardized piece of documentation for our D libraries, describing the compilers and runtimes needed, which we can link to from individual project READMEs.

README.rst
mihails-strasuns-sociomantic: This is a tough call. As soon as we actually start testing supported version range, we will need that data to be stored in `.neptune.yml` and not in readme. Because of that I am not even sure if we shall describe in standard something that is going to change in next few months.

mihails-strasuns-sociomantic: Relevant discussion thread https://github.com/sociomantic/backend/issues/331

mihails-strasuns-sociomantic (commented):
I am not sure about specifying dependencies in readme, everything else LGTM

joseph-wakeling-sociomantic:
> As soon as we actually start testing supported version range, we will need that data to be stored in .neptune.yml and not in readme.

If we take the submodule version as the minimum supported version (which is reasonable), then the range can be inferred automatically as between the submodule version, and the latest available release of that submodule within the same major version series.

Is there a reason why we might not want to have the submodule be taken as the minimum supported version?

mihails-strasuns-sociomantic:
> If we take the submodule version as the minimum supported version (which is reasonable)

Nope, this is not true. Some libraries update to later minor versions just to get specific bug fixes if their minimal supported version does not get bug fixes anymore. They still have to stay compilable with that old version though.

gavin-norman-sociomantic:
>As soon as we actually start testing supported version range, we will need that data to be stored in .neptune.yml and not in readme. Because of that I am not even sure if we shall describe in standard something that is going to change in next few months.

I would be happy to not include submodule dependencies in the README.

leandro-lucarella-sociomantic (approved):

gavin-norman-sociomantic:
Any thoughts on simply removing the minimum dependencies from the README? There's a high probability that what's listed isn't even correct, as it has only been maintained by (flaky) human diligence.

nemanja-boric-sociomantic:
Maybe just point user to https://github.com/sociomantic-tsunami/ocean/blob/f422350991c8b15367e9fa4cd859213c8a4c3203/docker/build ? This doesn't include dmd and rdmd, nor runtime, but all non-trivial dependencies are there.

nemanja-boric-sociomantic:
(and they are guaranteed to be completed at least to pass the tests).

gavin-norman-sociomantic:
No, the question was about submodule dependencies.

nemanja-boric-sociomantic:
I was going by the case that README will now describe both tangort and dmd (which are not complete list of package dependencies, as you can see from the docker script) and submodule dependency - makd.

As for the submodule dependencies, simply running `git submodule` after initializing them will suffice. However, that's bad for reason that normally user doesn't initialize submodule's submodules. I don't have a good answer. What I do think is that this needs somehow to be checked by the CI. Maybe `neptune.yml` is the answer?

gavin-norman-sociomantic:
That's what I mean. The current system is really flaky and can only be checked with a manual process. We need to implement something that can be tested automatically.

In the meantime, my suggestion is to just remove the flaky documentation.

README.rst
leandro-lucarella-sociomantic: I guess we can keep it like this until is moved to `.neptune.yml`. Or maybe better start writing it in `.neptune.yml` even if it's not checked yet (although with the risk that we realize get it wrong once doing the implementation and have to change it again anyways).

README.rst
leandro-lucarella-sociomantic: Probably we should point the users to look at the supported versions based on the test matrix, but it's true that normally we probably also support older versions too (by coincidence). Maybe is enough to say exactly that.

leandro-lucarella-sociomantic (approved):
Besides the small comments, LGTM.

mihails-strasuns-sociomantic:
I propose this `.neptune.yml` notation:

```
library: true
d2ready: true
dependencies:
    packages:
    - dmd-transitional: "2.070.2.s*", "2.071.2.s*"
    - libtangort-dmd-dev: "v1.6.0"
    libraries:
    - makd: "v1.6 - v2"
```

So packages section lists exact package versions as used by apt, but libraries section use simplified supported range syntax to be later used by CI.

leandro-lucarella-sociomantic:
Why not ranges for packages too? I feel like if this is not tested, and based on testing, it will get outdated pretty soon anyway...

gavin-norman-sociomantic:
Simpler alternative: https://github.com/sociomantic-tsunami/ocean/pull/396.

mihails-strasuns-sociomantic:
@leandro-lucarella-sociomantic 

> Why not ranges for packages too?

Package versions are less uniform and I am not sure how to implement it - especially considering packages right now can only be installed during image preparation and not via test script.

PR #339, opened 2017-11-03
Add formatting helper for SmartUnions

gavin-norman-sociomantic:
A small idea that just occurred to me.

### DIFF
--- a/relnotes/smartunionformatter.feature.md
+++ b/relnotes/smartunionformatter.feature.md
@@ -1,0 +1,8 @@
+### New `SmartUnion` formatting function
+
+`ocean.text.formatter.SmartUnion`
+
+The helper function `asActiveField` accepts a smart-union and returns a struct
+with a `toString` method, suitable for passing to the `Formatter`. This formats
+the value of the active field (if any), and (optionally) its name.
+
--- a/src/ocean/core/SmartUnion.d
+++ b/src/ocean/core/SmartUnion.d
@@ -3,6 +3,10 @@
     Template for a union that knows its active field and ensures the active
     field is read.
 
+    See_Also:
+        ocean.text.formatter.SmartUnion -- for helper functions to format a
+        SmartUnion to a string
+
     Copyright:
         Copyright (c) 2009-2016 Sociomantic Labs GmbH.
         All rights reserved.
--- a/src/ocean/text/formatter/SmartUnion.d
+++ b/src/ocean/text/formatter/SmartUnion.d
@@ -1,0 +1,184 @@
+/*******************************************************************************
+
+    Helper function for formatting a SmartUnion's active member to a string.
+
+    This function is not placed in the SmartUnion class itself so as to avoid
+    spreading the dependency on ocean.text.convert.Formatter throughout the
+    codebase.
+
+    Copyright:
+        Copyright (c) 2018 Sociomantic Labs GmbH.
+        All rights reserved.
+
+    License:
+        Boost Software License Version 1.0. See LICENSE_BOOST.txt for details.
+        Alternatively, this file may be distributed under the terms of the Tango
+        3-Clause BSD License (see LICENSE_BSD.txt for details).
+
+*******************************************************************************/
+
+module ocean.text.formatter.SmartUnion;
+
+import ocean.transition;
+import ocean.core.SmartUnion;
+import ocean.core.Verify;
+import ocean.text.convert.Formatter;
+
+/*******************************************************************************
+
+    Gets a wrapper struct with a toString method to format the provided smart
+    union.
+
+    Params:
+        SU = type of smart union to get formatting wrapper for
+        smart_union = smart union to get formatting wrapper for
+        include_name = flag to toggle formatting of the name of the active union
+            member. If true, format output will look like
+            "<activename>: value". If false, format output will look like
+            "value"
+
+*******************************************************************************/
+
+public SmartUnionFormatter!(SU) asActiveField ( SU ) ( SU smart_union,
+    bool include_name = false )
+{
+    return SmartUnionFormatter!(SU)(smart_union, include_name);
+}
+
+///
+unittest
+{
+    union U
+    {
+        int i;
+        cstring s;
+    }
+
+    SmartUnion!(U) su;
+
+    // format from ocean.text.convert.Formatter
+    format("union active member: {}", asActiveField(su));
+}
+
+/*******************************************************************************
+
+    Struct that wraps a smart union with a toString method, for Formatter
+    compatibility.
+
+    Params:
+        SU = type of smart union to format
+
+*******************************************************************************/
+
+private struct SmartUnionFormatter ( SU )
+{
+    import ocean.core.Traits : TemplateInstanceArgs;
+
+    static assert(is(TemplateInstanceArgs!(SmartUnion, SU)));
+
+    mixin TypeofThis;
+
+    /// Smart union to format.
+    private SU smart_union;
+
+    /// Flag to toggle formatting of the name of the active union member. If
+    /// true, format output will look like "<activename>: value". If false,
+    /// format output will look like "value".
+    private bool include_name;
+
+    /// Formatting sink delegate, passed to toString and used by
+    /// formatUnionMember.
+    static private void delegate ( cstring chunk ) sink;
+
+    /***************************************************************************
+
+        Formats the smart union as a string to the provided sink delegate
+        (suitable for use with ocean.text.convert.Formatter). The formatted text
+        depends on whether a field of the union is active or not and whether
+        the `include_name` member is true or false:
+            * No union field active; `include_name` is:
+                - true:  "<none>".
+                - false: "".
+            * Union field active; `include_name` is:
+                - true:  "<name>: value".
+                - false: "value".
+
+        Params:
+            sink = formatter sink delegate to use
+
+    ***************************************************************************/
+
+    public void toString ( void delegate ( cstring chunk ) sink )
+    {
+        if ( this.smart_union.active )
+        {
+            if ( this.include_name )
+                sformat(sink, "<{}>: ", this.smart_union.active_name);
+
+            This.sink = sink;
+            scope ( exit ) This.sink = null;
+
+            callWithActive!(formatUnionMember)(this.smart_union);
+        }
+        else
+        {
+            if ( this.include_name )
+                sformat(sink, "<{}>", this.smart_union.active_name);
+        }
+    }
+
+    /***************************************************************************
+
+        Formats the specified argument with the static sink delegate.
+
+        Note: this function has to be public, so that it can be accessed by the
+        callWithActive function in ocean.core.SmartUnion. It is not intended to
+        be called by users.
+
+        Params:
+            T = type of union member to format
+            union_member = union member to format
+
+    ***************************************************************************/
+
+    // FIXME_IN_D2: `package ocean.core`
+    static public void formatUnionMember ( T ) ( T union_member )
+    {
+        verify(This.sink !is null);
+        sformat(sink, "{}", union_member);
+    }
+}
+
+version ( UnitTest )
+{
+    import ocean.core.Test;
+}
+
+unittest
+{
+    union U
+    {
+        int i;
+        cstring s;
+    }
+
+    SmartUnion!(U) su;
+
+    // Inactive union.
+    test!("==")(format("{}", asActiveField(su)), "");
+
+    // Inactive union, plus name formatting.
+    test!("==")(format("{}", asActiveField(su, true)), "<none>");
+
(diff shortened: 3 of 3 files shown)


gavin-norman-sociomantic:
Notes:
1. I wonder whether we should format some information about the smart union itself, rather than just about the active field. e.g. (using the union in the unittest) instead of `"23"`, we could format something like `"SmartUnion: s1: 23"`. Thoughts?
2. Needs release notes, if we decide to merge this.

gavin-norman-sociomantic:
>I wonder whether we should format some information about the smart union itself, rather than just about the active field. e.g. (using the union in the unittest) instead of "23", we could format something like "SmartUnion: s1: 23".

Actually, I don't think we should do this. The name of the active field can already be accessed, via the `active_name` method, so if the user wants to format this information, they can do it themselves. e.g. `sformat(sink, "{}: {}", u.active_name, u)`.

mihails-strasuns-sociomantic:
I also don't like addition of non-trivial `toString` methods as those make everything depend on formatter (and `SmartUnion` is a primitive core module on its own). See also https://github.com/sociomantic-tsunami/ocean/pull/330

gavin-norman-sociomantic:
>I also don't like addition of non-trivial toString methods as those make everything depend on formatter (and SmartUnion is a primitive core module on its own). See also #330

I see what you mean. I don't really like the solution of having to have external functions to connect a container to the formatter :/ A third way would be to build support for containers into the formatter itself.

None of the solutions really seems ideal.

mihails-strasuns-sociomantic:
> I don't really like the solution of having to have external functions to connect a container to the formatter

It is the only reasonable option from the module system PoV I can think of. It may feel confusing from OOP mindset, but OOP sucks anyway so no big loss. `toString` is inherently leaky abstraction as there are always many useful ways to format non-trivial data structure.

Note that fomatter will be able to print most structs without any `toString` method, you only use wrappers to specify some specific pretty format.

gavin-norman-sociomantic:
Where would you suggest locating such a function? I see that, in #330, the formatting function (`asPrettyStr`) isn't in the same module as the container that it formats info about (`Time`). How is the user supposed to know this? It doesn't seem very user-friendly to me.

gavin-norman-sociomantic:
Ah I see. The container is in `ocean.time.Time` and the function to format it is in `ocean.text.convert`. That does make sense.

gavin-norman-sociomantic:
So the suggestion would be then to add `ocean.text.convert.SmartUnion`. (I don't like the package name `convert`, by the way...)

mihails-strasuns-sociomantic:
I see two options:

1) Keep using current structure and add more formatter modules to `ocean.text.convert` (like you have described). In that case moving all new formatting code + wrappers to new `ocean.text.formatter` package could be a way to make structure cleaner.

2) Start using convention of `ocean.core.SmartUnion` and `ocean.core.SmartUnionFormatting` being two modules side by side in the same package.

Former is better overall structure, latter is easier to spot immediately.

gavin-norman-sociomantic:
I'd go for 1. It's not so immediately obvious, but is a much cleaner structure and is easy once you're used to the idea. We can, of course, add documentation in the container packages saying "see also ocean.text.formatter".

mihails-strasuns-sociomantic:
Ok, please add to `ocean.text.convert` for now in that case and create separate issue about `formatter` package.

gavin-norman-sociomantic:
~~Ahh, of course, in this case (`SmartUnion`), it doesn't work. The formatting code requires access to private internals.~~ I think I can make it work.

nemanja-boric-sociomantic:
But what about this:

> Actually, I don't think we should do this. The name of the active field can already be accessed, via the active_name method, so if the user wants to format this information, they can do it themselves. e.g. sformat(sink, "{}: {}", u.active_name, u).

?

gavin-norman-sociomantic:
That was about printing `SmartUnion: s1: 23` vs `23`.

gavin-norman-sociomantic:
Updated. (Still no release notes. Will add before finalising.)

gavin-norman-sociomantic:
Updated: D2 stomping prevention fix.

mathias-lang-sociomantic:
> A small idea that just occurred to me.

That's what I suggested when you originally implemented it in swarm BTW :D
I think this approach removes most of the convenience of the formatter though, as you have to explicitly say you want a "pretty" string. It also completely kills generic code.

gavin-norman-sociomantic:
>I think this approach removes most of the convenience of the formatter though, as you have to explicitly say you want a "pretty" string. It also completely kills generic code.

I feel the same way.

gavin-norman-sociomantic:
>That's what I suggested when you originally implemented it in swarm BTW :D

I don't remember why I didn't follow your suggestion then...

gavin-norman-sociomantic:
@mathias-lang-sociomantic the "third way" would be to add support for smart union formatting _inside_ `Formatter`. Your thoughts on that?

mathias-lang-sociomantic:
@gavin-norman-sociomantic : I would be very opposed to it. It sounds very backward to modify a formatter to support a type, because there's many more type to support than formatting method.

The `toString` approach was specifically designed to not depend directly on the Formatter. The reason one has to import `Formatter` is because we need to convert a type to its string representation (like int => string). In the case of `SmartUnion`, you want a generic method, so there's no way around the Formatter.

However, I added my notes on how we can address the perceived problem of the Formatter being a dependency [in the related issue](https://github.com/sociomantic-tsunami/ocean/issues/340#issuecomment-341707137).

src/ocean/text/convert/SmartUnion.d
```diff
@@ -0,0 +1,196 @@
...
+        callWithActive function in ocean.core.SmartUnion. It is not intended to
+        be called by users.
+
+        Params:
+            T = type of union member to format
+            union_member = union member to format
+
+    ***************************************************************************/
+
+    static public void formatUnionMember ( T ) ( T union_member )
```
nemanja-boric-sociomantic: Maybe add comment `// package` here, so we can handle these cases one day.

gavin-norman-sociomantic:
Re: #340, thoughts on changing this PR to add the new module in `ocean.text.formatter` already?

gavin-norman-sociomantic:
Updated:
* Rebased.
* Added `package`.
* Fixed TODO and incorrect doc.
* Added release notes.

relnotes/smartunionformatter.feature.md
```diff
@@ -0,0 +1,8 @@
+### New `SmartUnion` formatting function
+
+`ocean.text.convert.SmartUnion`
+
+The helper function `formatActive` accepts a smart-union and returns a struct
```
mihails-strasuns-sociomantic: Maybe `asActive` to match https://github.com/sociomantic-tsunami/ocean/blob/v3.x.x/src/ocean/text/convert/DateTime_tango.d#L968 ?

mihails-strasuns-sociomantic:
> thoughts on changing this PR to add the new module in ocean.text.formatter already?

Depends on how fast would you like to have this functionality. Considering how long this PR was rotting, it is probably not hat urgent and we can proceed with `formatter` already.

relnotes/smartunionformatter.feature.md
gavin-norman-sociomantic: Hmm... that's not a name that makes any sense to the casual reader.

mihails-strasuns-sociomantic: Well, having `format` as part of name is definitely redundant and confusing as you can only ever use this wrapper with `(s)format` calls. It affects representation, not actually formats stuff.

I am not sure what would be the best naming pattern though.

gavin-norman-sociomantic: I'd suggest simply `activeField`, so the usage would look like:
```D
sformat(buf, "union active member: {}", activeField(su));
```

mihails-strasuns-sociomantic: How would you call the time one? I'd prefer to have consistent naming style for all formatter wrappers if possible.

gavin-norman-sociomantic: Maybe:
```D
sformat(buf, "the time is: {}", prettyString(time));
```

mihails-strasuns-sociomantic: Hm, how is that any clearer than `asActiveField` and `asPrettyString` though? From your naming one may expect that field/string gets returned eagerly and passed to formatter after that. At least that is how I read it.

gavin-norman-sociomantic: I don't really feel the `as` part adds much meaning, but I'm fine with calling it `asActiveField`, for the sake of consistency.

gavin-norman-sociomantic:
Updated:
* Changed function name to `asActiveField`.
* Moved to `ocean.text.formatter`.

relnotes/smartunionformatter.feature.md
mihails-strasuns-sociomantic: I find it useful to clearly denote "eager vs lazy" in either module names or function names, similar to how there is `split` and `splitter` in Phobos (ignoring stupid name choice itself :P)

src/ocean/text/formatter/SmartUnion.d
```diff
@@ -0,0 +1,205 @@
...
+
+    mixin TypeofThis;
+
+    /// Smart union to format.
+    private SU smart_union;
+
+    /// Flag to toggle formatting of the name of the active union member. If
+    /// true, format output will look like "<activename>: value". If false,
+    /// format output will look like "value".
+    private bool include_name;
```
mihails-strasuns-sociomantic, line 87: Is there really any reason to mark field private here, especially configuration ones?

gavin-norman-sociomantic: They're not intended to be accessible to the caller of `asActiveField`. I agree that modifying this config field in particular would be harmless, but it's not the intended usage of this module.

mihails-strasuns-sociomantic (approved):
LGTM! Would be nice to check this doesn't contradict whatever @mathias-lang-sociomantic had in mind too much.

src/ocean/core/SmartUnion.d
```diff
@@ -3,6 +3,10 @@
     Template for a union that knows its active field and ensures the active
     field is read.
 
+    See_Also:
+        ocean.text.convert.SmartUnion -- for helper functions to format a
```
mathias-lang-sociomantic: `s/convert/formatter`

src/ocean/text/formatter/SmartUnion.d
```diff
@@ -0,0 +1,205 @@
...
+    {
+        int i;
+        cstring s;
+    }
+
+    SmartUnion!(U) su;
+
+    mstring buf;
+    // sformat from ocean.text.convert.Formatter
+    sformat(buf, "union active member: {}", asActiveField(su));
```
mathias-lang-sociomantic: The output is not tested ?
Also might as well use `format` and remove the buffer.

src/ocean/text/formatter/SmartUnion.d
```diff
@@ -0,0 +1,205 @@
...
+        cstring s;
+    }
+
+    SmartUnion!(U) su;
+
+    // Inactive union.
+    {
+        mstring buf;
+        sformat(buf, "{}", asActiveField(su));
+        test!("==")(buf, "");
```
mathias-lang-sociomantic: No need to use a `buf` if you don't reuse it, just call `format` instead.

src/ocean/text/formatter/SmartUnion.d
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    // s active, plus name formatting.
+    {
+        mstring buf;
+        sformat(buf, "{}", asActiveField(su, true));
+        test!("==")(buf, "<s>: hello");
+    }
+}
+
```
mathias-lang-sociomantic: Trailing line

src/ocean/text/formatter/SmartUnion.d
mathias-lang-sociomantic: Also all could be up inline:
```
test!("==")(format("{}", asActiveField(su)), "");
```

This way no need for blocks either.

mathias-lang-sociomantic:
Styling fixes request, but the idea LGTM

src/ocean/text/formatter/SmartUnion.d
gavin-norman-sociomantic: It's a usage example, not a test.

gavin-norman-sociomantic:
Updated with tweaks suggested by @mathias-lang-sociomantic.

mihails-strasuns-sociomantic (approved):

gavin-norman-sociomantic:
Hmm no status from Travis...? I'll rebase and repush.

PR #343, opened 2017-11-03
Minor cleanup on multiple definitions of the same function

mathias-lang-sociomantic:
Just spotted I could remove 500 LoC just by using an import, figured it was a good deal.


src/ocean/text/utf/UtfUtil.d
```diff
@@ -34,13 +34,10 @@
 
 module ocean.text.utf.UtfUtil;
 
-
```
gavin-norman-sociomantic, line 37: It'd really help review if you didn't meld whitespace changes in with other types of change.

gavin-norman-sociomantic (approved):
Ha, well spotted!

mihails-strasuns-sociomantic (commented):
It looks like there are not test cases for unicode errors in these modules  - would it be simple enough to copy some from phobos?

mathias-lang-sociomantic:
The only test I could find was: https://github.com/dlang/druntime/blob/3485ff859a29ba44e7949bc49e62d5dd3a2a9ff0/src/rt/lifetime.d#L2070-L2096
Is it really worth it ?

mihails-strasuns-sociomantic:
I was thinking about something like https://github.com/dlang/phobos/blob/feab9a2f5225d7f3a5179f24ae3699e64455807c/std/utf.d#L1532-L1555 but if you feel it is too much trouble, nevermind.

mathias-lang-sociomantic:
Given this code is never ever touched, I'd say let's skip it. I'd rather spend time of getting rid of this code than testing it :)

mihails-strasuns-sociomantic:
Reasonable enough :)

PR #344, opened 2017-11-03
Minor fixes

mathias-lang-sociomantic:
Maybe that glib definition should be *somewhere* as it's also used in swarm: https://github.com/sociomantic-tsunami/swarm/blob/f7f9e268dc9501c08cff328498ff366ecadc0f15/src/swarm/neo/authentication/CredentialsFile.d#L53


mihails-strasuns-sociomantic (approved):
What is the problem with keeping it in swarm as is?

mathias-lang-sociomantic:
No problem, but we have a few glib definitions scattered here and there, so it might be better to group them... One day.

PR #345, opened 2017-11-03
 Make ZLIB_VERSION, Z_NULL, and ZLIB_VERNUM mangled constants

mathias-lang-sociomantic:
Before it was an extern(C) TLS variable, which clashed with Phobos which had the exact same code.

Fixed it in Phobos as well: https://github.com/dlang/phobos/pull/5844

### DIFF
--- a/src/ocean/util/compress/c/zlib.d
+++ b/src/ocean/util/compress/c/zlib.d
@@ -23,14 +23,14 @@ module ocean.util.compress.c.zlib;
 
 import ocean.transition;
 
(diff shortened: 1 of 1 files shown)


mathias-lang-sociomantic:
Updated to include `Z_NULL` (and another issue with Phobos found)

mathias-lang-sociomantic:
The ~ocean~ Phobos side of things was merged

mihails-strasuns-sociomantic (approved):

PR #347, opened 2017-11-09
Fix the broken D2 conversion of PullParser

don-clugston-sociomantic:
PullParser was converted to D2 incorrectly. Internally, the code does not modify the text which is being parsed. But the D2 conversion incorrectly used an `mstring` to store the text being parsed. This required a memory allocation for every cstring which was parsed, making the code essentially
unusable in D2.

The `name` and `value` member functions were incorrectly marked as returning mutable strings.

I'm not sure how the release notes should be written. This PR does not change the D1 behaviour in any way, only the (currently broken) D2 version is affected.

### DIFF
--- a/src/ocean/text/xml/Document.d
+++ b/src/ocean/text/xml/Document.d
@@ -2193,6 +2193,22 @@ unittest
 
     auto complex_result = printer.print(doc);
     test!("==")(complex_result, complex_doc);
+
+    // Check node renaming
+
+    auto renamed_doc = `
+<VAST version="3.0">
+  <InLine>
+    <AdTitle>VAST 3.0 Instream Test</AdTitle>
+    <Creatives>
+      <RenamedCreative id="123456" adId="654321"/>
+    </Creatives>
+  </InLine>
+</VAST>`;
+
+    creative.name("RenamedCreative");
+    auto renamed_result = printer.print(doc);
+    test!("==")(renamed_result, renamed_doc);
 }
 
 
--- a/src/ocean/text/xml/Document.d
+++ b/src/ocean/text/xml/Document.d
@@ -20,6 +20,7 @@ module ocean.text.xml.Document;
 import Array = ocean.core.Array;
 import ocean.transition;
 import ocean.core.Verify;
+import ocean.util.container.ConcatBuffer;
 
 package import ocean.text.xml.PullParser;
 
@@ -166,6 +167,15 @@ class Document(T_) : PullParser!(T_)
                                 freelists;
         private XmlPathT     xpath;
 
+
+        /***********************************************************************
+
+                Buffer for storing the text of modified nodes
+
+        ***********************************************************************/
+
+        private ConcatBuffer!(T_) node_text_buff;
+
         /***********************************************************************
 
                 Construct a DOM instance. The optional parameter indicates
@@ -183,6 +193,7 @@ class Document(T_) : PullParser!(T_)
                 newlist;
                 root = allocate;
                 root.id = XmlNodeType.Document;
+                this.node_text_buff = new ConcatBuffer!(T_);
         }
 
         /***********************************************************************
@@ -260,6 +271,7 @@ else
 }
                 newlist;
                 index = 1;
+                node_text_buff.clear();
 version(d)
 {
                 freelists = 0;          // needed to align the codegen!
@@ -428,12 +440,9 @@ else
                 p.lastChild =
                 p.firstAttr =
                 p.lastAttr = null;
-                p.rawValue.length = 0;
-                enableStomping(p.rawValue);
-                p.localName.length = 0;
-                enableStomping(p.localName);
-                p.prefixed.length = 0;
-                enableStomping(p.prefixed);
+                p.rawValue = null;
+                p.localName = null;
+                p.prefixed = null;
 }
                 return p;
         }
@@ -610,9 +619,9 @@ version (Filter)
                 public void*            user;           /// open for usage
                 package Document        doc;            // owning document
                 package XmlNodeType     id;             // node type
-                package MutT[]          prefixed;       // namespace
-                package MutT[]          localName;      // name
-                package MutT[]          rawValue;       // data value
+                package T[]             prefixed;       // namespace
+                package T[]             localName;      // name
+                package T[]             rawValue;       // data value
 
                 package Node            host,           // parent node
                                         prevSibling,    // prior
@@ -710,7 +719,7 @@ version (Filter)
 
                 Node prefix (T[] replace)
                 {
-                        Array.copy(this.prefixed, replace);
+                        this.prefixed = doc.node_text_buff.add(replace);
                         return this;
                 }
 
@@ -733,7 +742,7 @@ version (Filter)
 
                 Node name (T[] replace)
                 {
-                        Array.copy(this.localName, replace);
+                        this.localName = doc.node_text_buff.add(replace);
                         return this;
                 }
 
@@ -771,7 +780,7 @@ version(discrete)
                                      if (child.id is XmlNodeType.Data)
                                          return child.value (val);
 }
-                        Array.copy(this.rawValue, val);
+                        this.rawValue = doc.node_text_buff.add(val);
                         mutate;
                 }
 
@@ -1014,7 +1023,7 @@ version(discrete)
 }
 else
 {
-                        Array.copy(node.rawValue, value);
+                        this.rawValue = doc.node_text_buff.add(value);
 }
                         return node;
                 }
@@ -1163,8 +1172,8 @@ else
 
                 private Node set (T[] prefix, T[] local)
                 {
-                        Array.copy(this.localName, local);
-                        Array.copy(this.prefixed, prefix);
+                        this.localName = doc.node_text_buff.add(local);
+                        this.prefixed = doc.node_text_buff.add(prefix);
                         return this;
                 }
 
@@ -1177,7 +1186,7 @@ else
                 private Node create (XmlNodeType type, T[] value)
                 {
                         auto node = document.allocate;
-                        Array.copy(node.rawValue, value);
+                        node.rawValue = doc.node_text_buff.add(value);
                         node.id = type;
                         return node;
                 }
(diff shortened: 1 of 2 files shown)


don-clugston-sociomantic:
The problem is much larger, all of the `xml` modules are affected by this error.

gavin-norman-sociomantic:
>The name and value member functions were incorrectly marked as returning mutable strings.

Changing a return value from mutable to constant (i.e. reducing permissiveness) sounds like a breaking API change, unfortunately. (Vice versa would not be a breaking change.) I don't think the fact that this change only affects D2 builds should make any difference, from a strict SemVer perspective. Perhaps others might have different angles on this.

One workaround (to make this a new feature, rather than a breaking change) that comes to mind:
* Keep the existing `name` and `value` methods returning mutable strings, via `dup`.
* Add new methods that return constant strings.

gavin-norman-sociomantic:
Another way, if the problem is really widespread, might be to duplicate the whole modules, fix the const-ness as required, and deprecate the old modules (to be removed in the next major release).

don-clugston-sociomantic:
> One workaround (to make this a new feature, rather than a breaking change) that comes to mind:
*    Keep the existing name and value methods returning mutable strings, via dup.
*    Add new methods that return constant strings.

Yeah, I'm not using those functions anyway, so I've pushed with that change. Essentially the memory allocation that previously happened in the constructor, now happens in those functions instead.
So it's still broken, but the brokenness is now isolated to those functions.

gavin-norman-sociomantic:
Cool. You didn't add new const getters for those values. Are those not really necessary? i.e. is there already another way access the values?

gavin-norman-sociomantic:
By the way, without adding the const getters, this is a bug-fix, not even a new feature.

src/ocean/text/xml/PullParser.d
```diff
@@ -543,7 +543,7 @@ version (partialwhite)
 
         final Ch[] value()
         {
-                return rawValue;
+                return rawValue.dup;
```
nemanja-boric-sociomantic: Perhaps adding a note in docs about allocating behavior vs. accessing the field will be useful here.

don-clugston-sociomantic:
I was wrong, I do need the `name` and `value` members, and they do need to be const.

The problem is that `xml.Document` derives from this class, so it needs changes as well. And it is a mess. It has slices which sometimes point into the original document, but sometimes point into mutable buffers. It has some `enableStomping` calls which make the class look as though it avoids memory allocation, but actually it doesn't. 

Total mess.

src/ocean/text/xml/PullParser.d
don-clugston-sociomantic: I've changed my mind, I am 100% certain that is the wrong approach.

That's actually a severe breaking change (much more significant than making the return value const). And it would actually break D1 as well!

gavin-norman-sociomantic: Why is this a breaking change?

don-clugston-sociomantic: Because it means that previously working D1 code would leak memory like crazy.

We are not currently using `@nogc` but I think that we should consider all existing code which doesn't allocate memory as having an implicit `@nogc` on it. Changes which allocate memory should be considered as a breaking change.

gavin-norman-sociomantic: That's a pretty interesting point. Of course, changes that introduce new allocations are avoided, but I don't think they're strictly defined as breaking changes (from a semver perspective). But yeah, _if_ anyone is using these methods, it'd be a highly annoying change.

don-clugston-sociomantic: It's like the 'no array stomping' change in D. The idea that it's not a breaking change, is just wrong. I think given the importance of memory efficiency to our libraries, we need to have this extra nuance compared to normal semver. But note that if we were using `@nogc` it would actually be normal semver, I think.

In practice, it is important: if this change is made, then you need to change your code, quite extensively. Although existing code will compile and passes the unit tests, it will fail burn-in tests.

mihails-strasuns-sociomantic:
@don-clugston-sociomantic it would help a great deal if you could add some tests with `testNoAlloc` that are based on your test case (and will fail with current ocean HEAD). After that I will be able re-read whole package with this use case in mind and possibly suggest how to best handle migration.

don-clugston-sociomantic:
Very basic test added, which fails in existing ocean.

I have just noticed that `localName` and `rawValue` already exist, and do not involve memory allocation. So for my purposes, I could use those. In fact the existing `name` function could allocate even in D1, in the case where a prefix was set, because it uses concatenation.

But I don't think that the `name` and `value` functions should suddenly start allocating memory in D1.

don-clugston-sociomantic:
I have repushed this PR, so that it no longer changes the return types, and does not perform any extra memory allocations in D1.
I think that the correct path forward would be to deprecate and then remove the `value` function (since it's identical to `rawValue`, except that it has this buggy memory allocation), and likewise replace the `name` function with one which accepts a buffer to store the name into (since it is not possible to entirely avoid memory allocation, even if it uses a `const` return value).

So I have changed the `PullParser` unittest so that it uses `rawValue` instead.

src/ocean/text/xml/Document.d
```diff
@@ -610,9 +607,14 @@ version (Filter)
...
                 package XmlNodeType     id;             // node type
-                package MutT[]          prefixed;       // namespace
-                package MutT[]          localName;      // name
-                package MutT[]          rawValue;       // data value
+
+                private MutT[]          prefixedBuff;   // namespace buffer
+                private MutT[]          localNameBuff;  // name buffer
+                private MutT[]          rawValueBuff;   // data value buffer
+
+                package T[]             prefixed;       // namespace
```
mihails-strasuns-sociomantic: I looked into this a bit more carefully on a local checkout and have one immediate question - is there any reason why these 3 values have to be slices? Alternative would be to actually copy data into them, i.e. ` copy(rawValue, q [0 .. p - q])`. It feels a bit weird to have both slice and a buffer.

don-clugston-sociomantic: Yes, there is. The reason is that the slices are usually a slice into the original, const string and are extremely lightweight.
But, if the values are modified, there needs to be a buffer where the new values can be stored.
If we use the same member for both slice and buffer, the buffers won't be reused.

I think the original code was quite sloppy about this.

don-clugston-sociomantic: A nicer design might be to create a `ConcatBuffer` in the `Document`, and use that single buffer to store all of the modified pieces. But that's a rather more intrusive change in this PR, where I'm really just trying to fix a mistake in the D2 conversion of `PullParser`. I've never used `Document` so I'd prefer to minimize changes to it.

mihails-strasuns-sociomantic: I am suspicious that performance penalty for copying each of those fields into buffer is negligible on the scale of the whole XML parser. Would be interesting to benchmark.

But you are right of course that right now priority should be fixing original porting with minimal involved changes.

src/ocean/text/xml/PullParser.d
```diff
@@ -94,9 +94,9 @@ public enum XmlTokenType {Done, StartElement, Attribute, EndElement,
 class PullParser(Ch = char)
 {
         public int                      depth;
-        public Ch[]                     prefix;
-        public Ch[]                     rawValue;
-        public Ch[]                     localName;
+        public Const!(Ch)[]             prefix;
+        public Const!(Ch)[]             rawValue;
+        public Const!(Ch)[]             localName;
```
mihails-strasuns-sociomantic, line 99: Intention was that if you want to parse immutable data, you instantiate `PullParser!(Const!(char))`, not `PullParser!(char)`. Thus buffer code should use `MutT`, not the other way around.

src/ocean/text/xml/PullParser.d
```diff
@@ -539,24 +539,44 @@ version (partialwhite)
...
+                use rawValue instead.
+
         ***********************************************************************/
 
         final Ch[] value()
         {
-                return rawValue;
+                version (D_Version2)
+                {
+                    return rawValue.dup;
```
mihails-strasuns-sociomantic, line 551: Please don't introduce such performance differences between D1 and D2 intentionally - it makes profiling much harder later. If `.dup` is needed, it should be used on both.

However, if you keep in mind comment about `PullParser!(Const!(char))`, it may look like your original assumption mentioned in the commit body is wrong - this method does not inherently return mutable string.

src/ocean/text/xml/Document.d
```diff
@@ -260,6 +271,7 @@ else
 }
                 newlist;
                 index = 1;
+                node_text_buff.clear();
```
mihails-strasuns-sociomantic, line 274: Clearing concat buffer only when whole document is reset is not a very memory friendly pattern :(

mihails-strasuns-sociomantic (changes requested):
Going to check what is the problem with `PullParser!(Const!(char))`

mihails-strasuns-sociomantic:
I think this module has a much more fundamental issue. This code was originally written with assumption that `Ch` is always used with correct qualifier, not plain `char`: https://github.com/sociomantic-tsunami/ocean/blob/6bd4f1dea36003b6b4eb739ebab127aca2d88682/src/ocean/text/xml/PullParser.d#L94-L99

However here things go plain insane - https://github.com/sociomantic-tsunami/ocean/blob/6bd4f1dea36003b6b4eb739ebab127aca2d88682/src/ocean/text/xml/PullParser.d#L586-L608

It looks like during porting I forgot what I was doing half the way through and switched porting model. I would like to revisit the whole module myself if you don't mind waiting 1 or 2 days more.

src/ocean/text/xml/Document.d
don-clugston-sociomantic: I may be misreading the code, but I think that's no different to what happens already. This is the only place where nodes ever get added back to the freelist, as far as I can tell.

mihails-strasuns-sociomantic: AFAIR in your previous version there was a fixed buffer per field which got overwritten each time `name` was called (invalidating old references).

don-clugston-sociomantic:
>  It looks like during porting I forgot what I was doing half the way through and switched porting model. I would like to revisit the whole module myself if you don't mind waiting 1 or 2 days more.

OK. The practical problem is just that `.dup` in the constructor, which is a blocker for D2 migration. If it's possible to get rid of it by doing something more minimal, that'd be great.

mihails-strasuns-sociomantic:
Looking through git history I have found something interesting - it wasn't result of original porting that those fields stopped being slices, but 870e6115b43b9aa22940366da5d8cab921041b29

I have created https://github.com/sociomantic-tsunami/ocean/pull/372 as a simpler PR which continues with 870e6115b43b9aa22940366da5d8cab921041b29 approach, could you check if amount of memory it allocates is acceptable for you?

don-clugston-sociomantic:
That's very interesting. I personally don't use `Document` at all, so I can't really comment on the memory usage. It is only the memory allocation in `PullParser.reset` which matters to me, and that's fixed in your #372 PR. So I'm closing this one.

PR #348, opened 2017-11-09
Convert unreleased release notes to new format

gavin-norman-sociomantic:
See https://github.com/sociomantic-tsunami/neptune/blob/master/doc/library-contributor.rst

### DIFF
--- a/relnotes/stream-stuff.deprecation.md
+++ b/relnotes/stream-stuff.deprecation.md
@@ -1,3 +1,5 @@
-* `ocean.io.stream.Digester`, `ocean.io.stream.Endian`, `ocean.io.stream.Greedy`, `ocean.io.stream.Map`, `ocean.io.stream.Typed`, `ocean.io.stream.Utf`
+## Deprecate unused `io` modules
 
-  Those module are unused within Sociomantic. They are also  barely tested and antiquated, and thus have been deprecated.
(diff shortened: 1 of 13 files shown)


relnotes/stream-stuff.deprecation.md
```diff
@@ -1,3 +1,5 @@
-* `ocean.io.stream.Digester`, `ocean.io.stream.Endian`, `ocean.io.stream.Greedy`, `ocean.io.stream.Map`, `ocean.io.stream.Typed`, `ocean.io.stream.Utf`
+## Deprecate unused modules
```
leandro-lucarella-sociomantic: Maybe ```Deprecate unused `io` modules```? This title as it is kind of overlaps with the whole section title, it doesn't provide any further useful information.

Actually for deprecations it might even make sense to keep the old list format, as usually what you need to know is which modules were deprecated and the grouping of different modules based on how the deprecations were committed is not very important/useful. :thinking:

mihails-strasuns-sociomantic: Not all deprecated modules are unused though

gavin-norman-sociomantic: >Maybe Deprecate unused `io` modules?

:+1:

>Actually for deprecations it might even make sense to keep the old list format

I thought that too, at first, but then noticed that there are different types of deprecation. Despite what we see in this PR, the most common type is actually "Replace X with Y", I'd say, which makes a pretty useful quick summary.

gavin-norman-sociomantic: Updated.

leandro-lucarella-sociomantic (approved):

mihails-strasuns-sociomantic (approved):

PR #349, opened 2017-11-09
Implement `toContextDg` core utility

mihails-strasuns-sociomantic:


david-eckardt-sociomantic:
Another idea, even simpler:
```d
void delegate() toContextDg ( void function ( void* context ) func, void* context )
{
    void delegate() dg;
    dg.funcptr = cast(void function())func;
    dg.ptr = context;
    return dg;
}
```

mihails-strasuns-sociomantic:
I don't think that can work as ABI for calling regular function with one argument is different from delegate ABI.

david-eckardt-sociomantic:
I tested it, it works for me...

david-eckardt-sociomantic:
Fair enough if you don't want to rely on the calling convention for delegates to be the same as functions. I wonder why they should be different, though, but I'm not an ABI expert.

relnotes/context-dg.feature.md
```diff
@@ -0,0 +1,20 @@
...
+{
+    test!("==")(cast(size_t) context, 42);
+}
+
+void delegate() dg = toContextDg!(handler)(cast(void*) 42);
+dg();
+```
+
+This utility is intended for advanced performance optimization is not
+recommended for casual usage.
```
david-eckardt-sociomantic: "This utility is intended for advanced performance optimization. It is not recommended for casual usage."

src/ocean/core/TypeConvert.d
```diff
@@ -245,3 +245,65 @@ unittest
...
+
+    // Makes use of the fact that D ABI allows converting aggregate methods
+    // to delegates such that resulting delegate context pointer becomes
+    // aggregate `this`:
+
+    static struct Fake
+    {
+        void method ( )
+        {
+            void* context = cast(void*) this;
```
david-eckardt-sociomantic, line 282: Needs to be `&this` in D2.

src/ocean/core/TypeConvert.d
```diff
@@ -245,3 +245,65 @@ unittest
...
+
+            // do real work via provided F function:
+            F(context);
+        }
+    }
+
+    void delegate() dg = &Fake.init.method;
```
david-eckardt-sociomantic, line 289: If you could use `&((Fake*).init).method` the delegate will be created with a `null` context pointer.

src/ocean/core/TypeConvert.d
mihails-strasuns-sociomantic: `d1to2fix`

src/ocean/core/TypeConvert.d
mihails-strasuns-sociomantic: DMD seems to crash under certain conditions if you try creating a delegate with `null` context pointer.

mihails-strasuns-sociomantic: I mean, not DMD itself but generated code.

david-eckardt-sociomantic: Yeah, it's fine as it is.

mihails-strasuns-sociomantic:
> Fair enough if you don't want to rely on the calling convention for delegates to be the same as functions. I wonder why they should be different, though, but I'm not an ABI expert.

@don-clugston-sociomantic has mentioned that for delegates it is specifically in the `ECX`. For functions it does depends on what compiler you use and what is the type of the argument. In this case it is most likely passed via same register too, but it could be also on stack (D doesn't have standard/fixed ABI).

david-eckardt-sociomantic:
> (D doesn't have standard/fixed ABI)

What?!? I thought this was a key point for the design of D, to have a standard ABI which C++ doesn't have.

david-eckardt-sociomantic:
LGTM except for the one comment about documentation grammar.

mihails-strasuns-sociomantic:
> What?!? I thought this was a key point for the design of D, to have a standard ABI which C++ doesn't have.

Nope, AFAIK you can't even link libs built with different versions of DMD, not even mentioning LDC/GDC.

mihails-strasuns-sociomantic:
There is https://dlang.org/spec/abi.html but it only defines some very basic bits.

mihails-strasuns-sociomantic:
Updated and split whitespace change into separate commit

david-eckardt-sociomantic (approved):

relnotes/context-dg.feature.md
```diff
@@ -0,0 +1,20 @@
+# New utility to forge zero-allocation closures with one argument
```
gavin-norman-sociomantic, line 1: Should be `##` (heading 2), by the way.

gavin-norman-sociomantic (commented):
Sneaky :D

src/ocean/core/TypeConvert.d
```diff
@@ -245,3 +245,65 @@ unittest
...
+
+*******************************************************************************/
+
+void delegate() toContextDg ( alias F ) ( void* context )
+{
+    static assert (is(typeof(&F) == void function (void*)));
+
+    // Makes use of the fact that D ABI allows converting aggregate methods
+    // to delegates such that resulting delegate context pointer becomes
+    // aggregate `this`:
```
don-clugston-sociomantic, line 276: It makes uses of a second fact, as well, which you use a bit later.

Suggested wording:
``
This code makes use of two facts:
(1) The D ABI allows aggregate methods to be converted to delegates, such that the delegate context pointer
becomes the `this` pointer of the aggregate; and 
(2) The compiler supports explicit modification of the .ptr member of a delegate, without modifying the existing .functptr.
``

The second fact is actually pretty interesting, because it means the compiler knows exactly what you're doing here. The only reason for supporting that syntax in the compiler, is to allow hacks like this. Admittedly the spec doesn't mention that you can do this, but it's explicitly coded in the compiler.

don-clugston-sociomantic (commented):
LGTM.
I think this code is quite safe. My suggested comment would explain why.

src/ocean/core/TypeConvert.d
mihails-strasuns-sociomantic: Created fixup PR https://github.com/sociomantic-tsunami/ocean/pull/350

Issue #341, opened 2017-11-03
Redesign how resuming a `Task` works


mihails-strasuns-sociomantic:
Currently scheduler/task system has a fundamental design flaw which makes it possible to get into fiber memory corruption as soon as several tasks start resuming each other. It comes from misunderstanding the fact that fibers yield don't remember call stack:

```D
import core.thread;
import ocean.io.Stdout;

Fiber f1, f2;

void context1 ( )
{
    Stdout.formatln("-> context1");
    f2 = new Fiber(&context2, 10240);
    f2.call();
    Stdout.formatln("yield from context1");
    Fiber.yield();
    Stdout.formatln("<- context1");
}

void context2 ( )
{
    Stdout.formatln("-> context2");
    Stdout.formatln("yield from context2");
    Fiber.yield();
    Stdout.formatln("<- context2");
}

void main ( )
{
    f1 = new Fiber(&context1, 10240);
    f1.call();
    f2.call();
}
```

```
-> context1
-> context2
yield from context2
yield from context1
<- context2
```

Scheduler was written with an implicit assumption that in such snippet returning end of `context2` would get back to `context1` but this is not how fibers are defined to work - control always returns to whatever context called `call` method.

Fixing this will require non-trivial modifications to scheduler system and sadly can't go in point release. It can't wait for v4.0.0 either because of how critical bug is. I would suggest `v3.6.0+breaking` because of it.

mihails-strasuns-sociomantic:
FYI: I intend to fix it by making so that any resuming of the task will be done by the scheduler/epoll instead of direct `Fiber.call()` call. Will see how well does it go in practice though.

PR #353, opened 2017-11-10
Fix `drop` not working after `unregister`

mihails-strasuns-sociomantic:
Follow-up to b084f6b347fdc9495c74d600ed88924c648e274d which fixes the
problem with expiry reference being cleaned up on call to `unregister`
which is done when timeout hits. That resulted in not being able to
`drop` it later, which killed the very point `drop` was added.

### DIFF
--- a/src/ocean/time/timeout/TimeoutManager.d
+++ b/src/ocean/time/timeout/TimeoutManager.d
@@ -230,19 +230,20 @@ abstract class TimeoutManagerBase : ITimeoutManager
 
         /***********************************************************************
 
-            Unregisters the specified expiry. If the expiry is present in the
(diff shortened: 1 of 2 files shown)


mihails-strasuns-sociomantic:
Tested with a local development branch which re-enables scheduler/timer tests. Seems to fix the problem.

mihails-strasuns-sociomantic:
NB: this is necessary for https://github.com/sociomantic-tsunami/ocean/issues/341

gavin-norman-sociomantic (approved):
I didn't follow the whole story behind this, but the change itself seems fair enough.

nemanja-boric-sociomantic (approved):

PR #355, opened 2017-11-13
[+breaking] Several architectural tweaks to the scheduler

mihails-strasuns-sociomantic:
After some mental experiments I came to conclusion that
https://github.com/sociomantic-tsunami/ocean/issues/341 is not directly a
problem, and what was observed as such, was in fact result of memory corruption
caused by https://github.com/sociomantic-tsunami/ocean/issues/353

However, I also came to conclusion that there are possible other issues if it is
allowed to resume tasks from termination hooks, because it will leave "finished"
tasks running for prolonged amount of time. So I still used original idea to fix
that.

Lot of other changes came as a side effect from digging through the scheduler
sources, most notably introduction of `IScheduler` interface. Fortunately, most
of those don't make any impact on user-facing API so expected breakage impact is
very low. Still, it requires marking next minor release as `+breaking` per Neptune.


relnotes/isheduler.feature.md
```diff
@@ -0,0 +1,9 @@
+* `ocean.task.IScheduler`
+
+  New module defining all basic data types needed to interact with the scheduler
+  API. It is recommended that only module that needs to configure the scheduler
+  would import original `ocean.task.Scheduler` and all others should switch to
+  new light-weight one.
+
+  `ocean.task.IScheduler.theScheduler` will return the very same global
+  scheduler reference but downcast to `IScheduler` interface.
```
gavin-norman-sociomantic: I think it's an upcast actually.

relnotes/isheduler.feature.md
```diff
@@ -0,0 +1,9 @@
+* `ocean.task.IScheduler`
```
gavin-norman-sociomantic: Commit message:
"Provides new IScheduler module which allows to make use of scheduler
facilities without directly important all of involved implementation."

"important" -> "importing"?

relnotes/isheduler.feature.md
```diff
@@ -0,0 +1,9 @@
...
+
+  New module defining all basic data types needed to interact with the scheduler
+  API. It is recommended that only module that needs to configure the scheduler
```
gavin-norman-sociomantic: "modules that need to"

mihails-strasuns-sociomantic:
Updated:
- fix 2.071 build

src/ocean/task/IScheduler.d
```diff
@@ -241,6 +241,19 @@ public interface IScheduler
...
+        Orders scheduler to resume given task unconditionally after current
+        epoll cycle. Should be used instead of plain `Task.resume` from
+        termination hooks of other tasks.
+
+        Params:
+            task = task object to resume on next cycle
+
+    ***************************************************************************/
+
+    public void resume ( Task task );
```
gavin-norman-sociomantic: I feel this method should have a different name. `resume` is so ingrained to mean "switch execution immediately to the thing being resumed" that I think it's deceptive to use the word in a different way here.

Nothing really catchy comes to mind immediately, though. `resumeAfterProcessingEvents`? :/

src/ocean/task/IScheduler.d
```diff
@@ -241,6 +241,19 @@ public interface IScheduler
 
     /***************************************************************************
 
+        Orders scheduler to resume given task unconditionally after current
+        epoll cycle. Should be used instead of plain `Task.resume` from
```
gavin-norman-sociomantic: Is there any way to enforce this?

src/ocean/task/Scheduler.d
```diff
@@ -409,9 +409,12 @@ final class Scheduler : IScheduler
         assert (context !is null);
         assert (context !is task);
 
-        task.terminationHook(&context.resume);
+        // this methods stack is guaranteed to still be valid by the time
+        // task finishes, so we can reference `task` from delegate
+        task.terminationHook({ theScheduler.resume(context); });
```
gavin-norman-sociomantic: This commit seems more like "Implement `IScheduler.resume` method and adapt termination hooks to use it".

relnotes/scheduler-exceptions.migration.md
```diff
@@ -0,0 +1,21 @@
+# New default scheduler exception handler in tests
```
gavin-norman-sociomantic: `##`

relnotes/scheduler-exceptions.migration.md
```diff
@@ -0,0 +1,21 @@
...
+
+Previously, default exception handler in the scheduler was simply rethrowing any
+exception in both test and actual builds. This was most backwards compatible
```
gavin-norman-sociomantic: "simply rethrowing any exception (in both test builds and actual builds)"

relnotes/scheduler-exceptions.migration.md
```diff
@@ -0,0 +1,21 @@
...
+behaviour but with an unfortunate effect that unexpected exceptions in tests
+could be lost and were hard to debug in general.
+
+New default exception handler is defined to:
+
+  1) Print exception information
+  2) Call C `abort` to terminate the process immediately
```
gavin-norman-sociomantic: The way this is worded, it sounds like this is also the case in non-test builds.

relnotes/resume.migration.md
```diff
@@ -0,0 +1,8 @@
+# Disallow direct resume in termination hooks
```
gavin-norman-sociomantic: `##`

src/ocean/task/IScheduler.d
gavin-norman-sociomantic: I see this is done in a later commit.

relnotes/resume.migration.md
```diff
@@ -0,0 +1,8 @@
...
+
+It is now necessary to use `IScheduler.resume` method if one task has to be
```
gavin-norman-sociomantic: This commit should also edit the method comments added previously that said "you shouldn't do this". (It's now enforced, not merely prohibited by convention.)

relnotes/select-fiber-ext.deprecation.md
```diff
@@ -0,0 +1,5 @@
+# Deprecate transitional `SelectFiberSupport` task extenstion
```
gavin-norman-sociomantic: `##` :D

relnotes/select-fiber-ext.deprecation.md
```diff
@@ -0,0 +1,5 @@
...
+
+This extension was intended as a temporary hack to simplify initial switching to
+task system and is known to be rather unreliable/buggy. Currently task system is
+supported wide enough to make this hack unnecessary anymore.
```
gavin-norman-sociomantic: "widely"

relnotes/await-timeout.feature.md
```diff
@@ -0,0 +1,15 @@
+# Call `awaitOrTimeout` via scheduler interface
+
+Thanks to introducing `IScheduler` interface, it is not possible to call
```
gavin-norman-sociomantic: "it is now possible"

relnotes/await-timeout.feature.md
```diff
@@ -0,0 +1,15 @@
+# Call `awaitOrTimeout` via scheduler interface
```
gavin-norman-sociomantic: `##`

gavin-norman-sociomantic (commented):
Looks good generally.

Apart from the `##` thing, I think your release notes are missing the `name.of.affected.module` part.

src/ocean/task/IScheduler.d
```diff
@@ -0,0 +1,325 @@
+/*******************************************************************************
+
+    This module defines public API for `ocean.task.Scheduler` together with
+    all involved data types. It is intended to be used in other modules instead
+    of direct `ocean.task.Scheduler` import to untangle complex dependency chain
+    the latter brings.
+
+    Copyright:
+        Copyright (c) 2017 sociomantic Labs gmbH. All rights reserved.
```
nemanja-boric-sociomantic: you got capitalization wrong ;-) `sociomantic labs GmbH`

src/ocean/task/IScheduler.d
```diff
@@ -0,0 +1,325 @@
...
+
+import core.thread;
+
+import ocean.task.Task;
+import ocean.meta.traits.Indirections : hasIndirections;
+import ocean.io.select.EpollSelectDispatcher;
+
+/*******************************************************************************
+
+    Public methods of `ocean.task.Scheduler` stripped from implementation
```
nemanja-boric-sociomantic: It feels weird to mention the concrete implementation of the interface as the docs for the interface. Maybe you move the scheduler's docs here and then adjust scheduler's docs to be "implementation of IScheduler with ..."

src/ocean/task/IScheduler.d
```diff
@@ -0,0 +1,325 @@
...
+
+        /// Defines single mapping of `ClassInfo` to worker fiber pool in
+        /// configuration
+        public struct PoolDescription
+        {
+            /// result of `Task.classinfo` for task type which is to be handled
+            /// by this pool
+            ClassInfo task_kind;
+
+            /// worker fiber allocate stack size
```
nemanja-boric-sociomantic: `worker fiber's allocated stack size.`

src/ocean/task/IScheduler.d
```diff
@@ -0,0 +1,325 @@
...
+        size_t suspended_queue_busy;
+        size_t suspended_queue_total;
+        size_t worker_fiber_busy;
+        size_t worker_fiber_total;
+    }
+
+    /***************************************************************************
+
+        Getter for scheduler epoll instance. Necessary for integration with
+        `ISelectClient` utilities so that new select event can be registered.
```
nemanja-boric-sociomantic: `select event` or `select client`?

src/ocean/task/internal/FiberPoolWithQueue.d
```diff
@@ -68,7 +69,7 @@ public class FiberPoolWithQueue : FiberPool
     **************************************************************************/
 
     public this ( size_t queue_limit, size_t stack_size, size_t limit,
-        void delegate(istring, int) process_events )
+        void delegate () process_events )
```
nemanja-boric-sociomantic: These changes don't look connected with this commit?

src/ocean/task/IScheduler.d
nemanja-boric-sociomantic: `delayedResume`? `resumeDelayed`?

relnotes/scheduler-exceptions.migration.md
nemanja-boric-sociomantic: Yes, it should mention `version (UnitTest)` (also commit message: "Change the default scheduler exception handler in unittest buids").

src/ocean/task/Scheduler.d
```diff
@@ -861,6 +868,28 @@ public void initScheduler ( SchedulerConfiguration config,
...
+                    "Unhandled exception in task {} ({})",
+                    cast(void*) t, t.classinfo.name
+                );
+            }
+            else
+            {
+                Stdout.formatln("Unhandled exception in epoll/scheduler");
+            }
+
+            Stdout.formatln("\t{} ({}:{})", e.message(), e.file, e.line)
```
nemanja-boric-sociomantic: Maybe `Stderr` is better choice?

src/ocean/io/select/protocol/task/TaskSelectTransceiver_test.d
```diff
@@ -0,0 +1,111 @@
+/*******************************************************************************
+
+    Copyright:
+        Copyright (c) 2017 Sociomantic Labs GmbH.
```
nemanja-boric-sociomantic: bad copyright capitalisation, missing module header description if you want one.

src/ocean/task/IScheduler.d
```diff
@@ -202,19 +202,6 @@ public interface IScheduler
 
     /***************************************************************************
```
nemanja-boric-sociomantic: This commit can be easily squashed with the first IIUC.

Build.mak
```diff
@@ -59,25 +59,11 @@ TEST_FILTER_OUT += \
...
-$O/test-filesystemevent: override LDFLAGS += -lrt
-
-$O/test-selectlistener: override LDFLAGS += -lebtree
-
-$O/test-unixlistener: override LDFLAGS += -lebtree
-
-$O/test-loggerstats: override LDFLAGS += -lebtree
-
-$O/test-signalext: override LDFLAGS += -lebtree
+$O/test-%: override LDFLAGS += -lebtree
```
nemanja-boric-sociomantic, line 62: This change deserves separate commit :+1:.

relnotes/await-timeout.feature.md
```diff
@@ -0,0 +1,15 @@
...
+```D
+import ocean.task.IScheduler;
+theScheduler.awaitOrTimeout(task, timeout);
+```
+
+Internally it still does the same call to global function in
+`ocean.task.util.Timer`, but new alias allows for slightly more uniform API.
+
+NB: this means that any app using the scheduler is now required to add
+`-L-lebtree` to its linker flags.
```
nemanja-boric-sociomantic, line 15: Hm, so this is migration/feature note. Maybe add `migration` note with just this last paragraph as well.

src/ocean/task/IScheduler.d
mihails-strasuns-sociomantic: `delayedResume`, I am choosing you

mihails-strasuns-sociomantic:
Updated:
- rebased
- addressed all currently present comments

mihails-strasuns-sociomantic:

relnotes/resume.feature.md
```diff
@@ -0,0 +1,13 @@
+## Delayed resuming of a task
```
gavin-norman-sociomantic: Suggest: "New method to resume a task after the event loop cycle".

relnotes/resume.feature.md
```diff
@@ -0,0 +1,13 @@
...
+
+`ocean.task.IScheduler`
+
+`IScheduler` interface defines new method, `delayedResume`, which takes one task
+and adds it to the same resume queue as used by `processEvents` method. That
+means that such task will be resumed by the scheduler after current epoll cycle
+instead of immediate resuming by the current context.
```
gavin-norman-sociomantic: Is it worth explicitly mentioning that the task will be resumed _from_ the epoll context, too? (Also in method comments.)

src/ocean/task/Scheduler.d
```diff
@@ -544,6 +547,29 @@ final class Scheduler : IScheduler
...
 
+    /***************************************************************************
+
+        Orders scheduler to resume given task unconditionally after current
+        epoll cycle. Should be used instead of plain `Task.resume` from
+        termination hooks of other tasks.
+
+        Params:
+            task = task object to resume on next cycle
+
```
gavin-norman-sociomantic, line 579: `Throws:`

relnotes/scheduler-exceptions.migration.md
gavin-norman-sociomantic: >(also commit message: "Change the default scheduler exception handler in unittest buids").

Ping.

relnotes/scheduler-exceptions.migration.md
```diff
@@ -0,0 +1,21 @@
+## New default scheduler exception handler in tests
```
gavin-norman-sociomantic, line 1: This breaking change is slipped in because you're already making an important breaking change a few commits later, right?

Should the policy rather be to absolutely minimise breaking changes in minor branches? (Rather than a kind of free-for-all attitude, once one breaking change has been included ;)

relnotes/resume.migration.md
```diff
@@ -0,0 +1,8 @@
+## Disallow direct resume in termination hooks
+
+It is now necessary to use `IScheduler.delayedResume` method if one task has to
+be resumed as a result of termination of another task. Attempting direct resume
+will result in a `SanityException` being thrown.
+
+All ocean facilities have been already adjusted to switch accordingly (for
+example, `await`).
```
gavin-norman-sociomantic: Suggest ", but please pay attention to adjusting application code that does the same."

relnotes/await-timeout.migration.md
```diff
@@ -0,0 +1,4 @@
+## Using `Scheduler` requires linking ebtree
```
gavin-norman-sociomantic, line 1: This is another breaking change in the "not essential" category.

gavin-norman-sociomantic (commented):
A few small comments.

My only significant question is the policy on including non-essential breaking changes in this PR.

relnotes/await-timeout.migration.md
mihails-strasuns-sociomantic: This one indeed can be moved to v4.x.x (it is only needed to provide `awaitOrTimeout` via generic scheduler interface) but I took the opportunity to put it here because in practice it is almost not going to matter - pretty much any non-trivial app that uses scheduler also uses its timer utils and thus links ebtree.

relnotes/scheduler-exceptions.migration.md
mihails-strasuns-sociomantic: No, this one is actually quite related. With old default handler migrating main breaking change is much harder because exception from failed `verify` in `Task.resume` could be lost inside epoll or replaced by some other exception during the unwinding, making it much harder to debug any test failure caused by it.

Thus this change is to actually make another breaking change less annoying.

relnotes/resume.feature.md
mihails-strasuns-sociomantic: Isn't it already implied by "same resume queue as used by `processEvents`" and "resumed by the scheduler"?

mihails-strasuns-sociomantic:
Updated.

relnotes/scheduler-exceptions.migration.md
gavin-norman-sociomantic: Ok :+1:

relnotes/resume.feature.md
gavin-norman-sociomantic: Ah, "resumed by the scheduler", yes.

I'd remove the bit about `processEvents`, tbh. That's an implementation detail and irrelevant to the average user. That's what threw me off.

mihails-strasuns-sociomantic: Updated

gavin-norman-sociomantic (approved):
Do you have any thoughts on documenting (in the release notes) the reason why there are breaking changes in a minor release? (And the fact that they're of low impact and probably won't actually break anything.)

mihails-strasuns-sociomantic:
I am using only available exception case - https://github.com/sociomantic-tsunami/neptune/blob/master/doc/library-user.rst#exceptions : 

> If the addressed bug is critical (i.e. may result in memory corruption, wrong business logic, etc.), it will be released as a patch release. The release notes will clearly indicate that a breaking change has been made and the tag of the patch release will have +breaking appended.

mihails-strasuns-sociomantic:
(it only talks about patch releases but same principle applies to minor ones)

gavin-norman-sociomantic:
Right, I was just thinking that the automatically compiled release notes will look odd to users. "Why does this contain a Migration Instructions section!?" It'll need some lead-in / explanation.

mihails-strasuns-sociomantic:
Like a special first paragraph in mail notes?

gavin-norman-sociomantic:
Right, and in the gitub release, I'd suggest.

Interesting... this could actually be integrated into neptune-release, based on the presence of the `+breaking` in the tag. (For the future.)

mihails-strasuns-sociomantic:
Actually, the `+breaking` bit is currently not supported by `neptune-release` at all - I will have to do it manually this time :( (the whole release process)

gavin-norman-sociomantic:
Ah. That's too bad. See https://github.com/sociomantic-tsunami/neptune/issues/129.

mihails-strasuns-sociomantic:
So, with that neptune discussion in mind, is there anything pending for this PR?

gavin-norman-sociomantic:
Given that the whole release will need to be done manually, no.

PR #356, opened 2017-11-14
Reduce EpollProcess memory usage: use a FreelistAllocator

don-clugston-sociomantic:
The existing code assumed that creating new processes is sufficiently rare that memory efficiency is unimportant, but some apps create a large number of processes. (In my apps, several hundred processes are created every second).
To reduce GC usage, use a FreeListAllocator, so that memory is recycled.

### DIFF
--- a/relnotes/nogc-process.feature.md
+++ b/relnotes/nogc-process.feature.md
@@ -1,0 +1,6 @@
+## `EpollProcess` no longer allocates memory
+
+`ocean.io.select.client.EpollProcess`
+
+`EpollProcess` now uses a `FreeListAllocator` internally, so that starting a
+new process after a previous one has finished no longer allocates memory.
--- a/src/ocean/io/select/client/EpollProcess.d
+++ b/src/ocean/io/select/client/EpollProcess.d
@@ -177,13 +177,27 @@ public abstract class EpollProcess
 
         /***********************************************************************
 
+            Allocate the map from a free list, to reduce memory allocation.
+
+        ***********************************************************************/
+
+        import FreeListAllocator =
+            ocean.util.container.map.model.BucketElementFreeList;
+
+
+        /***********************************************************************
+
             Constructor.
 
         ***********************************************************************/
 
         public this ( )
         {
-            this.processes = new StandardKeyHashingMap!(EpollProcess, int)(20);
+            // Allocate the map using a free list
+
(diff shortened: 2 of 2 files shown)


gavin-norman-sociomantic:
Looks like a good enhancement! Do you think it'd be possible to add a small test to check that the memory is reused?

don-clugston-sociomantic:
> Do you think it'd be possible to add a small test to check that the memory is reused?

Tricky. I don't know how to do that, the existing code structure is not very test-friendly, and because it creates processes it's intrinsically not pure. I know that it is reused because after this commit I get no more calls to `gc_malloc` from this point.

gavin-norman-sociomantic:
Fair enough. If it's hassle, it's not essential.

gavin-norman-sociomantic:
I was thinking of a test program btw (`EpollProcess_test.d`), not a unittest, so purity isn't important.

don-clugston-sociomantic:
Even as an integration test, it's tricky. It's kind of a statistical property -- the memory usage will depend not only on the size of the map, but also on how many processes are active at any given time, and that's a hard thing to control. I guess some kind of performance measurement of an app could be done but it's would not be specific to this issue, and it would need to run for quite a long time to be able to detect this small effect. Certainly I can't see any way of doing a simple test.

mihails-strasuns-sociomantic (commented):
Will that memory still be scanned by GC?

don-clugston-sociomantic:
Yeah, there's nothing special about it, the only difference to the existing code is that it keeps reusing the memory it allocated before. Just like any other reused buffer, really.

mihails-strasuns-sociomantic (approved):
Ah, I see, it is still backed by GC for allocation of free list slots.

LGTM

mihails-strasuns-sociomantic:
Would you mind mentioning it in release notes please?

don-clugston-sociomantic:
What kind of release notes do we do for performance improvements like this (that don't affect user code in any way)? Would this class as a `feature` ?

mihails-strasuns-sociomantic:
Yeah, I would classify it as a feature - potentially far reaching performance improvement that may be of common interest.

don-clugston-sociomantic:
Release notes added.

mihails-strasuns-sociomantic (approved):

PR #360, opened 2017-11-14
Call libgcrypt and libgcrypt-error initialisation functions as shared

nemanja-boric-sociomantic:
In D2, these initialization functions should be called inside `shared
static` constructor, not inside static constructor, as the
initialization should not be done per thread, but once per application's
lifetime.


david-eckardt-sociomantic (approved):

mihails-strasuns-sociomantic (commented):
LGTM

PR #361, opened 2017-11-15
Remove ocean.text.convert.Layout, Layout_tango and core.RuntimeTraits

mathias-lang-sociomantic:
Just a WIP at the moment, a few points need to be addressed.
Last user I know of are `StaticTrace`, `PeriodicTrace` and `Layout` (the non `_tango` module).
I also need to review documentation and write release notes. Since I'm trying to deprecate what I can in v3.x.x, some commit might be of interest to @mihails-strasuns-sociomantic when he does the next v4 rebase.

Issue #206, opened 2017-08-01
Deprecate ocean.text.convert.{Format,Layout,Layout_tango}


mathias-lang-sociomantic:
Those modules have been replaced by `ocean.text.convert.Formatter` and should be deprecated as they don't play well with D2.
`Format` depends on `Layout_tango`, as does `Layout`.

Once https://github.com/sociomantic-tsunami/ocean/pull/203 and https://github.com/sociomantic-tsunami/ocean/pull/187 have been pulled, we will be able to deprecate `Layout_tango`. As 203 includes a breaking change, this will only happen in v4.0.0 at best.

Feel free to assign me.

mathias-lang-sociomantic:
Along with those modules, we should also kill `ocean.core.RuntimeTraits`

mihails-strasuns-sociomantic:
@mathias-lang-sociomantic this is now fully resolved in v4.x.x , correct?

mathias-lang-sociomantic:
It is!

PR #362, opened 2017-11-15
 Deprecate RTTI-specific functions in stream's Format

mathias-lang-sociomantic:
Deprecate a few methods as part of https://github.com/sociomantic-tsunami/ocean/pull/361

### DIFF
--- a/src/ocean/io/Stdout.d
+++ b/src/ocean/io/Stdout.d
@@ -57,8 +57,6 @@ public alias Stderr stderr; /// Alternative.
 
 static this ( )
 {
-    // note that a static-ctor inside Layout fails
-    // to be invoked before this is executed (bug)
     auto layout = Layout!(char).instance;
 
     Stdout = new TerminalOutput!(char)(layout, Cout.stream);
--- a/src/ocean/io/Stdout.d
+++ b/src/ocean/io/Stdout.d
@@ -57,10 +57,8 @@ public alias Stderr stderr; /// Alternative.
 
 static this ( )
 {
-    auto layout = Layout!(char).instance;
-
-    Stdout = new TerminalOutput!(char)(layout, Cout.stream);
-    Stderr = new TerminalOutput!(char)(layout, Cerr.stream);
+    Stdout = new TerminalOutput!(char)(Cout.stream);
+    Stderr = new TerminalOutput!(char)(Cerr.stream);
 
     Stdout.flush = !Cout.redirected;
     Stdout.redirect = Cout.redirected;
@@ -121,7 +119,7 @@ public class TerminalOutput ( T ) : FormatOutput!(T)
 
     public this (OutputStream output, Immut!(T)[] eol = Eol)
     {
-        this (Layout!(T).instance, output, eol);
+        super(output, eol);
     }
 
 
@@ -132,6 +130,7 @@ public class TerminalOutput ( T ) : FormatOutput!(T)
 
     ***************************************************************************/
 
+    deprecated("Using Layout with this class is deprecated - Remove first argument at call site")
     public this (Layout!(T) convert, OutputStream output, Immut!(T)[] eol = Eol)
     {
         super(convert, output, eol);
@@ -146,7 +145,7 @@ public class TerminalOutput ( T ) : FormatOutput!(T)
 
     public typeof(this) format ( Const!(T)[] fmt, ... )
     {
-        super.format(fmt, _arguments, _argptr);
+        super._transitional_format(fmt, _arguments, _argptr);
         return this;
     }
 
@@ -158,6 +157,7 @@ public class TerminalOutput ( T ) : FormatOutput!(T)
 
     ***************************************************************************/
 
+    deprecated("RTTI-specific function is deprecated, use a Formatter-compatible API")
     public typeof(this) format (Const!(T)[] fmt, TypeInfo[] arguments, ArgList args)
     {
         super.format(fmt, arguments, args);
@@ -174,8 +174,8 @@ public class TerminalOutput ( T ) : FormatOutput!(T)
 
     public typeof(this) formatln ( Const!(T)[] fmt, ... )
     {
-        super.formatln(fmt, _arguments, _argptr);
-        return this;
+        super._transitional_format(fmt, _arguments, _argptr);
+        return this.newline;
     }
 
 
@@ -186,6 +186,7 @@ public class TerminalOutput ( T ) : FormatOutput!(T)
 
     ***************************************************************************/
 
+    deprecated("RTTI-specific function is deprecated, use a Formatter-compatible API")
     public typeof(this) formatln (Const!(T)[] fmt, TypeInfo[] arguments, ArgList args)
     {
         super.formatln(fmt, arguments, args);
--- a/src/ocean/io/stream/Format.d
+++ b/src/ocean/io/stream/Format.d
@@ -292,21 +292,3 @@ class FormatOutput(T) : OutputFilter
                 return count;
         }
 }
-
-
-/*******************************************************************************
-
-*******************************************************************************/
-
-debug (Format)
-{
-        import ocean.io.device.Array;
-
-        void main()
-        {
-                auto print = new FormatOutput!(char) (new Array(1024, 1024));
-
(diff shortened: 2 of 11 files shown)


src/ocean/text/Arguments.d
```diff
@@ -611,6 +611,7 @@ import ocean.io.stream.Format : FormatOutput;
 import ocean.math.Math;
 import ocean.text.Util;
 import ocean.text.convert.Formatter;
+static import ocean.text.convert.Layout_tango; // TODO: Remove in v4.0.0
```
leandro-lucarella-sociomantic: Create an issue for this?

src/ocean/text/Arguments.d
```diff
@@ -1096,6 +1097,7 @@ public class Arguments
 
     ***************************************************************************/
 
+    deprecated("Use the overload that doesn't take a parameter (and use the Formatter")
```
leandro-lucarella-sociomantic: Missing an ending `)`

src/ocean/text/Arguments.d
```diff
@@ -1114,6 +1116,26 @@ public class Arguments
...
+    /// Ditto
+    public istring errors ()
+    {
+        mstring result;
+
+        foreach (arg; args)
+        {
+            if (arg.error)
+            {
+                // TODO: Replace with `sformat` in v4.0.0
```
leandro-lucarella-sociomantic, line 1136: Also issue

src/ocean/io/Stdout.d
```diff
@@ -148,7 +149,7 @@ public class TerminalOutput ( T ) : FormatOutput!(T)
 
     public typeof(this) format ( Const!(T)[] fmt, ... )
     {
-        super.format(fmt, _arguments, _argptr);
+        super.doNotUse(fmt, _arguments, _argptr);
```
leandro-lucarella-sociomantic: Say what? :thinking:

src/ocean/io/stream/Format.d
```diff
@@ -138,7 +127,14 @@ class FormatOutput(T) : OutputFilter
 
         final FormatOutput format (Const!(T)[] fmt, ...)
         {
-            return format(fmt, _arguments, _argptr);
+            this.convert(&this.emit, _arguments, _argptr, fmt);
+            return this;
+        }
+
+        /// Used by derived classes to avoid deprecations - Remove in v4.0.0
+        final void doNotUse (Const!(T)[] fmt, TypeInfo[] arguments, ArgList args)
```
leandro-lucarella-sociomantic: Also, create an issue not to forget and maybe give it a name that says what it does (besides making it obvious it shouldn't be used).

leandro-lucarella-sociomantic (commented):
The last commit message says: `* submodules/makd v1.9.1(8711da2)...v1.6.0(949d0ad) (50 commits)`. I guess it's wrong, right?

src/ocean/io/stream/Format.d
mathias-lang-sociomantic: I already raised a P.R. against v4.x.x to remove it. Also the comment says why it's called this way. However, it should be `protected`

src/ocean/text/Arguments.d
mathias-lang-sociomantic: https://github.com/sociomantic-tsunami/ocean/issues/206

src/ocean/text/Arguments.d
mathias-lang-sociomantic: Also covered by https://github.com/sociomantic-tsunami/ocean/issues/206

mathias-lang-sociomantic:
Updated:
- Removed wrong Makd downgrade
- Added missing `)`
- Added `protected` on `doNotUse`.

leandro-lucarella-sociomantic:
> Added protected on doNotUse.

Still not crazy about this (I meant name it something like `_transitional_format` wouldn't be enough / better?)

mathias-lang-sociomantic:
Well I'm not set on the name. I just consider it more important to say "Do not use this" than to say what it's for... Because it's not meant to be used, so its purpose had no value for external users.
But as I said, I am not going to argue about it, so I changed the name to `_transitional_format`.

leandro-lucarella-sociomantic:
> Well I'm not set on the name. I just consider it more important to say "Do not use this" than to say what it's for... Because it's not meant to be used, so its purpose had no value for external users.

But for people that need to read/maintain the code (not library users but library developers) is still important to be able to understand the code :)

src/ocean/text/Arguments.d
gavin-norman-sociomantic: You can also include the keyword "deprecated" in the comment, so we'll find it when searching for old crap to remove.

src/ocean/text/Arguments.d
```diff
@@ -1114,6 +1116,26 @@ public class Arguments
         return result;
     }
 
+    /// Ditto
```
gavin-norman-sociomantic: Not ditto, as the params are different.

src/ocean/io/Stdout.d
```diff
@@ -134,6 +134,7 @@ public class TerminalOutput ( T ) : FormatOutput!(T)
 
     ***************************************************************************/
 
+    deprecated("Using Layout with this class is deprecated")
```
gavin-norman-sociomantic: Suggested course of action?

src/ocean/io/stream/Format.d
```diff
@@ -120,6 +108,7 @@ class FormatOutput(T) : OutputFilter
 
         **********************************************************************/
 
+        deprecated("Using Layout with this class is deprecated")
```
gavin-norman-sociomantic: Suggested course of action?

src/ocean/text/Arguments.d
mathias-lang-sociomantic: Then by definition you can never use `ditto` for the same overload set.
AFAIK ddoc supports that usage (and it's widely used in Phobos), but maybe we should document in which case we can use ditto then, since expectations seems to vary.

gavin-norman-sociomantic: Seems so. My expectation is that literally the same comment would be applicable to the `ditto`ed thing.

mathias-lang-sociomantic:
Fixed according to @gavin-norman-sociomantic 's feedback.

src/ocean/text/Arguments.d
mathias-lang-sociomantic: Well actually most the the description was wrong as well, so Ditto was not applicable here.

src/ocean/io/Stdout.d
```diff
@@ -59,10 +59,10 @@ static this ( )
 {
     // note that a static-ctor inside Layout fails
     // to be invoked before this is executed (bug)
```
mihails-strasuns-sociomantic: is comment still applicable? i would expect not, if new code works

src/ocean/io/stream/Format.d
```diff
@@ -138,7 +127,14 @@ class FormatOutput(T) : OutputFilter
 
         final FormatOutput format (Const!(T)[] fmt, ...)
         {
-            return format(fmt, _arguments, _argptr);
+            this.convert(&this.emit, _arguments, _argptr, fmt);
+            return this;
+        }
+
+        /// Used by derived classes to avoid deprecations - Remove in v4.0.0
```
mihails-strasuns-sociomantic: Please mention "deprecated" substring here too.

mihails-strasuns-sociomantic (commented):
Few minor comments

src/ocean/io/Stdout.d
mathias-lang-sociomantic: The only static ctor I could find is here: https://github.com/sociomantic-tsunami/ocean/blob/edf92dc7eea6869b8e54d99fe11f9a452a8ab1fe/src/ocean/text/convert/Layout_tango.d#L1033-L1036
I'll remove the comment

mathias-lang-sociomantic:
Updated

mihails-strasuns-sociomantic (approved):
LGTM, @gavin-norman-sociomantic please merge if you have no further requests

PR #366, opened 2017-11-20
Allow both non-interactive and interactive sessions over the unix socket

nemanja-boric-sociomantic:
Most usage scenarios over the unix domain socket doesn't require
interactive session, but they consist only of echoing required data over
the socket. For the API to stay simple, UnixListener and related modules
now allow for the old-fashion echo-only command handlers, which only
have a single delegate argument.

### DIFF
--- a/src/ocean/net/server/unix/UnixConnectionHandler.d
+++ b/src/ocean/net/server/unix/UnixConnectionHandler.d
@@ -62,12 +62,39 @@ import ocean.core.array.Mutation : copy;
 /// Provides basic command handling functionality for unix socket commands.
 public class BasicCommandHandler
 {
-    /// Alias for a command handler delegate.
+    /// Alias for an interactive command handler delegate.
     public alias void delegate ( cstring,  void delegate (cstring),
-            void delegate (ref mstring)) Handler;
+            void delegate (ref mstring)) InteractiveHandler;
 
-    /// Map of command name to handler response delegate.
+    /// Alias for a non-interactive command handler delegate.
+    public alias void delegate ( cstring, void delegate (cstring) ) Handler;
+
+    /// Map of command name to interactive handler response delegate.
+    public InteractiveHandler[istring] interactive_handlers;
+
+    /// Map of a command name to non-interactive handlers delegate
     public Handler[istring] handlers;
+
+    /***************************************************************************
+
+        Constructor
+
+        Note that handlers and interactive handlers' command names may overlap.
+        In that case, the interactive handler is given the priority.
+
+        Params:
+            handlers = Array of command string to handler delegate.
+            interactive_handlers = Array of command string to interactive handler
+                delegate.
+
+    ***************************************************************************/
+
+    public this ( Handler[istring] handlers,
+            InteractiveHandler[istring] interactive_handlers )
+    {
+        this.handlers = handlers;
+        this.interactive_handlers = interactive_handlers;
+    }
 
     /***************************************************************************
 
@@ -100,9 +127,14 @@ public class BasicCommandHandler
         void delegate ( cstring ) send_response,
         void delegate (ref mstring) wait_reply)
     {
-        if (auto handler = command in this.handlers)
+
+        if (auto handler = command in this.interactive_handlers)
         {
             (*handler)(args, send_response, wait_reply);
+        }
+        else if (auto handler = command in this.handlers)
+        {
+            (*handler)(args, send_response);
         }
         else
         {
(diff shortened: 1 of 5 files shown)


nemanja-boric-sociomantic:
Interesting that this now makes the original change non-breaking change :thinking:

src/ocean/net/server/unix/UnixConnectionHandler.d
```diff
@@ -100,7 +124,12 @@ public class BasicCommandHandler
         void delegate ( cstring ) send_response,
         void delegate (ref mstring) wait_reply)
     {
+
         if (auto handler = command in this.handlers)
+        {
+            (*handler)(args, send_response);
+        }
+        else if (auto handler = command in this.interactive_handlers)
```
mathias-lang-sociomantic: As the interactive session is a more "advanced" API, shouldn't it preferred ?
Also the order in which they are processed should be documented somewhere.

nemanja-boric-sociomantic: I think what I really want to do is to iterate over keys of both maps in the constructor and verify that there are no duplicates. `O(mn)` complexity should not matter in case like this (I hope).

src/ocean/net/server/unix/UnixConnectionHandler.d
```diff
@@ -130,7 +183,8 @@ public class UnixConnectionHandler : UnixSocketConnectionHandler!(BasicCommandHa
                   BasicCommandHandler.Handler[istring] handlers,
                   istring address_path )
     {
-        super(finalize_dg, epoll, new BasicCommandHandler(handlers),
+        super(finalize_dg, epoll,
+            new BasicCommandHandler(handlers),
```
mathias-lang-sociomantic: Unrelated :angel:

mathias-lang-sociomantic:
> Interesting that this now makes the original change non-breaking change :thinking:

Do you feel like backporting ? :D

src/ocean/net/server/unix/UnixConnectionHandler.d
mathias-lang-sociomantic: That was something else I was wondering if we could do, but I'm usually on the opinion that we should be as lax as practical. So we could have a natural overriding scheme. But I don't have a strong opinion here.

nemanja-boric-sociomantic: Just out of the interest, why do you think that the more advanced API should have the priority in the overriding? I would thought it's the reverse.

nemanja-boric-sociomantic:
Not really :-). It's not something that we are quite desperate to get right now, I think it can wait a bit.

src/ocean/net/server/unix/UnixConnectionHandler.d
mathias-lang-sociomantic: Since the more advanced API is a superset of the simpler API, I expect users of the more advanced API to support a superset of the features supported by the simpler API, because it's usually trivial to wrap the simpler API.

nemanja-boric-sociomantic: Yeah, I think what you're saying makes sense. I'll update the ordering and the documentation.

nemanja-boric-sociomantic:
Updated.

src/ocean/net/server/unix/UnixConnectionHandler.d
```diff
@@ -62,13 +62,40 @@ import ocean.core.array.Mutation : copy;
...
 {
-    /// Alias for a command handler delegate.
+    /// Alias for an interactive command handler delegate.
     public alias void delegate ( cstring,  void delegate (cstring),
-            void delegate (ref mstring)) Handler;
+            void delegate (ref mstring)) InteractiveHandler;
 
-    /// Map of command name to handler response delegate.
+    /// Alias for a non-interactive command handler delegate.
+    public alias void delegate ( cstring, void delegate (cstring) ) Handler;
```
mihails-strasuns-sociomantic, line 70: You change definition of a public alias, how this is not a breaking change?

nemanja-boric-sociomantic: Because it changes it back to what it was in the v3.x.x.

nemanja-boric-sociomantic: https://github.com/sociomantic-tsunami/ocean/blob/6a5371e43c1e28d8d99a676c67566306ae791e7e/src/ocean/net/server/unix/UnixConnectionHandler.d#L65

mathias-lang-sociomantic: It's v4.x.x, not released yet. It's changing it back to the definition we have in v3.x.x

EDIT: Damn too slow

mihails-strasuns-sociomantic: Ah, I see. Would it be less confusing to fixup it into original commit upon next rebase?

nemanja-boric-sociomantic: Yeah, perhaps that's what we could do. I'll update the commit title.

nemanja-boric-sociomantic: Updated.

mihails-strasuns-sociomantic (approved):

PR #367, opened 2017-11-22
Deprecate opCall alias to deprecated print

mihails-strasuns-sociomantic:
Found when rebasing v4.x.x and trying to clean deprecated things.

### DIFF
--- a/src/ocean/io/stream/Format.d
+++ b/src/ocean/io/stream/Format.d
@@ -81,6 +81,7 @@ class FormatOutput(T) : OutputFilter
         private Layout!(T)      convert;
         private bool            flushLines;
 
+        deprecated("Use the 'format' method instead, or 'flush'")
         public alias print      opCall;         /// opCall -> print
         public alias newline    nl;             /// nl -> newline
 
--- a/relnotes/stdout-formatter.deprecation.md
+++ b/relnotes/stdout-formatter.deprecation.md
@@ -6,4 +6,5 @@ In both those module, specific support for RTTI has be removed to prepare for
 the transition from Layout_tango to Formatter.  This includes `format` and
 `formatln` methods accepting the `TypeInfo[]` / `void*[]` duo, the `print`
 function which was redundant with `format` and of little use, the `layout`
-setter and getter, and the `Layout`-accepting constructors.
+setter and getter, and the `Layout`-accepting constructors. Same applies to
+`opCall` as it was aliasing the `print` method.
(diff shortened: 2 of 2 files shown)


mihails-strasuns-sociomantic:
@mathias-lang-sociomantic also how shall I proceed with removing `_transitional_format`? AFAICS it is used in non-deprecated methods. Should it be replaced with plain `super.format`?

src/ocean/io/stream/Format.d
```diff
@@ -81,6 +81,7 @@ class FormatOutput(T) : OutputFilter
         private Layout!(T)      convert;
         private bool            flushLines;
 
+        deprecated("Use the 'format' method instead, or 'flush'")
```
mathias-lang-sociomantic, line 84: Actually I didn't do it on purpose, since `print` is already deprecated, using `opCall` *should* trigger the deprecation as well... Doesn't it ?

mihails-strasuns-sociomantic: It does, but it breaks after "remove all deprecated symbols" procedure I do on rebase.

mathias-lang-sociomantic: Remove the alias as well then ? :D

mihails-strasuns-sociomantic: Yes, which is why I mark alias deprecated explicitly here too - so that it becomes obvious and acknowledged, and not something I secretly do in my branch :)

mathias-lang-sociomantic (approved):

mihails-strasuns-sociomantic:
@mathias-lang-sociomantic https://github.com/sociomantic-tsunami/ocean/pull/367#issuecomment-346324443 this question still stands - if I try to replace `_transitional_format` with plain format, it breaks formatting:

```D
import ocean.io.Stdout;
import ocean.transition;

void main()
{
    cstring hello = "Hello World!";
    Stdout.formatln("Hello {}", hello);
}

// rdmd --compiler=dmd1 -I./src xxx.d 
// Hello [char[]]
```

mathias-lang-sociomantic:
@mihails-strasuns-sociomantic : Yeah it still uses `Tango_Layout`, I can submit a fixup P.R. later. Once v4.x.x is rebased I'll rebase https://github.com/sociomantic-tsunami/ocean/pull/361

mihails-strasuns-sociomantic:
Self-merging as this is just infra triviality.

PR #369, opened 2017-11-22
Fix ocean.io.Path.isWritable

nemanja-boric-sociomantic:
The previous implementation of isWritable was broken, as it was using
stat function inappropriately, combining it with a O_RDONLY flag from
the open function, which also happens to has value of 0 so it would
always report is writable. On top of this, even if it worked by some
accident, it would still completely ignore the effective/real uid and
the file ownership.
Although now used access doesn't check the real uid/gid, but the
effective one (meaning the succeeding open could still fail/pass), it's
what we actually want most of the time.

### DIFF
--- a/src/ocean/io/Path.d
+++ b/src/ocean/io/Path.d
@@ -302,9 +302,7 @@ package struct FS
 
                 static bool isWritable (cstring name)
                 {
-                        stat_t stats = void;
-
-                        return (getInfo(name, stats) & O_RDONLY) is 0;
+                        return posix.access(name.ptr, W_OK) == 0;
                 }
 
                 /***************************************************************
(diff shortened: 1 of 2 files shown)


src/ocean/io/Path.d
```diff
@@ -302,9 +302,7 @@ package struct FS
 
                 static bool isWritable (cstring name)
                 {
-                        stat_t stats = void;
-
-                        return (getInfo(name, stats) & O_RDONLY) is 0;
+                        return posix.access(name.ptr, W_OK) == 0;
```
mihails-strasuns-sociomantic, line 305: is name guaranteed to be 0-terminated here?

nemanja-boric-sociomantic: Yes, as in all these methods in the same `package` struct.

mihails-strasuns-sociomantic (approved):

PR #372, opened 2017-11-24
Fix memory usage of PullParser in D2 builds

mihails-strasuns-sociomantic:
Alternative approach to https://github.com/sociomantic-tsunami/ocean/pull/347

Now `PullParser` operates strictly on slices and does no allocations and
`Document` copies those slices into nodes when necessary.

### DIFF
--- a/src/ocean/text/xml/PullParser.d
+++ b/src/ocean/text/xml/PullParser.d
@@ -770,4 +770,11 @@ unittest
 {
     auto itr = new PullParser!(char)(testXML);
     testParser (itr);
+
+    // Parsing new text (or even the same one) should not involve any further
+    // memory allocation
+    testNoAlloc({
+        itr.reset(testXML);
+        testParser(itr);
+    }());
 }
--- a/src/ocean/text/xml/PullParser.d
+++ b/src/ocean/text/xml/PullParser.d
@@ -108,8 +108,10 @@ public enum XmlTokenType
 
 *******************************************************************************/
 
-class PullParser(Ch = char)
+class PullParser(ChMut = char)
 {
+    alias Const!(ChMut)  Ch;
+
     public int           depth;
     public Ch[]          prefix;
     public Ch[]          rawValue;
@@ -126,7 +128,7 @@ class PullParser(Ch = char)
 
      ***********************************************************************/
 
-    this(Immut!(Ch)[] content = null)
+    this(Ch[] content = null)
     {
         reset (content);
     }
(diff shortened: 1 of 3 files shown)


src/ocean/text/xml/PullParser.d
```diff
@@ -1,17 +1,13 @@
...
-            Some parts copyright (c) 2009-2016 Sociomantic Labs GmbH.
-            All rights reserved.
-
-        License:
-            Tango Dual License: 3-Clause BSD License / Academic Free License v3.0.
-            See LICENSE_TANGO.txt for details.
-
-        Version: Initial release: February 2008
-
-        Authors: Aaron, Kris
```
gavin-norman-sociomantic: I thought we'd deliberately left in the tango authors info?

src/ocean/text/xml/PullParser.d
```diff
@@ -680,10 +670,10 @@ class PullParser(Ch = char)
 
 package struct XmlText(Ch)
 {
-    package Ch*     end;
-    package size_t  len;
-    package Ch[]    text;
-    package Ch*     point;
+    package Ch*      end;
```
gavin-norman-sociomantic: Should go in the previous commit.

gavin-norman-sociomantic (commented):
I'm not really familiar with the XML parser, so won't comment on the details unless there is a need for an extra opinion, in addition to Ben / Don.

don-clugston-sociomantic:
I think the approach you've taken here is basically right. I don't think it ever makes sense to have a parser that gives mutable slices. The changes to `Document` do mean that it allocates a lot more memory, and does a lot more copying -- every tag in the document will be copied -- but since it allocates one node for every XML element anyway I don't think it's really a problem.
I've never used `Document` but the `PullParser` changes work for me. I will close my PRs on this topic.

I just have one comment, and it's optional.

Since you are cleaning up the formatting of the code, I think you should also delete the untested code in the undocumented `version` statements. There's this horrendous one in Document.d
```d
version(d)
{
                freelists = 0;          // needed to align the codegen!
}
```
I don't know if this is a typo of `version(discrete)` or if they really had `version=d;` (!)

ben-palmer-sociomantic (approved):
I agree with Dons comments. I also tested with my code and found no problems.

src/ocean/text/xml/PullParser.d
mihails-strasuns-sociomantic: I don't remember details but I remember that SVN logs has shown that real authors are quite different from whatever is left in DDOC. And git history is still present anyway. @leandro-lucarella-sociomantic do you know how important is this part legal-wise?

mihails-strasuns-sociomantic: (restored just in case)

mihails-strasuns-sociomantic:
Updated. I must say, amount of `version` code that is likely never used in practice worries me a bit.

don-clugston-sociomantic (approved):
LGTM

nemanja-boric-sociomantic (approved):
LGTM

PR #373, opened 2017-11-25
Remove all `dispose` methods from ocean

mihails-strasuns-sociomantic:
AFAICS all of changes but one are uncontroversial (any plain call to `delete` is
point less and can be removed). The bit about ebtree is something I would like
to get external opinion for - it looks like it _may_ affect ebtree own memory
deallocation down the line. However, trying to turn it into destructor causes
tests to crash.


src/ocean/util/container/AppendBuffer.d
```diff
@@ -1263,26 +1263,6 @@ private abstract class AppendBufferImpl: IAppendBufferBase
...
-            Called immediately when this instance is deleted.
-            (Must be protected to prevent an invariant from failing.)
-
-        ***********************************************************************/
-
-        protected override void dispose ( )
-        {
-            if (this.content)
-            {
-                this.deleteContent(this.content);
```
mathias-lang-sociomantic, line 1274: We should also probably deprecate this method

src/ocean/util/container/map/model/BucketElementMallocAllocator.d
```diff
@@ -264,19 +264,15 @@ public class BucketElementMallocAllocator (Bucket) : IAllocator
...
-            *******************************************************************/
+        ***********************************************************************/
 
-            protected override void dispose ( )
-            {
-                free(this.elements.ptr);
-            }
+        ~this ( )
+        {
+            free(this.elements.ptr);
```
mathias-lang-sociomantic, line 275: `dispose` is not called for GC-allocated objects, right ? So that was potentially leaking memory before ?

mathias-lang-sociomantic (commented):
Typo (meningless) in first commit message

src/ocean/util/container/map/model/BucketElementMallocAllocator.d
david-eckardt-sociomantic: It is called in [`rt_finalize`](https://github.com/sociomantic-tsunami/tangort/blob/1878c87eb69b212a6a2958e2e10f10d946c88e0c/src/rt/lifetime.d#L614) so it should be called for GC-allocated objects.

mathias-lang-sociomantic: Except if `det` is `false`. And when called from the GC [it is](https://github.com/sociomantic-tsunami/tangort/blob/32ad03f41da038b6e315931c551ddfe4cd0d26eb/src/gc/cdgc/gc.d#L1151-L1153). Only when called through `_d_delclass` it will be called with `true`.

david-eckardt-sociomantic:
Removing `IEBTree.dispose` should be fine.

`IEBTree.dispose` seems meaningless to me: It merely calls `clear`, which marks the ebtree empty, but doesn't deallocate the nodes. If the nodes are GC-allocated then they are referenced by `IEBTree.root`, which is collected, so they will be collected, too. However, if the nodes are not GC-allocated then they will be lost and leak memory, and `IEBTree.dispose` won't prevent that from happening. So far this hasn't been reported to be a problem, though.

src/ocean/util/container/map/model/BucketElementMallocAllocator.d
david-eckardt-sociomantic: Oh, I see, you're right.

mihails-strasuns-sociomantic: Yep, I think we do, though it rarely matters in practice as such maps live through all app lifetime.

mihails-strasuns-sociomantic:
@david-eckardt-sociomantic where is the allocator for ebtree nodes configured?

david-eckardt-sociomantic:
> where is the allocator for ebtree nodes configured?

The `EBTree*` classes have two constructors, one that creates an `ocean.util.container.ebtree.nodepool.NodePool!(Node)` and a second that accepts a custom `INodePool` implementation.

mathias-lang-sociomantic:
Needs a rebase

mihails-strasuns-sociomantic:
Updated, should be good to go now that CI is fixed :)

mihails-strasuns-sociomantic:

src/ocean/io/compress/Lzo.d
```diff
@@ -226,21 +226,6 @@ class Lzo
...
-    {
-        /***********************************************************************
-
-            Destructor
-
-        ***********************************************************************/
-
-        override void dispose ( )
-        {
-            delete this.workmem;
```
gavin-norman-sociomantic, line 241: Haha, oh, I see what you mean about "meaningless"!

gavin-norman-sociomantic (approved):

PR #374, opened 2017-11-26
Don't error our on deprecation with upstream compiler

mihails-strasuns-sociomantic:
This is necessary to be able to test ocean as part of upstream
regression test suite.

### DIFF
--- a/Build.mak
+++ b/Build.mak
@@ -1,9 +1,12 @@
 # Treat deprecations as errors to ensure ocean doesn't use own deprecated
-# symbols internally
-ifeq ($(DVER),1)
-override DFLAGS := $(filter-out -di,$(DFLAGS)) -v2 -v2=-static-arr-params -v2=-volatile
-else
-override DFLAGS += -de
+# symbols internally. Disable it on explicit flag to make possible regression
+# testing in D upstream.
+ifneq ($(ALLOW_DEPRECATIONS),1)
+	ifeq ($(DVER),2)
+	override DFLAGS += -de
+	else
+	override DFLAGS := $(filter-out -di,$(DFLAGS)) -v2 -v2=-static-arr-params -v2=-volatile
+	endif
 endif
 
 # Makd auto-detects if Ocean's test runner should be used based on submodules,
--- a/Build.mak
+++ b/Build.mak
@@ -5,7 +5,7 @@ ifneq ($(ALLOW_DEPRECATIONS),1)
 	ifeq ($(DVER),2)
 	override DFLAGS += -de
 	else
-	override DFLAGS := $(filter-out -di,$(DFLAGS)) -v2 -v2=-static-arr-params -v2=-volatile
+	override DFLAGS := $(filter-out -di,$(DFLAGS))
 	endif
 endif
 
--- a/Build.mak
+++ b/Build.mak
@@ -1,8 +1,6 @@
 ifeq ($(DVER),1)
 override DFLAGS := $(filter-out -di,$(DFLAGS)) -v2 -v2=-static-arr-params -v2=-volatile
 else
-# Open source Makd uses dmd by default
-DC ?= dmd
 override DFLAGS += -de
 endif
 
--- a/Build.mak
+++ b/Build.mak
@@ -1,3 +1,5 @@
+# Treat deprecations as errors to ensure ocean doesn't use own deprecated
+# symbols internally
 ifeq ($(DVER),1)
 override DFLAGS := $(filter-out -di,$(DFLAGS)) -v2 -v2=-static-arr-params -v2=-volatile
 else
--- a/Build.mak
+++ b/Build.mak
@@ -1,17 +1,3 @@
-# This will make D2 unittests fail if stomping prevention is triggered
-export ASSERT_ON_STOMPING_PREVENTION=1
-
-override DFLAGS += -w -version=GLIBC
-
-# Makd auto-detects if Ocean's test runner should be used based on submodules,
-# but we -or course- don't have Ocean as a submodule, so we set it explicitly.
-TEST_RUNNER_MODULE := ocean.core.UnitTestRunner
-
-# Enable coverage report in CI
-ifdef CI
-COVFLAG:=-cov
-endif
-
 ifeq ($(DVER),1)
 override DFLAGS := $(filter-out -di,$(DFLAGS)) -v2 -v2=-static-arr-params -v2=-volatile
 else
@@ -19,6 +5,10 @@ else
 DC ?= dmd
 override DFLAGS += -de
 endif
+
+# Makd auto-detects if Ocean's test runner should be used based on submodules,
+# but we -or course- don't have Ocean as a submodule, so we set it explicitly.
+TEST_RUNNER_MODULE := ocean.core.UnitTestRunner
 
 # Remove coverage files
 clean += .*.lst
--- a/Config.mak
+++ b/Config.mak
@@ -1,0 +1,9 @@
+# This will make D2 unittests fail if stomping prevention is triggered
+export ASSERT_ON_STOMPING_PREVENTION=1
+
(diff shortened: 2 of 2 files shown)


nemanja-boric-sociomantic:
Was moving the stuff from `Build.mak` to `Config.mak` necessary? The commit message could explain the reason.

mihails-strasuns-sociomantic:
> Was moving the stuff from Build.mak to Config.mak necessary?

No, I did it only because I have noticed it. Reason is that it is recommended by MakD documentation :P

nemanja-boric-sociomantic:
Fair enough. Just mention it in the commit message (as it will surely confuse the many) and LGTM.

mathias-lang-sociomantic:
Exactly what Nemanja said, add a message to the first commit and LGTM.

Config.mak
```diff
@@ -3,11 +3,15 @@ export ASSERT_ON_STOMPING_PREVENTION=1
 
 override DFLAGS += -w -version=GLIBC
 
-ifeq ($(DVER),1)
-override DFLAGS := $(filter-out -di,$(DFLAGS)) -v2 -v2=-static-arr-params -v2=-volatile
-else
+ifeq ($(DVER),2)
 # Open source Makd uses dmd by default
 DC ?= dmd
```
leandro-lucarella-sociomantic: I don't get this, if we are in D2, then assign the same as makd uses by default?

leandro-lucarella-sociomantic (changes requested):
One comment, plus what @nemanja-boric-sociomantic said, plus please also add more clarifications on the other commit message saying exactly why is this needed for upstream test. I don't really get what's going on with these changes or why they are necessary.

leandro-lucarella-sociomantic:
Like "if DC is blah, we do bleh to avoid X doing Y" or something on those lines.

mihails-strasuns-sociomantic:
Updated with more descriptive commits

mihails-strasuns-sociomantic:
@leandro-lucarella-sociomantic if we can get this in today, I will be able to include it in v4.0.0-alpha.1 and start testing upstream ^_^

Config.mak
nemanja-boric-sociomantic: Ping @mihails-strasuns-sociomantic (I know it's not your fault ;-) ). This is the left over when we had `DC ?= dmd-transitional` here IIRC.

Config.mak
```diff
@@ -1,17 +1,19 @@
...
+# Common D compiler flags
 override DFLAGS += -w -version=GLIBC
 
+# Treat deprecations as errors to ensure ocean doesn't use own deprecated
+# symbols internally
 ifeq ($(DVER),1)
 override DFLAGS := $(filter-out -di,$(DFLAGS)) -v2 -v2=-static-arr-params -v2=-volatile
 else
-# Open source Makd uses dmd by default
-DC ?= dmd
```
nemanja-boric-sociomantic: Ah, you sorted it out.

nemanja-boric-sociomantic (approved):

Config.mak
leandro-lucarella-sociomantic: Ah, right, it was just moved in this commit.

Config.mak
```diff
@@ -0,0 +1,17 @@
+# This will make D2 unittests fail if stomping prevention is triggered
+export ASSERT_ON_STOMPING_PREVENTION=1
+
+override DFLAGS += -w -version=GLIBC
+
+ifeq ($(DVER),1)
```
leandro-lucarella-sociomantic: Something to have in mind here, this will not do what you want if `DVER` is not passed explicitly to `make`, because `DVER` is defined **after** `Config.mak` was read (look at the docs, it says there are only a few variables available to `Config.mak`). I think this should be kept in `Build.mak`.

Config.mak
```diff
@@ -0,0 +1,17 @@
...
+override DFLAGS := $(filter-out -di,$(DFLAGS)) -v2 -v2=-static-arr-params -v2=-volatile
```
leandro-lucarella-sociomantic: Same for `DFLAGS`, you can define the defaults in here, but you can't `filter-out` because it's not defined at this stage.

Config.mak
```diff
@@ -6,12 +6,15 @@ export ASSERT_ON_STOMPING_PREVENTION=1
 override DFLAGS += -w -version=GLIBC
 
 # Treat deprecations as errors to ensure ocean doesn't use own deprecated
-# symbols internally
-ifeq ($(DVER),1)
-override DFLAGS := $(filter-out -di,$(DFLAGS)) -v2 -v2=-static-arr-params -v2=-volatile
-else
+# symbols internally. Only exception is upstream D2 compiler which may introduce
+# new language deprecations ocean was not yet adjusted to.
```
leandro-lucarella-sociomantic: This also seems like a very niche case (the one testing the development compiler). What about adding some option `DMD_DEVEL` or something, to indicate we are testing with a in-development compiler, because if we would like to test with a final and stable upstream dmd, we would want to error on deprecations, right?

Config.mak
```diff
@@ -6,12 +6,15 @@ export ASSERT_ON_STOMPING_PREVENTION=1
...
+ifeq ($(DVER),2)
+ifeq ($(DC),dmd-transitional)
 override DFLAGS += -de
 endif
+else
```
leandro-lucarella-sociomantic: Maybe introduce some indentation to make more clear which `else` belongs to each `ifeq`?

leandro-lucarella-sociomantic (changes requested):

leandro-lucarella-sociomantic:
> I will be able to include it in v4.0.0-alpha.1

Can we **please** use `-alpha1` so it's consistent with the `-rcX` releases? :innocent:

Config.mak
mihails-strasuns-sociomantic: I don't expect it matter in practice, but sounds reasonable.

mihails-strasuns-sociomantic:
Updated.

mihails-strasuns-sociomantic:
> Can we please use -alpha1 so it's consistent with the -rcX releases?

For the reference to observers - https://github.com/sociomantic-tsunami/neptune/issues/132 :)

Build.mak
```diff
@@ -1,17 +1,3 @@
-# This will make D2 unittests fail if stomping prevention is triggered
```
leandro-lucarella-sociomantic, line 1: Typo in commit message "Move port ..." should be "part" I guess.

Build.mak
```diff
@@ -1,8 +1,6 @@
 ifeq ($(DVER),1)
 override DFLAGS := $(filter-out -di,$(DFLAGS)) -v2 -v2=-static-arr-params -v2=-volatile
 else
-# Open source Makd uses dmd by default
-DC ?= dmd
```
leandro-lucarella-sociomantic: Typo in commit message "confusin" -> "confusing", maybe add that it's removed because is the makd default.

Build.mak
```diff
@@ -1,9 +1,12 @@
 # Treat deprecations as errors to ensure ocean doesn't use own deprecated
-# symbols internally
-ifeq ($(DVER),1)
-override DFLAGS := $(filter-out -di,$(DFLAGS)) -v2 -v2=-static-arr-params -v2=-volatile
-else
-override DFLAGS += -de
+# symbols internally. Disable it on explicit flag to make possible regression
+# testing in D upstream.
+ifndef ALLOW_DEPRECATIONS
```
leandro-lucarella-sociomantic: I won't say this is a blocker, but I would compare against `ALLOW_DEPRECATIONS=1` to avoid confusion if someone thinks it can explicitly disable deprecations by using `ALLOW_DEPRECATIONS=0`, which is a common practice in Makd.

leandro-lucarella-sociomantic (approved):
Minus minor comments, look good to me, feel free to self-merge once the typos are fixed!

PR #375, opened 2017-11-26
Add epoll instance / main Task to app base class

mihails-strasuns-sociomantic:
Should lift some boilerplate burden from app developers.

### DIFF
--- a/relnotes/taskext.feature.md
+++ b/relnotes/taskext.feature.md
@@ -1,0 +1,70 @@
+## Built-in task and scheduler support in DaemonApp
+
+`ocean.util.app.DaemonApp` `ocean.util.app.ext.TaskExt`
+
+It is now possible to make `DaemonApp` initialize scheduler automatically using
+standard app configuration file. To do so, it is sufficient to enable it via
+`DaemonApp` optional settings:
+
+```D
+class MyApp : DaemonApp
+{
+  this ( )
+  {
+    DaemonApp.OptionalSettings settings;
+    settings.use_task_ext = true;
+    super("name", "desc", VersionInfo.init, settings);
+  }
+```
+
+This extension is disabled by default to avoid clashing with existing
+application code which is already handling scheduler initialization in some way.
+
+If `TaskExt` is enabled, overriden `run` method will be called within a task
+context and there is no need to create one manually or start scheduler event
+loop:
+
+```D
+class MyApp : DaemonApp
+{
+  this ( )
+  {
+    DaemonApp.OptionalSettings settings;
+    settings.use_task_ext = true;
+    super("name", "desc", VersionInfo.init, settings);
+  }
+
+  override int run ( Arguments args, ConfigParser config )
+  {
+    assert(Task.getThis() !is null);
+    assert(isSchedulerUsed());
+
+    scope (exit)
+    {
+      // must be called at some point, registered DaemonApp events won't let
+      // event loop exit on its own:
+      theScheduler.epoll.shutdown();
+      // theScheduler.shutdown(); // alternative to also hard kill
+                                  // all remaining tasks
+    }
+
+    return 0;
+  }
+```
+
+NB: using `TaskExt` results in `startEventHandling` being called automatically,
+    do not call it again from the `run` method!
+
+Expected configuration file format matches `SchedulerConfiguration` struct
+fields (all field are optional, defaults will be used if not defined):
+
+```
+[SCHEDULER]
+worker_fiber_stack_size = 102400
+worker_fiber_limit = 5
+task_queue_limit = 10
+suspended_task_limit = 16
+specialized_pools =
+    pkg.mod.MyTask:1024
+    pkg.mod.MyOtherTask:2048
+```
--- a/src/ocean/util/app/DaemonApp.d
+++ b/src/ocean/util/app/DaemonApp.d
@@ -73,6 +73,7 @@ public abstract class DaemonApp : Application,
     import ocean.util.app.ext.ReopenableFilesExt;
     import ocean.util.app.ext.PidLockExt;
     import ocean.util.app.ext.UnixSocketExt;
+    import ocean.util.app.ext.TaskExt;
     import ocean.util.app.ExitException;
     import ocean.util.log.Log;
     import NewLog = ocean.util.log.Logger;
@@ -193,6 +194,14 @@ public abstract class DaemonApp : Application,
 
     /***************************************************************************
 
+        Extension to start `run` method inside a task.
+
+    ***************************************************************************/
+
+    public TaskExt task_ext;
+
+    /***************************************************************************
+
         Cpu and memory collector instance.
 
     ***************************************************************************/
@@ -309,6 +318,15 @@ public abstract class DaemonApp : Application,
 
         /// Delegate for LogExt that instantiates a `Appender.Layout` from a name
         Appender.Layout delegate (cstring name) make_layout;
+
+        /***********************************************************************
+
+            By default TaskExt is disabled to prevent breaking change for
+            applications already configuring scheduler on their own.
+
+        ***********************************************************************/
+
+        bool use_task_ext;
     }
 
     /***************************************************************************
@@ -383,6 +401,12 @@ public abstract class DaemonApp : Application,
         this.unix_socket_ext = new UnixSocketExt();
         this.config_ext.registerExtension(this.unix_socket_ext);
         this.registerExtension(this.unix_socket_ext);
+
+        if (settings.use_task_ext)
+        {
+            this.task_ext = new TaskExt();
+            this.config_ext.registerExtension(this.task_ext);
+        }
 
         // Create and register repoenable files extension
         this.reopenable_files_ext = new ReopenableFilesExt();
@@ -460,7 +484,13 @@ public abstract class DaemonApp : Application,
 
     override protected int run ( istring[] args )
     {
-        return this.run(this.args, this.config);
+        if (this.task_ext is null)
+            return this.run(this.args, this.config);
+
+        this.startEventHandling(theScheduler.epoll());
+        return this.task_ext.run({
+            return this.run(this.args, this.config);
+        });
     }
 
     /***************************************************************************
--- a/src/ocean/util/app/ext/TaskExt.d
+++ b/src/ocean/util/app/ext/TaskExt.d
@@ -1,0 +1,191 @@
+/*******************************************************************************
+
+    Extension to automate typical task scheduler setup.
+
+    Reads scheduler configuration from app config from [SCHEDULER] section,
+    initializes global scheduler instance using it and provides a method to
+    start main application method inside a task.
+
+    Expected configuration file format (all fields are optional):
+
+    [SCHEDULER]
+    worker_fiber_stack_size = 102400
+    worker_fiber_limit = 5
+    task_queue_limit = 10
+    suspended_task_limit = 16
+    specialized_pools =
+        pkg.mod.MyTask:1024
+        pkg.mod.MyOtherTask:2048
+
+    Copyright:
+        Copyright (c) 2017 sociomantic labs GmbH.
+        All rights reserved.
+
+    License:
+        Boost Software License Version 1.0. See LICENSE_BOOST.txt for details.
+        Alternatively, this file may be distributed under the terms of the Tango
+        3-Clause BSD License (see LICENSE_BSD.txt for details).
+
+*******************************************************************************/
+
+module ocean.util.app.ext.TaskExt;
+
+import ocean.transition;
+import ocean.core.array.Search : find;
+import ocean.core.Enforce;
+import ocean.text.convert.Integer;
+import ocean.util.app.ext.model.IConfigExtExtension;
+import ocean.task.Scheduler;
+import ocean.task.Task;
+import ocean.util.config.ConfigParser;
+import ocean.util.config.ConfigFiller;
+import ocean.meta.codegen.Identifier;
+
+/// ditto
+class TaskExt : IConfigExtExtension
+{
+    /***************************************************************************
+
+        Parse the configuration file options to get the scheduler configuration
+
+        Params:
+            app = the application instance
+            parser = configuration instance
+
+    ***************************************************************************/
+
+    public override void processConfig ( IApplication app, ConfigParser parser )
+    {
+        SchedulerConfiguration scheduler_config;
+
+        const category = "SCHEDULER";
+
+        foreach (idx, ref field; scheduler_config.tupleof)
+        {
+            static if (fieldIdentifier!(SchedulerConfiguration, idx)
+                != "specialized_pools")
+            {
+                field = parser.get(
+                    category,
+                    fieldIdentifier!(SchedulerConfiguration, idx),
+                    field
+                );
+            }
+        }
+
+        auto specialized_pools = parser.getList!(istring)(
+            category, "specialized_pools", null);
+
+        foreach (line; specialized_pools)
+        {
+            if (line.length == 0)
+                continue;
+
+            auto idx = find(line, ':');
+            enforce(
+                idx < line.length,
+                "Malformed configuration for scheduler"
+            );
+
+            size_t size;
+            enforce(
+                toInteger(line[idx+1 .. $], size),
+                "Malformed configuration for scheduler"
+            );
+
+            scheduler_config.specialized_pools ~=
+                SchedulerConfiguration.PoolDescription(line[0 .. idx], size);
+        }
+
+        initScheduler(scheduler_config);
+    }
+
+    /***************************************************************************
+
+        Wraps given delegate returning app exit status code in a new allocated
+        task object and immediately starts the scheduler event loop to handle
+        it.
+
+        Params:
+            dg = delegate forwarding to the app entry point
+
+        Returns:
+            app return status code
+
+    ***************************************************************************/
+
+    public int run ( int delegate () dg )
+    {
+        int result = -1;
+
+        auto task = new class Task {
+            override void run ( )
+            {
+                result = dg();
+            }
+        };
+
+        theScheduler.queue(task);
+        theScheduler.eventLoop();
+
+        return result;
+    }
+
+    /***************************************************************************
+
+        Extension order. Doesn't matter as long as it happens after LogExt.
+
+        Returns:
+            the extension order
+
+    ***************************************************************************/
+
+    public override int order ( )
+    {
+        return 0;
+    }
+
+    /***************************************************************************
+
+        Unused IConfigExtExtension method.
+
+        We just need to provide an "empty" implementation to satisfy the
+        interface.
+
+        Params:
+            app = the application instance
+            config = configuration instance
+
+    ***************************************************************************/
+
+    public override void preParseConfig ( IApplication app, ConfigParser config )
+    {
+        // Unused
+    }
+
+
+    /***************************************************************************
+
+        Unused IConfigExtExtension method.
+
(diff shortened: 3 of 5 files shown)


src/ocean/util/app/ext/TaskExt.d
```diff
@@ -0,0 +1,195 @@
...
+        }
+
+        // can't fill SchedulerConfiguration directly because it contains
+        // `PoolDescription` array which has to be provided via code
+
+        SchedulerConfiguration scheduler_config;
+        PartialSchedulerConfiguration partial_config;
+        // fill with init values
+        foreach (idx, ref elem; partial_config.tupleof)
+            elem = scheduler_config.tupleof[idx];
```
nemanja-boric-sociomantic: Perhaps we should `assert` that the names of the fields are the same, otherwise bugs can cripple in if you change `SchedulerConfiguration` definition, but still have the matching field types.

src/ocean/util/app/ext/TaskExt.d
```diff
@@ -0,0 +1,195 @@
+/*******************************************************************************
+
+    Extension to automate typical task scheduler setup.
+
+    Reads scheduler configuration from app config from [SCHEDULER] section,
+    initializes global scheduler instance using it and provides a method to
+    start main application method inside a task.
+
+    Copyright:
+        Copyright (c) 2007 sociomantic labs GmbH.
```
nemanja-boric-sociomantic: 2017

src/ocean/util/app/DaemonApp.d
```diff
@@ -460,7 +472,13 @@ public abstract class DaemonApp : Application,
 
     override protected int run ( istring[] args )
     {
-        return this.run(this.args, this.config);
+        if (!this.task_ext.enabled)
+            return this.run(this.args, this.config);
+
+        this.startEventHandling(theScheduler.epoll());
```
nemanja-boric-sociomantic: There should be a release notes note that this is not required with `TaskExt`

test/taskext/main.d
```diff
@@ -0,0 +1,52 @@
+import ocean.transition;
```
nemanja-boric-sociomantic, line 16: Missing module info.

relnotes/taskext.feature.md
```diff
@@ -0,0 +1,64 @@
...
+{
+    this ( )
+    {
+        super("name", "desc", VersionInfo.init);
+        this.task_ext.enable(); // NB: must be after `super` call
+    }
+```
+
+This extension is disabled by default to avoid clashing with existing
+application code which is already handling scheduler initialization in some way.
```
leandro-lucarella-sociomantic: Are daemonapps always going to use an scheduler? If so I guess the default should be changed in the next major? If so, maybe add an issue so we don't forget?

mihails-strasuns-sociomantic: Eventually, but AFAIK we have plenty of apps that haven't switched to task system yet. Also there is a problem with providing `PoolDescription` configuration which requires manual intervention anyway. I am not convinced there is any benefit in going for this direction.

src/ocean/util/app/ext/TaskExt.d
mihails-strasuns-sociomantic: Even better - use reflection to access required fields by name ;)

mihails-strasuns-sociomantic:
Updated per @nemanja-boric-sociomantic comments

src/ocean/util/app/ext/TaskExt.d
```diff
@@ -0,0 +1,203 @@
...
+    ***************************************************************************/
+
+    public int run ( int delegate () dg )
+    {
+        int result = -1;
+
+        auto task = new class Task {
+            override void run ( )
+            {
+                result = dg();
```
daniel-zullo, line 124: What would it happen if the `dg()` throws?

mihails-strasuns-sociomantic: Same as with any other throwing task - depends on the state of the https://github.com/sociomantic-tsunami/ocean/blob/3282758547765ac4dcceca1c81a42a49e8f138e5/src/ocean/task/Scheduler.d#L164-L175

By default it simply rethrows the exception and that will propagate to whatever context has resumed the task. I have updated the test to show one such case.

src/ocean/util/app/ext/TaskExt.d
```diff
@@ -0,0 +1,203 @@
...
+
+    /***************************************************************************
+
+        Wraps given delegate returning app exit status code in a new allocated
+        task object and immediately starts the scheduler event loop to handle
+        it.
+
+        Params:
+            dg = delegate forwarding to the app entry point
+
```
daniel-zullo, line 111: `Returns:` is missing

test/taskext/main.d
```diff
@@ -0,0 +1,79 @@
+/*******************************************************************************
+
+    Test for ocean.util.app.ext.TaskExt
+
+    Copyright:
+        Copyright (c) 2017 sociomantic labs GmbH.
+        All rights reserved.
```
daniel-zullo, line 7: Shouldn't it be written in one line?
`Copyright: Copyright (c) 2017 sociomantic labs GmbH. All rights reserved`

src/ocean/util/app/ext/TaskExt.d
```diff
@@ -0,0 +1,203 @@
...
+    Extension to automate typical task scheduler setup.
+
+    Reads scheduler configuration from app config from [SCHEDULER] section,
+    initializes global scheduler instance using it and provides a method to
+    start main application method inside a task.
+
+    Copyright:
+        Copyright (c) 2017 sociomantic labs GmbH.
+        All rights reserved.
```
daniel-zullo: ditto
`Copyright: Copyright (c) 2017 sociomantic labs GmbH. All rights reserved`

daniel-zullo (commented):
Just a few cosmetic comments

relnotes/taskext.feature.md
```diff
@@ -0,0 +1,67 @@
...
+  this ( )
+  {
+    super("name", "desc", VersionInfo.init);
+    this.task_ext.enable();
+  }
+
+  override int run ( Arguments args, ConfigParser config )
+  {
+    assert(Task.getThis() !is null);
+    assert(isSchedulerUsed());
```
daniel-zullo, line 40: Shouldn't we use `verify()` instead for this cases?

mihails-strasuns-sociomantic: Interesting question. In actual app code you shouldn't have any of these checks as they are 100% guaranteed to be true and present here only for demonstration. And for example purpose I use `assert` to minimize amount of assumed imported context. Don't feel particularly strong either way though.

test/taskext/main.d
mihails-strasuns-sociomantic: DDOC should render it identically.

relnotes/taskext.feature.md
daniel-zullo: I see your points. 
(It might also be good to promote the usage of `verify()` through the examples though)

test/taskext/main.d
daniel-zullo: Fair enough

relnotes/taskext.feature.md
```diff
@@ -0,0 +1,67 @@
...
+standard app configuration file. To do so, it is sufficient call
+`task_ext.enable` in the constructor:
+
+```D
+class MyApp : DaemonApp
+{
+    this ( )
+    {
+        super("name", "desc", VersionInfo.init);
+        this.task_ext.enable(); // NB: must be after `super` call
```
gavin-norman-sociomantic: A more "normal" seeming way would be to add a flag (`bool enable_task_ext = false;`) to the options struct passed to `DaemonApp`'s ctor. Was there a reason why this wouldn't work?

relnotes/taskext.feature.md
```diff
@@ -0,0 +1,67 @@
...
+
+It is also possible to supply specialized fiber pool description as an argument
+to `enable`. If present, it will be forwarded to `initScheduler` call without
+any changes.
+
+```D
+with (SchedulerConfiguration)
+{
+  this.task_ext.enable([
+    PoolDescription(SomeTask.classinfo, 10240)
```
gavin-norman-sociomantic: I _guess_ this could also be included as a field of the options struct...?

src/ocean/util/app/ext/TaskExt.d
```diff
@@ -0,0 +1,203 @@
...
+            size_t worker_fiber_limit =
+                SchedulerConfiguration.worker_fiber_limit.init;
+            size_t task_queue_limit =
+                SchedulerConfiguration.task_queue_limit.init;
+            size_t suspended_task_limit =
+                SchedulerConfiguration.suspended_task_limit.init;
+        }
+
+        // can't fill SchedulerConfiguration directly because it contains
+        // `PoolDescription` array which has to be provided via code
```
gavin-norman-sociomantic: Hmm, that's kind of annoying :/

gavin-norman-sociomantic (commented):
Looks good! A few small comments...

src/ocean/util/app/ext/TaskExt.d
mihails-strasuns-sociomantic: One idea that just came to my mind is that we can use fully-qualified class name which is converted into `ClassInfo` later internally. That way it can be stored in config file.

gavin-norman-sociomantic: Ahh I was wondering if there's a runtime way to go from a string to a `ClassInfo`. Would be cool if there is.

mihails-strasuns-sociomantic: There is https://github.com/sociomantic-tsunami/tangort/blob/32ad03f41da038b6e315931c551ddfe4cd0d26eb/src/rt/include/object.di#L102 but it has the same limitation as all runtime reflection facilities - it completely ignores all templated classes.

gavin-norman-sociomantic: Oof. Even an alias to a templated class?

mihails-strasuns-sociomantic: Yep. Only way to make it work is to make a dummy non-templated class that inherits from a templated one.

gavin-norman-sociomantic: Ok so there's at least a workaround. Do you think the trouble (fully qualified names / potential workarounds required) is worth it?

mihails-strasuns-sociomantic: Not yet sure. Being able to adjust fiber stack sizes via config file sounds quite desirable when you think about something like connection handler in propulsor. But introducing brand new feature with such inherent flaw is a bit awkward.

mihails-strasuns-sociomantic:
Updated to use full text configuration and `DaemonApp.OptionalSettings` - I am certain you will like this version more @gavin-norman-sociomantic :)

src/ocean/util/app/ext/TaskExt.d
nemanja-boric-sociomantic: still missing `Return:`

relnotes/taskext.feature.md
```diff
@@ -0,0 +1,70 @@
...
+fields (all field are optional, defaults will be used if not defined):
+
+```
+[SCHEDULER]
+worker_fiber_stack_size = 102400
+worker_fiber_limit = 5
+task_queue_limit = 10
+suspended_task_limit = 16
+specialized_pools =
+    pkg.mod.MyTask:1024
```
gavin-norman-sociomantic, line 68: Nice!

src/ocean/util/app/DaemonApp.d
```diff
@@ -384,6 +402,12 @@ public abstract class DaemonApp : Application,
         this.config_ext.registerExtension(this.unix_socket_ext);
         this.registerExtension(this.unix_socket_ext);
 
+        if (settings.use_task_ext)
+        {
+            this.task_ext = new TaskExt();
+            this.config_ext.registerExtension(this.task_ext);
```
gavin-norman-sociomantic, line 408: I think you need `this.registerExtension(this.task_ext);` here too. (Though I forget what it does exactly.)

test/taskext/main.d
```diff
@@ -0,0 +1,94 @@
...
+        scope (success)
+            remove();
+        scope (failure)
+            exitSandbox();
+
+        File.set("etc/config.ini", "[SCHEDULER]
+worker_fiber_limit = 3
+specialized_pools =
+    main.AnotherTask:10240
+    whatever:1024
```
gavin-norman-sociomantic, line 84: What does this line in the config file do?

gavin-norman-sociomantic (commented):
Yes, this is looking good :+1:

test/taskext/main.d
mihails-strasuns-sociomantic: Nothing, just checking that one can parse multi-line value

mihails-strasuns-sociomantic:
```diff
diff --git a/src/ocean/util/app/DaemonApp.d b/src/ocean/util/app/DaemonApp.d
index ded2505..c0e1d81 100644
--- a/src/ocean/util/app/DaemonApp.d
+++ b/src/ocean/util/app/DaemonApp.d
@@ -406,6 +406,7 @@ public abstract class DaemonApp : Application,
         {
             this.task_ext = new TaskExt();
             this.config_ext.registerExtension(this.task_ext);
+            this.registerExtension(this.task_ext);
         }
 
         // Create and register repoenable files extension
diff --git a/src/ocean/util/app/ext/TaskExt.d b/src/ocean/util/app/ext/TaskExt.d
index 3646c1d..6f3cde0 100644
--- a/src/ocean/util/app/ext/TaskExt.d
+++ b/src/ocean/util/app/ext/TaskExt.d
@@ -109,6 +109,9 @@ class TaskExt : IConfigExtExtension
         Params:
             dg = delegate forwarding to the app entry point
 
+        Returns:
+            app return status code
+
     ***************************************************************************/
 
     public int run ( int delegate () dg )
```

mihails-strasuns-sociomantic:
Rolled back `this.registerExtension(this.task_ext)` as I am not implementing `IApplicationExtension`

mihails-strasuns-sociomantic:
Reverted `this.registerExtension(this.task_ext)` as implementing `IApplicationExtension` for `TaskExt` adds nothing but 4 more dummy method overrides.

gavin-norman-sociomantic:
>Reverted this.registerExtension(this.task_ext) as implementing IApplicationExtension for TaskExt adds nothing but 4 more dummy method overrides.

Ha, interesting. Wait, though... doesn't this make it impossible for other extensions to query for the presence of the task extension and don't we need that capability in order to be able to hook epoll-using extensions into it?

mihails-strasuns-sociomantic:
> Wait, though... doesn't this make it impossible for other extensions to query for the presence of the task extension and don't we need that capability in order to be able to hook epoll-using extensions into it?

I don't see how that would matter here. All https://github.com/sociomantic-tsunami/ocean/blob/5c01dc0adee97b833054a8cd1efd0ca29c8d8cea/src/ocean/util/app/model/ExtensibleClassMixin.d does is storing array of interfaces and allows querying if specific class is present among them. And every extension that needs to be hooked in defines own array of extensions + own interface. As such, only place where you could query presence of specific extension is the one that has access to `DaemonApp` instance and at that point you can do so by simply checking `this.task_ext is null`.

And it doesn't matter in the end because only check epoll-using extensions need is `if (isSchedulerUsed()) epoll = theScheduler.epoll()`

gavin-norman-sociomantic:
Fair enough. It's a long time since I've looked at the fine details of the app framework.

gavin-norman-sociomantic (approved):

relnotes/taskext.feature.md
```diff
@@ -0,0 +1,70 @@
...
+    super("name", "desc", VersionInfo.init, settings);
+  }
+```
+
+This extension is disabled by default to avoid clashing with existing
+application code which is already handling scheduler initialization in some way.
+
+If `TaskExt` is enabled, overriden `run` method will be called within a task
+context and there is no need to create one manually or start scheduler event
+loop:
```
joseph-wakeling-sociomantic, line 25: I didn't realize this was under development, but just to raise a few key concerns here:

* having `run` automatically run inside a task means that there is no longer any reliable way to do things in `run` before the scheduler (and hence, epoll) event loop starts.  This may have some implications in terms of separation of concerns between when the app's data structures and functionality are set up, versus when they start _doing_ something.

* having `run` work within a task makes changes to how exceptions thrown from within it are handled, even more so if (for example) suspend and resume are involved.

Given that it's (for now) an optional change, I dare say we'll work these things out.  But both these things could have some implications for app transitions to use this new functionality.

relnotes/taskext.feature.md
joseph-wakeling-sociomantic: @mihails-strasuns-sociomantic Forgive me if I'm just missing things, but I don't see any description anywhere either in these release notes or in code about what this `specialized_pools` config actually means.  Can you explain ... ?

mihails-strasuns-sociomantic: This is another feature coming in 3.6.0 - https://github.com/sociomantic-tsunami/ocean/pull/293

joseph-wakeling-sociomantic: Nice, I like it!  FWIW it might be helpful if the release notes explained the "magic numbers" in the example, as it wasn't clear (until I read the detailed `PoolDescription` code) that these were stack sizes.  I do like that VERY much as a feature, though, because customizing stack size requirements in this way is likely very useful.

relnotes/taskext.feature.md
```diff
@@ -0,0 +1,70 @@
...
+  {
+    assert(Task.getThis() !is null);
+    assert(isSchedulerUsed());
+
+    scope (exit)
+    {
+      // must be called at some point, registered DaemonApp events won't let
+      // event loop exit on its own:
+      theScheduler.epoll.shutdown();
+      // theScheduler.shutdown(); // alternative to also hard kill
```
joseph-wakeling-sociomantic, line 47: Does the use of these `scope (exit)` options assume that at some point in the `run` method, there will be some kind of epoll-using thing that happens that blocks until the app is done?

I'm thinking of the use-case of a `DaemonApp`, where the `run` method might spin off a number of things that repeatedly schedule various epoll events, but which doesn't do anything that _blocks_.  In that case one might exit the run method and therefore trigger the `shutdown` call, when actually one wants the eventloop to keep running.

mihails-strasuns-sociomantic: Any `DaemonApp` has such epoll using things from timer ext and periodic stats log. If you want to keep everything running (and have some own condition for terminating things), you can simply remove shutdown line - in that case main task will exit but everything else will keep running.

joseph-wakeling-sociomantic: Fair enough, but it might be worth considering whether the release notes could be tweaked a bit.  This documented example might be a bit misleading, as I suspect quite a lot of practical application uses don't have anything that blocks long-term -- they just spin up a bunch of different things that keep triggering new tasks and/or epoll events.

So, I'm not sure that `scope (exit) theScheduler.epoll.shutdown();` is advisable in many circumstances, although it's obviously useful to have it listed as a possible option.

mihails-strasuns-sociomantic: Example was only to show difference with old `run` method which would end simultaneously with the end of an app. Any phrasing suggestions for release notes?

joseph-wakeling-sociomantic: I'll try to think of something, and if I come up with an idea, submit a PR.

PR #378, opened 2017-11-27
Add `Time.fromUnixTime` method

### LINKED ISSUE #188: Add `Time.fromUnixTime` to `ocean.time.Time`
Add a static function to `Time` named for example `fromUnixTime` that converts a `time_t` to a `Time`. There seems to be no function of such functionality in ocean.

nemanja-boric-sociomantic:
`Time` structure now contains static method `fromUnixTime` which creates
`Time` structure instance corresponding to the unix time (relative to 1/1/1970).

Fixes #188

### DIFF
--- a/src/ocean/time/Time.d
+++ b/src/ocean/time/Time.d
@@ -23,6 +23,14 @@ module ocean.time.Time;
 import ocean.transition;
 import ocean.text.convert.DateTime_tango;
 
+import core.stdc.time: time_t;
+
+version (UnitTest)
+{
+    import core.stdc.time;
+    import ocean.core.Test;
+}
+
 /******************************************************************************
 
     This struct represents a length of time.  The underlying representation is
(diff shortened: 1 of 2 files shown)


src/ocean/time/Time.d
```diff
@@ -697,6 +703,23 @@ struct Time
...
+
+            Params:
+                unix_time = number of seconds since 1/1/1970
+
+            Returns:
+                Time instance corresponding the given unix time.
+
+        ***********************************************************************/
+
+        static Time fromUnixTime (long unix_time)
```
david-eckardt-sociomantic: Maybe use `time_t`? This method is useful for interfacing with POSIX.

nemanja-boric-sociomantic: Ah, good point, I have no idea how I missed that.

nemanja-boric-sociomantic: Updated.

david-eckardt-sociomantic (approved):

PR #379, opened 2017-11-27
Run PidLock extension before LogExt

### LINKED ISSUE #107: PidLock extension should run before loggers
In case logging system doesn't have enough permission to write to the log file, it can happen that
the application fails on permission denied exception, before trying to look for the other application instance.

While this  effectively prevents the application to run, it would be nice if the `PidLock` extension would check if the another instance is already running, before any other systems are started.

nemanja-boric-sociomantic:
PidLock extension require ConfigExt (to read the path to the PidLock
file) but it should be run before LogExt, since the LogExt can try to
perform sideeffects which should not be allowed if another instance of
the application is already running.

Fixes #107

### DIFF
--- a/src/ocean/util/app/ext/PidLockExt.d
+++ b/src/ocean/util/app/ext/PidLockExt.d
@@ -77,7 +77,8 @@ class PidLockExt : IConfigExtExtension, IApplicationExtension
 
     /***************************************************************************
 
-        Order doesn't matter, so return default -> 0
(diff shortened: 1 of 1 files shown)


leandro-lucarella-sociomantic (approved):

leandro-lucarella-sociomantic:
I guess is hard to do a test for this, right?

mihails-strasuns-sociomantic (approved):
Needs rebase

nemanja-boric-sociomantic:
rebased.

PR #381, opened 2017-11-27
Make `IByteQueue.push` accept an `in` buffer argument

david-eckardt-sociomantic:
This method does not modify the input so it should accept `const` data.

Trying to apply the same to [`IQueue.push`](https://github.com/sociomantic-tsunami/ocean/blob/v3.4.x/src/ocean/util/container/queue/model/IQueue.d#L75) I found it causes a dilemma in `FixedRingQueue.push`, which comes down to
```d
bool push ( in T element )
{
    // ...
    *element_in_queue = element;
}
```
This is impossible if `T` is a class. If `typeof(element_in_queue)` is `T*` then the `const` from `element` would be lost. But if `typeof(element_in_queue)` is `const(T)*` then one cannot assign to `*element_in_queue`.

### DIFF
--- a/src/ocean/util/container/queue/FixedRingQueue.d
+++ b/src/ocean/util/container/queue/FixedRingQueue.d
@@ -32,6 +32,8 @@ import ocean.util.container.mem.MemManager;
 
 version(UnitTest) import ocean.core.Test;
 
+import ocean.transition;
+
 
 /*******************************************************************************
 
@@ -185,7 +187,7 @@ class FixedByteRingQueue : FixedRingQueueBase!(IByteQueue)
 
     ***************************************************************************/
 
-    bool push ( ubyte[] element )
+    bool push ( in ubyte[] element )
     {
         verify (element.length == super.element_size, "element size mismatch");
 
@@ -193,7 +195,8 @@ class FixedByteRingQueue : FixedRingQueueBase!(IByteQueue)
 
         if (element_in_queue)
         {
-            element_in_queue[] = element[];
+            Const!(void)[] element_ = element;
+            element_in_queue[] = element_[];
             return true;
         }
         else
--- a/src/ocean/util/container/queue/FlexibleRingQueue.d
+++ b/src/ocean/util/container/queue/FlexibleRingQueue.d
@@ -751,7 +751,7 @@ class FlexibleByteRingQueue : IRingQueue!(IByteQueue)
 
     ***************************************************************************/
 
-    private static void validate ( ExportMetadata meta, void[] data )
+    private static void validate ( in ExportMetadata meta, in void[] data )
     {
         if (meta.items)
         {
@@ -770,7 +770,7 @@ class FlexibleByteRingQueue : IRingQueue!(IByteQueue)
 
             size_t pos = 0;
 
-            for (typeof(meta.items) i = 0; i < meta.items; i++)
+            for (uint i = 0; i < meta.items; i++)
             {
                 verify(pos <= data.length);
 
@@ -789,7 +789,7 @@ class FlexibleByteRingQueue : IRingQueue!(IByteQueue)
                             itoa(start))
                     );
 
-                    auto header = cast(Header*)data[start .. pos].ptr;
+                    auto header = cast(Const!(Header)*)data[start .. pos].ptr;
 
                     enforce!(ValidationError)(
                         (data.length - pos) >= header.length,
--- a/src/ocean/util/container/queue/FlexibleRingQueue.d
+++ b/src/ocean/util/container/queue/FlexibleRingQueue.d
@@ -171,7 +171,7 @@ class FlexibleByteRingQueue : IRingQueue!(IByteQueue)
 
     ***************************************************************************/
 
-    public bool push ( ubyte[] item )
+    public bool push ( in ubyte[] item )
     {
         auto data = this.push(item.length);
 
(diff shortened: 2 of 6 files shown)


src/ocean/util/container/queue/FlexibleRingQueue.d
```diff
@@ -783,7 +783,7 @@ class FlexibleByteRingQueue : IRingQueue!(IByteQueue)
 
             size_t pos = 0;
 
-            for (typeof(meta.items) i = 0; i < meta.items; i++)
+            for (uint i = 0; i < meta.items; i++)
```
mathias-lang-sociomantic, line 773: `size_t`

src/ocean/util/container/queue/FixedRingQueue.d
```diff
@@ -196,7 +196,7 @@ class FixedByteRingQueue : FixedRingQueueBase!(IByteQueue)
 
         if (element_in_queue)
         {
-            element_in_queue[] = element[];
+            (cast(ubyte[])element_in_queue)[] = element[];
```
mathias-lang-sociomantic: This alone is a big no-no

mathias-lang-sociomantic (changes requested):
First I don't think that should be in a patch release. 

Second, why do you need this ? If you need to accept `const` data downstream, you can override the method with `in`.

src/ocean/util/container/queue/FlexibleRingQueue.d
david-eckardt-sociomantic: `typeof(meta.items)` is `uint`.

src/ocean/util/container/queue/FixedRingQueue.d
david-eckardt-sociomantic: What do you suggest instead?

david-eckardt-sociomantic: For clarity: `typeof(element_in_queue)` is `void[]`.

nemanja-boric-sociomantic:
> First I don't think that should be in a patch release.

Out of the interest, what's the reasoning behind this? Historically, we did the D2 const-fixes that just relaxes what the API can accept as bugfixes.

david-eckardt-sociomantic:
> Second, why do you need this ?

I want to pass `const(void)[]` data to `FlexibleByteRingQueue.push`. Just adding `in` there results in an error:
_class `ocean.util.container.queue.FlexibleRingQueue.FlexibleByteRingQueue` interface function `bool push(ubyte[] item)` is not implemented_
So I have to change the interface method and all its implementations, too (and they should accept `const` anyway because they make a copy of the data).

> If you need to accept const data downstream, you can override the method with `in`.

How would I pass it to `super.push` in the override? The problem would be the same.

src/ocean/util/container/queue/FixedRingQueue.d
mihails-strasuns-sociomantic:
```D
void[] element_ = element[];
element_in_queue[] = element_[];
```

mihails-strasuns-sociomantic:
Without test it is not clear what use case you have tried to enable. What is your T? `FlexibleByteRingQueue!(char[])` or `FlexibleByteRingQueue!(const(char)[])`?

src/ocean/util/container/queue/FixedRingQueue.d
david-eckardt-sociomantic: This results in _Error: cannot implicitly convert expression `(element[])` of type `const(ubyte)[]` to `void[]`._

A detailed description of what I'm doing is
```d
// The function argument.
const(ubyte[]) element;
// Get the copy destination buffer, its type is mutable `void[]`.
void[] element_in_queue = super.push_();
// Cast the copy destination buffer from mutable `void[]` to mutable `ubyte[]`.
ubyte[] element_in_queue_ = cast(ubyte[])element_in_queue;
// Copy the array content of `element` into the destination buffer.
element_in_queue_[] = element[];
```
In the actual code I do the last two steps in one line:
```d
(cast(ubyte[])element_in_queue)[] = element[];
```
I don't see which of these steps is wrong.

david-eckardt-sociomantic:
> Without test it is not clear what use case you have tried to enable. What is your `T`? `FlexibleByteRingQueue!(char[])` or `FlexibleByteRingQueue!(const(char)[])`?

`FlexibleByteRingQueue` is not a template, `push` and `pop` use `ubyte[]`.

src/ocean/util/container/queue/FixedRingQueue.d
mihails-strasuns-sociomantic: `Const!(void)[] element_ = element[]` in that case.

What you are doing wrong is using a cast. Every single cast is an armed nuclear warhead - if you can avoid having one in your cupboard, you probably should do so.

david-eckardt-sociomantic: I see, thanks for the clarification.

david-eckardt-sociomantic:
Updated, removing the contentious array cast.

david-eckardt-sociomantic:
Can this be a patch release for the reasons described above (https://github.com/sociomantic-tsunami/ocean/pull/381#issuecomment-347234724)?

mathias-lang-sociomantic:
> Out of the interest, what's the reasoning behind this? Historically, we did the D2 const-fixes that just relaxes what the API can accept as bugfixes.

I think we have enough applications D2-compatible to reconsider this policy. Also adding `const` to a free function parameter's is usually okay, but adding it to a method can lead to overrides not being overrides anymore, thus compilation will halt.

david-eckardt-sociomantic:
Changed the base to v4.x.x. Travis doesn't change it so the build now fails, I have to open a new PR.

david-eckardt-sociomantic:
New PR: #382

PR #382, opened 2017-11-28
Make `IByteQueue.push` accept an `in` buffer argument (remake)

david-eckardt-sociomantic:
This method does not modify the input so it should accept const data.

Remake of #381.

### DIFF
--- a/relnotes/queue-push-const.migration.md
+++ b/relnotes/queue-push-const.migration.md
@@ -1,0 +1,25 @@
+## Queue I/O buffer arguments now use `void[]` instead of `ubyte[]` and `const`
+
+* `ocean.util.container.queue.model.IByteQueue`,
+  `ocean.util.container.queue.model.IUntypedQueue`,
+  `ocean.util.container.queue.FixedRingQueue`,
+  `ocean.util.container.queue.FlexibleFileQueue`,
+  `ocean.util.container.queue.FlexibleRingQueue`,
+  `ocean.util.container.queue.NotifyingQueue`,
+  `ocean.util.container.queue.QueueChain`
+
+  The types of the arguments and return types for queue input or output buffers
+  have been changed to `void[]` and use `const` where applicable. The particular
+  methods are
+
+  - `bool push ( ubyte[] )` changed to `bool push ( in void[] )`
+  - `ubyte[] push ( size_t )` changed to `void[] push ( size_t )`
+  - `bool push ( IUntypedQueue, void[] )` changed to
+    `bool push ( IUntypedQueue, in void[] )`
+  - `ubyte[] pop ( )` changed to `Const!(void)[] pop ( )`
+  - `bool pop ( ubyte[] )` changed to `bool pop ( void[] )`
+  - `ubyte[] peek ( )` changed to `Const!(void)[] peek ( )`
+  - `size_t pushSize ( ubyte[] )` changed to `size_t pushSize ( in void[] )`
+  - `bool willFit ( ubyte[] )` changed to `bool willFit ( in void[] )`
+  - `void save ( void delegate ( void[] , void[], void[] ) )` changed to
+    `void save ( void delegate ( in void[], in void[], in void[] ) )`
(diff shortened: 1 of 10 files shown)


mihails-strasuns-sociomantic (changes requested):
Migration notes and unit test please ;)

mathias-lang-sociomantic:
And rebase

david-eckardt-sociomantic:
Updated.

relnotes/queue-push-const.migration.md
```diff
@@ -7,3 +7,9 @@
 
   The type of the `item` argument of `push` has been changed from `ubyte[]` to
   `in ubyte[]` to allow for passing `const` buffers.
+
+
+* `ocean.util.container.queue.model.IIUntypedQueue`
+
+The type of the `t` argument of `push` has been changed from `void[]` to
```
mathias-lang-sociomantic: Indent

mathias-lang-sociomantic (approved):
LGTM

relnotes/queue-push-const.migration.md
david-eckardt-sociomantic: Oops

david-eckardt-sociomantic:
Updated.

nemanja-boric-sociomantic (approved):

nemanja-boric-sociomantic:
@mihails-strasuns-sociomantic ?

src/ocean/util/container/queue/FixedRingQueue.d
```diff
@@ -185,15 +187,16 @@ class FixedByteRingQueue : FixedRingQueueBase!(IByteQueue)
 
     ***************************************************************************/
 
-    bool push ( ubyte[] element )
+    bool push ( in ubyte[] element )
```
mihails-strasuns-sociomantic: I guess only question remaining from my side is why this takes `in ubyte[]` and not `in void[]` - you wouldn't need to do all those casts in tests otherwise.

Considering this is a breaking change targeting v4.x.x, maybe it is time to go void? :)

david-eckardt-sociomantic: I would like that, but it would be a rather intrusive change at a greater scope. It would in addition include
- `IByteQueue`
  - `ubyte[] push ( size_t size )`
  - `ubyte[] pop ( )`
  - `ubyte[] peek ( )`
- `NotifyingQueue`
  - `bool push ( size_t size, void delegate ( ubyte[] ) filler )`

It will most certainly break much more user code than just adding `in`. Right I'm torn here, `void[]` is much better IMHO.

david-eckardt-sociomantic: We could deprecate `IByteQueue` and all other `*Byte*Queue`s in favour of `*Data*Queue` which is `const`-correct and uses `void[]`. That way the `const`-correctnes would even got in v3.x.x.

mihails-strasuns-sociomantic: Sounds cool, though if we go this way, I would even leave `Byte*` part undeprecated first - it propagates up to swarm and forth which will annoy everyone again with dozens of unfixable deprecations.

david-eckardt-sociomantic: There is a complication with class naming. The classes in question are `FlexibleFileQueue`, `NotifyingQueue` and `QueueChain`. These classes depend on `IByteQueue` and need `IDataQueue` clones. What would be the best procedure here? I would like to avoid to add a useless `Data` infix to the class names.

david-eckardt-sociomantic: I guess I'm going to make it a breaking change in the next major release then.

mihails-strasuns-sociomantic: I am utterly irresponsible about naming things, ask @gavin-norman-sociomantic :trollface: 

As for breaking change - I would aim for adding new stuff in v3.x.x, deprecating old one at some point after 4.0.0 and removing completely in v5.x.x. This way we will have simple transitional point of v4.0.0-v4.N.0 during which both old and new API is known to be available.

david-eckardt-sociomantic:
Updated, in addition
- changing `ubyte[]` to `void[]`,
- making `push` and `peek` return `const` arrays,
- `const`-correcting the `FlexibleByteQueue.save` delegate.

mihails-strasuns-sociomantic (approved):

mihails-strasuns-sociomantic:
With next ocean v4.0.0 alpha will be interesting to check how this affects swarm.

PR #383, opened 2017-11-29
Check if CI env var was not set

mihails-strasuns-sociomantic:


nemanja-boric-sociomantic:
Interesting that v4.x.x base produces coverage #382

mihails-strasuns-sociomantic:
Not it has same broken flag issue.

mihails-strasuns-sociomantic:
(but `CI` is indeed already present)

PR #386, opened 2017-11-29
Wait until uptime part of /proc/uptime progresses

nemanja-boric-sociomantic:
The test that waits for the progression of the uptime
should confirm that the uptime part of /proc/uptime has progressed,
and not the idle.

### DIFF
--- a/test/sysstats/main.d
+++ b/test/sysstats/main.d
@@ -58,7 +58,7 @@ class MyApp : DaemonApp
         auto timer = new TimerEvent(
                 {
                     // wait until uptime advances, clock might be slower on VMs
-                    if (ProcVFS.getProcUptime() == uptime)
+                    if (ProcVFS.getProcUptime().uptime == uptime.uptime)
                         return true;
                     stats = this.sys_stats.collect();
                     this.epoll.shutdown();


nemanja-boric-sociomantic:
Rebased.

mihails-strasuns-sociomantic (approved):
Is there anyone but you who knows this code? :) I am fine with blind trust otherwise.

nemanja-boric-sociomantic:
I think @leandro-lucarella-sociomantic and @MartinNowak could be another set of eyes.

MartinNowak (approved):

leandro-lucarella-sociomantic:
I don't get this change, are you actually testing that `/proc/uptime` works?

Burgos:
Yeah, because it doesn't always work :-). Apparently, when running on Jenkins slaves lxcfs rounds uptime to whole seconds. So we're waiting until uptime difference is recorded. See https://github.com/dlang/ci/issues/88 to see how it fails.

leandro-lucarella-sociomantic:
:man_facepalming:

test/sysstats/main.d
```diff
@@ -58,7 +58,7 @@ class MyApp : DaemonApp
         auto timer = new TimerEvent(
                 {
                     // wait until uptime advances, clock might be slower on VMs
-                    if (ProcVFS.getProcUptime() == uptime)
+                    if (ProcVFS.getProcUptime().uptime == uptime.uptime)
```
leandro-lucarella-sociomantic, line 61: What if this never changes, will the test never exit?

nemanja-boric-sociomantic: I think it's fair not to optimize for the broken case, but to rely on the timeout on CI. Because, even the rounding is happening, you're still _using_ some CPU time in the timer handler/setting up the timer, so it should eventually happen.

mihails-strasuns-sociomantic: I would be more concerned about virtualized systems that always report uptime to be some fake value. But we can wait until such is actually encountered :)

leandro-lucarella-sociomantic (approved):

PR #389, opened 2017-11-30
fix uptime test wait-loop

MartinNowak:
- test must avoid divide-by-zero for both uptime and idle time

### DIFF
--- a/test/sysstats/main.d
+++ b/test/sysstats/main.d
@@ -58,7 +58,8 @@ class MyApp : DaemonApp
         auto timer = new TimerEvent(
                 {
                     // wait until uptime advances, clock might be slower on VMs
-                    if (ProcVFS.getProcUptime() == uptime)
+                    auto uptime2 = ProcVFS.getProcUptime();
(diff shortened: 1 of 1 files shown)


test/sysstats/main.d
```diff
@@ -58,7 +58,8 @@ class MyApp : DaemonApp
         auto timer = new TimerEvent(
                 {
                     // wait until uptime advances, clock might be slower on VMs
-                    if (ProcVFS.getProcUptime() == uptime)
+                    auto uptime2 = ProcVFS.getProcUptime();
+                    if (uptime.uptime == uptime2.uptime || uptime.idle == uptime2.idle)
```
nemanja-boric-sociomantic, line 62: No, just the `uptime` part is important - idle is not used for any calculations: https://github.com/sociomantic-tsunami/ocean/blob/v3.x.x/src/ocean/sys/Stats.d#L171

Btw, there's already #386

MartinNowak: thx

PR #391, opened 2017-12-04
Use names as keys for custom worker fiber pools

mihails-strasuns-sociomantic:
Contrary to original `ClassInfo` based approach, this allows to define
such configuration in a regular config file.

This does not count as a breaking change because relevant functionality
was not released yet.

### DIFF
--- a/relnotes/fiber-pool.feature.md
+++ b/relnotes/fiber-pool.feature.md
@@ -6,14 +6,19 @@ It is now possible to define dedicated worker fiber pools to process tasks of
 specific type:
 
 ```D
+module mymod;
+
+class MyTask : Task { /* ... */ }
+class MyOtherTask : Task { /* ... */ }
+
 SchedulerConfiguration config;
 
 with (config)
 {
-specialized_pools = [
-  PoolDescription(MyTask.classinfo, 10240),
-  PoolDescription(MyOtherTask.classinfo, 2048),
-];
+    specialized_pools = [
+        PoolDescription("mymod.MyTask", 10240),
+        PoolDescription(MyOtherTask.classinfo.name, 2048),
+    ];
 }
 
 initScheduler(config);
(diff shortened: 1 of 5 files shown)


mihails-strasuns-sociomantic:
@gavin-norman-sociomantic like this?

relnotes/fiber-pool.feature.md
```diff
@@ -6,14 +6,19 @@ It is now possible to define dedicated worker fiber pools to process tasks of
...
 
 with (config)
 {
-specialized_pools = [
-  PoolDescription(MyTask.classinfo, 10240),
-  PoolDescription(MyOtherTask.classinfo, 2048),
-];
+    specialized_pools = [
+        PoolDescription("mymod.MyTask", 10240),
+        PoolDescription(MyOtherTask.classinfo.name, 2048),
```
gavin-norman-sociomantic, line 20: Ahh nice!

gavin-norman-sociomantic (commented):
So this gets around the problem with template classes not being lookup-able?

mihails-strasuns-sociomantic:
> So this gets around the problem with template classes not being lookup-able?

No, this will remain a problem with only workaround I have mentioned.

mihails-strasuns-sociomantic:
If that was not clear, this PR switches to string-based pool config so that it can be read from config file later.

gavin-norman-sociomantic:
Ah I see, so it's just to enable reading `PoolDescription`s directly from the config file?

mihails-strasuns-sociomantic:
Yep

mihails-strasuns-sociomantic:
So, @gavin-norman-sociomantic ?

gavin-norman-sociomantic:
>No, this will remain a problem with only workaround I have mentioned.

It's probably worth documenting this somewhere.

mihails-strasuns-sociomantic:
Actually no, nevermind - with this PR the problem is not present in any way, because config is stored under string keys now and there is no need to ever do runtime search for a class. And templated classes do have a valid `classinfo.name`, they just can't be found via runtime reflection.

gavin-norman-sociomantic:
Ok that's what I thought originally. Good!

mihails-strasuns-sociomantic:
So how about merge? :D

gavin-norman-sociomantic (approved):

PR #395, opened 2017-12-05
 Add documentation on the formatting specifier 

### LINKED ISSUE #43: Document supported format string in Formatter
It's a common request, and would be nice to have comprehensive docs on that.
The format is inherited from `Layout`, and is currently DMD-style documentated (read the code).

mathias-lang-sociomantic:
Fix #43

### DIFF
--- a/src/ocean/text/convert/Formatter.d
+++ b/src/ocean/text/convert/Formatter.d
@@ -13,6 +13,54 @@
     - `snformat` is equivalent to `std.format.sformat`
     - `sformat` is roughly equivalent to `std.format.formattedWrite`
 
+    Format_specifier:
+
+    The Formatter uses a format specification similar to C#/.NET over
+    the traditional `printf` style.
+    As a result, the most simple usage is to call:
+    ---
+    format("This value will be default formatted: {}", value);
+    ---
+
+    More specific formatting options are however available.
+
+    The format specifier is defined as follows:
+
+    '{'[INDEX][WIDTH_CHAR[ALIGN_LEFT_CHAR][ALIGN_WIDTH][' '*]][':'][FORMAT_STRING]'}'
+
+    In more details:
+    - `INDEX` is the positive, decimal and 0 based index of the argument
+      to format.
+    - `WIDTH_CHAR` is either ',' (comma) if a minimum width is requested,
+      in which case the output will be padded with spaces,
+      or '.' if a maximum width is requested, in which case the output will be
+      cropped and cropping will be noted by the presence of "..."
+    - `ALIGN_LEFT_CHAR` is '-'. If present, padding / cropping will be done
+      on the left side of the string, otherwise it will be the right side
+      This can only be used after a `WIDTH_CHAR`.
+    - `ALIGN_WIDTH` is the positive, decimal width the argument should have.
+    - ':' can optionally be used to separate the index / width specification
+      from the format string. So `{}`, `{:}` and `{0:X}` are all valid.
+    - `FORMAT_STRING` is an argument-defined format string
+
+    Format_string:
+
+    The format string defines how the argument will be formatted, and thus
+    is dependent on the argument type.
+
+    Currently the following formatting strings are supported:
+        - 'X' or 'x' are used for hexadecimal formatting of the output.
+          'X' outputs uppercase, 'x' will output lowercase.
+          Applies to integer types and pointers, and will also output
+          the hexadecimal.
+        - 'e' for floating point type will display exponential notation.
+          Using a number will set the precision, so for example the string
+          `"{:2}"` with argument `0.123456` will output `"0.12"`
+          Finally, '.' will prevent padding.
+    Unrecognized formatting strings should be ignored. For composed types,
+    the formatting string is passed along, so using `X` on a `struct` or an
+    array will display any integer / pointer members in uppercase hexadecimal.
+
     Copyright:
         Copyright (c) 2009-2016 Sociomantic Labs GmbH.
         Some parts (marked explicitly) copyright Kris and/or Larsivi.
--- a/src/ocean/text/convert/Formatter.d
+++ b/src/ocean/text/convert/Formatter.d
@@ -1,13 +1,17 @@
 /*******************************************************************************
 
-    Compile-time version of `Layout_tango`, allows one to format structs,
-    and supports `Typedef` both in D1 and D2.
+    Lightweight, memory friendly string formatting module
 
     This module provides 4 possible semantics:
     - For pedestrian usage which doesn't care about allocations, see `format`
     - For allocation-friendly semantic where the data is output either to
       a sink or to a `ref char[]`, see the `sformat` overloads
     - To ensure absolutely no allocation happens, see `snformat`
+
+    Users of Phobos' `std.format` will find many similarities in the API:
+    - `format` is equivalent to `std.format.format`
+    - `snformat` is equivalent to `std.format.sformat`
+    - `sformat` is roughly equivalent to `std.format.formattedWrite`
 
     Copyright:
         Copyright (c) 2009-2016 Sociomantic Labs GmbH.


nemanja-boric-sociomantic:
What a time to be alive.

mihails-strasuns-sociomantic (commented):
<3

Do you mind separately listing differences from old tango layout? Basically bits of info from https://github.com/sociomantic/ocean/commit/9ba517cf958d2e2f50f5d49c7c131cda5be82b42

src/ocean/text/convert/Formatter.d
```diff
@@ -13,6 +13,52 @@
...
+    More specific formatting options are however available.
+
+    The format specifier is defined as follows:
+
+    '{'[INDEX][WIDTH_CHAR[ALIGN_LEFT_CHAR][ALIGN_WIDTH][\s*]][':'][FORMAT_STRING]'}'
+
+    In more details:
+    - `INDEX` is the positive, decimal and 0 based index of the argument
+      to format.
+    - `WIDTH_CHAR` is either ',' (coma) if a minimum width is requested,
```
gavin-norman-sociomantic: comma

src/ocean/text/convert/Formatter.d
```diff
@@ -13,6 +13,52 @@
...
+    As a result, the most simple usage is to call:
+    ---
+    format("This value will be default formatted: {}", value);
+    ---
+
+    More specific formatting options are however available.
+
+    The format specifier is defined as follows:
+
+    '{'[INDEX][WIDTH_CHAR[ALIGN_LEFT_CHAR][ALIGN_WIDTH][\s*]][':'][FORMAT_STRING]'}'
```
gavin-norman-sociomantic: It's not clear to me what `[\s*]` means. Some kind of regex, I presume, but we might want to make this doc readable to people who don't have such knowledge.

gavin-norman-sociomantic: Why `[:]` but `[ALIGN_LEFT_CHAR]` (not `[-]`)?

src/ocean/text/convert/Formatter.d
```diff
@@ -13,6 +13,52 @@
...
+    - ':' can optionally be used to separate the index / width specification
+      from the format string. So `{}`, `{:}` and `{0:X}` are all valid.
+    - `FORMAT_STRING` is an argument-defined format string
+
+    Format_string:
+
+    The format string defines how the argument will be formatted, and thus
+    is dependent on the argument type.
+
+    Currently little formatting strings are supported, e.g.:
```
gavin-norman-sociomantic: little -> few

gavin-norman-sociomantic: Actually, just say "the following". (No need to judge whether the number is few / many.)

src/ocean/text/convert/Formatter.d
```diff
@@ -13,6 +13,52 @@
...
+        - 'X' or 'x' are used for hexadecimal formatting of the output.
+          'X' outputs uppercase, 'x' will output lowercase.
+          Applies to integer types and pointers, and will also output
+          the hexadecimal.
+        - 'e' for floating point type will display exponential notation.
+          Using a number will set the precision, and '.' will prevent padding.
```
gavin-norman-sociomantic: >Using a number

What does this mean?

src/ocean/text/convert/Formatter.d
```diff
@@ -13,6 +13,52 @@
...
+    Unrecognized formatting strings should be ignored. For composed types,
```
gavin-norman-sociomantic: "should be"?

gavin-norman-sociomantic (commented):
Thank you!!!

mathias-lang-sociomantic:
The only difference in formatting is on pointers. I would prefer to avoid any reference to something we're removing in documentation (release notes are fine) because as time passes it will be less and less relevant.
However I could add a documented unittest with default formatting at the top so it appears in the generated module doc if you want (some of that is already covered in the tests, but they are not so visible)

mihails-strasuns-sociomantic:
I am mostly asking because we intend to switch to new formatter in `Stdout` & Co in v4.x.x so the topic of expected difference will become hot again for a while. Alternatively, it can be copied in relnotes for that v4.x.x PR.

src/ocean/text/convert/Formatter.d
mathias-lang-sociomantic: Yeah `\s` is the whitespace class. However that's slightly wrong since only spaces are skipped. will correct with `[' '*]`, how does it sound ?

mathias-lang-sociomantic: Because `ALIGN_LEFT_CHAR` has an impact on the output (thus I refer to it later) but ':' doesn't

src/ocean/text/convert/Formatter.d
mathias-lang-sociomantic: If you pass "42" the precision will be 42

src/ocean/text/convert/Formatter.d
gavin-norman-sociomantic: Yeah `[' '*]` is a bit more intuitively obvious.

src/ocean/text/convert/Formatter.d
gavin-norman-sociomantic: Pass where? Probably needs an example.

mathias-lang-sociomantic:
@mihails-strasuns-sociomantic : Then maybe it should be in #361 ?

mihails-strasuns-sociomantic:
I don't mind either way, just put it somewhere :)

src/ocean/text/convert/Formatter.d
mathias-lang-sociomantic: Ultimately, it's dependent on the `toString` implementation. One could choose to throw an exception.

mathias-lang-sociomantic:
Updated:

```diff
diff --git a/src/ocean/text/convert/Formatter.d b/src/ocean/text/convert/Formatter.d
index ae38e9e1..267d824b 100644
--- a/src/ocean/text/convert/Formatter.d
+++ b/src/ocean/text/convert/Formatter.d
@@ -26,12 +26,12 @@
 
     The format specifier is defined as follows:
 
-    '{'[INDEX][WIDTH_CHAR[ALIGN_LEFT_CHAR][ALIGN_WIDTH][\s*]][':'][FORMAT_STRING]'}'
+    '{'[INDEX][WIDTH_CHAR[ALIGN_LEFT_CHAR][ALIGN_WIDTH][' '*]][':'][FORMAT_STRING]'}'
 
     In more details:
     - `INDEX` is the positive, decimal and 0 based index of the argument
       to format.
-    - `WIDTH_CHAR` is either ',' (coma) if a minimum width is requested,
+    - `WIDTH_CHAR` is either ',' (comma) if a minimum width is requested,
       in which case the output will be padded with spaces,
       or '.' if a maximum width is requested, in which case the output will be
       cropped and cropping will be noted by the presence of "..."
@@ -48,13 +48,15 @@
     The format string defines how the argument will be formatted, and thus
     is dependent on the argument type.
 
-    Currently little formatting strings are supported, e.g.:
+    Currently the following formatting strings are supported:
         - 'X' or 'x' are used for hexadecimal formatting of the output.
           'X' outputs uppercase, 'x' will output lowercase.
           Applies to integer types and pointers, and will also output
           the hexadecimal.
         - 'e' for floating point type will display exponential notation.
-          Using a number will set the precision, and '.' will prevent padding.
+          Using a number will set the precision, so for example the string
+          `"{:2}"` with argument `0.123456` will output `"0.12"`
+          Finally, '.' will prevent padding.
     Unrecognized formatting strings should be ignored. For composed types,
     the formatting string is passed along, so using `X` on a `struct` or an
     array will display any integer / pointer members in uppercase hexadecimal.
```

mihails-strasuns-sociomantic (approved):
Leaving it in @gavin-norman-sociomantic hands

gavin-norman-sociomantic (approved):

Issue #405, opened 2017-12-08
v3.6.0 introduces symbol clashes between hasIndirections definitions


joseph-wakeling-sociomantic:
Building an application where `ocean.core.Traits.hasIndirections` is used inside one module with v3.6.0 I now get a symbol clash with `ocean.meta.traits.Indirections.hasIndirections`:

```
Error: ocean.meta.traits.Indirections.hasIndirections(T...) at
./submodules/ocean/src/ocean/meta/traits/Indirections.d(51)
conflicts with ocean.core.Traits.hasIndirections(T...) at
./submodules/ocean/src/ocean/core/Traits.d(228)
```

My guess would be that the `ocean.meta.traits.Indirections` module is somehow accidentally publicly imported via some other module, in a way that wasn't previously the case.  This is arguably a breaking change (albeit an unintended one) so would be nice to fix if possible.

joseph-wakeling-sociomantic:
It looks like it's down to commit 264378d68f4f8f8cd2535a0aa2b2e0e217260e33, which adjusts the traits import used in `ocean.task.Scheduler`.  Specifically, I would imagine the selective import of `hasIndirections` is causing it to be exposed publicly to any module that imports `ocean.task.Scheduler`.

joseph-wakeling-sociomantic:
As a note: I only observe the problem in D2 builds.  Reverting commit 264378d68f4f8f8cd2535a0aa2b2e0e217260e33 means I no longer see the build error.

mihails-strasuns-sociomantic:
Sorry, but I am inclined to close this as wont-fix. It is a known and in general unfixable problem that introducing new symbols can cause downstream breakage in D - we have made a conscious decision to ignore that in our neptune rules because otherwise every single addition of a symbol would require a major release. This is unfortunate but I don't see what can be done about in.

joseph-wakeling-sociomantic:
Actually it looks like a fairly simple fix: https://github.com/sociomantic-tsunami/ocean/pull/407

mihails-strasuns-sociomantic:
I know that this can be simply fixed - I argue that it should not be done on principle, because cases like that are not considered breaking changes. Where is the line between case that should be fixed and the one that should not?

joseph-wakeling-sociomantic:
> Where is the line between case that should be fixed and the one that should not?

If the fix was complicated or problematic, then I would agree it might not be worth addressing it.  But the fix in this case is super-simple with no apparent negatives, so why not take it?

[...]

PR #407, opened 2017-12-11
Replace selective hasIndirections import in ocean.task.Scheduler

joseph-wakeling-sociomantic:
This selective import causes the `hasIndirections` symbol to be seen as public in D2 builds of modules that import `ocean.task.Scheduler`, and hence can cause symbol clashes with the identically-named symbol from `ocean.core.Traits`.

Fixes https://github.com/sociomantic-tsunami/ocean/issues/405.

### DIFF
--- a/src/ocean/task/Scheduler.d
+++ b/src/ocean/task/Scheduler.d
@@ -28,7 +28,7 @@ import ocean.core.Enforce;
 import ocean.core.Verify;
 import ocean.io.select.EpollSelectDispatcher;
 import ocean.util.container.queue.FixedRingQueue;
-import ocean.meta.traits.Indirections : hasIndirections;
+import ocean.meta.traits.Indirections;
 
 import ocean.task.Task;
 import ocean.task.IScheduler;


joseph-wakeling-sociomantic:
I can confirm this solves the problem in the particular case where the symbol clash was first observed.  I don't think it should cause any problems in general: since the non-selective import should be properly private, it shouldn't create clashes except for internal ones in `ocean.task.Scheduler`.

mathias-lang-sociomantic (approved):
And this is why selective imports are still broken...

nemanja-boric-sociomantic:
> And this is why selective imports are still broken...

Is this the case for the compilers after v2.070?

mathias-lang-sociomantic:
Sadly yes

EDIT: In this case it wouldn't show, because the symbol would be private, but selective import still create aliases in the module.

mihails-strasuns-sociomantic:
@nemanja-boric-sociomantic to clarify what @mathias-lang-sociomantic has said, it is still possible to break your app by changing `import aaa;` into `import aaa : sym;` because latter is interpreted as locally defined symbol when symbol lookup has to be done.

mihails-strasuns-sociomantic (approved):

mihails-strasuns-sociomantic:
Damn, wrong branch.

mihails-strasuns-sociomantic:
Picked into 3.6.x as e202c9818f34854f9d52725c072f1c925a06d792

joseph-wakeling-sociomantic:
Oh damn!  I deliberately based it off v3.6.x but must have mixed up the target branch :-\  Sorry about that.

PR #409, opened 2017-12-11
Workaround broken 2.077.1 packaging

mihails-strasuns-sociomantic:


mathias-lang-sociomantic:
What's up with Codecov ? Also, swarm will need this, and turtle, and dlsproto :(

mihails-strasuns-sociomantic:
I already added workaround for turtle (different one) earlier, but yeah, swarm and dlsproto are in queue.

> What's up with Codecov

I wish I knew. Looks like at some point of switching to `beaver dlang codecov` command it got funny ideas but haven't got to bisecting it yet.

leandro-lucarella-sociomantic:
It seems like there is one more package that needs to be specified explicitly for some reason: `libphobos2-77=2.077.0-0`, which could complicate the workaround. :unamused:

mihails-strasuns-sociomantic:
Any ideas if we can do anything without changing beaver in that case? If not, I would disable 2.077 testing for now.

leandro-lucarella-sociomantic:
I tried to look for options to force automatic resolution of those dependency problems (is really annoying that it basically knows how to fix it, just install the damn version you need to install), but I failed :(

We could do a "dirty" patch release of beaver for this and then rollback when/if d-apt is fixed in this case too...

mihails-strasuns-sociomantic:
Disabling tests until we come up with long-term solution (or the problem is addressed in D-APT itself)

PR #410, opened 2017-12-12
Switch back to manual codecov.sh call

mihails-strasuns-sociomantic:
Checking if it will fix double-comment and malformed report problem we have now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ install: beaver dlang install
 
 script: beaver dlang make
 
-after_success: beaver run ci/codecov.sh
+after_success: ci/codecov_wrapper.sh
 
 jobs:
     include:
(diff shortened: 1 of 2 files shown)


leandro-lucarella-sociomantic:
> Checking if it will fix double-comment and malformed report problem we have now.

Malformed report? Maybe you can do a test here with the same arguments beaver uses to see if there is a problem with the script version or to identify if it's some argument that breaks it?

leandro-lucarella-sociomantic (commented):
Just a comment to "ACK" the review request :D

mihails-strasuns-sociomantic:
> Maybe you can do a test here with the same arguments beaver uses to see if there is a problem with the script version or to identify if it's some argument that breaks it?

Yep, now that we see that reverting to old command fixes the issue I will try adding tweaks from beaver one by one.

mihails-strasuns-sociomantic:
Closing as problem was confirmed to be on codecov side

PR #413, opened 2017-12-13
Add module to shallow reset a struct

gautam-kotian-sociomantic:
The main motivation for introducing this functionality (and not simply
    using the existing DeepReset functionality) is that DeepReset recurses
    into and individually resets every element of each dynamic array present
    in the struct. However, this additional work is sometimes overkill for
    certain performance critical applications where just resetting the
    lengths of the dynamic arrays to zero suffices.

### DIFF
--- a/src/ocean/util/DeepReset.d
+++ b/src/ocean/util/DeepReset.d
@@ -1,12 +1,10 @@
 /*******************************************************************************
 
-    Utility to recursively reset fields of struct to their .init value while
-    preserving array pointers (their length is set to 0 but memory is kept
-    available for further reusage)
+    Utility to recursively reset fields of an aggregate to their .init value
+    while preserving array pointers (their length is set to 0 but memory is kept
+    available for further reuse)
 
-    Copyright:
-        Copyright (c) 2009-2016 Sociomantic Labs GmbH.
-        All rights reserved.
+    Copyright: Copyright (c) 2017 sociomantic labs GmbH. All rights reserved
 
     License:
         Boost Software License Version 1.0. See LICENSE_BOOST.txt for details.
@@ -25,7 +23,7 @@ version(UnitTest) import ocean.core.Test;
 
 /*******************************************************************************
 
-    Template to determine the correct DeepReset function to call dependent on
+    Template to determine the correct DeepReset function to call depending on
     the type given.
 
     Params:
@@ -75,7 +73,7 @@ public template DeepReset ( T )
     length to 0.
 
     Params:
-        T = type of array to deep copy
+        T = type of array to deep reset
         dst = destination array
 
 *******************************************************************************/
@@ -96,7 +94,7 @@ public void DynamicArrayDeepReset ( T ) ( ref T[] dst )
     array.
 
     Params:
-        T = type of array to deep copy
+        T = type of array to deep reset
         dst = destination array
 
 *******************************************************************************/
@@ -113,7 +111,7 @@ public void StaticArrayDeepReset ( T ) ( T[] dst )
     Deep reset function for arrays.
 
     Params:
-        T = type of array to deep copy
+        T = type of array to deep reset
         dst = destination array
 
 *******************************************************************************/
@@ -122,7 +120,7 @@ private void ArrayDeepReset ( T ) ( ref T[] dst )
 {
     static if ( isAssocArrayType!(T) )
     {
-        // TODO: copy associative arrays
+        // TODO: reset associative arrays
         pragma(msg, "Warning: deep reset of associative arrays not yet implemented");
     }
     else static if ( is(T S : S[]) )
@@ -170,7 +168,7 @@ private void ArrayDeepReset ( T ) ( ref T[] dst )
     Deep reset function for structs.
 
     Params:
-        T = type of struct to deep copy
+        T = type of struct to deep reset
         dst = destination struct
 
 *******************************************************************************/
@@ -188,7 +186,7 @@ public void StructDeepReset ( T ) ( ref T dst )
     {
         static if ( isAssocArrayType!(typeof(member)) )
         {
-            // TODO: copy associative arrays
+            // TODO: reset associative arrays
             pragma(msg, "Warning: deep reset of associative arrays not yet implemented");
         }
         else static if ( is(typeof(member) S : S[]) )
@@ -224,7 +222,7 @@ public void StructDeepReset ( T ) ( ref T dst )
     Deep reset function for dynamic class instances.
 
     Params:
-        T = type of class to deep copy
+        T = type of class to deep reset
         dst = destination instance
 
 *******************************************************************************/
@@ -240,7 +238,7 @@ public void ClassDeepReset ( T ) ( ref T dst )
     {
         static if ( isAssocArrayType!(typeof(member)) )
         {
-            // TODO: copy associative arrays
+            // TODO: reset associative arrays
             pragma(msg, "Warning: deep reset of associative arrays not yet implemented");
         }
         else static if ( is(typeof(member) S : S[]) )
@@ -268,8 +266,8 @@ public void ClassDeepReset ( T ) ( ref T dst )
         }
     }
 
-    // Recurse into super any classes
-    static if ( is(T S == super ) )
+    // Recurse into any super classes
+    static if ( is(T S == super) )
     {
         foreach ( V; S )
         {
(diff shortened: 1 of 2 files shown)


src/ocean/util/StructShallowReset.d
```diff
@@ -0,0 +1,127 @@
...
+
+unittest
+{
+    struct A
+    {
+        int w;
+        int[] x;
+        char y;
+        char[] z;
+    }
```
nemanja-boric-sociomantic, line 104: Please add a test with the nested struct.

src/ocean/util/DeepReset.d
```diff
@@ -1,12 +1,10 @@
...
-    preserving array pointers (their length is set to 0 but memory is kept
-    available for further reusage)
+    Utility to recursively reset fields of an aggregate to their .init value
+    while preserving array pointers (their length is set to 0 but memory is kept
+    available for further reuse)
 
-    Copyright:
-        Copyright (c) 2009-2016 Sociomantic Labs GmbH.
-        All rights reserved.
+    Copyright: Copyright (c) 2017 sociomantic labs GmbH. All rights reserved
```
gavin-norman-sociomantic, line 7: I believe you have to keep the original date...?

src/ocean/util/StructShallowReset.d
```diff
@@ -0,0 +1,127 @@
...
+    Copyright: Copyright (c) 2017 sociomantic labs GmbH. All rights reserved
+
+    License:
+        Boost Software License Version 1.0. See LICENSE_BOOST.txt for details.
+        Alternatively, this file may be distributed under the terms of the Tango
+        3-Clause BSD License (see LICENSE_BSD.txt for details).
+
+*******************************************************************************/
+
+module ocean.util.StructShallowReset;
```
gavin-norman-sociomantic, line 24: For parity with `DeepReset`, it might make more sense to name this `ShallowReset` and internally `static assert` that it only supports structs. (I guess this limitation exists for the sake of simplicity... not having to recurse into fields of super classes?)

gavin-norman-sociomantic (commented):
A general thought: an alternative (and more flexible) approach would be to adapt the existing `DeepReset` templates to also accept a policy argument determining whether arrays, sub-aggregates, etc should be recursed into or not. The default would be `true`, for all cases, but it would then be trivial to do a shallow reset of exactly the desired type, using the same code.

david-eckardt-sociomantic:
Using [`hasIndirections`](https://github.com/sociomantic-tsunami/ocean/blob/v3.x.x/src/ocean/core/Traits.d#L228) you can support all value types i.e. sub-structs/unions or static arrays that are values and don't contain reference types (i.e. dynamic/associative arrays, class objects, pointers/delegates etc.).
```d
foreach ( i, ref member; dst.tupleof )
{
    static if ( is(typeof(member) Element == Element[]) ) // dynamic array
    {
        // Make sure the array elements don't contain indirections i.e. are values.
        static assert (!hasIndirections!(Element));
        member.length = 0;
        enableStomping(member); // See below
    }
    else
    {
        // The struct field is not a dynamic array; other indirections are not supported.
        // This makes sure member is a value.
        static assert (!hasIndirections!(typeof(member)));
        member = member.init;
    }
}
```
When setting `member.length = 0` you also have to watch out for array stomping prevention, which can be triggered the next time the array length is increased again. Calling `enableStomping(member)` prevents that from happening.

src/ocean/util/DeepReset.d
gautam-kotian-sociomantic: There's only a single year in our "standard" copyright statement present in https://github.com/sociomantic/backend/blob/master/doc/coding/d-style-guide.rst#modules. So should this be set to 2009?

nemanja-boric-sociomantic: `2009-2017` is the solution.

gavin-norman-sociomantic: >There's only a single year in our "standard" copyright statement

Ah, that's an example for something that's been in copyright since 2017. For things that have been copyrighted for longer, you should specify the year range, as Nemanja says.

src/ocean/util/StructShallowReset.d
gautam-kotian-sociomantic: >I guess this limitation exists for the sake of simplicity

Yep. And the fact that we currently don't need it for anything other than structs.

>For parity with DeepReset, it might make more sense to name this ShallowReset and internally static assert that it only supports structs. 

Yes, that can be done. The main reason the `Struct` was added to the module name was to make it clear from the get-go that only structs are supported (to further discourage someone from using this rather than `DeepReset`). But I guess this will be irrelevant if we go the path of adding a policy argument to `DeepReset` itself.

nemanja-boric-sociomantic:
>  A general thought: an alternative (and more flexible) approach would be to adapt the existing DeepReset templates to also accept a policy argument determining whether arrays, sub-aggregates, etc should be recursed into or not. The default would be true, for all cases, but it would then be trivial to do a shallow reset of exactly the desired type, using the same code.

Perhaps a better approach would be to rename policy-based `DeepReset` to `Reset` and instantiate it as `DeepReset` and `ShallowReset` with the proper policies.

gavin-norman-sociomantic:
>Perhaps a better approach would be to rename policy-based DeepReset to Reset and instantiate it as DeepReset and ShallowReset with the proper policies.

That would make sense.

gautam-kotian-sociomantic:
>A general thought: an alternative (and more flexible) approach would be adapt the existing DeepReset templates to also accept a policy argument determining arrays, sub-aggregates, etc should be recursed into or not. The default would be true, for all cases, but it would then be trivial to do a shallow reset of exactly the desired type, using the same code.

I considered doing this earlier, albeit with a single boolean flag that would control whether the reset is deep or shallow. Individually controlling the type of reset performed for different types would of course be more flexible, but would also substantially increase complexity.

For now, it may be better to control the type of reset with a single boolean variable, and then in future make it more flexible if necessary. What do you think?

The other question that arises if we introduce a policy argument is what should the module be called. It seems wrong to continue calling it `DeepReset` when the reset can sometimes be shallow.

nemanja-boric-sociomantic:
What about my suggestion above: it should be renamed to `Reset` (or similar) and then the `DeepReset` can import it and instantiate it as an `alias DeepReset` (so it stays backward compatible) and you can add `ShallowReset` module.

gautam-kotian-sociomantic:
>What about my suggestion above: it should be renamed to Reset (or similar) and then the DeepReset can import it and instantiate it as an alias DeepReset (so it stays backward compatible) and you can add ShallowReset module.

You answered part of my question before I'd finished asking it :)
I agree with your suggestion.

gavin-norman-sociomantic:
>I considered doing this earlier, albeit with a single boolean flag that would control whether the reset is deep or shallow. Individually controlling the type of reset performed for different types would of course be more flexible, but would also substantially increase complexity.

To be honest, I think the increase in complexity would be minimal. But, as your current use case is a simple recurse / do not recurse boolean, we can go with that. As long as it's implemented in a way that makes it easy to extend.

e.g.:
```D
struct ResetPolicy
{
  bool recurse;
  // we could easily add more fields here, in the future
}
```

>The other question that arises if we introduce a policy argument is what should the module be called

I'd create a new module `Reset` and replace everything in `DeepReset` with an alias to the code in new module. (I don't think we need a separate `ShallowReset` module just containing an alias.)

gautam-kotian-sociomantic:
>Using hasIndirections you can support all value types i.e. sub-structs/unions or static arrays that are values and don't contain reference types (i.e. dynamic/associative arrays, class objects, pointers/delegates etc.).

Thank you for that suggestion, it does seem like a cleaner approach. I'll try and implement it that way.

gavin-norman-sociomantic:
>Thank you for that suggestion, it does seem like a cleaner approach. I'll try and implement it that way.

Is that suggestion relevant to the policy idea, though? (I was assuming they were orthogonal.)

mihails-strasuns-sociomantic:
Ping

mihails-strasuns-sociomantic:
(I can take this over if you don't have time)

gautam-kotian-sociomantic:
>(I can take this over if you don't have time)

Thanks Mihails, I'd highly appreciate that. But please check with Gavin before you start. I think he made a basic attempt at it, but ran into some problems.

PR #418, opened 2017-12-14
Fix bug in Lzo.decompressSafe

gavin-norman-sociomantic:


src/ocean/io/compress/Lzo.d
```diff
@@ -243,6 +243,23 @@ class Lzo
     }
 }
 
+/// Simple compress / decompress example.
+unittest
+{
+    auto lzo = new Lzo;
+    auto original = "compress this!".dup;
```
david-eckardt-sociomantic: The `.dup` should not be needed.

src/ocean/io/compress/Lzo.d
```diff
@@ -243,6 +243,23 @@ class Lzo
...
+
+    // Compress.
+    auto compressed = new void[lzo.maxCompressedLength(original.length)];
```
david-eckardt-sociomantic: Not very important, but without the `.dup` above `compressed` and `uncompressed` can be static `ubyte[n]` arrays.

src/ocean/io/compress/Lzo.d
```diff
@@ -243,6 +243,23 @@ class Lzo
...
+    compressed.length = lzo.compress(original, compressed);
+
+    // Decompress (with overrun safety checks).
+    auto uncompressed = new void[original.length];
+    lzo.decompressSafe(compressed, uncompressed);
```
david-eckardt-sociomantic: Test that it returns `uncompressed.length`.

src/ocean/io/compress/Lzo.d
gavin-norman-sociomantic: Right, true, it doesn't need to be mutable.

src/ocean/io/compress/Lzo.d
gavin-norman-sociomantic: I think that's less intuitive as a usage example.

src/ocean/io/compress/Lzo.d
gavin-norman-sociomantic: As this is primarily intended to be a usage example (documented unittest), I'm not sure about that. In the example given, the return value is irrelevant, as we've already ensured that the decompress destination buffer is exactly the right size.

src/ocean/io/compress/Lzo.d
gavin-norman-sociomantic: Updated.

src/ocean/io/compress/Lzo.d
david-eckardt-sociomantic: Maybe add a comment, otherwise it may look as if the return value is irrelevant.

gavin-norman-sociomantic:
Updated: reworked usage example to demonstrate the common case of receiving a compressed buffer over the network.

src/ocean/io/compress/Lzo.d
```diff
@@ -243,6 +243,41 @@ class Lzo
...
+    // LZO instance. (Usually a single one can be shared globally.)
+    auto lzo = new Lzo;
+
+    // Typical function to decompress a buffer that was received over the
+    // network. (i.e. where the buffer and the expected length of the
+    // decompressed data are both received from an external source.)
+    void decompress ( in void[] src, size_t expected_decompressed_length,
+        ref void[] dst )
+    {
+        dst.length = expected_decompressed_length;
```
nemanja-boric-sociomantic, line 258: _maybe_ `enableStomping` ?

nemanja-boric-sociomantic (approved):

david-eckardt-sociomantic (approved):

src/ocean/io/compress/Lzo.d
gavin-norman-sociomantic: :+1: Updated.

PR #421, opened 2017-12-18
More `ocean.meta`

mihails-strasuns-sociomantic:


mihails-strasuns-sociomantic:
Split into commits and ready for review. This contains utilities I have found necessary when switching ocean internally to use `ocean.meta` for v4 branch.

src/ocean/meta/codegen/CTFE.d
```diff
@@ -30,15 +30,15 @@ import ocean.meta.types.Qualifiers;
...
 {
     if (i == 0)
         return "0";
 
     if (i < 0)
-        return "-" ~ toString(cast(ulong) -i);
+        return "-" ~ toStringImpl(cast(ulong) -i);
 
-    return toString(cast(ulong) i);
+    return toStringImpl(cast(ulong) i);
```
mathias-lang-sociomantic: I highly doubt this will work. Try:
```
    pragma(msg, toString(int.min), " -- ", int.min);
    pragma(msg, toString(short.min), " -- ", short.min);
```

mathias-lang-sociomantic: Side note: We should make the `Formatter` CTFE-able to simplify this kind of things

src/ocean/meta/traits/Aggregates.d
```diff
@@ -0,0 +1,97 @@
...
+
+import ocean.meta.traits.Basic;
+import ocean.meta.types.Qualifiers;
+
+/*******************************************************************************
+
+    Checks for presence of method/field with specified name in aggregate.
+
+    In D1 most common idiom is to simply check for `is(typeof(T.something))` but
+    in D2 it can backfire because of UFCS as global names are checked too
```
mathias-lang-sociomantic: "... so we use `__traits(addMember)`. Reading the documentation and note the code would let one to believe this is problematic in D2.

src/ocean/meta/traits/Aggregates.d
```diff
@@ -0,0 +1,97 @@
...
+unittest
+{
+    struct S
+    {
+        int foo ( double ) { return 0; }
+    }
+
+    static assert ( hasMethod!(S, "foo", int delegate(double)));
+    static assert (!hasMethod!(S, "foo", void delegate(double)));
+    static assert (!hasMethod!(S, "foobar", int delegate(double)));
```
mathias-lang-sociomantic: Can you add a test where the requested name is a member ? Ditto for a (`static`) function.

src/ocean/meta/types/Arrays.d
```diff
@@ -107,3 +107,30 @@ unittest
...
+    Params:
+        T = any type
+
+    Returns:
+        If T is a static array, evaluates to matching dynamic array. Othwerwise
+        evaluates to just T.
+
+*******************************************************************************/
+
+public template SliceIfStaticArray ( T )
```
mathias-lang-sociomantic: That looks... Utterly specific

src/ocean/meta/types/Function.d
```diff
@@ -0,0 +1,143 @@
...
+    static assert (Params.length == 3);
+    static assert (is(Params[0] == int));
+    static assert (is(Params[1] == double));
+    static assert (is(Params[2] == char[]));
+}
+
+/*******************************************************************************
+
+    Params:
+        TCallable = callable type (function/delegate/function pointer) or
```
mathias-lang-sociomantic: Wrong name

src/ocean/meta/types/Arrays.d
mihails-strasuns-sociomantic: I was very surprised to see how widespread it is in ocean. Not being able to return static arrays from functions makes some far reaching impact on generic code.

src/ocean/meta/codegen/CTFE.d
mihails-strasuns-sociomantic: It worked but minus sign and cast had to be exchanged :) Added tests for that.

mihails-strasuns-sociomantic:
Updated, rebased

src/ocean/meta/codegen/CTFE.d
```diff
@@ -30,15 +30,15 @@ import ocean.meta.types.Qualifiers;
 
 *******************************************************************************/
 
-istring toString ( long i )
+istring toString ( T ) ( T i )
```
gavin-norman-sociomantic, line 33: By the way, this is a bit of a weird function / module name for something that only handles integers.

src/ocean/meta/types/Function.d
```diff
@@ -0,0 +1,143 @@
...
+
+module ocean.meta.types.Function;
+
+import ocean.meta.AliasSeq;
+import ocean.meta.traits.Basic;
+import ocean.meta.types.Typedef;
+
+/*******************************************************************************
+
+    Evaluates to a compile-time sequence representing the parameters of Fn
```
gavin-norman-sociomantic: "parameters of Callable" ?

src/ocean/meta/types/Arrays.d
```diff
@@ -107,3 +107,30 @@ unittest
...
+{
+    static if (isArrayType!(T) == ArrayKind.Static)
+        alias ElementTypeOf!(T)[] SliceIfStaticArray;
+    else
+        alias T SliceIfStaticArray;
+}
+
+///
```
gavin-norman-sociomantic: This is a test, really, not a usage example. It'd be interesting to see an actual usage example (i.e. a situation where this template is useful), as its usefulness is not immediately obvious.

gavin-norman-sociomantic (commented):
Just a couple of small points.

src/ocean/meta/codegen/CTFE.d
mihails-strasuns-sociomantic: Anyone is welcome to implement support for other types if needed :P

mihails-strasuns-sociomantic:
I have decided to take a somewhat different approach with `SliceIfStaticArray` - now it is branded as purely D2 migration helper and added examples shows such use case.

src/ocean/meta/codegen/CTFE.d
```diff
@@ -51,15 +51,26 @@ unittest
     static assert (toString(14L) == "14");
     static assert (toString(long.max) == "9223372036854775807");
     static assert (toString(long.min) == "-9223372036854775808");
+    static assert (toString(int.min) == "-2147483648");
+    static assert (toString(int.max) == "2147483647");
+    static assert (toString(short.min) == "-32768");
+    static assert (toString(short.max) == "32767");
```
david-eckardt-sociomantic: For the sake of completeness check `ulong.max`, too. Values greater than `long.max` used to be a notorious source of trouble, and it's good to see they are supported now.

src/ocean/meta/types/Function.d
david-eckardt-sociomantic: > "parameters of Callable" ?

:+1:

src/ocean/meta/types/Function.d
```diff
@@ -0,0 +1,143 @@
...
+
+    Params:
+        Callable = callable type (function/delegate/function pointer) or
+            function symbol
+
+*******************************************************************************/
+
+public template ParametersOf ( Callable... )
+{
+    static assert (Callable.length == 1);
```
david-eckardt-sociomantic, line 40: Why is it a variadic template and not `ParametersOf ( Callable )`?

src/ocean/meta/types/Function.d
```diff
@@ -0,0 +1,143 @@
...
+    alias ParametersOf!(typeof(dg)) Params;
+
+    static assert (Params.length == 3);
+    static assert (is(Params[0] == int));
+    static assert (is(Params[1] == double));
+    static assert (is(Params[2] == char[]));
+}
+
+/*******************************************************************************
+
```
david-eckardt-sociomantic, line 91: Description missing.

src/ocean/meta/types/Function.d
```diff
@@ -0,0 +1,143 @@
...
+            } ()) ReturnTypeOf;
+    }
+    else
+    {
+        alias typeof({
+                ParametersOf!(Callable[0]) args;
+                Callable[0] call;
+                return call(args);
+            } ()) ReturnTypeOf;
+    }
```
david-eckardt-sociomantic, line 122: Couldn't this be done using `static if (is(Callable[0] R == return))`?

src/ocean/meta/types/Function.d
david-eckardt-sociomantic: Ah, it is probably to allow for accepting a function symbol as a template argument. Still this needs a brief note to avoid confusion.

src/ocean/meta/types/Function.d
mihails-strasuns-sociomantic: Only if it is a function AFAIR (so, for example, won't work for `opCall` object)

src/ocean/meta/types/Function.d
david-eckardt-sociomantic: This is fixed in the "Move meta Function module to correct package" commit, maybe amend this commit instead.

mihails-strasuns-sociomantic:
Updated per @david-eckardt-sociomantic comments

david-eckardt-sociomantic (approved):

PR #424, opened 2017-12-21
Simplify --version and add --build-info

### LINKED ISSUE #249: Make --version just print the version and add --build-info
Coming from https://github.com/sociomantic-tsunami/ocean/pull/247#issuecomment-329248597.

After that PR is merged, I will like to rebump the version thingy for v4.

My idea is to:

* Change `--version` to just print the version, as most programs do.
* Add a `--build-info` to print the whole 2MiB long version information.
* Make `--build-info` use a more *tabular* format, only treating the `version` key specially, and then just print all the rest of the items sorted by key to simplify things, and using a more multi-line `key=value` format.
* Do the same formatting change for logging (i.e. keep the full build info), but keep it as a one-line (multi-lines in the log might be weird).

Nothing will be changed on the MakD side, all the version/build information will come as usual in the `version_info`, but only the `version` key will be treated specially.

mihails-strasuns-sociomantic:
Fixes #249

### DIFF
--- a/relnotes/version.migration.md
+++ b/relnotes/version.migration.md
@@ -1,0 +1,35 @@
+## Changes in --version flag handling and version log
+
+`VersionArgsExt` used in both `CliApp` and `DaemonApp` now implements
+`--version` flag to only print basic version information, without any detailed
+build description.
+
+Instead, new `--build-info` flag is introduced that prints all detailed
+information that used to be present in old `--version` output. However, format
+of that output was changed to use plain `key=value` pairs, one per each line. It
+also doesn't treat any keys specially, printing all data found in the supplied
+`VersionInfo`.
+
+Old output:
+
+```
+$ ./app --version
+app version v1.0 (compiled by 'author' on today with dmd1 using lib1:v10.0 lib2:v0.5)
+```
+
+New output:
+
+```
+$ ./app --version
+app version v1.0
+$ ./app --build-info
+app version v1.0
+build_author=author
+build_date=today
+compiler=dmd1
+lib_lib1=v10.0
+lib_lib2=v0.5
+```
+
+Information logged to `version.log` will match the one printed by `--build-info`
+but will use comma-separated `key=value` pairs instead of multi-line.
--- a/src/ocean/util/app/ext/VersionArgsExt.d
+++ b/src/ocean/util/app/ext/VersionArgsExt.d
@@ -37,7 +37,7 @@ import ocean.transition;
 import ocean.util.log.Logger;
 import ocean.util.log.AppendFile;
 import ocean.util.log.LayoutDate;
-import ocean.core.Array: sort;
+import ocean.core.array.Mutation /* : moveToEnd, sort */;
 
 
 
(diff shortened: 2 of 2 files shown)


relnotes/version.migration.md
```diff
@@ -0,0 +1,14 @@
...
+build description.
+
+Instead, new `--build-info` flag is introduced that prints all detailed
+information that used to be present in old `--version` output. However, format
+of that output was changed to use plain `key=value` pairs, one per each line. It
+also doesn't treat any keys specially, printing all data found in the supplied
+`VersionInfo`.
+
+Information logged to `version.log` will match the one printed by `--build-info`
+but will use comma-separated `key=value` pairs instead of multi-line.
```
leandro-lucarella-sociomantic: Can you include some example here with an *before* and *after*?

src/ocean/util/app/ext/VersionArgsExt.d
```diff
@@ -119,92 +119,71 @@ class VersionArgsExt : IApplicationExtension, IArgumentsExtExtension,
...
+            return app_name ~ " version " ~ *v;
+        else
+            return app_name ~ " unkown version";
     }
 
-
     /***************************************************************************
 
-        Get the program's name and full version information as a string.
+        Get the program's name and basic version information as a string.
```
leandro-lucarella-sociomantic: Why this change? Isn't `--build-info` giving the full build information?

src/ocean/util/app/ext/VersionArgsExt.d
```diff
@@ -230,7 +209,10 @@ class VersionArgsExt : IApplicationExtension, IArgumentsExtExtension,
 
     public override void setupArgs ( IApplication app, Arguments args )
     {
-        args("version").params(0).help("show version information and exit");
+        args("version").params(0)
+            .help("show version information and exit");
+        args("build-info").params(0)
+            .help("show detailed build information and exit");
```
leandro-lucarella-sociomantic, line 215: Should these arguments conflict? If not maybe it should be mentioned that if both are found, then `--build-info` will be used (if I understand correctly, give the `if` order, that's what will happen).

src/ocean/util/app/ext/VersionArgsExt.d
```diff
@@ -334,7 +307,7 @@ class VersionArgsExt : IApplicationExtension, IArgumentsExtExtension,
             return;
         }
 
-        this.ver_log.info(getVersionString(app.name, this.ver));
+        this.ver_log.info(getBuildInfoString(app.name, this.ver, true));
```
leandro-lucarella-sociomantic, line 310: :heart:

leandro-lucarella-sociomantic (commented):
Besides the minor comments, LGTM.

src/ocean/util/app/ext/VersionArgsExt.d
mihails-strasuns-sociomantic: `--build-info` is an exact superset of `--version` (prints the same stuff + more of its own), so it makes sense to make you can supply both

mihails-strasuns-sociomantic:
Updated

src/ocean/util/app/ext/VersionArgsExt.d
leandro-lucarella-sociomantic: For me it would make more sense to be able to use both it wasn't a super-set (so using both would print more info than only using one). For me exactly because it is a superset is that using both would look super weird. But it's just a minor thing...

leandro-lucarella-sociomantic (approved):

leandro-lucarella-sociomantic:
Test needs update though.

```
./src/ocean/util/app/ext/VersionArgsExt.d:485: test failure: expression 'test version v1.0, build_author=me, build_date=today, compiler=dmd3, extra=useful, lib_awesome=v10.0, lib_sucks=v0.5, more=info == test version v1.0
```

mihails-strasuns-sociomantic:
Green

PR #433, opened 2018-01-12
Fix release notes

leandro-lucarella-sociomantic:
A mixed combination of old and new formatting is used in release notes,
not only in different files but even also in the same file. There are
also a couple of other formatting issues and some other minor things to
fix.

This commit should make everything more consistent and clean.


mihails-strasuns-sociomantic (approved):
Thanks! I noticed that too but you have saved me from fixing it myself :angel:

PR #434, opened 2018-01-16
Issues with registering log files with ReopenableFilesExt

mihails-strasuns-sociomantic:
@gautam-kotian-sociomantic has said:

> It was observed in one of our internal projects that there were a couple of oddities when registering log 
> files with the `ReopenableFilesExt` extension. Specifically:
>
> - Some log files were registered multiple times (most notably `root.log`)
> - `version.log` was _not_ registered at all

### DIFF
--- a/src/ocean/util/app/ext/LogExt.d
+++ b/src/ocean/util/app/ext/LogExt.d
@@ -152,18 +152,28 @@ class LogExt : IConfigExtExtension
         auto log_config = ConfigFiller.iterate!(LogUtil.Config)("LOG", config);
         auto log_meta_config = ConfigFiller.fill!(LogUtil.MetaConfig)("LOG", config);
 
+        enable_loose_parsing(conf_ext.loose_config_parsing);
+
+        // deprecated: tracks already registered files to prevent double
+        // registration by both old and new logger system. Can be removed in
+        // ocean v4 together with the old logger.
+        File[cstring] registered_files;
+
         Appender appender ( istring file, LogUtil.Layout layout )
         {
-            auto stream = new File(file, File.WriteAppending);
-            if ( auto reopenable_files_ext =
-                (cast(Application)app).getExtension!(ReopenableFilesExt) )
+            if (!(file in registered_files))
             {
-                reopenable_files_ext.register(stream);
+                auto stream = new File(file, File.WriteAppending);
+                if ( auto reopenable_files_ext =
+                    (cast(Application)app).getExtension!(ReopenableFilesExt) )
+                {
+                    reopenable_files_ext.register(stream);
+                }
+                registered_files[file] = stream;
             }
-            return new AppendStream(stream, true, layout);
-        }
 
-        enable_loose_parsing(conf_ext.loose_config_parsing);
+            return new AppendStream(registered_files[file], true, layout);
+        }
 
         configureOldLoggersLogExtHack(log_config, log_meta_config, &appender,
             this.layout_maker, this.use_insert_appender);


mihails-strasuns-sociomantic:
Regarding `root.log` - it looks like https://github.com/sociomantic-tsunami/ocean/blob/v3.x.x/src/ocean/util/app/ext/LogExt.d#L168-L172 may be a guilty part. It configured both old and new logging system and thus registers file on reopenable extension twice. @mathias-lang-sociomantic can you confirm?

mihails-strasuns-sociomantic:
As for `version.log` it is configured here https://github.com/sociomantic-tsunami/ocean/blob/v3.x.x/src/ocean/util/app/ext/VersionArgsExt.d#L297-L309 and seems like it is working as intended - this file is only written to during app startup, thus there is nothing to rotate.

mihails-strasuns-sociomantic:
Patch like

```diff
--- a/src/ocean/util/app/ext/LogExt.d
+++ b/src/ocean/util/app/ext/LogExt.d
@@ -152,15 +152,24 @@ class LogExt : IConfigExtExtension
         auto log_config = ConfigFiller.iterate!(LogUtil.Config)("LOG", config);
         auto log_meta_config = ConfigFiller.fill!(LogUtil.MetaConfig)("LOG", config);
 
+        File[cstring] appenders;
+
         Appender appender ( istring file, LogUtil.Layout layout )
         {
-            auto stream = new File(file, File.WriteAppending);
-            if ( auto reopenable_files_ext =
-                (cast(Application)app).getExtension!(ReopenableFilesExt) )
+            auto pstream = file in appenders;
+
+            if (pstream is null)
             {
-                reopenable_files_ext.register(stream);
+                auto stream = new File(file, File.WriteAppending);
+                appenders[file] = stream;
+                if ( auto reopenable_files_ext =
+                    (cast(Application)app).getExtension!(ReopenableFilesExt) )
+                {
+                    reopenable_files_ext.register(stream);
+                }
             }
-            return new AppendStream(stream, true, layout);
+
+            return new AppendStream(appenders[file], true, layout);
         }
```

Seems to fix `root.log` issue indeed. Waiting for @mathias-lang-sociomantic confirmation and will make an ocean patch release.

mathias-lang-sociomantic:
Oh I completely overlooked registration when doing those changes. Yeah that sounds correct.

mihails-strasuns-sociomantic:
Fix attached

src/ocean/util/app/ext/LogExt.d
```diff
@@ -152,18 +152,27 @@ class LogExt : IConfigExtExtension
         auto log_config = ConfigFiller.iterate!(LogUtil.Config)("LOG", config);
         auto log_meta_config = ConfigFiller.fill!(LogUtil.MetaConfig)("LOG", config);
 
+        enable_loose_parsing(conf_ext.loose_config_parsing);
+
+        // tracks already registered files to prevent double registration
+        // by both old and new logger system. Can be removed in ocean v4
+        File[cstring] registered_files;
```
nemanja-boric-sociomantic: Please add `deprecated` somewhere in the comment, so we can grep for it when removing deprecations ;-)

src/ocean/util/app/ext/LogExt.d
```diff
@@ -152,18 +152,27 @@ class LogExt : IConfigExtExtension
         auto log_config = ConfigFiller.iterate!(LogUtil.Config)("LOG", config);
         auto log_meta_config = ConfigFiller.fill!(LogUtil.MetaConfig)("LOG", config);
 
+        enable_loose_parsing(conf_ext.loose_config_parsing);
```
mathias-lang-sociomantic, line 155: Any reason to move this ?

mihails-strasuns-sociomantic: Yep, I have spent some minutes thinking why does it matter to call this right before logger configuration, realized it doesn't matter and moved it to the top to prevent further confusion :angel:

mathias-lang-sociomantic: I'm fine with the move, I find it clearer also, was just wondering if there was an impact on the code I was not seeing.

mathias-lang-sociomantic (approved):

gavin-norman-sociomantic (approved):

PR #441, opened 2018-01-19
Add BinaryHistogram

gavin-norman-sociomantic:
An immigrant from swarm.

### DIFF
--- a/relnotes/histogram.feature.md
+++ b/relnotes/histogram.feature.md
@@ -1,0 +1,9 @@
+### New binary logarithmic histogram for integer values
+
+`ocean.math.BinaryHistogram`
+
+The new struct template `BinaryHistogram` is useful for tracking stats on the
+distribution of any kind of integer values, separated into power of two bins.
+The struct provides a method for convenient passing of the aggregated stats to a
+`StatsLog` instance.
+
--- a/src/ocean/math/BinaryHistogram.d
+++ b/src/ocean/math/BinaryHistogram.d
@@ -1,0 +1,393 @@
+/*******************************************************************************
+
+    An integer, binary logarithmic statistics counter
+
+    Collects the following statistical information:
+    - a binary logarithmic histogram with bins from ≥1 to <2^MaxPow2
+      (see template documentation, below), one bin per power of two, plus one
+      bin for each 0 and ≥2^MaxPow2,
+    - the total number of entries and the aggregated total amount.
+
+    To reset all counters to zero use `BinaryHistogram.init`.
+
+    Copyright:
+        Copyright (c) 2017-2018 sociomantic labs GmbH. All rights reserved.
+
+    License:
+        Boost Software License Version 1.0. See LICENSE.txt for details.
+
+*******************************************************************************/
+
+module ocean.math.BinaryHistogram;
+
+import ocean.transition;
+import ocean.core.Verify;
+
+/*******************************************************************************
+
+    An integer, binary logarithmic statistics counter
+
+    Params:
+        MaxPow2 = the maximum power of two that is tracked in the lower bins of
+            the counter. Any values >= 2^MaxPow2 will be aggregated in the upper
+            bin.
+        Suffix = suffix for the names of the fields of the Bins accessor struct
+
+*******************************************************************************/
+
+public struct BinaryHistogram ( uint MaxPow2, istring Suffix = "" )
+{
+    // Values above the range of ulong are not supported.
+    static assert(MaxPow2 < 64,
+        "Buckets for 2^64-bit values or larger unsupported");
+
+    import core.bitop: bsr;
+
+    /// The number of bins. (The specified number of powers of two, plus one bin
+    /// for 0B and one for ≥1<<MaxPow2.)
+    private const NumBins = MaxPow2 + 2;
+
+    /// The aggregated total from all calls to `add()`.
+    public ulong total;
+
+    /// The total number of calls to `add()`.
+    public uint count;
+
+    /***************************************************************************
+
+        The bins of the byte count histogram. Given a number, the index i of the
+        bin to increment is calculated as follows:
+
+        - If 1 ≤ n and n < 2^MaxPow2: i = floor(log_2(n)) + 1,
+        - otherwise, if t < 1: i = 0,
+        - otherwise, t ≥ 2^MaxPow2: i = NumBins - 1.
+
+        For example, for MaxPow2 == 16, the following bins exist:
+        0: 0,
+        1:       1, 2:  2 -   3, 3:   4 -   7, 4:   8 -  15,  5:  16 -   31,
+        6: 32 - 63, 7: 64 - 127, 8: 128 - 255, 9: 256 - 511, 10: 512 - 1023,
+        11: 1ki -  (2ki-1), 12:  2ki -  (4ki-1), 13:  4ki -  (8ki-1),
+        14: 8ki - (16ki-1), 15: 16ki - (32ki-1), 16: 32ki - (64ki-1),
+        17: 64ki - ∞
+
+    ***************************************************************************/
+
+    private uint[NumBins] bins;
+
+    /***************************************************************************
+
+        Struct with one uint field per bin (see this.bins), named as follows
+        (assuming Suffix == ""):
+            from_0, from_1, from_2, from_4, from_8,
+            from_16, ...,
+            from_1Ki, from_2Ki, ...,
+            etc
+
+        Useful, for example, for logging the whole histogram. An instance of
+        this struct is returned by `stats()`.
+
+    ***************************************************************************/
+
+    private struct Bins
+    {
+        import CTFE = ocean.meta.codegen.CTFE : toString;
+
+        /***********************************************************************
+
+            Interprets the passed bins array as a Bins instance.
+
+            Params:
+                array = bins array to reinterpret
+
+            Returns:
+                the passed bins array as a Bins instance
+
+        ***********************************************************************/
+
+        public static Bins fromArray ( typeof(BinaryHistogram.bins) array )
+        {
+            return *(cast(Bins*)array.ptr);
+        }
+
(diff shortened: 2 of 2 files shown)


src/ocean/math/BinaryHistogram.d
```diff
@@ -0,0 +1,384 @@
...
+            the counter. Any values higher than 1<<MaxPow2 will be aggregated in
+            the upper bin
+        Suffix = suffix for the names of the fields of the Bins accessor struct
+
+*******************************************************************************/
+
+public struct BinaryHistogram ( uint MaxPow2, istring Suffix = "" )
+{
+    import core.bitop: bsr;
+    import ocean.core.Traits : FieldName;
```
mihails-strasuns-sociomantic: `ocean.meta.codegen.Identifiers : fieldIdentifier` :)

src/ocean/math/BinaryHistogram.d
```diff
@@ -0,0 +1,384 @@
...
+
+    /// The number of bins. (The specified number of powers of two, plus one bin
+    /// for 0B and one for ≥1<<MaxPow2.)
+    private const NumBins = MaxPow2 + 2;
```
mihails-strasuns-sociomantic: Could be useful to expose as public

src/ocean/math/BinaryHistogram.d
```diff
@@ -0,0 +1,384 @@
...
+            etc
+
+        Useful, for example, for logging the whole histogram. An instance of
+        this struct is returned by `stats()`.
+
+    ***************************************************************************/
+
+    private struct Bins
+    {
+        import ocean.core.Traits : ctfe_i2a;
```
mihails-strasuns-sociomantic: `ocean.meta.codegen.CTFE : toString`

src/ocean/math/BinaryHistogram.d
```diff
@@ -0,0 +1,384 @@
...
+                array = bins array to reinterpret
+
+            Returns:
+                the passed bins array as a Bins instance
+
+        ***********************************************************************/
+
+        public static Bins fromArray ( typeof(BinaryHistogram.bins) array )
+        {
+            return *(cast(Bins*)array.ptr);
```
mihails-strasuns-sociomantic, line 109: It is not immediately obvious to me why do you need store the same data as both `Bins` struct and `ubyte` array.

src/ocean/math/BinaryHistogram.d
```diff
@@ -0,0 +1,384 @@
...
+        Returns:
+            n
+
+    ***************************************************************************/
+
+    public ulong record ( ulong n )
+    {
+        this.count++;
+        this.total += n;
+        this.bins[n? (n < (1UL << 16))? bsr(n) + 1 : $ - 1 : 0]++;
```
mihails-strasuns-sociomantic: The same question - if onyl `Bins` struct was used, this code could be rewritten like this:

```D
n = n ? (n < (1UL << 16))? bsr(n) + 1 : $ - 1 : 0;
foreach (idx, ref bin; this.bins.tupleof)
{
    if (idx == n)
        bin += 1;
}
```

It is more verbose but this way we get rid of error-prone reinterpret casting. It will also make possible to use bins of different types if ever needed.

src/ocean/math/BinaryHistogram.d
gavin-norman-sociomantic: Good question. Originally, the bins array was also public. I made it private because I couldn't see a usage of it. You're right, though: the final step would be to remove it entirely.

Perhaps @david-eckardt-sociomantic has some insight into this, as the original author.

src/ocean/math/BinaryHistogram.d
david-eckardt-sociomantic: Run-time indexing with O(0) time complexity through `bsr` is the reason for using an array.
Should we use a `foreach` loop there is no need for using ` bsr` any more, it could be done in the loop:
```d
if (n >= 1UL << this.bins.tupleof.length)
{
    this.bins.tupleof[$ - 1]++;
    return;
}

foreach_reverse (idx, ref bin; this.bins.tupleof[1 .. $ - 1])
    if (n & (1UL << idx))
    {
        bin++;
        return;
    }

this.bins.tupleof[0]++;
```

david-eckardt-sociomantic: Another note, included in the previous comment: The constant `16` needs to be adapted to depend on `NumBins`.

david-eckardt-sociomantic (changes requested):
In `record` change the constant `16` to depend on `NumBins`.

src/ocean/math/BinaryHistogram.d
```diff
@@ -0,0 +1,384 @@
...
+    /***************************************************************************
+
+        The bins of the byte count histogram. Given a number, the index i of the
+        bin to increment is calculated as follows:
+
+        - If 1 ≤ n and n < 2^MaxPow2: i = floor(log_2(n)) + 1,
+        - otherwise, if t < 1: i = 0,
+        - otherwise, t ≥ 2^MaxPow2: i = NumBins - 1.
+
+        For example, for MaxPow2 == 2, the following bins exist:
```
david-eckardt-sociomantic: Isn't that `MaxPow2 == 16`?

src/ocean/math/BinaryHistogram.d
david-eckardt-sociomantic: I don't mean to bikeshed, but... I'd be inclined to leave it as it is. The iteration is a stupid way of indexing.

src/ocean/math/BinaryHistogram.d
```diff
@@ -0,0 +1,384 @@
...
+
+        Params:
+            n = the value to add to the histogram
+
+        Returns:
+            n
+
+    ***************************************************************************/
+
+    public ulong record ( ulong n )
```
gavin-norman-sociomantic: Looking again, I'm not sure about this method name. "Record" is a verb and a noun, so I think it's a bit confusing. Maybe simply `add`?

src/ocean/math/BinaryHistogram.d
```diff
@@ -0,0 +1,384 @@
...
+import ocean.core.Verify;
+
+/*******************************************************************************
+
+    An integer, binary logarithmic statistics counter
+
+    Params:
+        MaxPow2 = the maximum power of two that is tracked in the lower bins of
+            the counter. Any values higher than 1<<MaxPow2 will be aggregated in
+            the upper bin
```
david-eckardt-sociomantic: State the allowed maximum for `MaxPow2`, imposed by the limited number of order-of-maginute suffixes (the highest being `Ei`).

david-eckardt-sociomantic: Maybe add a unit test for `MaxPow2 == 0`.

src/ocean/math/BinaryHistogram.d
```diff
@@ -0,0 +1,384 @@
...
+                TODO
+
+            Returns:
+                code for the series of fields for bins up to the specified
+                maximum power of 2. (See unittest for examples.)
+
+        ***********************************************************************/
+
+        private static istring divisionBinVariables ( uint max_power,
+            istring suffix )
```
david-eckardt-sociomantic: The parameters aren't actually needed, you could use `MaxPower` and `Suffix` directly (or move the function to module scope).

src/ocean/math/BinaryHistogram.d
david-eckardt-sociomantic: Good idea, or `count`?

david-eckardt-sociomantic: `opAddAssign` :trollface:

gavin-norman-sociomantic:
Updated.

src/ocean/math/BinaryHistogram.d
```diff
@@ -0,0 +1,395 @@
...
+            the counter. Any values higher than 1<<MaxPow2 will be aggregated in
+            the upper bin.
+        Suffix = suffix for the names of the fields of the Bins accessor struct
+
+*******************************************************************************/
+
+public struct BinaryHistogram ( uint MaxPow2, istring Suffix = "" )
+{
+    // The left shift operator (used internally) only supports right operands of
+    // < 64.
```
david-eckardt-sociomantic: It is not the shift operator but the range of `ulong`.

src/ocean/math/BinaryHistogram.d
```diff
@@ -0,0 +1,393 @@
+/*******************************************************************************
+
+    An integer, binary logarithmic statistics counter
+
+    Collects the following statistical information:
+    - a binary logarithmic histogram with bins from ≥1 to <1<<MaxPow2
+      (see template documentation, below), one bin per power of two, plus one
+      bin for each 0 and ≥1<<MaxPow2,
```
david-eckardt-sociomantic: A small thing: Replace `1<<x` with `2^x`, this is about powers of 2.

gavin-norman-sociomantic: Updated.

david-eckardt-sociomantic (approved):

nemanja-boric-sociomantic:
Needs rebase.

gavin-norman-sociomantic:
Rebased on v3.x.x.

mihails-strasuns-sociomantic (approved):

relnotes/histogram.feature.md
```diff
@@ -0,0 +1,9 @@
+## New binary logarithmic histogram for integer values
```
mihails-strasuns-sociomantic: New header format :) `###`

mihails-strasuns-sociomantic (changes requested):

relnotes/histogram.feature.md
gavin-norman-sociomantic: :+1: updated.

mihails-strasuns-sociomantic (approved):

PR #442, opened 2018-01-19
A whole bunch of cleanup

mathias-lang-sociomantic:
Couldn't find a more descriptive name for this PR.
I was mostly removing dead code / always true `version` block. Did a bit of cleanup in the indent / documentation while doing so.
Also deprecated `ocean.io.UnicodeFile`, which no-one uses.

### DIFF
--- a/relnotes/unicodefile.deprecation.md
+++ b/relnotes/unicodefile.deprecation.md
@@ -1,0 +1,3 @@
+## Deprecated ocean.io.UnicodeFile
+
+This module is not used anywhere, and is neither tested nor maintained.
--- a/src/ocean/io/Console.d
+++ b/src/ocean/io/Console.d
@@ -27,10 +27,8 @@ import ocean.sys.Common;
 import ocean.io.device.Device,
        ocean.io.stream.Buffered;
 
-version (Posix)
-{
-    import core.sys.posix.unistd: isatty;
-}
+import core.sys.posix.unistd: isatty;
+
 
 /*******************************************************************************
 
--- a/src/ocean/io/FileSystem.d
+++ b/src/ocean/io/FileSystem.d
@@ -29,19 +29,12 @@ import ocean.core.ExceptionDefinitions;
 
 import ocean.io.Path : standard, native;
 
-/*******************************************************************************
+import ocean.stdc.string;
+import core.sys.posix.unistd;
+import core.sys.posix.sys.statvfs;
 
-*******************************************************************************/
-
-version (Posix)
-        {
-        import ocean.stdc.string;
-        import core.sys.posix.unistd,
-                       core.sys.posix.sys.statvfs;
-
-        import ocean.io.device.File;
-        import Integer = ocean.text.convert.Integer_tango;
-        }
+import ocean.io.device.File;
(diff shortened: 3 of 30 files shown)


src/ocean/io/Console.d
```diff
@@ -498,16 +498,3 @@ static ~this()
...
-
-/******************************************************************************
-
- ******************************************************************************/
-
-debug (Console)
-{
-    void main()
-    {
-        Cout ("hello world").newline;
```
leandro-lucarella-sociomantic: I love the removal of old ununsed code, but it would be nice if these tests are not *lost* (I know, since they are never ran by anyone technically they are already lost, but still). I would add the tests as unittests when possible and when not as *integrationtest*s, if it's not too much trouble (i.e. only "saving" the low hanging fruit).

leandro-lucarella-sociomantic (approved):

mathias-lang-sociomantic:
@leandro-lucarella-sociomantic : I stripped out the commit which removed the debug block. I understand your argument, and it makes sense. I had a similar feeling while stripping them out. Something that surprised me is that the container modules had little tests... I'll look into how we can improve that, and will submit another PR

src/ocean/io/FileSystem.d
```diff
@@ -53,139 +46,124 @@ version (Posix)
 
 struct FileSystem
 {
-        /***********************************************************************
```
gavin-norman-sociomantic: Could you separate all the indentation fixes? :grimacing:

mathias-lang-sociomantic: I hope you meant "indentation fixes not in version" blocks !
Sure, will do.

relnotes/unicodefile.deprecation.md
```diff
@@ -0,0 +1,3 @@
+## Deprecated ocean.io.UnicodeFile
+
+This module is not used anywhere, and is not tested nor maintained.
```
mihails-strasuns-sociomantic: `nor` requires `neither`, right @gavin-norman-sociomantic ?

mathias-lang-sociomantic:
Splitted the unrelated changes in their own commit. BTW, if you value your sanity, use `?w=1` in the URL, e.g. https://github.com/sociomantic-tsunami/ocean/pull/442/files?w=1 (works with commits too).

mihails-strasuns-sociomantic (approved):

relnotes/unicodefile.deprecation.md
gavin-norman-sociomantic: Strictly speaking, yes. It should say "and is neither tested nor maintained". (What Mathias wrote sounds fine, on a casual level, though.)

mathias-lang-sociomantic: You learn something new every day! Fixed the wording, thanks!

mihails-strasuns-sociomantic:
Conflicts with other merged cleanup PR now

mathias-lang-sociomantic:
Fixed it

mathias-lang-sociomantic:
Ping

mihails-strasuns-sociomantic (approved):

PR #446, opened 2018-01-22
 Deprecate DateTime's formatWide

mathias-lang-sociomantic:
It's unused and likely buggy (since it does not support cstring for ex)


mathias-lang-sociomantic:
Rebased

mihails-strasuns-sociomantic (approved):

PR #448, opened 2018-01-26
Update LibFeatures module for v4.0.0

gavin-norman-sociomantic:
This was forgotten, right?


mathias-lang-sociomantic:
Forgetting LibFeatures ? That would never happen :D

mihails-strasuns-sociomantic (approved):

Issue #451, opened 2018-02-05
Flawed design of `Scheduler.exception_handler`


mihails-strasuns-sociomantic:
https://github.com/sociomantic-tsunami/ocean/blob/v3.x.x/src/ocean/task/Scheduler.d#L175 is a public field and expected to be customizable. However it will only work before the event loop was started because it is passes by value https://github.com/sociomantic-tsunami/ocean/blob/v3.x.x/src/ocean/task/Scheduler.d#L545-L550 here. This will cause mysterious failures if someone attempts to change the exception handler after event loop was started.

gavin-norman-sociomantic:
So this is unrelated to https://github.com/sociomantic-tsunami/turtle/pull/36, right? (You just noticed this while investigating that?)

mihails-strasuns-sociomantic:
It is semi-related in a sense that this issue is why turtle point release is needed and similar fix in the app code won't work (as they can't modify handler before event loop starts).

don-clugston-sociomantic:
There is another problem which I have noticed. I am not sure if this is an independent issue, but it is a design problem with the exception handler.

The default exception handler catches the exception and rethrows it. I think this assumes that the calling fiber has a try-catch block which will handle that exception. 
But if the exception is not caught, there'll be a core dump. Unfortunately, the stack for the fiber which caused the exception has already been unwound. So the stack trace will begin from the moment when the exception is rethrown. This contains no useful information.

So, that's pretty annoying. By default, core dumps for uncaught exceptions are completely useless. But, the reason why it behaves this way is interesting.
Fundamentally, the idea of switching stacks while processing exceptions would need to be handled at a lower level. The runtime needs to know about fibers when it is walking the stack.

I don't know how it works in Linux but I do understand the Windows method, and I presume they're pretty similar. Exception handling has two passes, one where you walk the stack trying to find the target `catch` block, and then a second time when you walk the stack, unrolling it and executing the catch block. Switching stacks needs to be done in the first pass, but this code tries to do it only in the second. 

Catching and rethrowing looks superficially similar but it isn't quite right.

david-eckardt-sociomantic:
[...]

PR #454, opened 2018-02-07
Minimal fixes for exception_handler

mihails-strasuns-sociomantic:
Does not try to address fundamental flaws discussed in https://github.com/sociomantic-tsunami/ocean/issues/451 but
makes it at least possible to reset handler back to null after event loop was
already started.

### DIFF
--- a/src/ocean/io/select/EpollSelectDispatcher.d
+++ b/src/ocean/io/select/EpollSelectDispatcher.d
@@ -151,7 +151,7 @@ public class EpollSelectDispatcher : IEpollSelectDispatcherInfo
 
     ***************************************************************************/
 
-    private void delegate (Exception) unhandled_exception_hook;
+    private bool delegate (Exception) unhandled_exception_hook;
 
     /***************************************************************************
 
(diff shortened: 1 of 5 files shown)


src/ocean/task/Scheduler.d
```diff
@@ -689,6 +688,24 @@ final class Scheduler : IScheduler
                 throw e;
         }
     }
+
+    /***************************************************************************
+
+        Wraps configured `exception_handler` into API that doesn't refer to
+        tasks and thus is usable by EpollSelectDispatcher
+
```
daniel-zullo, line 696: `Params` and `Returns` documentation is missing

src/ocean/io/select/EpollSelectDispatcher.d
```diff
@@ -712,8 +712,11 @@ public class EpollSelectDispatcher : IEpollSelectDispatcherInfo
             }
             catch (Exception e)
             {
-                if (unhandled_exception_hook !is null)
-                    unhandled_exception_hook(e);
+                if (  (unhandled_exception_hook !is null)
+                    && unhandled_exception_hook(e))
+                {
+                    // do nothing
+                }
```
daniel-zullo: What about
```D
if ( unhandled_exception_hook is null || !unhandled_exception_hook(e) )
{
    throw e;
}
```
instead?

mihails-strasuns-sociomantic:
Updated per @daniel-zullo-sociomantic comments

PR #457, opened 2018-02-12
Update documentation for serialize()

daniel-zullo:
The method `serialize()` needs to update the documentation
specially because the return value might be just a slice of
the buffer passed as a ref parameter.

### DIFF
--- a/src/ocean/util/serialize/contiguous/Serializer.d
+++ b/src/ocean/util/serialize/contiguous/Serializer.d
@@ -54,13 +54,23 @@ struct Serializer
 
     /***************************************************************************
 
-        Serializes the data in s
+        Serializes the data in src.
+
+        Note that the serialization process may expand the dst buffer more than
+        ultimately necessary, but it always returns a slice into the buffer
+        that covers bytes of the struct instance itself, not including any of
+        indirections (that are placed after).
+        Therefore calling applications should be careful to use the returned
+        buffer rather than the dst buffer directly.
 
         Params:
             S = type of the struct to dump
             src = struct to serialize
             dst = buffer to write to. It is only extended if needed and
                 never shrunk
+
+        Returns:
+            the slice to the serialize data
 
     ***************************************************************************/
 


src/ocean/util/serialize/contiguous/Serializer.d
```diff
@@ -54,14 +54,21 @@ struct Serializer
 
     /***************************************************************************
 
-        Serializes the data in s
+        Serializes the data in src.
+
+        Note that the serialization process may expand the dst buffer more than
+        ultimately necessary, but always returns a slice into the buffer of
+        only the required length.
```
mihails-strasuns-sociomantic: To be more specific, it return slice that covers bytes of struct instance itself, not including any of indirections (that are placed after). Can you please mention that?

daniel-zullo: Yes, sure. I'll add your suggestion and also add the line `Calling applications should be careful to use the returned buffer rather than the dst buffer directly`  if it sounds good to you.

daniel-zullo:
Updated documentation based on a suggestion from Mihails.

<details>
<summary>Previous head was 0cad3d6, new head at fe33511. Diff between heads are:</summary>

```diff
diff --git a/src/ocean/util/serialize/contiguous/Serializer.d b/src/ocean/util/serialize/contiguous/Serializer.d
index 3ca1add..203c3b7 100644
--- a/src/ocean/util/serialize/contiguous/Serializer.d
+++ b/src/ocean/util/serialize/contiguous/Serializer.d
@@ -57,8 +57,11 @@ struct Serializer
         Serializes the data in src.
 
         Note that the serialization process may expand the dst buffer more than
-        ultimately necessary, but always returns a slice into the buffer of
-        only the required length.
+        ultimately necessary, but it always returns a slice into the buffer
+        that covers bytes of the struct instance itself, not including any of
+        indirections (that are placed after).
+        Therefore calling applications should be careful to use the returned
+        buffer rather than the dst buffer directly.
 
         Params:
             S = type of the struct to dump
```
</details>

mihails-strasuns-sociomantic (approved):

daniel-zullo:
I clicked on the `update branch` and added a new commit to merge v3.x.x. I'll rebase my branch instead and delete the commit added.

daniel-zullo:
Rebased onto latest v3.x.x

PR #458, opened 2018-02-13
 Allow removing the client from the selected set in EpollDispatcher.unregister

nemanja-boric-sociomantic:
`EpollSelectDispatcher.unregister` now accepts an optional flag
(`remove_from_selected_set`) which will remove the client from the
selected set if it's in there. This will allow other ISelectClients to
safely destroy other ISelectClients from their handle method
(previously, this was unsafe because even unregistering the client
would just remove it from the all next epoll cycles, but not from the
other one, if the client's fd reported that it's ready to be handled).

The assumption taken in this PR is that the branch predictor will make most of the branches in the `SelectedKeysHandler.opCall` no-ops, or that they have very little cost compared with the IO that is done, so let's try this before taking alternative more complicated approaches.

### DIFF
--- a/integrationtest/unregisterclient/main.d
+++ b/integrationtest/unregisterclient/main.d
@@ -1,0 +1,88 @@
+/*******************************************************************************
+
+    Test for EpollSelectDispatcher.unregister behaviour
+
+    Copyright:
+        Copyright (c) 2018 sociomantic labs GmbH.
+        All rights reserved.
+
+    License:
+        Boost Software License Version 1.0. See LICENSE_BOOST.txt for details.
+        Alternatively, this file may be distributed under the terms of the Tango
+        3-Clause BSD License (see LICENSE_BSD.txt for details).
+
+*******************************************************************************/
+
+module integrationtest.unregisterclient.main;
+
+import ocean.transition;
+
+import ocean.io.select.EpollSelectDispatcher;
+import ocean.io.select.client.SelectEvent;
+import ocean.core.Test;
+
+private EpollSelectDispatcher epoll;
+
+private SelectEvent event1, event2;
+private int count;
+
+// Corrupts the SelectEvent object. This should make test
+// segfault at access
+private void corruptEventObject (SelectEvent client)
+{
+    version (D_Version2)
+    {
+        auto initializer = client.classinfo.initializer();
+    }
+    else
+    {
+        auto initializer = client.classinfo.init;
+    }
+
+    ubyte* ptr = cast(ubyte*)client;
+    ptr[0..initializer.length] = 0;
+}
+
+version(UnitTest) {} else
+void main(istring[] args)
+{
+    bool handler1()
+    {
+        .count++;
+
+        // unregister and delete the other instance
+        .epoll.unregister(.event2, true);
+        corruptEventObject(.event2);
+
+        return false;
+    }
+
+    bool handler2()
+    {
+        .count++;
+
+        // unregister and delete the other instance
+        .epoll.unregister(.event1, true);
+        corruptEventObject(.event1);
+
+        return false;
+    }
+
+    .epoll = new EpollSelectDispatcher();
+
+    .event1 = new SelectEvent(&handler1);
+    .event2 = new SelectEvent(&handler2);
+
+    .epoll.register(.event1);
+    .epoll.register(.event2);
+
+    // trigger them both
+    .event1.trigger();
+    .event2.trigger();
+
+    // Do a select loop
+    .epoll.eventLoop();
+
+    // Make sure only one was handled.
+    test!("==")(.count, 1);
+}
--- a/relnotes/unregister.feature.md
+++ b/relnotes/unregister.feature.md
@@ -1,0 +1,9 @@
+### Allow removing the client from the selected set in EpollSelectDispatcher.unregister
+
+`ocean.io.select.EpollSelectDispatcher`
+
+`EpollSelectDispatcher.unregister` now accepts an optional flag
+(`remove_from_selected_set`) which will remove the client from the selected set
+if it's in there. This guarantees that the unregistered client's handle method
+will not be subsequently called by the selector. The client may thus be safely
+destroyed after unregistering.
--- a/src/ocean/io/select/EpollSelectDispatcher.d
+++ b/src/ocean/io/select/EpollSelectDispatcher.d
@@ -136,6 +136,15 @@ public class EpollSelectDispatcher : IEpollSelectDispatcherInfo
 
     /***************************************************************************
 
+        List of the clients that we're currently handling. Always slices
+        this.events.
+
+    ***************************************************************************/
+
+    private epoll_event_t[] selected_set;
+
+    /***************************************************************************
+
         Re-usable errno exception
 
      **************************************************************************/
@@ -791,9 +800,11 @@ public class EpollSelectDispatcher : IEpollSelectDispatcherInfo
 
                 version ( EpollCounters ) if ( n == 0 ) this.counters.timeouts++;
 
-                auto selected_set = events[0 .. n];
+                this.selected_set = this.events[0 .. n];
+                scope (exit)
+                    this.selected_set = null;
 
-                this.handle(selected_set, this.unhandled_exception_hook);
+                this.handle(this.selected_set, this.unhandled_exception_hook);
 
                 return n;
             }
--- a/src/ocean/io/select/EpollSelectDispatcher.d
+++ b/src/ocean/io/select/EpollSelectDispatcher.d
@@ -50,6 +50,8 @@ import ocean.io.select.selector.RegisteredClients;
 
 import ocean.core.Array : copy;
 
+import ocean.core.array.Search;
+
 import ocean.util.container.AppendBuffer;
 
 import ocean.time.timeout.model.ITimeoutManager;
@@ -367,6 +369,11 @@ public class EpollSelectDispatcher : IEpollSelectDispatcherInfo
 
        Params:
             client = client to unregister
+            remove_from_selected_set = if true, removes the client from the
+                selected set that may be currently being iterated over. This
+                guarantees that the unregistered client's handle method will not
+                be subsequently called by the selector. The client may thus be
+                safely destroyed after unregistering.
 
        Returns:
             0 if everything worked as expected or the error code (errno) as a
@@ -382,13 +389,36 @@ public class EpollSelectDispatcher : IEpollSelectDispatcherInfo
 
      **************************************************************************/
 
-    public int unregister ( ISelectClient client )
+    public int unregister ( ISelectClient client,
+        bool remove_from_selected_set = false )
     {
         try if (client.is_registered)
         {
             scope (success)
             {
                 this.registered_clients -= client;
+
+                if (remove_from_selected_set)
+                {
+                    /// Predicate for finding the ISelectClient inside
+                    /// array of epoll_event_t entries.
+                    bool entry_to_client (Const!(epoll_event_t) entry)
+                    {
+                        return entry.data.ptr == cast(void*)client;
+                    }
+
+                    auto index = this.selected_set.findIf(&entry_to_client);
+                    // Instead of removing the array entry, we'll just invalidate
+                    // it. This is to avoid problems with both the fact that
+                    // SelectedKeysHandler might be foreach-iterating over
+                    // this array at this time and the fact that shrinking
+                    // the slice owned by EpollSelectDispatcher wouldn't shrink
+                    // the slices owned by SelectedKeysHandler.
+                    if (index < this.selected_set.length)
+                    {
+                        this.selected_set[index] = epoll_event_t.init;
+                    }
+                }
             }
 
             if (!this.epollCtl(epoll.CtlOp.EPOLL_CTL_DEL, client.fileHandle,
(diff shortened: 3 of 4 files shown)


src/ocean/io/select/selector/SelectedKeysHandler.d
```diff
@@ -68,7 +68,7 @@ class SelectedKeysHandler: ISelectedKeysHandler
 
     ***************************************************************************/
 
-    public alias int delegate ( ISelectClient client ) UnregisterDg;
+    public alias int delegate ( ISelectClient client, bool = false ) UnregisterDg;
```
david-eckardt-sociomantic: This compiles?!? Anyway, a parameter name would be lovely.

src/ocean/io/select/EpollSelectDispatcher.d
```diff
@@ -382,13 +387,27 @@ public class EpollSelectDispatcher : IEpollSelectDispatcherInfo
...
+                if (remove_from_selected_set)
+                {
+                    scope match_client_ptr = (Const!(epoll_event_t) entry) {
+                        return entry.data.ptr == cast(void*)client;
+                    };
+
+                    auto index = this.selected_set.findIf(match_client_ptr);
+                    if (index < this.selected_set.length)
+                    {
+                        this.selected_set[index] = epoll_event_t.init;
```
david-eckardt-sociomantic: Maybe you could remove the array element i.e.
```d
this.selected_set[index] = this.selected_set[$ - 1];
this.selected_set = this.selected_set[0 .. $ - 1];
```
?
Also, just `foreach (selected_client; this.selected_set)` is probably less code (a matter of taste, though).

nemanja-boric-sociomantic: Removing the array element is problematic because of two things:

1. When this is running, there's a good chance that `SelectedKeysHandler` is already doing `foreach (key; selected_keys)`, so we need to be very careful about changing the array while iterating.

2. How the things are setup at the moment, `EpollSelectDispatcher` has the array that both `EpollSelectDispatcher` and `SelectedKeysHandler` slice. Changing the `this.selected_set` slice will not reflect on the slices in `SelectedKeysHandler`, so we need to add the method to the `SelectedKeysHandler` to remove this array from it's slice, but also this will have to change the array stored here.

Since the only penalty here is checking with the null, and which would always pass, except for very obscure cases where the `ISelectClient` instances have the scoped lifetime, I don't think this is worth it.

david-eckardt-sociomantic: > Maybe you could remove the array element

As discussed verbally, this would interfere with the iteration in `SelectedKeysHandler.opCall` so never mind.

src/ocean/io/select/EpollSelectDispatcher.d
```diff
@@ -791,9 +800,9 @@ public class EpollSelectDispatcher : IEpollSelectDispatcherInfo
 
                 version ( EpollCounters ) if ( n == 0 ) this.counters.timeouts++;
 
-                auto selected_set = events[0 .. n];
+                this.selected_set = events[0 .. n];
 
-                this.handle(selected_set, this.unhandled_exception_hook);
+                this.handle(this.selected_set, this.unhandled_exception_hook);
```
david-eckardt-sociomantic: Maybe set `this.selected_set = null` here?

nemanja-boric-sociomantic: Yes, thank you!

nemanja-boric-sociomantic:
Updated with `scope (exit) this.selected_set = null`

src/ocean/io/select/EpollSelectDispatcher.d
```diff
@@ -791,9 +800,9 @@ public class EpollSelectDispatcher : IEpollSelectDispatcherInfo
 
                 version ( EpollCounters ) if ( n == 0 ) this.counters.timeouts++;
 
-                auto selected_set = events[0 .. n];
+                this.selected_set = events[0 .. n];
```
gavin-norman-sociomantic: Should this actually be `this.events`, too?

src/ocean/io/select/EpollSelectDispatcher.d
```diff
@@ -367,6 +369,9 @@ public class EpollSelectDispatcher : IEpollSelectDispatcherInfo
 
        Params:
             client = client to unregister
+            remove_from_selected_set = if true, removes the client from the
```
gavin-norman-sociomantic, line 372: Release notes: new feature.

src/ocean/io/select/EpollSelectDispatcher.d
```diff
@@ -367,6 +369,9 @@ public class EpollSelectDispatcher : IEpollSelectDispatcherInfo
...
+                selected_set that we could be currently iterating. This will allow
```
gavin-norman-sociomantic: Small tweak / expansion:
"if true, removes the client from the selected set that may be currently being iterated over. This guarantees that the unregistered client's handle method will not be subsequently called by the selector. The client may thus be safely destroyed after unregistering."

src/ocean/io/select/EpollSelectDispatcher.d
```diff
@@ -382,13 +387,27 @@ public class EpollSelectDispatcher : IEpollSelectDispatcherInfo
...
     {
         try if (client.is_registered)
         {
             scope (success)
             {
                 this.registered_clients -= client;
+
+                if (remove_from_selected_set)
+                {
+                    scope match_client_ptr = (Const!(epoll_event_t) entry) {
```
gavin-norman-sociomantic: What's the reason for this unusual syntax, as opposed to just making this a normal nested function?

src/ocean/io/select/EpollSelectDispatcher.d
gavin-norman-sociomantic: Maybe add a comment that the item is not actually removed from the set, but set to null.

src/ocean/io/select/selector/SelectedKeysHandler.d
```diff
@@ -122,7 +122,10 @@ class SelectedKeysHandler: ISelectedKeysHandler
     {
         foreach (key; selected_set)
         {
-            this.handleSelectedKey(key, unhandled_exception_hook);
+            if (key != epoll_event_t.init)
```
gavin-norman-sociomantic: Worth adding a comment explaining why an entry could be null.

integrationtest/unregisterclient/main.d
```diff
@@ -0,0 +1,123 @@
...
+            auto initializer = client.classinfo.init;
+        }
+
+        ubyte* ptr = cast(ubyte*)client;
+        ptr[0..initializer.length] = 0;
+    }
+
+    private bool handler1()
+    {
+        count++;
```
gavin-norman-sociomantic: `this.` :)

integrationtest/unregisterclient/main.d
```diff
@@ -0,0 +1,123 @@
...
+        // unregister and delete the other instance
+        this.epoll.unregister(this.event2, true);
+        corruptEventObject(this.event2);
+
+        return false;
+    }
+
+    private bool handler2()
+    {
+        count++;
```
gavin-norman-sociomantic: `this.` :)

integrationtest/unregisterclient/main.d
```diff
@@ -0,0 +1,123 @@
...
+        ptr[0..initializer.length] = 0;
+    }
+
+    private bool handler1()
+    {
+        count++;
+
+        // unregister and delete the other instance
+        this.epoll.unregister(this.event2, true);
+        corruptEventObject(this.event2);
```
gavin-norman-sociomantic: `this.` :)

integrationtest/unregisterclient/main.d
```diff
@@ -0,0 +1,123 @@
...
+        return false;
+    }
+
+    private bool handler2()
+    {
+        count++;
+
+        // unregister and delete the other instance
+        this.epoll.unregister(this.event1, true);
+        corruptEventObject(this.event1);
```
gavin-norman-sociomantic: `this.` :)

integrationtest/unregisterclient/main.d
```diff
@@ -0,0 +1,123 @@
...
+
+import ocean.io.device.File;
+import ocean.io.select.EpollSelectDispatcher;
+import ocean.io.select.client.SelectEvent;
+import ocean.util.app.DaemonApp;
+import ocean.util.test.DirectorySandbox;
+import ocean.core.Test;
+
+/// Main application class
+class EpollTestApp : DaemonApp
```
gavin-norman-sociomantic: Why is this a daemon app btw? You don't seem to be using any of its facilities.

gavin-norman-sociomantic:
Even though it contains new features, this ought to be in a patch release, as it's required to fix a segfault bug in an application.

src/ocean/io/select/EpollSelectDispatcher.d
david-eckardt-sociomantic: `scope`?

gavin-norman-sociomantic: Yeah. I've never seen anyone declare a nested function like that. I'm probably missing something.

david-eckardt-sociomantic: To be honest, me neither. I wonder when one should use `scope` explicitly an when not.

nemanja-boric-sociomantic: If you feel this is the unusual syntax, you're probably right, I'll change it to become the nested function.

nemanja-boric-sociomantic:
> Even though it contains new features, this ought to be in a patch release, as it's required to fix a segfault bug in an application.

Good point.

integrationtest/unregisterclient/main.d
nemanja-boric-sociomantic: It's just the easiest way for me to write these tests - just copy the boilerplate from another test. I'll make it non-daemon app.

mihails-strasuns-sociomantic:
This is similar to TimerSet/TimeoutManager problem we had earlier with task system, right?

nemanja-boric-sociomantic:
Yes, it's pretty much the same problem.

integrationtest/unregisterclient/main.d
gavin-norman-sociomantic: I mean, do you need an app class at all?

nemanja-boric-sociomantic: No, that's what I meant by making it non-daemon app - I'll remove all classes.

nemanja-boric-sociomantic:
Updated with @gavin-norman-sociomantic 's comments + rebase to v3.6.x

src/ocean/io/select/EpollSelectDispatcher.d
```diff
@@ -800,7 +830,9 @@ public class EpollSelectDispatcher : IEpollSelectDispatcherInfo
 
                 version ( EpollCounters ) if ( n == 0 ) this.counters.timeouts++;
 
-                this.selected_set = events[0 .. n];
+                this.selected_set = this.events[0 .. n];
+                scope (exit)
+                    this.selected_set = null;
```
gavin-norman-sociomantic: I think these three lines should be in the first commit.

nemanja-boric-sociomantic: Correct, I messed up the fixup.

nemanja-boric-sociomantic: Forgot to mention, this is moved to the right commit.

relnotes/unregister.feature.md
```diff
@@ -0,0 +1,9 @@
+### Allow removing the client from the selected set in EpollDispatcher.unregister
```
gavin-norman-sociomantic: `EpollSelectDispatcher.unregister`

relnotes/unregister.feature.md
```diff
@@ -0,0 +1,9 @@
...
+
```
gavin-norman-sociomantic: Should also list the names of affected modules. (`ocean.io.select.EpollSelectDispatcher`, in this case.)

relnotes/unregister.feature.md
```diff
@@ -0,0 +1,9 @@
...
+`EpollSelectDispatcher.unregister` now accepts an optional flag
+(`remove_from_selected_set`) which will remove the client from the selected set
+if it's in there. This will allow other `ISelectClient`s to safely destroy
+other `ISelectClient`s from their handle method (previously, this was unsafe
```
gavin-norman-sociomantic: Maybe reword the section in (). It's quite unclear.

(You could probably copy some of the doc from the `unregister` method, say.)

src/ocean/io/select/EpollSelectDispatcher.d
```diff
@@ -382,13 +389,36 @@ public class EpollSelectDispatcher : IEpollSelectDispatcherInfo
...
+                    {
+                        return entry.data.ptr == cast(void*)client;
+                    }
+
+                    auto index = this.selected_set.findIf(&entry_to_client);
+                    // Instead of removing the array entry, we'll just invalidate
+                    // it. This is to avoid problems with both the fact that
+                    // SelectedKeysHandler might be foreach-iterating over
+                    // this array at this time and the fact that shrinking
+                    // the slice owned by EpollSelectDispatcher wouldn't shink
```
gavin-norman-sociomantic: Typo shink.

nemanja-boric-sociomantic:
Updated.

gavin-norman-sociomantic (approved):

gavin-norman-sociomantic:
@david-eckardt-sociomantic please merge if you approve.

PR #459, opened 2018-02-19
Recursive aggregate value visitor + new reset

mihails-strasuns-sociomantic:
This should be used to reimplement things liek DeepReset and maybe even
serializer eventually. Its core implementation is very simple, just dumping it
here for a quick Yes/No review of a concept.

### DIFF
--- a/relnotes/metavalues.feature.md
+++ b/relnotes/metavalues.feature.md
@@ -4,3 +4,22 @@ Hosts utilities that use heavy compile-time reflection to do some automatic
 operation on runtime values. Its core is `ocean.meta.values.VisitValue` module
 which allows to recursively visit arbitrary variable with user-defined
 processing logic.
+
+`ocean.meta.values.Reset`
+
+Provides single function, `reset`, which is intended to replace
+`ocean.util.DeepReset` while being both more performant and using simpler
+implementation. It has two main differences from `DeepReset`:
+
+1) Simplified API - any value can be reset by simply calling `reset` function
+   on it, no matter if it is a single integer, array or deeply nested struct
+   instance.
+
+2) No recursion through reset dynamic arrays. Original `DeepReset` was going through
+   every element of reset again to recursively reset it too and that was very
+   wasteful because such data couldn't be reused again as runtime sets array
+   elements to 0 on its own when length is increased. New `reset` only recurses
+   into nested structs and static arrays, but not dynamic arrays.
+
+It is worth mentioning that all tests from old `DeepReset` still pass with new
+utility which proves that effective observable functionality stays the same.
--- a/relnotes/metavalues.feature.md
+++ b/relnotes/metavalues.feature.md
@@ -1,0 +1,6 @@
+### New `ocean.meta.values` package
+
+Hosts utilities that use heavy compile-time reflection to do some automatic
+operation on runtime values. Its core is `ocean.meta.values.VisitValue` module
+which allows to recursively visit arbitrary variable with user-defined
+processing logic.
--- a/src/ocean/meta/values/Reset.d
+++ b/src/ocean/meta/values/Reset.d
@@ -1,0 +1,92 @@
+/*******************************************************************************
+
+    Utility that resets value state in a way that its memory can be reused again
+    as much as possible. Resets arrays to 0 length and values to their init
+    state.
+
+    Copyright:
+        Copyright (c) 2018 sociomantic labs GmbH.
+        All rights reserved.
+
+    License:
+        Boost Software License Version 1.0. See LICENSE_BOOST.txt for details.
+        Alternatively, this file may be distributed under the terms of the Tango
+        3-Clause BSD License (see LICENSE_BSD.txt for details).
+
+*******************************************************************************/
+
+module ocean.meta.values.Reset;
+
+import ocean.transition;
+
+import ocean.meta.values.VisitValue;
+import ocean.meta.traits.Basic;
+
+import ocean.core.Test;
+
+/*******************************************************************************
+
+    Utility that resets value state in a way that its memory can be reused again
+    as much as possible. Resets arrays to 0 length and values to their init
+    state.
+
+    Non-class references are ignored. For classes it tries to recurse
+    into fields of the class instance without nullifying actual reference.
+
+    Params:
+        value = value to reset
+
+*******************************************************************************/
+
+public void reset ( T ) ( ref T value )
+{
+    Reset visitor;
+    visitValue(value, visitor);
+}
+
+///
+unittest
+{
+    // Can reset single value:
+    int x = 42;
+    reset(x);
+    test!("==")(x, 0);
+
+    // Or some array:
+    int[] arr = [ 1, 2, 3 ];
+    reset(arr);
+    test!("==")(arr.length, 0);
+    test!("!=")(arr.ptr, null);
+
+    // Or some aggregate recursively:
+    struct S
+    {
+        int x;
+        mstring buf;
+    }
+
+    auto s = S(42, "abcd".dup);
+    reset(s);
+    test!("==")(s.x, 0);
+    test!("==")(s.buf.length, 0);
+}
+
+/// Visitor for ocean.meta.values.VisitValue
+private struct Reset
+{
+    // has to be public to be usable from VisitValue module
+    public bool visit ( T ) ( T* value )
+    {
+        static if (isArrayType!(T) == ArrayKind.Dynamic)
+        {
+            (*value).length = 0;
+            enableStomping(*value);
+        }
+        else static if (isPrimitiveType!(T))
+        {
+            *value = T.init;
+        }
+
+        return is(T == class) || !isReferenceType!(T);
+    }
+}
--- a/src/ocean/meta/values/Reset_test.d
+++ b/src/ocean/meta/values/Reset_test.d
@@ -1,0 +1,153 @@
+module ocean.meta.values.Reset_test;
+
+import ocean.meta.values.Reset;
+import ocean.core.array.Mutation;
+import ocean.core.Test;
+
+unittest
+{
+    struct TestStruct
+    {
+        int a;
+        char[] b;
+        int[7] c;
+
+        public struct SubStruct
+        {
+            int d;
+            char[] e;
+            char[][] f;
+            int[7] g;
+
+            public struct SubSubStruct
+            {
+                int h;
+                char[] i;
+                char[][] j;
+                int[7] k;
+
+                void InitStructure()
+                {
+                    this.h = -52;
+                    this.i.copy("even even more test text");
+                    this.j.length = 3;
+                    this.j[0].copy("abc");
+                    this.j[1].copy("def");
+                    this.j[2].copy("ghi");
+                    foreach ( ref item; this.k )
+                    {
+                        item = 120000;
+                    }
+                }
+            }
+
+            void InitStructure()
+            {
+                this.d = 32;
+                this.e.copy("even more test text");
+
+                this.f.length = 1;
+                this.f[0].copy("abc");
+                foreach ( ref item; this.g )
+                {
+                    item = 32400;
+                }
+            }
+
+            SubSubStruct[] sub_sub_struct;
+            SubSubStruct[1] sub_sub_static_array;
+        }
+
+        SubStruct sub_struct;
+
+        SubStruct[] sub_struct_array;
+        int[][1] dynamic_static_array;
+    }
+
+    TestStruct test_struct;
+    test_struct.a = 7;
+    test_struct.b.copy("some test");
+    foreach ( i, ref item; test_struct.c )
+    {
+        item = 64800;
+    }
+
+    TestStruct.SubStruct sub_struct;
+    sub_struct.InitStructure;
+    test_struct.sub_struct = sub_struct;
+    test_struct.sub_struct_array ~= sub_struct;
+    test_struct.sub_struct_array ~= sub_struct;
+
+
+    TestStruct.SubStruct.SubSubStruct sub_sub_struct;
+    sub_sub_struct.InitStructure;
+    test_struct.sub_struct_array[0].sub_sub_struct ~= sub_sub_struct;
+    test_struct.sub_struct_array[1].sub_sub_struct ~= sub_sub_struct;
+    test_struct.sub_struct_array[1].sub_sub_struct ~= sub_sub_struct;
+    test_struct.sub_struct.sub_sub_struct ~= sub_sub_struct;
+    test_struct.sub_struct.sub_sub_struct ~= sub_sub_struct;
+    test_struct.sub_struct.sub_sub_static_array[0] = sub_sub_struct;
+
+    test_struct.dynamic_static_array[0] = [ 1, 2, 3 ];
+
+    reset(test_struct);
+
+    test!("==")(test_struct.a, 0);
+    test!("==")(test_struct.b, ""[]);
+    foreach ( item; test_struct.c )
+    {
+        test!("==")(item, 0);
+    }
+
+    test!("==")(test_struct.sub_struct_array.length, 0);
+
+    test!("==")(test_struct.sub_struct.d, 0);
+    test!("==")(test_struct.sub_struct.e, ""[]);
+    test!("==")(test_struct.sub_struct.f.length, 0);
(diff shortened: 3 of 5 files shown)


src/ocean/meta/values/VisitValue_test.d
```diff
@@ -0,0 +1,56 @@
...
+import ocean.core.Test;
+
+import ocean.io.Stdout;
+
+struct CountIntegers
+{
+    long sum;
+
+    /// Returns: 'true' if recursion shall continue for this type
+    bool visit ( T ) ( T* value )
```
mihails-strasuns-sociomantic, line 26: FYI: values are passed via pointer and not ref because of D1 static arrays

src/ocean/meta/values/VisitValue.d
```diff
@@ -0,0 +1,102 @@
...
+        if (!deeper)
+            return;
+
+        static if (isPrimitiveType!(T))
+        {
+            // do nothing, already processed
+        }
+        else static if (isTypedef!(T))
+        {
+            auto reinterp = cast(TypedefBaseType!(T)*) &value;
```
gavin-norman-sociomantic: Should the `&` be there? (Ditto for `enum` handling.)

src/ocean/meta/values/VisitValue.d
```diff
@@ -0,0 +1,102 @@
...
+            this.visitAll(value);
```
gavin-norman-sociomantic: Visit `reinterp`? (Ditto for `enum` handling.)

gavin-norman-sociomantic (commented):
Looks good!

The only thing that comes to mind is: could the usage be simplified for cases where the visitor _doesn't_ need to store any internal state?

mihails-strasuns-sociomantic:
> The only thing that comes to mind is: could the usage be simplified for cases where the visitor doesn't need to store any internal state?

How much simpler do you have in mind? :) Without state you would use pretty much the same `struct NoState { bool visit ( T ) ( T* value ) { } }`. Of course one could also support free function for that purpose but it doesn't seem to be of much gain.

gavin-norman-sociomantic:
Yeah, I was thinking it'd be nice to support free functions or delegates.

mihails-strasuns-sociomantic:
Not WIP anymore, includes reset reimplementation and tests.

mihails-strasuns-sociomantic:
@gavin-norman-sociomantic I'd really rather not support free functions right now. It requires adding quite some `static if` blocks to implementation and only benefit is saving 3-5 lines of code in other library utilities.

gavin-norman-sociomantic:
Fair enough. It's not a big deal.

mihails-strasuns-sociomantic:
Updated:
- fix 2.071 build

src/ocean/meta/values/VisitValue.d
```diff
@@ -0,0 +1,140 @@
...
+
+import ocean.meta.traits.Basic;
+import ocean.meta.types.Arrays;
+import ocean.meta.types.Typedef;
+
+/*******************************************************************************
+
+    Recursively iterates over `value` fields and calls `visitor.visit` for each
+    one of those.
+
```
gavin-norman-sociomantic: For clarity, I'd add:
```
`visitor.visit` is expected to be a template with the following signature:
    bool visit ( T ) ( T* value )
The return value indicates whether the value should be recursed into or not.
```

src/ocean/meta/values/VisitValue_test.d
```diff
@@ -0,0 +1,72 @@
...
+*******************************************************************************/
+
+module ocean.meta.values.VisitValue_test;
+
+import ocean.meta.values.VisitValue;
+import ocean.meta.traits.Basic;
+import ocean.meta.types.Typedef;
+import ocean.core.Test;
+
+struct CountIntegers
```
gavin-norman-sociomantic: `SumIntegers` would be clearer.

src/ocean/meta/values/VisitValue.d
```diff
@@ -0,0 +1,140 @@
...
+    Example:
+
+    ---
+    struct CountIntegersVisitor
+    {
+        /// Visitor can have any internal state it may need
+        long sum;
```
gavin-norman-sociomantic: `count` would be clearer.

relnotes/metavalues.feature.md
```diff
@@ -4,3 +4,22 @@ Hosts utilities that use heavy compile-time reflection to do some automatic
...
+
+Provides single function, `reset`, which is intended to replace
+`ocean.util.DeepReset` while being both more performant and using simpler
+implementation. It has two main differences from `DeepReset`:
+
+1) Simplified API - any value can be reset by simply calling `reset` function
+   on it, no matter if it is a single integer, array or deeply nested struct
+   instance.
+
+2) No recursion through reset values. Original `DeepReset` was going through
```
gavin-norman-sociomantic: "No recursion through reset dynamic arrays"?

src/ocean/meta/values/Reset_test.d
```diff
@@ -0,0 +1,183 @@
...
+    reset(test_class);
+    test!("==")(test_class.a, 0);
+    test!("==")(test_class.b.length, 0);
+    test!("==")(test_class.c[1], 0);
+    test!("!is")(cast(void*)test_class.s, null);
+    test!("==")(test_class.s.d, 0);
+    test!("==")(test_class.s.e.length, 0);
+}
+
+unittest
```
gavin-norman-sociomantic: What does this test check that the other one doesn't? A comment would be helpful.

src/ocean/meta/values/Reset.d
```diff
@@ -0,0 +1,92 @@
...
+import ocean.meta.values.VisitValue;
+import ocean.meta.traits.Basic;
+
+import ocean.core.Test;
+
+/*******************************************************************************
+
+    Utility that resets value state in a way that its memory can be reused again
+    as much as possible. Resets arrays to 0 length and values to their init
+    state.
```
gavin-norman-sociomantic, line 31: Maybe also worth mentioning that non-class references are skipped?

src/ocean/meta/values/VisitValue.d
mihails-strasuns-sociomantic: That contradicts with a previous statement ;)

src/ocean/meta/values/Reset_test.d
mihails-strasuns-sociomantic: I removed it - it was simply copied from a different place but it doesn't check anything extra.

mihails-strasuns-sociomantic:
Rebased, updated

mihails-strasuns-sociomantic:
Green @gavin-norman-sociomantic

src/ocean/meta/values/VisitValue.d
gavin-norman-sociomantic: No, these are two different situations and (IMO) you have the naming backwards.

src/ocean/meta/values/Reset.d
```diff
@@ -0,0 +1,95 @@
...
+
+    Non-class references are ignored. For classes it tried to recurse
```
gavin-norman-sociomantic: "tries"

src/ocean/meta/values/VisitValue.d
gavin-norman-sociomantic: `CountIntegersVisitor` (here) -> `count`. `SumIntegers` (below) -> `sum`.

Though actually, looking at the logic of this example, it's summing, not counting, so it's the struct name that's wrong, not the field.

gavin-norman-sociomantic: By the way, any reason for this to not be a documented unittest? (Are they not rendered for templates?)

mihails-strasuns-sociomantic: > By the way, any reason for this to not be a documented unittest? (Are they not rendered for templates?)

You can't define templates in D1 inside function scopes. Which means most important part of the example won't actually be part of it.

gavin-norman-sociomantic: Ah yes :+1:

mihails-strasuns-sociomantic:
Updated:
- fixed naming in example

src/ocean/meta/values/VisitValue.d
```diff
@@ -0,0 +1,155 @@
...
+/*******************************************************************************
+
+    Implementation of visiting logic. Some types are currently not supported
+    because it is not clear what would recursing mean for them.
+
+*******************************************************************************/
+
+private struct VisitImpl ( Visitor )
+{
+    // Refer to the persistent instance of visitor struct in case implementation
```
gavin-norman-sociomantic: Non-standard internal comment format. (Also for `visitAll`.)

src/ocean/meta/values/VisitValue.d
```diff
@@ -0,0 +1,155 @@
...
+        else static if (isAggregateType!(T))
+        {
+            static if (is(T == class))
+                bool ignore = *value is null;
+            else
+                bool ignore = false;
+
+            if (!ignore)
+            {
+                foreach (ref field; (*value).tupleof)
```
gavin-norman-sociomantic: What about super class fields? e.g. https://github.com/sociomantic-tsunami/ocean/blob/v3.x.x/src/ocean/util/DeepReset.d#L271-L281

src/ocean/meta/values/VisitValue.d
mihails-strasuns-sociomantic: What do you mean? This is not a ddoc comment.

gavin-norman-sociomantic: Right. It should be.

src/ocean/meta/values/VisitValue.d
mihails-strasuns-sociomantic: Good catch! I should add tests for that too

src/ocean/meta/values/VisitValue.d
mihails-strasuns-sociomantic: Why? We don't have strict requirement to document private implementation details in a same way, do we?

gavin-norman-sociomantic: I've never been aware of any explicit exception.

src/ocean/meta/values/VisitValue.d
```diff
@@ -0,0 +1,155 @@
...
+
+    Generic utility allowing to recursively visit every field in an aggregate
+    and possibly those indirectly reachable from them.
+
+    Intended as a common implementation base for various more domain-specific
+    utilities.
+
+    Copyright:
+        Copyright (c) 2018 Sociomantic Labs GmbH.
+        All rights reserved.
```
nemanja-boric-sociomantic: `sociomantic labs` lowercase (I know, I know...)

mihails-strasuns-sociomantic:
Updated:
- fix derived class visiting
- ddoc for visit impl

gavin-norman-sociomantic (approved):

src/ocean/meta/values/VisitValue.d
```diff
@@ -0,0 +1,183 @@
...
+            {
+                bool ignore = *value is null;
+
+                // Recurse into super class fields
+                static if (is(T Bases == super))
+                {
+                    foreach (Base; Bases)
+                    {
+                        static if (Base.init.tupleof.length)
+                            this.visitAll(cast(Base*) value);
```
david-eckardt-sociomantic: This cast looks a bit dodgy to me, and an explicit cast can be avoided:
```d
Base base_value = *value; // implicit upcast
this.visitAll(&base_value);
```

src/ocean/meta/values/VisitValue.d
```diff
@@ -0,0 +1,183 @@
...
+                    this.visitAll(&elem);
+            }
+            else
+            {
+                static assert (false, "AA currently not supported");
+            }
+        }
+        else static if (isFunctionType!(T))
+        {
+            // ignored for now
```
david-eckardt-sociomantic: "for now" -- what would be the alternative to ignoring function types (other than rejecting them)?

src/ocean/meta/values/Reset.d
```diff
@@ -0,0 +1,95 @@
...
+        }
+        else static if (isPrimitiveType!(T))
+        {
+            *value = T.init;
+        }
+
+        static if (isReferenceType!(T) && !is(T == class))
+            return false;
+        else
+            return true;
```
david-eckardt-sociomantic: `return (isReferenceType!(T) && !is(T == class));`?

src/ocean/meta/values/Reset_test.d
```diff
@@ -0,0 +1,142 @@
...
+
+                this.f.length = 1;
+                this.f[0].copy("abc");
+                foreach ( ref item; this.g )
+                {
+                    item = 32400;
+                }
+            }
+
+            SubSubStruct[] sub_sub_struct;
```
david-eckardt-sociomantic, line 57: Suggest testing a static array of dynamic arrays and structs.

src/ocean/meta/values/Reset_test.d
```diff
@@ -0,0 +1,142 @@
...
+        char[] b;
+        int[2] c;
+
+        public class SubClass
+        {
+            int d;
+            char[] e;
+        }
+
+        SubClass s;
```
david-eckardt-sociomantic: Suggest testing a static array of a class.

src/ocean/meta/values/VisitValue.d
```diff
@@ -0,0 +1,183 @@
...
+            return;
+
+        static if (isPrimitiveType!(T))
+        {
+            // do nothing, already processed
+        }
+        else static if (isTypedef!(T))
+        {
+            auto reinterp = cast(TypedefBaseType!(T)*) value;
+            this.visitAll(reinterp);
```
david-eckardt-sociomantic, line 115: Why not `this.visitAll(cast(TypedefBaseType!(T)*) value);` like for classes below? (Same for enums below.)

src/ocean/meta/values/VisitValue.d
mihails-strasuns-sociomantic: This is exactly why "for now" - I can't imagine any iseful processing semantics, but maybe someone will request one :)

src/ocean/meta/values/Reset_test.d
nemanja-boric-sociomantic: these two comments from David are overlooked it seems.

src/ocean/meta/values/Reset.d
```diff
@@ -0,0 +1,95 @@
+/*******************************************************************************
+
+    Utility that resets value state in a way that its memory can be reused again
+    as much as possible. Resets arrays to 0 length and values to their init
+    state.
+
+    Copyright:
+        Copyright (c) 2018 Sociomantic Labs GmbH.
```
nemanja-boric-sociomantic: ditto `sociomantic labs`

src/ocean/meta/values/VisitValue.d
```diff
@@ -0,0 +1,183 @@
...
+                    foreach (Base; Bases)
+                    {
+                        static if (Base.init.tupleof.length)
+                            this.visitAll(cast(Base*) value);
+                    }
+                }
+            }
+            else
+            {
+                bool ignore = false;
```
nemanja-boric-sociomantic: These `static if` fake scopes are very deceptive /rant

nemanja-boric-sociomantic: I suggest moving `bool ignore` above `static if` and just setting it to true in the `static if (is(T == class))`, so you avoid fake scope and double definition of the `ignore` and you get rid of `static else` branch.

src/ocean/meta/values/VisitValue.d
mihails-strasuns-sociomantic: Because I wanted to stress the fact this is reinterpret cast and not a "smart" D cast.

mihails-strasuns-sociomantic: (I switched class version to your proposal so it is more consistent now)

mihails-strasuns-sociomantic:
Updated per @david-eckardt-sociomantic and @nemanja-boric-sociomantic comments

david-eckardt-sociomantic (approved):

PR #460, opened 2018-02-21
DaemonApp: Align the stats timer call with the interval

mathias-lang-sociomantic:
> This will make stats easier to compare between two apps / multiple instances, provided the clocks are in sync.

CC @scott-gibson-sociomantic who had the idea following a discussion with @andrej-mitrovic-sociomantic

### DIFF
--- a/relnotes/aligned-stats.feature.md
+++ b/relnotes/aligned-stats.feature.md
@@ -1,0 +1,9 @@
+### Stats timer is now triggered at a predictable moment
+
+`ocean.util.app.DaemonApp`
+
+In order to help with multiple instance apps, or similar apps needing to compare their stats,
+the initial call to `onStatsTimer` will not be dependent on when the application was started anymore,
+but will be aligned on the interval provided.
+For example, if an application starts at 12:00:04 with an interval of 30 (the default),
+the first call to `onStatsTimer` will happen at 12:00:30 instead of 12:00:34 (currently).
--- a/src/ocean/util/app/DaemonApp.d
+++ b/src/ocean/util/app/DaemonApp.d
@@ -45,8 +45,11 @@ import ocean.util.app.ext.model.ILogExtExtension;
 import ocean.util.app.ext.model.ISignalExtExtension;
 
 import ocean.transition;
+version (UnitTest) import ocean.core.Test;
 import ocean.core.Verify;
 import ocean.task.IScheduler;
+
+import core.stdc.time;
 
 /// ditto
 public abstract class DaemonApp : Application,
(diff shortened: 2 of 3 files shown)


relnotes/aligned-stats.feature.md
```diff
@@ -0,0 +1,9 @@
+### Stats timer is now triggered at a predictable moment
+
+`ocean.util.app.DaemonApp`
+
+In order to help with multiple instance apps, or similar apps needing to compare their stats,
+the initial call to `onStatsTimer` will not be dependent on when the application was started anymore,
+but will be aligned on the interval provided.
+For example, if an application starts at 12:00:04 with an interval of 30 (the default),
+the first call to `onStatsTimer` will happen at 12:00:30 instead of 12:00:34 (currently).
```
nemanja-boric-sociomantic, line 9: That's a good idea :+1:

src/ocean/util/app/DaemonApp.d
```diff
@@ -460,7 +463,8 @@ public abstract class DaemonApp : Application,
         this.registerExtension(this.timer_ext);
 
         // Register stats timer with epoll
-        this.timer_ext.register(&this.statsTimer, this.stats_ext.config.interval);
+        ulong offset = timeToNextInterval(this.stats_ext.config.interval);
```
nemanja-boric-sociomantic: `this.` :angel:

src/ocean/util/app/DaemonApp.d
```diff
@@ -469,6 +473,29 @@ public abstract class DaemonApp : Application,
         this.unix_socket_ext.initializeSocket(this.epoll);
     }
 
+    /***************************************************************************
+
+        Returns:
```
gavin-norman-sociomantic: Params.

src/ocean/util/app/DaemonApp.d
nemanja-boric-sociomantic: `offset` could be probably named a bit more descriptive, indicating that this is the offset only for the first firing time point.

gavin-norman-sociomantic (commented):
Seems like a good idea.

src/ocean/util/app/DaemonApp.d
mathias-lang-sociomantic: It's `static` :P

nemanja-boric-sociomantic: In the same moment I got this email, that suddenly appeared to me :-) :+1:

mathias-lang-sociomantic:
Updated:

```diff
diff --git a/src/ocean/util/app/DaemonApp.d b/src/ocean/util/app/DaemonApp.d
index 8709ea7d..f56cadf6 100644
--- a/src/ocean/util/app/DaemonApp.d
+++ b/src/ocean/util/app/DaemonApp.d
@@ -463,8 +463,9 @@ public abstract class DaemonApp : Application,
         this.registerExtension(this.timer_ext);
 
         // Register stats timer with epoll
-        ulong offset = timeToNextInterval(this.stats_ext.config.interval);
-        this.timer_ext.register(&this.statsTimer, offset, this.stats_ext.config.interval);
+        ulong initial_offset = timeToNextInterval(this.stats_ext.config.interval);
+        this.timer_ext.register(
+            &this.statsTimer, initial_offset, this.stats_ext.config.interval);
 
         // Register signal event handler with epoll
         this.epoll.register(this.signal_ext.selectClient());
@@ -475,6 +476,10 @@ public abstract class DaemonApp : Application,
 
     /***************************************************************************
 
+        Params:
+            interval = interval used for calling the stats timer
+            current  = current time, default to now (`time(null)`)
+
         Returns:
             function to calculate the amount of time to wait until the next
             interval is reached.
```

mathias-lang-sociomantic:
Ping ?

gavin-norman-sociomantic (approved):

joseph-wakeling-sociomantic:
> This will make stats easier to compare between two apps / multiple instances, provided the clocks are in sync.

I'm not entirely sure this is a valid assumption.  It depends after all on how much time it takes to calculate and output the stats.  Suppose e.g. it takes 10 seconds just to calculate the stats that are to be output ... ?

nemanja-boric-sociomantic:
The problem of the stats taking too long is covered in https://github.com/sociomantic-tsunami/ocean/pull/78 (the one that got merged april last year, I confused this PR with that one ;-) )

joseph-wakeling-sociomantic:
Indeed, but that's not what I meant: the problem is not drift in when statslog entries get written, but "truth" in the alignment of stats values.

e.g. imagine that you have a statslog entry for :00 where the actual stats didn't finish getting generated 'til :10.  The idea that these stats are representative of the state of reality at :00 is ... dubious.

Now, of course, this remains true for statslog entities written at arbitrary 30-second intervals (rather than precise :00 and :30 points).  But in that case there's at least an implicit understanding that stats of different apps are not so precisely comparable.

My concern is that the exact :00 and :30 lineup gives an impression of comparability that _doesn't actually reflect reality_.

scott-gibson-sociomantic:
If i'm reading this correctly, the stats log line is written out when `.flush` is called on the StatsLog instance. This means the timestamp of the stats log line should be the time of the flush, not the time of the stats timer call. This means the time to generate the stats is accounted for.
```
    public void flush ( )
    {
        this.logger.info(this.buffer[]);
        this.add_separator = false;
        this.buffer.clear();
    }
```

scott-gibson-sociomantic:
The point behind this change was if we had 10 app instances and they were restarted over 10 second intervals to ensure no downtime, that eventually all stats timers would align (assuming all server times are synced)

PR #465, opened 2018-02-27
Give a pointer to addCustomHeaders in docs

andrew-stevenson-sociomantic:
As well as indicating it must be done, give a user a hint as to how it
should be done.

The docs could probably be better yet but this one seemed crying out for fixing once I was faced with the question myself.


mihails-strasuns-sociomantic:
This should target v3.x.x instead, shouldn't it?

mihails-strasuns-sociomantic (approved):
Rebased on v3.x.x and force pushed to @andrew-stevenson-sociomantic branch.

PR #473, opened 2018-03-19
Remove meaningless destructor

mihails-strasuns-sociomantic:
No point in setting length of internal field to 0 upon destruction of
host object


gavin-norman-sociomantic (approved):
Indeed, seems pointless.

gavin-norman-sociomantic:
Oh I just noticed this is done as part of a patch release. How come?

mihails-strasuns-sociomantic:
Well, it isn't a feature and it causes invalid memory operation error in nautica when compiled with `-profile=gc` :)

gavin-norman-sociomantic:
I see. (Normally it'd count as a refactoring.)

PR #474, opened 2018-03-20
Throw exception instead of error in ParamSet

### LINKED ISSUE #471: Remove catch of ArrayBoundsCheckException in QueryParams
https://github.com/sociomantic-tsunami/ocean/blob/fb5f7928e46e3d75c2a251f0dc54c901e3392f9d/src/ocean/net/util/ParamSet.d#L152-L163

mihails-strasuns-sociomantic:
Fixes #471

### DIFF
--- a/src/ocean/net/util/ParamSet.d
+++ b/src/ocean/net/util/ParamSet.d
@@ -39,10 +39,9 @@ import ocean.core.TypeConvert;
 import ocean.core.Verify;
 
 import ocean.text.util.SplitIterator: ISplitIterator;
+import ocean.text.convert.Formatter;
 
 import core.stdc.ctype:  tolower;
-
-import ocean.core.ExceptionDefinitions: ArrayBoundsException;
 
 version(UnitTest) import ocean.core.Test;
 
@@ -151,15 +150,14 @@ class ParamSet
 
     cstring opIndex ( cstring key )
     {
-        try
-        {
-            return this.paramset[this.tolower(key)].val;
-        }
-        catch (ArrayBoundsException e)
-        {
-            e.msg ~= " [\"" ~ key ~ "\"]";
-            throw e;
-        }
+        key = this.tolower(key);
+        auto value = key in this.paramset;
+        verify(
+            value !is null,
+            format("Key '{}' is not found in ParamSet", key)
+        );
+
+        return value.val;
     }
 
     /**************************************************************************


gavin-norman-sociomantic (approved):
Makes sense to me. Maybe someone in charge of the HTTP server code could double check.

mihails-strasuns-sociomantic:
That would be @ben-palmer-sociomantic @andrew-stevenson-sociomantic ?

src/ocean/net/util/ParamSet.d
```diff
@@ -151,15 +150,14 @@ class ParamSet
...
-        }
-        catch (ArrayBoundsException e)
-        {
-            e.msg ~= " [\"" ~ key ~ "\"]";
-            throw e;
-        }
+        key = this.tolower(key);
+        auto value = key in this.paramset;
+        verify(
+            value,
```
david-eckardt-sociomantic: `value !is null`

david-eckardt-sociomantic:
@stefan-brus-sociomantic, too.

ben-palmer-sociomantic:
So this will preserve the D1 behaviour right? I haven't started testing with D2 builds much yet.

stefan-brus-sociomantic:
Same here. So `ArrayBoundsException` is an error, and not actually an exception?

also FYI @scott-gibson-sociomantic

mihails-strasuns-sociomantic:
@ben-palmer-sociomantic it will preserve it in a sense that exception will be thrown in both debug and release builds and it can be caught. But exception type/name will be different.

mihails-strasuns-sociomantic:
@stefan-brus-sociomantic yes https://github.com/sociomantic-tsunami/ocean/blob/v3.x.x/src/ocean/core/ExceptionDefinitions.d#L63-L68

I intend to deprecate them all in next ocean minor, as using those is not Ok for migration stages 3/4.

david-eckardt-sociomantic:
> So `ArrayBoundsException` is an error, and not actually an exception?

Yes, it is actually [`RangeError`](https://dlang.org/phobos/core_exception.html#.RangeError).

mihails-strasuns-sociomantic:
Good to merge?

mihails-strasuns-sociomantic:
Ping

src/ocean/net/util/ParamSet.d
```diff
@@ -151,15 +150,14 @@ class ParamSet
...
-        catch (ArrayBoundsException e)
-        {
-            e.msg ~= " [\"" ~ key ~ "\"]";
-            throw e;
-        }
+        key = this.tolower(key);
+        auto value = key in this.paramset;
+        verify(
+            value !is null,
+            format("Key '{}' is not found in ParamSet", key)
```
nemanja-boric-sociomantic, line 157: Unrelated to this PR, and it's probably not too important, but this allocates, while the `verify` uses reusable instance. Perhaps this could be misused, if the remote party attack us with the requests to hit this path. Is there any way that we format-enable the `Verify` so it can reuse the message buffer?

The answer is probably "no", since we tend not to import anything in `Verify` module.

nemanja-boric-sociomantic (approved):

vasileios-kyrillidis-sociomantic:
Andrew is on vacation and back on Monday. I don't see this used by us to be honest (maybe I didn't check thoroughly enough) so I believe it should be fine.

src/ocean/net/util/ParamSet.d
mihails-strasuns-sociomantic: So this is not under control of the app? If yes, I should replace `verify` here with custom exception type. `verify` is expected to never be hit in production, as it basically replaces `assert`.

nemanja-boric-sociomantic: That is a good point. Two applications that I've look into (search for the `QueryParams` and `QueryParamSet`) are fine with this being an assertion, not enforce, since they fill the `ParamSet` with the empty value for every key, then parse the query filling as much values as they can and then they lookup the keys that are guaranteed to be in the set.

david-eckardt-sociomantic: > So this is not under control of the app?

It is under control of the application. The application knows which keys are in the set. 
The method documentation says, "key must be one of the parameter keys passed on instantiation or added by a subclass." Like AA indexing the application should call this method only if it can for sure pass one of these as a parameter, otherwise it should use `opIn_r`.

david-eckardt-sociomantic (approved):

PR #476, opened 2018-03-23
Remove unused (and wrong) extern (C) declarations

mathias-lang-sociomantic:


mathias-lang-sociomantic:
Updated

mihails-strasuns-sociomantic (approved):
Technically a bug but safer to merge it in minor release, consequences are not very obvious.

PR #479, opened 2018-03-28
Deprecate ocean.transition : getMsg

mathias-lang-sociomantic:
```
As mentioned in the release notes, it always resolves to `Exception.message()` nowadays.
```

@mihails-strasuns-sociomantic mentioned that this is now possible.
However, I recommend not to merge before libraries are adapted, otherwise there will be a cascade of useless deprecation messages.

### DIFF
--- a/relnotes/getMsg.deprecation.md
+++ b/relnotes/getMsg.deprecation.md
@@ -1,0 +1,6 @@
+### `getMsg` is deprecated in favor of `Exception.message()`
+
+`ocean.transition`
+
+The `getMsg()` utility was introduced to abstract away the differences between runtimes.
+Nowadays (since `v2.077.0`), it always resolves to `Exception.message()` and can be removed.
--- a/src/ocean/transition.d
+++ b/src/ocean/transition.d
@@ -554,6 +554,7 @@ unittest
 
 *******************************************************************************/
 
+deprecated("Use Exception.message() directly")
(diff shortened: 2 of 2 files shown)


mathias-lang-sociomantic:
XREF https://github.com/dlang/druntime/commit/e7e49cd206dec5324f9f4baf56099e211d8f9c9e

mihails-strasuns-sociomantic:
@mathias-lang-sociomantic can you split switching to `message` into a separate PR so that this one remains only a deprecation? We can merge the former right away.

mathias-lang-sociomantic:
https://github.com/sociomantic-tsunami/ocean/pull/480

mathias-lang-sociomantic:
Rebased

mihails-strasuns-sociomantic:
@sociomantic-tsunami/core-team any objections against merging this in 3.9.0?

nemanja-boric-sociomantic:
I think that should be fine, as long as we can use `.message` in downstream libraries that support ocean v3.x.x (to avoid needless deprecation spam). I'll check that.

nemanja-boric-sociomantic:
I think this should be fine. There are few places in swarm/protos where we use `getMsg`, but that can be changed in the patch release, since the oldest compiler we support is dmd-transitional v2.070, which comes with `Throwable.message`.

mihails-strasuns-sociomantic (approved):

PR #481, opened 2018-04-04
Deprecate error aliases

### LINKED ISSUE #475: Deprecate aliases to `*Error`
See https://github.com/sociomantic-tsunami/ocean/issues/471 and https://github.com/sociomantic-tsunami/ocean/pull/474

Original aliases https://github.com/sociomantic-tsunami/ocean/blob/v3.x.x/src/ocean/core/ExceptionDefinitions.d#L63-L68 helped to do stage 2 migration faster but now that we move towards stage 3/4 actively, all usage of these has to be eliminated.

mihails-strasuns-sociomantic:
Fixes #475

### DIFF
--- a/relnotes/error.deprecation.md
+++ b/relnotes/error.deprecation.md
@@ -1,0 +1,16 @@
+### Deprecate aliases in `ExceptionDefinions` that forward to errors
+
+Symbols like `AssertException` or `OutOfMemoryException` are in fact `Error`
+derivatives in D2. During original ocean porting compatibility aliases for
+those were added to `ocean.core.ExceptionDefinions` module to simplify stage 2
+migration. However now that we put most effort into stages 3 and 4, this
+becomes a liability as catching those errors is never correct when D2 build is
+deployed.
+
+For example, throwing/catching `ArrayBoundsExceptions` should be replaced by
+explicitly checking if element is present in array before access and throwing
+custom exception type instead. Similar approach can be applied to most
+deprecated symbols.
+
+`AssertError` is special here because after conversion to `verify` it should not
+ever be thrown by libraries and thus can be simply removed.
--- a/src/ocean/core/UnitTestRunner.d
+++ b/src/ocean/core/UnitTestRunner.d
@@ -77,7 +77,6 @@ else
     import core.sys.posix.sys.time: timersub;
 }
 
-import ocean.core.ExceptionDefinitions : AssertException;
 import ocean.io.Stdout: Stdout, Stderr;
 import ocean.io.stream.Format: FormatOutput;
 import ocean.io.stream.TextFile: TextFileOutput;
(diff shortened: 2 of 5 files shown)


relnotes/error.deprecation.md
```diff
@@ -0,0 +1,8 @@
+### Deprecate aliases in `ExceptionDefinions` that forward to errors
+
+Symbols like `AssertException` or `OutOfMemoryException` are in fact `Error`
+derivatives in D2. During original ocean porting compatibility aliases for
+those were added to `ocean.core.ExceptionDefinions` module to simplify stage 2
+migration. However now that we put most effort into stages 3 and 4, this
+becomes a liability as catching those errors is never correct when D2 build is
+deployed.
```
mathias-lang-sociomantic, line 8: A pointer in how to deal with it would be nice.

mihails-strasuns-sociomantic: Pretty tough call to give generic guidelines for this. But I will give it a try :)

leandro-lucarella-sociomantic: It might be pretty extreme, but what about instead of deprecating them, making them behave more like D2, like putting a call to `abort()` in the constructor?

src/ocean/core/UnitTestRunner.d
```diff
@@ -651,13 +650,6 @@ private scope class UnitTestRunner
                 err = sformat(err, "{}:{}: test failure: {}", e.file, e.line, e.message());
             return Result.Fail;
         }
-        catch (AssertException e)
-        {
-            version (D_Version2)
-                e.toString((d) { err ~= d; });
-            else
-                err = sformat(err, "{}:{}: assert error: {}", e.file, e.line, e.message());
-        }
```
leandro-lucarella-sociomantic, line 660: This sucks... (just ranting, it is what it is)

relnotes/error.deprecation.md
mihails-strasuns-sociomantic: @leandro-lucarella-sociomantic but this is exactly why we are introducing things like `verify`- because `abort` behavior is not acceptable for some of our apps :)

mihails-strasuns-sociomantic:
Updated with more explanation

relnotes/error.deprecation.md
leandro-lucarella-sociomantic: Yes, I know. All I'm saying is if we keep them with an `abort()`, then we are bringing D1 closer to D2. People will still need to move to `verify()` because errors are broken in D2.

leandro-lucarella-sociomantic (approved):
But besides the suggestion, LGTM. <del>Travis is not happy though</del> (it's just codecove, I assume is misbehaving?).

mihails-strasuns-sociomantic:
Codecov is not misbehaving - only executable code lines in this PR are those that call `onOutOfMemoryError` and we don't test that :) So it is honest 0%

relnotes/error.deprecation.md
mihails-strasuns-sociomantic: > then we are bringing D1 closer to D2

But if no one wants that D2 behaviour, what would be the point? :D

leandro-lucarella-sociomantic: Less migration efforts at first (of course eventually you have to address it). I'm fine with merging as is BTW, it was just a comment.

PR #484, opened 2018-04-06
Get back the old EpollSelectDispatcher.eventLoop which accepts old de…

nemanja-boric-sociomantic:
…legate

The eventLoop delegate was changed by accident to return bool,
introducing a breaking change.


nemanja-boric-sociomantic:
Sadly, this doesn't fix the bug we've noticed elsewhere (the user will override this method, but they will call `eventLoop()` with no arguments and the `EpollSelectDispatcher` accepting `bool delegate(Exception) = null` will be chosen as a better match), but it's still something.

nemanja-boric-sociomantic:
I'll close this, reopen if you think it's worth the trouble.

PR #490, opened 2018-04-12
Register the version.log file with the ReopenableExt

### LINKED ISSUE #437: Rotating version.log files
Quoting from https://github.com/sociomantic-tsunami/ocean/pull/434#issuecomment-357923520

>As for version.log it is configured here https://github.com/sociomantic-tsunami/ocean/blob/v3.x.x/src/ocean/util/app/ext/VersionArgsExt.d#L297-L309 and seems like it is working as intended - this file is only written to during app startup, thus there is nothing to rotate.

Normally, the logrotate configurations of apps are wildcard-based (i.e. `/path/to/app/directory/log/*.log`). This means that from logrotate's point-of-view, `version.log` is also a candidate for rotation.

`version.log` is indeed written to only during app startup, so under normal circumstances, it would fill up rather slowly. But if it ever gets rotated, then the app would never reopen this file - meaning that the app will keep writing the version to the now rotated `version.log.1` file.

This is why I feel that at least for the sake of completeness, `version.log` should also be registered with `ReopenableFilesExt`.

nemanja-boric-sociomantic:
The `version.log` is now registered with the `ReopenableFilesExt` to avoid
extra logic in the logrotate when rotating the `version.log` file.

Fixes #437

### DIFF
--- a/relnotes/versionrotate.feature.md
+++ b/relnotes/versionrotate.feature.md
@@ -1,0 +1,6 @@
+### Register version.log with the ReopenableFilesExt
+
+* `ocean.util.app.ext.VersionArgsExt`
+
+  The `version.log` is now registered with the `ReopenableFilesExt` to avoid
+  extra logic in the logrotate when rotating the `version.log` file.
--- a/src/ocean/util/app/ext/VersionArgsExt.d
+++ b/src/ocean/util/app/ext/VersionArgsExt.d
@@ -20,11 +20,13 @@ module ocean.util.app.ext.VersionArgsExt;
 
 public import ocean.util.app.ext.VersionInfo;
 
+import ocean.io.device.File;
 import ocean.util.app.model.IApplicationExtension;
 import ocean.util.app.ext.model.IArgumentsExtExtension;
 import ocean.util.app.ext.model.ILogExtExtension;
 import ocean.util.app.ext.LogExt;
 import ocean.util.app.ext.ConfigExt;
+import ocean.util.app.ext.ReopenableFilesExt;
 import ocean.util.app.ext.UnixSocketExt;
 import ocean.util.app.Application;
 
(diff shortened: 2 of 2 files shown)


mihails-strasuns-sociomantic:
What was the extra logic? I remember we discussed this and decided there was no reason to reopen version.log as it is written once during startup only.

nemanja-boric-sociomantic:
With the GDPR, all applications rotate all the logfiles now - weekly - even if they do not grow. Now, the `logrotate` configuration file would have something like this:

```
/srv/app/log/*.log {

postsript
    echo "reload_files $1" | nc -U /srv/app/app.socket | grep -q "ACK"
endscript
```

So this will trigger rotating version.log at least once a week and the command to the
socket will be issued, and the application will respond with an error code for the version.log.

If we want to avoid this, we either need to avoid `version.log` file inside the glob (if possible) or to test the filename of the logfile to rotate in the postrotate script.

mihails-strasuns-sociomantic:
> and the application will respond with an error code for the version.log

Aha, I have missed this part. Is it really necessary though? I mean, wouldn't it make more sense to just respond with some non-error status for any log file that isn't registered for reopening, exactly for cases like this?

nemanja-boric-sociomantic:
That, though, compromises the error handling - responding with the error code will trigger the emails that the maintainers will receive, and it usually indicates an error because there's something in the log directory that the application doesn't own. I'm not sure, to be honest, I can't say for sure without real world experience on this approach to show what's the bigger hustle.

mihails-strasuns-sociomantic (approved):
I see, that indeed makes this PR most simple solution for now.

gavin-norman-sociomantic (approved):

gavin-norman-sociomantic:
>0% of diff hit (target 80%) 

I guess this indicates that we don't have any tests for the app framework.

nemanja-boric-sociomantic:
TBH, the codecov is so screwed that I stopped caring. I can't even open the report right now (and it has been like that for ages (cc @leandro-lucarella-sociomantic ))

gavin-norman-sociomantic:
Oh well. LGTM, anyway.

PR #495, opened 2018-04-19
Compile ocean with -debug=ISelectClient in CI

nemanja-boric-sociomantic:
Approximately every time someone needs to use debug=ISelectClient to
debug their application, they find it broken. This patch compiles Ocean
in Travis with this debug flag set.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ env:
         - DMD=1.081.* F=production
         - DMD=1.081.* F=devel
         - DMD=1.081.* F=production DFLAGS=-release
+        - DMD=1.081.* F=devel DFLAGS=-debug=ISelectClient
         - DMD=2.070.2.s* F=production
         - DMD=2.070.2.s* F=devel
         - DMD=2.071.2.s* F=production


nemanja-boric-sociomantic:
I was a bit suspicious of this approach (used for the `-release` flag here: https://github.com/sociomantic-tsunami/ocean/blob/v3.x.x/.travis.yml#L25) so I've tested it with a broken commit and it doesn't work. (Meaning that our test for `-release` doesn't work as well, I guess).

cc @mihails-strasuns-sociomantic @leandro-lucarella-sociomantic

nemanja-boric-sociomantic:
OK, last commit works, beaver PR incoming.

nemanja-boric-sociomantic:
https://github.com/sociomantic-tsunami/beaver/pull/51

mihails-strasuns-sociomantic:
Why just `ISelectClient`? If we go that way, more efficient to have one matrix element that enables all common debug flags at once.

nemanja-boric-sociomantic:
> Why just ISelectClient? If we go that way, more efficient to have one matrix element that enables all common debug flags at once.

Yes, we can enable all debug flags at once, that's a good idea.

Updated beaver to v0.5.0.

leandro-lucarella-sociomantic:
Travis says NO.

nemanja-boric-sociomantic:
That is good, because I've added a commit that breaks `ISelectClient` so I know it actually doesn't compile. I'll rebase it now.

nemanja-boric-sociomantic:
Updated.

mihails-strasuns-sociomantic (approved):

PR #497, opened 2018-04-26
D2: fix compilation error when Map key is a class

mihails-strasuns-sociomantic:
Hacky workaround for regression introduced via https://github.com/sociomantic-tsunami/ocean/commit/1787cf2f98a4896193ca85fa39a496e194c88d44

Formally it may introduce Undefined Behaviour if underlying class
instance was immutable but we know to never do that :)

### DIFF
--- a/src/ocean/util/container/map/model/StandardHash.d
+++ b/src/ocean/util/container/map/model/StandardHash.d
@@ -80,7 +80,12 @@ struct StandardHash
                          ~ "classes/interfaces/structs/unions implementing "
                          ~ "toHash() supported, not '" ~ K.stringof ~ '\'');
 
-            return key.toHash();
+            // HACK: we don't want to start implementing all our class `toHash`
+            // methods as `const` yet, have to cast to silence druntime
+            static if (is(K == class) || is(K == interface))
+                return (cast(Unqual!(K)) key).toHash();
+            else
+                return key.toHash();
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/ocean/util/container/map/model/StandardHash.d
```diff
@@ -80,7 +80,12 @@ struct StandardHash
                          ~ "classes/interfaces/structs/unions implementing "
                          ~ "toHash() supported, not '" ~ K.stringof ~ '\'');
 
-            return key.toHash();
+            // HACK: we don't want to start implementing all our class `toHash`
+            // methods as `const` yet, have to cast to silence druntime
+            static if (is(K == class) || is(K == interface))
+                return (cast(Unqual!(K)) key).toHash();
```
stefan-koch-sociomantic, line 86: this is D2 only right ?
please use `cast()` instead of `cast(Unqual!(K))`

nemanja-boric-sociomantic: No, this code compiles both in D1 and (after conversion) in D2. I don't think `cast()` works in D1.

stefan-koch-sociomantic: true it does not work in D1.
Maybe in a version(D_Version2) ?
It does cut down on unwanted template-instances.

mihails-strasuns-sociomantic: No.

gavin-norman-sociomantic (approved):

Issue #500, opened 2018-05-03
Provide method that runs before the Scheduler's event loop in TaskExt 


nemanja-boric-sociomantic:
At the moment, `DaemonApp` will start event handling and then it will schedule the `run` task
and start the scheduler:

https://github.com/sociomantic-tsunami/ocean/blob/b200e555190591538c44ca789c5dc26fc2540928/src/ocean/util/app/DaemonApp.d#L542-L545

https://github.com/sociomantic-tsunami/ocean/blob/b200e555190591538c44ca789c5dc26fc2540928/src/ocean/util/app/ext/TaskExt.d#L129

The problem with this is that, given that `processConfig` runs after the constructor, the `run` method is the only appropriate place to initialize the application's components. That leads to trouble if the `onStatsTimer` or any other event (like `SignalExt` or `UnixSocketExt`) signals and run its callback assuming the application is fully initialized while the `run` hasn't yet run or hasn't finished and had a context switch (the issue is easy to see if you run your application close to 0 or 30 seconds mark: https://github.com/sociomantic-tsunami/ocean/blob/b200e555190591538c44ca789c5dc26fc2540928/src/ocean/util/app/DaemonApp.d#L509-L512 , but any other registered event source may trigger this bug).

Requiring explicitly from the user to call `startEventHandling` is error prone, since it can be easily forgotten. The alternative could be to allow extra method to be overriden to put the initialization code, or to override `run` and call `DaemonApp`'s run after the initialization has completed.

joseph-wakeling-sociomantic:
What I would suggest/request is that the method concerned should be possible to use with all resources that one would have prior to `this.startEventHandling` and `theScheduler.eventLoop` being called.  So e.g. one should be able to use `theScheduler.epoll` (to instantiate swarm clients), etc. etc.

joseph-wakeling-sociomantic:
HOWEVER, I wonder if it might be worth just rolling back the change to introduce :00 and :30 stats timing.  The reason is simply that the "guarantee" it provides is not very strong: writing and/or calculating stats can take time, and it is entirely possible that this will mess up the "promise" of the fixed stats times.

nemanja-boric-sociomantic:
That's not a robust solution because the stats timer is not the only event that you start in `startEventHandling`: also the signal extension and the unix domain socket extension are started and getting any signal or connecting any client (it could easily be the monitor application) will invoke the callbacks leading to the same problem.

[...]

PR #501, opened 2018-05-04
Increase Application own order among extensions

mihails-strasuns-sociomantic:
To address https://github.com/sociomantic-tsunami/ocean/issues/500

### DIFF
--- a/relnotes/order.feature.md
+++ b/relnotes/order.feature.md
@@ -1,0 +1,7 @@
+### Changed order of app extensions
+
+Previously several extensions used in DaemonApp had same order as application
+itself meaning it was not possible to rely on `processConfig` method being
+run after `processConfig` of those extensions. That proved to be an extra
+burden, particularly with `TaskExt` and now all extensions are ordered before
+the application itself unless it is strictly necessary otherwise.
--- a/src/ocean/util/app/ext/RefuseRootExt.d
+++ b/src/ocean/util/app/ext/RefuseRootExt.d
@@ -33,7 +33,7 @@ class RefuseRootExt : IArgumentsExtExtension
 {
     /***************************************************************************
 
-        Order doesn't matter, so return default -> 0
+        Order doesn't matter, so return anything less than default app 0
 
         Returns:
             the extension order
@@ -42,7 +42,7 @@ class RefuseRootExt : IArgumentsExtExtension
 
     override int order ()
     {
-        return 0;
+        return -1;
     }
 
     /***************************************************************************
--- a/src/ocean/util/app/ext/ReopenableFilesExt.d
+++ b/src/ocean/util/app/ext/ReopenableFilesExt.d
@@ -290,7 +290,7 @@ public class ReopenableFilesExt : IApplicationExtension, ISignalExtExtension
 
     override public int order ( )
     {
-        return 0;
+        return -1;
     }
 
 
--- a/src/ocean/util/app/ext/TaskExt.d
+++ b/src/ocean/util/app/ext/TaskExt.d
@@ -144,7 +144,7 @@ class TaskExt : IConfigExtExtension
 
     public override int order ( )
     {
-        return 0;
+        return -1;
     }
 
     /***************************************************************************
--- a/src/ocean/util/app/ext/TimerExt.d
+++ b/src/ocean/util/app/ext/TimerExt.d
@@ -314,7 +314,7 @@ public class TimerExt : IApplicationExtension
 
     public override int order ( )
     {
-        return 0;
+        return -1;
     }
 
     /***************************************************************************
--- a/src/ocean/util/app/ext/UnixSocketExt.d
+++ b/src/ocean/util/app/ext/UnixSocketExt.d
@@ -242,7 +242,7 @@ public class UnixSocketExt : IApplicationExtension, IConfigExtExtension
 
     public override int order ( )
     {
-        return 0;
+        return -1;
     }
 
     /***************************************************************************


mihails-strasuns-sociomantic:
FYI @joseph-wakeling-sociomantic

gavin-norman-sociomantic:
Could you add a brief explanation in the commit message of why this change is required?

mihails-strasuns-sociomantic:
Sure, probably worth release notes too. For now you can check linked issue though :)

joseph-wakeling-sociomantic:
Thanks!  Just tried this out with an app moving much of its resource instantiation to `processConfig`, which fails without this patch, and works with it.

joseph-wakeling-sociomantic:
Agree that some "why" is needed in commit message and release notes.

leandro-lucarella-sociomantic (commented):
Why does it matter? Extensions have a priority which should set the order in which they are executed. Or is this purely for the initialization phase?

leandro-lucarella-sociomantic:
(maybe the commit message could clarify that)

mihails-strasuns-sociomantic:
> Extensions have a priority which should set the order in which they are executed

How is this supposed to work? Specific case where this matters for #500 is https://github.com/sociomantic-tsunami/ocean/blob/8a7db8ce9fd9eb48d8bca17969ff547651e7d5cf/src/ocean/util/app/ext/ConfigExt.d#L317-L320 - for it to work as expected, application has to be last in that extension array.

joseph-wakeling-sociomantic:
@leandro-lucarella-sociomantic consider the case where e.g. you want to instantiate something before event scheduling starts (e.g. something that will be relied on by `onStatsTimer`), and that thing relies on having epoll available.

In principle the natural place to do this is `processConfig`, but in practice right now that's not possible, because `theScheduler.epoll` will not be initialized before `processConfig` is called.

leandro-lucarella-sociomantic:
Extension will be sorted according to `order()` when they are `registerExtension()`ed.

https://github.com/sociomantic-tsunami/ocean/blob/8a7db8ce9fd9eb48d8bca17969ff547651e7d5cf/src/ocean/util/app/model/ExtensibleClassMixin.d#L94-L104

leandro-lucarella-sociomantic:
So I think the correct fix for this is make your application have a very high `order()` if you want it to execute last. But that will violate some assumptions are made that applications themselves have `order()` 0 (not sure if anything will actually break). The ordering of extensions is probably the trickiest and more annoying part of the app framework. :(

leandro-lucarella-sociomantic:
Maybe some scheduler stuff have to be moved to the `preRun()` hook instead of `run()`?

leandro-lucarella-sociomantic:
https://github.com/sociomantic-tsunami/ocean/blob/8a7db8ce9fd9eb48d8bca17969ff547651e7d5cf/src/ocean/util/app/model/IApplicationExtension.d#L49-L59

joseph-wakeling-sociomantic:
@leandro-lucarella-sociomantic the key thing here is that we need to have some way to instantiate app class resources _after_ config has been parsed and `theScheduler` has been initialized, but _before_ event scheduling actually starts.

Until https://github.com/sociomantic-tsunami/ocean/pull/460 we could just do this kind of stuff in the beginning of `run`, because so long as the stuff got instantiated in less than 30 seconds, there was no issue (which is a flaky solution, but worked in practice).  But now we have actual segfaults happening because `onStatsTimer` tries to use uninstantiated class fields, which can't be instantiated in the app constructor.

mihails-strasuns-sociomantic:
> But that will violate some assumptions are made that applications themselves have order() 0

That sounds very wrong to me and what I was trying to fix in this PR. Downstream shouldn't care at all what is internal implementation detail of `DaemonApp`, by the time control reaches user code everything needs to be initialized and ready whenever possible. So application should have something like `int.max` order.

mihails-strasuns-sociomantic:
@joseph-wakeling-sociomantic the issue here is not specific to scheduler or config - I want to establish strict guarantee that all extension code runs before app code, one that holds true in all possible situations.

joseph-wakeling-sociomantic:
@mihails-strasuns-sociomantic understood and agreed.  Just wanted to make clear the specific problems that we are having to deal with.

leandro-lucarella-sociomantic:
> @leandro-lucarella-sociomantic the key thing here is that we need to have some way to instantiate app class resources after config has been parsed and theScheduler has been initialized, but before event scheduling actually starts.

This sounds a lot like we need a new extension point.

leandro-lucarella-sociomantic:
> @joseph-wakeling-sociomantic the issue here is not specific to scheduler or config - I want to establish strict guarantee that all extension code runs before app code, one that holds true in all possible situations.

Last time I looked at this code was probably when I wrote it :D, but I think this might not work. I don't remember the details, but I think there are some cases where extension code needs to run after the app, but you can check the `order()` for all extensions to see if that's true. In any case, if it works, the fix would be just to increase the `order()` of the app, not change where the `registerExtension` happens.

Still, with cases like this, when you need something to happen between X and Y no matter which other extensions are used, you probably want to define a new extension point/hook. That can even be done inside the extension iteself, like the config or args extensions do.

joseph-wakeling-sociomantic:
I'm starting to feel that @gavin-norman-sociomantic has a definite point that we are overcomplicating our app class framework and that we should simplify down to our actual use cases ... ;-)

Still, I'm happy with any option here as long as we have a way forward.  What I'm worried about is that right now, we can have arbitrary segfaults with our applications if they start up too close to the :00 or :30 points.  Could it be worth rolling that back temporarily until we have our solution in place?

(Memories of Tom: "Never roll back!!"...)

mihails-strasuns-sociomantic:
@leandro-lucarella-sociomantic I checked all extensions and the only one right now that uses order above 0 is `VersionArgsExt`. Reading its code I see that the only reason why it may matter is to terminate argument parsing early if `--version` is specified (and not print any errors from app own parsing if present). Not sure if this really any important.

Everything else will work exactly the same with `int.max` order for app.

leandro-lucarella-sociomantic:
OK, so do you have any plans going forward then?

mihails-strasuns-sociomantic:
Pushed proposed fix.

src/ocean/util/app/Application.d
```diff
@@ -334,13 +334,16 @@ class Application : IApplication
...
+        it is convenient for applications to assume their methods run last
+        but still leaves the space for rare few extension to set even higher
+        order.
 
     ***************************************************************************/
 
     public override int order ( )
     {
-        return 0;
+        return 1_000_000;
```
leandro-lucarella-sociomantic: I don't like it. I like that the application has order 0, is the "center" of the framework. Why changing this at all?

leandro-lucarella-sociomantic (changes requested):

src/ocean/util/app/Application.d
leandro-lucarella-sociomantic: I see there are a lot of extensions using order 0, I think the correct fix would be to make those extensions use some negative value that makes sense (here is when it gets tricky, to get an order that make the extensions play well between themselves).

leandro-lucarella-sociomantic:
https://github.com/sociomantic-tsunami/ocean/blob/b200e555190591538c44ca789c5dc26fc2540928/src/ocean/util/app/Application.d#L39-L43

src/ocean/util/app/Application.d
joseph-wakeling-sociomantic: @leandro-lucarella-sociomantic we could adopt the BASIC line numbering principle of increments of 10 or more, to leave room for other priorities to be inserted ... :trollface: 

Note that 0 is explicitly described in the `IExtension` doc as the default "if order doesn't matter" setting:
https://github.com/sociomantic-tsunami/ocean/blob/v4.1.1/src/ocean/util/app/model/IExtension.d#L35-L36

So if this is true, it is _not_ the case that the correct `order` for the app extension is 0, because for the app extension the order _does_ matter.

A related question is if it can possibly be true (or reasonable) that an extension could/should have a higher `order` than the application extension.

joseph-wakeling-sociomantic: Note, I'm not anti the idea of 0 being reinterpreted as application-extension priority, so individual extensions should specify priority in line with whether they need to be ready before or after the application extension.  But that means we have to change the meaning of the `order` return value.

joseph-wakeling-sociomantic: Reading the docs of `Application.d` it looks like we have a clash here between what is specified in `IExtension.order` doc and what is specified in `Application` doc.

mihails-strasuns-sociomantic: > I don't like it. I like that the application has order 0, is the "center" of the framework. Why changing this at all?

Because `0` is the default value for extension order and we need application to have lower order by default. And relying on virtue of developers to always modify order value of new added extensions is not going to work.

We can keep app at 0 if base `order` method return value for all extensions is defined to `-1_000` or something like that.

leandro-lucarella-sociomantic: There is no default, you have to always explicitly set one. To be honest, I can understand that maybe the docs are difficult to navigate, but it's all there. I think this is a docs-only issue, so maybe we can do that, recommend some value for extensions (like `-1_000`), saying that extensions want to almost always run before the application.

mihails-strasuns-sociomantic: That sounds exactly the same as setting app to `1_000` but (considerably) more error-prone. Why is this so important? Application framework is convoluted enough that at this point you can safely assume no on reads docs when writing extensions, thus suitable defaults are extremely important.

leandro-lucarella-sociomantic: It's not important but minimizes the chances that we fuck up something else. For example your PR didn't update the Application docs. There might be other areas where the App having order of 0 is assumed. I don't see a strong point to change it, and without a strong point, I prefer to minimize the chances of fuck ups.

leandro-lucarella-sociomantic: Specially for a patch-release for me this makes NO sense. If it were a major (or maybe a minor) then maybe. I don't know if other apps might be relying on this order.

mihails-strasuns-sociomantic: Right now it is semi-undefined (in a sense that most extensions use order 0 and resulting order ends up being "reverse order of register calls") - hard to rely on this.

Will you be OK with having this in a minor?

joseph-wakeling-sociomantic:
If this is going to be contentious, perhaps we should bump it to a later release and allow v3.7.2 to go out ... ?  There is useful stuff in there.

src/ocean/util/app/Application.d
leandro-lucarella-sociomantic: > Right now it is semi-undefined (in a sense that most extensions use order 0 and resulting order ends up being "reverse order of register calls") - hard to rely on this.

Well, for those extensions order was important, so I would say is a bug in those extensions and those extensions should be fixed (change the order to some negative value). For me that's the fix that should go in a patch release.

> Will you be OK with having this in a minor?

Change the Application order to something bigger than 0? I'll be OK-ish, but I strongly would recommend against it, because is a silent change on behaviour for any project deriving from `Application` that could have unpredictable and hard to track consequences. I think that sort of changes belong to a major.

leandro-lucarella-sociomantic: @gavin-norman-sociomantic it would be useful to have your POV on this.

gavin-norman-sociomantic: Given that:
1. I think we should seriously look into a major refactoring of the internals of the app framework soon,
2. it's pretty unlikely that more extensions will be added any time soon,

I think it makes the most sense to keep the meaning of 0 as the natural middle point (app initialisation), and move all extensions that should happen before then to a negative order.

mihails-strasuns-sociomantic:
Updated

relnotes/order.feature.md
```diff
@@ -0,0 +1,7 @@
+### Changed order of app extensions
+
+Previously several extensions used in DaemonApp had same order as application
```
gavin-norman-sociomantic, line 3: Maybe put this description in the commit message too?

mihails-strasuns-sociomantic: Sure, done

leandro-lucarella-sociomantic (approved):

PR #502, opened 2018-05-07
v5.0.0: DMD 2.080 support

mihails-strasuns-sociomantic:


mihails-strasuns-sociomantic:
For now just showing changes needed to support 2.080 - not sure if we can tweak to make suitable for v4.x.x branch but feels unlikely. FYI @sociomantic-tsunami/core-team

mathias-baumann-sociomantic:
I am a bit worried about the removal of the `delete` commands. Couldn't this cause the GC to run & collect for apps that try to avoid the GC and its stop-the-world behavior?

mihails-strasuns-sociomantic:
No, one can't rely on `delete` to help with preventing GC collections anyway - and our regular practice is to simply not allocate at all.

From what I see, most code that currently uses is either already subtly wrong, or has potential to become such.

mihails-strasuns-sociomantic:
What it can do though is to make existing collections take longer - I will have to measure that one one of our GC-heavy apps.

src/ocean/text/convert/UnicodeBom.d
```diff
@@ -381,13 +381,13 @@ class BomSniffer
     private static Const!(Info[]) lookup = [
         {Utf8,  Encoding.Unknown,  null,        true,  false, false, Encoding.UTF_8},
         {Utf8,  Encoding.UTF_8N,   null,        true,  false, false, Encoding.UTF_8},
-        {Utf8,  Encoding.UTF_8,    x"efbbbf",   false},
+        {Utf8,  Encoding.UTF_8,    arrayOf!(char)(0xEF, 0xBB, 0xBF),   false},
         {Utf16, Encoding.UTF_16,   null,        true,  false, false, Encoding.UTF_16BE},
-        {Utf16, Encoding.UTF_16BE, x"feff",     false, true, true},
-        {Utf16, Encoding.UTF_16LE, x"fffe",     false, true},
+        {Utf16, Encoding.UTF_16BE, arrayOf!(char)(0xFE, 0xFF), false, true, true},
+        {Utf16, Encoding.UTF_16LE, arrayOf!(char)(0xFF, 0xFE), false, true},
```
leandro-lucarella-sociomantic, line 387: I know is not really important, but this really cripples readability in some cases, isn't there any function to just keep using the string literal (like `hex("ffef")`)?

src/ocean/io/device/TempFile.d
```diff
@@ -275,8 +276,8 @@ class TempFile : File
             istring prefix=DEFAULT_PREFIX,
             istring suffix=DEFAULT_SUFFIX)
     {
-        auto junk = new char[length];
-        scope(exit) delete junk;
+        auto junk = (cast(char*) malloc(length))[0 .. length];
+        scope(exit) free(junk.ptr);
```
leandro-lucarella-sociomantic, line 280: Maybe documents this allocates memory? I guess since one probably creates a name to do I/O is not a big issue, but still... On the other side, creating a random name without creating the file atomically is insecure, so maybe we should just deprecate this function really :)

src/ocean/math/internal/BiguintCore.d
```diff
@@ -498,7 +498,6 @@ static uint modInt(BigUint x, uint y) {
         uint [] wasteful = new BigDigit[x.data.length];
         wasteful[] = x.data[];
         uint rem = multibyteDivAssign(wasteful, y, 0);
-        delete wasteful;
```
leandro-lucarella-sociomantic, line 501: What about using malloc here too? I see the comment above says this is horribly inefficient anyway, but I'm not sure if trolling the GC is helping either.

leandro-lucarella-sociomantic (commented):
In general I'm not too happy to push more work to the GC when we had deterministic frees. I understand it comes with a risk (that's why the GC is there), but when we had already tested and working code... In any case, I know it is probably more a philosophical issue rather a practical one, but it would be nice to do some actual performance testing.

mihails-strasuns-sociomantic:
I would actually consider if we can deprecate part of affected code completely (like BigInt modules) - those are really bad performance-wise and better not be used at all.

leandro-lucarella-sociomantic:
That works for me.

src/ocean/text/convert/UnicodeBom.d
jacob-carlborg: `std.conv.hexString` [1], it pays to read the changelog :smiley:.

[1] https://dlang.org/changelog/2.079.0.html#hexstrings

david-eckardt-sociomantic: What about `"\xFF\xFE"`? Not as nice to read as `x"fffe"` but no library required.

jacob-carlborg: The implementation of `hexString` [1] is pretty trivial and could be copied to Ocean to avoid dependencies.

[1] https://github.com/dlang/phobos/blob/27a875709dbbb607795a562b4c20bebe3d38f81e/std/conv.d#L6255

mihails-strasuns-sociomantic: Yes, but current workaround is even more trivial. It is not like this functionality is used any regularly.

PR #504, opened 2018-05-08
Use Throwable as a type constraint for testThrown

### LINKED ISSUE #503: testThrown does not compile with AssertError
```
 Error: template instance testThrown!(AssertError) does not match template declaration testThrown(E : Exception = Exception)(lazy void expr, bool strict = true, istring file = __FILE__, int line = __LINE__)
```

maybe this should be the solution: 
```
public void testThrown ( E : Throwable = Exception )
```

bogdan-szabo-sociomantic:
fix #503

### DIFF
--- a/src/ocean/core/Test.d
+++ b/src/ocean/core/Test.d
@@ -130,7 +130,7 @@ unittest
 
 ******************************************************************************/
 
-public void testThrown ( E : Exception = Exception ) ( lazy void expr,
+public void testThrown ( E : Throwable = Exception ) ( lazy void expr,
     bool strict = true, istring file = __FILE__, int line = __LINE__ )
 {
     bool was_thrown = false;


mihails-strasuns-sociomantic:
This currently works as intended. Under no circumstance ever you should catch `AssertError` - if your current code does so, this indicates bug in either application or tests.

andrej-mitrovic-sociomantic:
Hmm.. isn't there a use-case for testing logical errors? It's used in Phobos in some cases for testing for `AssertError`s. Although admittedly there's not that many tests, most of them use Exceptions.

andrej-mitrovic-sociomantic:
Edit: we talked about this IRL and we ended up not needing this after all. Though I'm still curious if logical errors should be tested (in general and not really related to this PR).

mihails-strasuns-sociomantic:
@andrej-mitrovic-sociomantic well, that means that Phobos is violating D own spec. You can't catch anything that derives from `Error` and expect app to remain in defined state, never.

We can test logical errors if `verify` is used instead of `assert` as it throws `SanityException`

PR #505, opened 2018-05-11
Specify exact d1to2fix version to use

mihails-strasuns-sociomantic:
Currently used docker base image has too old d1to2fix fix version
installed which affects auto-generated +d2 tag content.

### DIFF
--- a/docker/build
+++ b/docker/build
@@ -3,6 +3,7 @@ set -xeu
 
 # Install dependencies
 apt-get install -y \
+    d1to2fix=0.10.* \
     libglib2.0-dev \
     libpcre3-dev \
     libxml2-dev \


docker/build
```diff
@@ -3,6 +3,7 @@ set -xeu
 
 # Install dependencies
 apt-get install -y \
+    d1to2fix=0.10.3* \
```
leandro-lucarella-sociomantic: Oh, I don't think 0.10.x is present in any cachalot image, does it have important changes to justify a cachalot release? Is it backwards incompatible or can it be done as a "patch" release of cachalot?

Also, why the `.3*`, that would match `0.10.3` and `0.10.34` but not `0.10.4`, which sounds really weird.

mihails-strasuns-sociomantic:
Oops, I started to bisect existing +d2 tags to see where it differs from generated version and found that this is not d1to2fix problem after all - it is makd problem. Makd submodule has to be relatively recent to pass flags to d1to2fix correctly.

leandro-lucarella-sociomantic:
Ouch! Which is the minimal minor version of Makd needed for this? Are you updating this/creating a new PR?

leandro-lucarella-sociomantic:
Ping?

mihails-strasuns-sociomantic:
> Ouch! Which is the minimal minor version of Makd needed for this?

1.9.x AFAICS. In 3.7.x base makd v2.1.1 is already used so there is nothing to upgrade in this PR. I would just merge it as a precaution towards any d1to2fix changes in future.

mihails-strasuns-sociomantic:
Updated:
- better wildcard

PR #506, opened 2018-05-11
Fix closure for D2 builds

daniel-zullo:
Using closure causes GC allocation in D2 builds.
The problem was found setting the flag -vgc for DFLAGS.

### DIFF
--- a/src/ocean/io/select/EpollSelectDispatcher.d
+++ b/src/ocean/io/select/EpollSelectDispatcher.d
@@ -402,12 +402,11 @@ public class EpollSelectDispatcher : IEpollSelectDispatcherInfo
                 {
                     /// Predicate for finding the ISelectClient inside
                     /// array of epoll_event_t entries.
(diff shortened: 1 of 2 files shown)


daniel-zullo:
There is another closure here:
https://github.com/sociomantic-tsunami/ocean/blob/v3.6.x/src/ocean/task/util/Timer.d#L105-L106

But I'm not entirely sure if it's right to make it `scope` in this special case.

david-eckardt-sociomantic:
> But I'm not entirely sure if it's right to make it `scope` in this special case.

Make it `scope`, please: Later in the function the delegate is unregistered before the function returns, and this is documented with code comments.

daniel-zullo:
Updated based on a suggestion from David. And rebased it against v3.7.x

david-eckardt-sociomantic (approved):

daniel-zullo:
Thanks! :+1:

david-eckardt-sociomantic:
Thank you!

PR #509, opened 2018-05-15
Add closure allocation check to Travis

mihails-strasuns-sociomantic:


mihails-strasuns-sociomantic:
Ready to go

ci/closures.sh
```diff
@@ -0,0 +1,15 @@
...
+export F=devel
+# Enables printing of all potential GC allocation sources to stdout
+export DFLAGS=-vgc
+
+# Prepare docker image and sources
+beaver dlang install
+beaver dlang make d2conv
+
+# Ensure that there are no lines about closure allocations in the output
+! 2>&1 beaver dlang make fasttest | grep "closure"
```
leandro-lucarella-sociomantic: Just cosmetic, but the `!` gets a bit lost if you do the redirection at the start, why not
```sh
! beaver dlang make fasttest 2>&1 | grep "closure"
```

leandro-lucarella-sociomantic (commented):
I think the script to check should go to beaver though, for re usability purposes

leandro-lucarella-sociomantic:
Mist fail?

ci/closures.sh
mihails-strasuns-sociomantic: Sure, no preferences here.

mihails-strasuns-sociomantic:
Removed commit used to check that check can indeed fail. Should be green now.

mihails-strasuns-sociomantic:
> I think the script to check should go to beaver though, for re usability purposes

It is pretty small though. I am a bit concerned about adding levels of indirection for 4 line scripts.

leandro-lucarella-sociomantic:
> It is pretty small though. I am a bit concerned about adding levels of indirection for 4 line scripts.

Well, on the other hand if this is copy&pasted to every project and then one adjustment needs to be made, we need to spread it to all projects manually.

joseph-wakeling-sociomantic:
> Well, on the other hand if this is copy&pasted to every project and then one adjustment needs to be made, we need to spread it to all projects manually.

Bear in mind that not all projects use beaver yet, whereas just about all D projects use ocean.  So it's likely more discoverable/useful if it's provided by the latter for now.  (We can always reorganize that later as beaver usage becomes more comprehensive.)

nemanja-boric-sociomantic:
Though the script in ocean already depends on the `beaver` anyway, so we may have it there as well.

joseph-wakeling-sociomantic:
Yea, just noticed :-\  Is there a recommendation for how to integrate this closure check into non-beaver-using libraries and/or applications?

nemanja-boric-sociomantic:
The biggest issue is adapting your CI for it. If you're able to execute arbitrary command on your CI (like in travis), then it's nothing else than:

1. Convert to D2 code
2. Run `!(DFLAGS=-vgc make fasttest 2>&1 | grep -q closure)` or similar

mihails-strasuns-sociomantic:
> grep -q closure

I would actually recommend using it without `-q` - this way you don't need to rerun tests locally to see what caused failure ;)

leandro-lucarella-sociomantic:
> Yea, just noticed :-\ Is there a recommendation for how to integrate this closure check into non-beaver-using libraries and/or applications?

This will be integrated in the new pipeline jobs/library. See https://github.com/sociomantic/krill/pull/799

Issue #511, opened 2018-05-16
AppendFile uses the file instance to register itself, while other appenders use ClassInfo


nemanja-boric-sociomantic:
`AppendFile` is special in a way that the registering is done by the name, and not by the 
classinfo: https://github.com/sociomantic-tsunami/ocean/blob/d7e97a4c55a56cb0fc330a663754c7983deb4e7c/src/ocean/util/log/AppendFile.d#L53

Other appenders use `classinfo` to register themselves:

https://github.com/sociomantic-tsunami/ocean/blob/d7e97a4c55a56cb0fc330a663754c7983deb4e7c/src/ocean/util/log/Appender.d#L228

The latter does match the documentation, but I'm not sure what the consequences are:

https://github.com/sociomantic-tsunami/ocean/blob/d7e97a4c55a56cb0fc330a663754c7983deb4e7c/src/ocean/util/log/Appender.d#L111-L114

mihails-strasuns:
As long as actual instances are unique, I don't think there is any problem here.

nemanja-boric-sociomantic:
I wouldn't characterize this as a bug anymore. I tried diving into the logging system, but it's really hard to figure out will the proposed change have any further impact and knowing this framework, it most likely it has.

PR #512, opened 2018-05-16
Revert back to AppendFile in VersionArgsExt

nemanja-boric-sociomantic:
There's something weird with the logger hierarchy when it comes to the difference between `AppendStream` and `AppendFile` (see #511), and that change is too risky and big for the bug fix release IMHO. So, this reverts the VersionArgsExt to use AppendFile, but with keeping it registered with the `ReopenableFilesExt`.


mihails-strasuns-sociomantic (approved):
`AppendFile` is magic :X

PR #515, opened 2018-05-18
Make UnixListener command response a delegate in a integration test

### LINKED ISSUE #514: unixlistener integrationtest fails on the dlang jenkins
```
/var/lib/jenkins/dlang_projects@3/sociomantic-tsunami/ocean/integrationtest/unixlistener/main.d(281): Error: incompatible types for `(handleEcho) : (handleShutdown)`: `void function(const(char)[] args, void delegate(const(char)[] response) send_response) @system` and `void delegate(const(char)[] args, void delegate(const(char)[] response) send_response) @system`
```

Got broken when merging v3.x.x. in v4.x.x in the release process. We didn't see it because of the D-APT failure.

nemanja-boric-sociomantic:
UnixListener uses delegates for the commands, and the provided command
didn't use the outer context, so it was deducted to be a function
instead. This explicitly sets it to be delegate.

Fixes #514

### DIFF
--- a/integrationtest/unixlistener/main.d
+++ b/integrationtest/unixlistener/main.d
@@ -270,7 +270,7 @@ int main ( )
         };
 
         // Simple echo command, non-interactive
-        scope handleEcho = ( cstring args,
+        scope handleEcho = delegate ( cstring args,
                 void delegate (cstring response) send_response)
         {
             send_response(args);


gavin-norman-sociomantic:
>the provided command didn't use scope

What do you mean here? The variable is declared as `scope handleEcho`...?

nemanja-boric-sociomantic:
Ah, sorry - I was meaning to say - it wasn't using outer context so it was deduced to be a `function`, not `delegate`. Will reword.

gavin-norman-sociomantic (approved):

PR #520, opened 2018-05-30
Fix performance regression in task timer utils

mihails-strasuns-sociomantic:


src/ocean/task/util/Timer.d
```diff
@@ -200,13 +205,16 @@ private struct EventData
 
 private class TaskTimerSet : TimerSet!(EventData)
 {
-    invariant ( )
+    debug (TaskTimerInvariant)
```
nemanja-boric-sociomantic, line 208: Oh my, that was a nasty slip!

nemanja-boric-sociomantic (approved):

PR #521, opened 2018-05-30
Remove note about upstream D2 incompatibility from the readme

nemanja-boric-sociomantic:
This was resolved in the recent compilers and in the ocean's v4.x.x
branch.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -22,18 +22,7 @@ D2 Compatibility
 
 By default all development in Ocean is done in D1, but using a subset that is
 almost D2 comptaible (and with the help of d1to2fix_, it can be fully converted
-to D2).
-
-That said, for now Ocean is only intended to work with D 2.070.x, but even that
-is not possible at the moment because of some changes needed in the upstream
-compiler that are still pending. Because of this a patched *transitional*
-compiler is needed.  The patches needed to compile the ``dmd-transitional``
-compiler are located in the `dedicated repository
-<https://github.com/sociomantic-tsunami/dmd-transitional>`_.
-
-We are working with upstream to get this issue sorted out as soon as possible.
-To track progress or read more details, please subscribe to `issue #9
-<https://github.com/sociomantic-tsunami/ocean/issues/9>`_.
+to D2, and you checkout ``+d2`` tags which contain D2 code).
 
 Also, as far as D2 upstream get serious about hassle-free upgrading to new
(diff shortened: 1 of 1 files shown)


mihails-strasuns-sociomantic (approved):

README.rst
```diff
@@ -22,18 +22,7 @@ D2 Compatibility
...
-is not possible at the moment because of some changes needed in the upstream
-compiler that are still pending. Because of this a patched *transitional*
-compiler is needed.  The patches needed to compile the ``dmd-transitional``
-compiler are located in the `dedicated repository
-<https://github.com/sociomantic-tsunami/dmd-transitional>`_.
-
-We are working with upstream to get this issue sorted out as soon as possible.
-To track progress or read more details, please subscribe to `issue #9
-<https://github.com/sociomantic-tsunami/ocean/issues/9>`_.
+to D2, and you checkout ``+d2`` tags which contain D2 code).
```
leandro-lucarella-sociomantic, line 25: > By default all development in Ocean is done in D1, but using a subset that is
> almost D2 comptaible (and with the help of d1to2fix_, it can be fully converted
> to D2, and you checkout ``+d2`` tags which contain D2 code).

Is it me or does this sound a bit funny? I can't parse it. Did you mean "fully converted to D2. There are also tags with pre-converted code (i.e. they have D2 code) which are automatically converted tags for every release. These tags end with `+d2`." or something like that?

README.rst
```diff
@@ -22,18 +22,7 @@ D2 Compatibility
...
 
 Also, as far as D2 upstream get serious about hassle-free upgrading to new
 compiler versions, expect Ocean to lag behind latest upstream releases for
```
leandro-lucarella-sociomantic, line 28: Maybe we can remove this paragraph too, now that Ocean is being tested with new upstream compilers, right?

nemanja-boric-sociomantic: I was thinking about it, but then, we have only a PR to move the _unmaintained_ future branch to dmd v2.080, and I'm not sure if we're going to just merge it before v2.081 is out.

On the other side, Ocean test suite passes with the dmd master (with the deprecations allowed).

README.rst
nemanja-boric-sociomantic: That's what I wanted to say, yes ;-)

nemanja-boric-sociomantic: I will address it again, let's just agree on the bottom issue.

PR #522, opened 2018-05-31
Implement support for awaiting already started task

mihails-strasuns-sociomantic:
@matthias-wende-sociomantic this should make your original example "just work". If you don't want to wait for the next ocean release, the following function can be used for emulating same behavior:

```D
void myAwait ( Task task )
{
    auto context = Task.getThis();
    task.terminationHook({
            if (context.suspended())
                theScheduler.delayedResume(context);
    });
    if (!task.finished())
            context.suspend();
}
```

### DIFF
--- a/relnotes/await.feature.md
+++ b/relnotes/await.feature.md
@@ -1,0 +1,20 @@
+### Awaiting on already running task
+
+It is now possible to write code like this:
+
+```D
+  auto sub = new SubTask;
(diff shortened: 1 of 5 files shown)


matthias-wende-sociomantic (commented):
Lgtm. Thanks!

nemanja-boric-sociomantic (approved):

PR #525, opened 2018-06-01
Don't schedule the timer interval to that very moment

nemanja-boric-sociomantic:
Due to the application's misdesign, it might happen
that not all resources that are used in `onStatsTimer` are
initialized at the point when the `startEventHandling` is
called. This leads then to the issue when the Application
starts at the time which would be the aligned time point
and there's no wait between next onStatsTimer call.

This ensures that there's at least one second of the
delay between onStatsTimer and startEventHandling.

### DIFF
--- a/relnotes/timer.migration.md
+++ b/relnotes/timer.migration.md
@@ -1,0 +1,8 @@
+### onStatsTimer will never trigger after zero seconds
+
+`ocean.util.app.DaemonApp`
+
+Previously, to align the stats on 0/30s mark, `onStatsTimer` could fire
+as soon as `startEventlHandling` is called. Ocean now delayes the first
+onStatsTimer call, meaning that it will be called in `[1, interval]`, instead
+of `[0, interval)` range.
--- a/src/ocean/util/app/DaemonApp.d
+++ b/src/ocean/util/app/DaemonApp.d
@@ -507,7 +507,7 @@ public abstract class DaemonApp : Application,
 
     private static ulong timeToNextInterval (ulong interval, time_t current = time(null))
     {
-        return (current % interval) ? (interval - (current % interval)) : 0;
+        return (current % interval) ? (interval - (current % interval)) : interval;
     }
 
     unittest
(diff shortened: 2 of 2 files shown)


mihails-strasuns-sociomantic:
Are you sure about putting this into point release? Can someone depend on old behaviour?

nemanja-boric-sociomantic:
I would be really surprised that anybody depends on this.

leandro-lucarella-sociomantic:
Would it be the first time you are surprised? Even if we would want to risk it, I think this deserves a `+breaking` and a release note to warn the users.

nemanja-boric-sociomantic:
>Would it be the first time you are surprised?

Good point. Since the users need to fix their applications anyway, I would move this into v5.x.x milestone, to avoid `+breaking`.

nemanja-boric-sociomantic:
Updated.

mihails-strasuns-sociomantic (approved):

mihails-strasuns-sociomantic:
Looks like v5.x.x head is broken, I will have a look.

joseph-wakeling-sociomantic:
I thought that this was being addressed by changing priority of extensions so that we could rely on stuff like `processConfig` happening first ... ?

nemanja-boric-sociomantic:
I see this as an orthogonal concern - the applications that do not initialize everything before they use it in `onStatsTimer` are broken. If the ocean needs to be fixed (and it needs) so they can initialize it in `processConfig`, so be it. However, I fail to see the good reason for allowing `0` as a start interval - in the worst case, we are going to miss that time point because application is still initializing. 

Note - I don't see this as a bugfix for the segmentation fault issue (the fix for that needs to be deterministic), that's why I postponed this just as a addition to v5.0.0.

joseph-wakeling-sociomantic:
Fair enough.  Sounds good.

mihails-strasuns-sociomantic:
@nemanja-boric-sociomantic can you rebase please?

nemanja-boric-sociomantic:
Done.

PR #527, opened 2018-06-05
Always set the date in the HTTP response header

gautam-kotian-sociomantic:
If this is not done, then only the first HTTP response will have the
correct date and time. All subsequent responses will just reuse the
value in that first response.

### DIFF
--- a/src/ocean/net/http/HttpResponse.d
+++ b/src/ocean/net/http/HttpResponse.d
@@ -310,8 +310,7 @@ class HttpResponse : HttpHeader
 
     /**************************************************************************
 
-        Sets the Date message header to the current wall clock time if it is not
-        already set.
+        Sets the Date message header to the current wall clock time.
 
      **************************************************************************/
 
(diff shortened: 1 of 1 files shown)


gautam-kotian-sociomantic:
I still need to investigate why this old date issue is not seen in case of sonar.

leandro-lucarella-sociomantic:
Travis says NO.

leandro-lucarella-sociomantic:
Also, isn't this a breaking (behaviour) change?

mihails-strasuns-sociomantic:
Also base branch should be 3.7.x now :)

mihails-strasuns-sociomantic:
Ping @gautam-kotian-sociomantic

gautam-kotian-sociomantic:
Sorry I don't think I will have time to look into this next week. I will try and get back to it the week after next. It is not a high priority bugfix, so it's ok to wait a bit longer.

PR #531, opened 2018-06-13
Replace alloca usage with static heap buffer

mihails-strasuns-sociomantic:
Makes possible to compile this module with -profile in D1

@joseph-wakeling-sociomantic this should fix your error I think

### DIFF
--- a/src/ocean/util/digest/Digest.d
+++ b/src/ocean/util/digest/Digest.d
@@ -23,8 +23,6 @@ import ocean.transition;
 
 import ocean.core.Verify;
 
-import core.stdc.stdlib : alloca;
-
 version(UnitTest) import ocean.core.Test;
 
 /*******************************************************************************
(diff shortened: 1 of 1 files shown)


nemanja-boric-sociomantic (approved):
Finally somebody that cares enough to fix this.

joseph-wakeling-sociomantic:
Cheers, I'll give it a go.

david-eckardt-sociomantic:
The irony is that for all supported hash algorithms the size of the required buffer is known at compile-time, it just can be passed to this method at run-time only.
Another note, although it's too late now: IMHO a fixed-size array is just as good as `alloca`:
```d
ubyte[2048] buf; // or whatever the longest supported hash needs
ubyte[] ret = binaryDigest(buf[0 .. digestSize()]);
```
Full unittest coverage will confirm that `buf.length` is sufficient. Using `alloca` has no advantage over a fixed-size array: While with `alloca` the buffer will always be long enough, it comes at the cost of a possible stack overflow, so the worst-case scenario needs always to be taken into account. Therefore there is no advantage if `alloca` allocates less stack memory than the worst case would be.

nemanja-boric-sociomantic:
With fixed array or alloca, as long as you keep a large buffer on the stack, we have a potential problem, since you normally can't guarantee how deep your fiber is going to be when calling this method. 

The benefit of keeping the buffer not on the stack (though it has its price) is to avoid for it to play a major role in stackoverflow. This is what I suggested for the new `Formatter` at the time and it turned out it was a major success (no more stackoverflows from `Layout` are spotted in the wild).

david-eckardt-sociomantic:
Fair enough, a dynamically allocated buffer has this advantage.

nemanja-boric-sociomantic:
Though, storing a maximum fixed-size buffer on the heap and slicing it avoids querying GC and `assumeSafeAppend` calls, which brings us best of two worlds.

PR #534, opened 2018-06-15
Extract the Unix socket command system from the application

nemanja-boric-sociomantic:
The only convenient way to use the Unix domain socket as a control point
was through the `UnixSocketExt`, which was limiting this to a single
socket per application. Now the `UnixSocketExt` is only instantiating and
configuring the instance of `UnixSocketControlInterface`, which can
also be instantiated and configured manually as many times as needed.

### DIFF
--- a/relnotes/unixcommand.feature.md
+++ b/relnotes/unixcommand.feature.md
@@ -1,0 +1,7 @@
+### Extract the Unix socket command system from the application
+
+The only convenient way to use the Unix domain socket as a control point was
+through the `UnixSocketExt`, which was limiting this to a single socket per
+application. Now the `UnixSocketExt` is only instantiating and configuring the
+instance of `UnixSocketListener` and `ComandsRegistry`, which can also be
+instantiated and configured manually as many times as needed.
--- a/src/ocean/net/server/unix/CommandRegistry.d
+++ b/src/ocean/net/server/unix/CommandRegistry.d
@@ -1,0 +1,135 @@
+/*******************************************************************************
+
+    Utility for handling requests via a unix socket.
+
+    Commands can be registered to the registry and when the command is received
+    the provided delegate will be called. Any provided arguments will be split
+    by the space character and provided as an array.
+
+    Copyright:
+        Copyright (c) 2009-2018 Sociomantic Labs GmbH.
+        All rights reserved.
+
+    License:
+        Boost Software License Version 1.0. See LICENSE_BOOST.txt for details.
+        Alternatively, this file may be distributed under the terms of the Tango
+        3-Clause BSD License (see LICENSE_BSD.txt for details).
+
+*******************************************************************************/
+
+module ocean.net.server.unix.CommandRegistry;
+
+import ocean.transition;
+
+/// ditto
+public class CommandsRegistry
+{
+    import ocean.core.array.Mutation: filterInPlace;
+    import ocean.core.array.Transformation: split;
+    import ocean.core.Buffer;
+    import ocean.core.Enforce;
+    import ocean.core.Verify;
+    import ocean.text.convert.Integer;
+
+    /// Alias for our interactive handler delegate.
+    public alias void delegate ( cstring[],
+            void delegate (cstring),
+            void delegate (ref mstring) ) InteractiveHandler;
+
+    /// Alias for our non-interactive handler delegate.
+    public alias void delegate ( cstring[],
+            void delegate (cstring) ) Handler;
+
+    /// Our registered map of interactive handlers by command.
+    private InteractiveHandler[istring] interactive_handlers;
+
+    /// Our registered map of handlers by command.
+    private Handler[istring] handlers;
+
+    /// Stores the command arguments split by " ";
+    private cstring[] args_buf;
+
+    /***************************************************************************
+
+        Receives all commands from the socket and splits the command by " ".
+        If a matching command is registered then the command will be called with
+        the remaining arguments. This method will be called by the
+        UnixSocketListener whenever a command is received.
+
+        Params:
+            command = The command received by the unix socket.
+            args = The arguments provided with the command.
+            send_response = Delegate to call with response string.
+            wait_reply = Delegate to call to obtain the reply from the user
+
+    ***************************************************************************/
+
+    public void handle ( cstring command, cstring args,
+                 void delegate (cstring) send_response,
+                 void delegate (ref mstring buf) wait_reply )
+    {
+        if (auto handler = command in this.interactive_handlers)
+        {
+            split(args, " ", this.args_buf);
+            scope predicate = (cstring v) { return !v.length; };
+            auto arguments = this.args_buf[0..filterInPlace(this.args_buf[], predicate)];
+            (*handler)(arguments, send_response, wait_reply);
+        }
+        else if (auto handler = command in this.handlers)
+        {
+            split(args, " ", this.args_buf);
+            scope predicate = (cstring v) { return !v.length; };
+            auto arguments = this.args_buf[0..filterInPlace(this.args_buf[], predicate)];
+            (*handler)(arguments, send_response);
+        }
+        else
+        {
+            send_response("Command not found\n");
+        }
+    }
+
+    /***************************************************************************
+
+        Register a command and interactive handler to the unix listener.
+
+        Params:
+            command = The command to listen for in the socket listener.
+            handler = The interactive handler to call when command is received.
+
+    ***************************************************************************/
+
+    public void addHandler ( istring command,
+        InteractiveHandler handler )
+    {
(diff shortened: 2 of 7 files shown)


src/ocean/util/UnixSocketCommandInterface.d
```diff
@@ -0,0 +1,161 @@
+/*******************************************************************************
```
gavin-norman-sociomantic: I question the choice of `util` for this module. As it's inextricably connected with `ocean.net.server.unix.UnixListener`, wouldn't it make more sense to put it in the same package? (I know it's not actually related to network I/O, but the same goes for loads of the stuff in `ocean.net.http`, for example.)

src/ocean/util/UnixSocketCommandInterface.d
```diff
@@ -0,0 +1,161 @@
...
+module ocean.util.UnixSocketCommandInterface;
+
+import ocean.transition;
+import ocean.core.Verify;
+
+import ocean.core.Enforce;
+import ocean.io.select.EpollSelectDispatcher;
+
+/// ditto
+public class UnixSocketCommandInterface
```
gavin-norman-sociomantic: Any reason for this to be a class?

src/ocean/util/UnixSocketCommandInterface.d
```diff
@@ -0,0 +1,161 @@
...
+
+
+    /***************************************************************************
+
+        Binds to the path and listens for the new commands
+
+        Params:
+            epoll = Epoll instance.
+            path = path to bind the socket to
+            mode = mode to apply to the socket
```
gavin-norman-sociomantic: It'd be nice to include a link or documentation for which values are possible and what they mean.

src/ocean/util/UnixSocketCommandInterface.d
```diff
@@ -0,0 +1,161 @@
...
+            enforce(chmod(StringC.toCString(mpath), this.mode) == 0,
+                    "Couldn't change UnixSocket mode.");
+        }
+
+        epoll.register(this.unix_listener);
+    }
+
+    /***************************************************************************
+
+        Receives all commands from the socket and splits the command by " ".
```
gavin-norman-sociomantic: "all commands" makes it sound like the string can contain multiple commands. The code doesn't look like that, though.

src/ocean/util/UnixSocketCommandInterface.d
```diff
@@ -0,0 +1,161 @@
...
+        Binds to the path and listens for the new commands
+
+        Params:
+            epoll = Epoll instance.
+            path = path to bind the socket to
+            mode = mode to apply to the socket
+
+    ***************************************************************************/
+
+    public void listen ( EpollSelectDispatcher epoll,
```
gavin-norman-sociomantic: I feel slightly uncomfortable about this class being basically about string parsing and handler lookup, but then this one method (ok, `shutdown` too) involves networking. Any thoughts?

src/ocean/util/UnixSocketCommandInterface.d
nemanja-boric-sociomantic: It didn't feel all right for me to be in `ocean.net.*`, since it does not have much to do with the network itself - this just presents itself as way to add and remove the handlers and run them. And it's really the utility module IMHO. I somehow feel the difference between `http` and this - while the former is full blown http server, this is just a small utility.

I'm open for any other proposals, though, as I don't feel strongly about this.

src/ocean/util/UnixSocketCommandInterface.d
nemanja-boric-sociomantic: I'll explain this better - this is just a normal UNIX permission mode (and I default to the reasonable value - `0600`).

src/ocean/util/UnixSocketCommandInterface.d
nemanja-boric-sociomantic: Yeah, that's the copy/paste from the old comment.

src/ocean/util/UnixSocketCommandInterface.d
nemanja-boric-sociomantic: No, good catch :+1:

src/ocean/util/UnixSocketCommandInterface.d
nemanja-boric-sociomantic: So, the exiting API of the `UnixSocketListener` is that it accepts some object that has a `handle` method and it calls that method to handle the request. So, that bit of the class (along with the `addHandler`, etc) can surely be moved to a separate struct, and from there we can go two ways:

1. Either we still maintain `UnixSocketCommandInterface` as a convenient way to create a socket and handle the commands (we just instantiate), or
2. We just leave the handler bits. However, then user falls into a trap where they need to figure out, honestly, not so trivial `UnixSocketListener`'s API and how to plug these bits together. 

Over these two, I would prefer 1, but I honestly don't see downsides to keep everything in a single struct. However, I was wrong before about the modularization (see the reason for this PR as a reference), so I am keen to change things.

src/ocean/util/UnixSocketCommandInterface.d
nemanja-boric-sociomantic: Sad, the bit with the connection handlers needs to be a class, since the `UnixListener` takes a reference of the current instance. Making this a struct would copy it and `addHandler`/`removeHandler` would no longer reflect on the actual connection handler.

src/ocean/util/UnixSocketCommandInterface.d
nemanja-boric-sociomantic: I've implemented the approach described in 1 - I've split the handler's registry functionality into a separate class, and created the handy helper to combine the `UnixListener` and `CommandsRegistry` API.

src/ocean/util/UnixSocketCommandInterface.d
```diff
@@ -0,0 +1,219 @@
...
+            cstring path, int mode = -1)
+    {
+        this.path = path;
+        this.mode = mode;
+
+        this.unix_listener =
+            new typeof(this.unix_listener)(this.path, epoll, this.commands_registry);
+
+        if (this.mode >= 0)
```
gavin-norman-sociomantic: Could it make any sense to move this mode setting into `UnixSocketListener`? If that were done, I think we could do away with `UnixSocketCommandInterface`. Usage would look like:
```D
auto commands = new CommandsRegistry;
auto unix_listener = new UnixListener(this.path, mode, epoll, commands);
epoll.register(unix_listener);
```

nemanja-boric-sociomantic: Yeah good idea, it is a property of unix socket and it should be there.

nemanja-boric-sociomantic:
Rebased on top of v4.x.x, changed milestone and addressed comments.

src/ocean/net/server/unix/CommandRegistry.d
```diff
@@ -0,0 +1,134 @@
...
+            split(args, " ", this.args_buf);
+            scope predicate = (cstring v) { return v.length; };
+            auto arguments = filter(this.args_buf[], predicate, this.args_buf);
+            (*handler)(arguments, send_response, wait_reply);
+        }
+        else if (auto handler = command in this.handlers)
+        {
+            split(args, " ", this.args_buf);
+            scope predicate = (cstring v) { return v.length; };
+            auto arguments = filter(this.args_buf[], predicate, this.args_buf);
```
mihails-strasuns-sociomantic: This will allocate on each unix socket command, surely we don't want that? https://github.com/sociomantic-tsunami/ocean/blob/v3.x.x/src/ocean/core/array/Transformation.d#L375

Same for filter/split calls few lines above

mihails-strasuns-sociomantic (commented):
Very nice! One memory allocation remark though

src/ocean/net/server/unix/CommandRegistry.d
nemanja-boric-sociomantic: Ah good catch! This is old code so I didn't pay attention.

relnotes/unixcommand.feature.md
```diff
@@ -0,0 +1,7 @@
+### Extract the Unix socket command system from the application
+
+The only convenient way to use the Unix domain socket as a control point was
```
gavin-norman-sociomantic, line 3: Commit message needs to be updated. (Mentions `UnixSocketControlInterface`.)

src/ocean/net/server/unix/CommandRegistry.d
nemanja-boric-sociomantic: Very nice how this method actually receives a buffer but then ruins it all :-)

mihails-strasuns-sociomantic (approved):

PR #536, opened 2018-06-19
Add bcontains and bsubset functions for Array

burak-tutanlar-sociomantic:


relnotes/array_bcontains.feature.md
```diff
@@ -0,0 +1,8 @@
+### Add bcontains which uses bsearch
```
mihails-strasuns-sociomantic: In cases like that I think it is more convenient for users of ocean to have one `### New array search utilities` entry which lists all added functions

src/ocean/core/array/Search.d
```diff
@@ -1013,6 +1013,101 @@ unittest
...
+        T = type of array element
+
+    Params:
+        a = array A
+        b = array B
+
+    Returns:
+        false if there is any item of array B, which is not an item of array A
+
+    In:
```
mihails-strasuns-sociomantic, line 1042: There is no actual contract in code, copy-paste?

src/ocean/core/array/Search.d
```diff
@@ -1013,6 +1013,101 @@ unittest
...
+        T = type of array element
+
+    Params:
+        array = array to search
+        match = element to search for
+
+    Returns:
+        true if the element was found in the array
+
+    In:
```
mihails-strasuns-sociomantic, line 1091: Ditto lack of contract

src/ocean/core/array/Search.d
```diff
@@ -1013,6 +1013,101 @@ unittest
...
+*******************************************************************************/
+
+public bool bsubset ( T ) ( T[] a, T[] b )
+{
+    if ( a.length == 0 )
+        return b.length == 0;
+
+    foreach ( ref T item; b )
+    {
+        if ( !bcontains(a, item) )
```
mihails-strasuns-sociomantic, line 1054: Note that this is `O(N * log N)`. It can be optimized to `O(N)` if both A and B are required to be sorted.

Not sure if this is of any importance for use case you have in mind.

src/ocean/core/Array.d
```diff
@@ -75,13 +75,14 @@ unittest
...
-    alias bsearch!(long) bs;
-
-    long[] arr = [1, 2, 3,  5, 8, 13, 21];
+    long[] arr = [1, 2, 3, 5, 8, 13, 21];
 
     size_t n;
 
-    test (bs(arr, 5, n));
+    test(bsearch(arr, 5L, n));
+    test(n == 3);
```
mihails-strasuns-sociomantic: `test!("==")`

mihails-strasuns-sociomantic (changes requested):

src/ocean/core/array/Search.d
burak-tutanlar-sociomantic: I copy-pasted from it `bsearch` because this function is using `bcontains` which uses `bsearch`. So the same warning goes for this too.

src/ocean/core/array/Search.d
burak-tutanlar-sociomantic: AFAIU `mismatch` which is implemented in this module is doing that already if the arrays are both sorted and don't have duplicate items. I needed a function which handles the case when I don't have the array B as sorted.

burak-tutanlar-sociomantic:
Updated.

src/ocean/core/array/Search.d
mihails-strasuns-sociomantic: I think it would be more maintainable to just say in regular documentation "See `bsearch` for input constraints" or something like that.

src/ocean/core/array/Search.d
mihails-strasuns-sociomantic: Aha, maybe cross-reference these 2 functions in their documentation?

src/ocean/core/array/Search.d
```diff
@@ -1013,6 +1013,101 @@ unittest
     test!("==")( countIf( "gbbbi"[], ( char c ) { return c == 'd'; } ), 0 );
 }
 
+/*******************************************************************************
+
+    Checks if array B is a subset of array A. Array A needs to be sorted.
+    Array B doesn't need to be sorted and might have duplicate elements. It
+    checks if array A contains each item of array B by using bcontains function.
```
nemanja-boric-sociomantic: `s/by using bcontains function/using binary search`

src/ocean/core/array/Search.d
nemanja-boric-sociomantic: Yes, I would point to `mismatch` from this functions' docs and vice versa.

src/ocean/core/array/Search.d
```diff
@@ -1013,6 +1013,101 @@ unittest
     test!("==")( countIf( "gbbbi"[], ( char c ) { return c == 'd'; } ), 0 );
 }
 
+/*******************************************************************************
+
+    Checks if array B is a subset of array A. Array A needs to be sorted.
```
nemanja-boric-sociomantic: Should specify the sort order for A (like in the function below).

burak-tutanlar-sociomantic:
Updated.

mihails-strasuns-sociomantic (approved):

Issue #542, opened 2018-06-26
DaemonApp's usage example shouldn't register SIGTERM


gkotian:
DaemonApp [always registers SIGTERM](https://github.com/sociomantic-tsunami/ocean/blob/v3.x.x/src/ocean/util/app/DaemonApp.d#L390-L391) in the list of handled signals, but the usage example [seems to suggest](https://github.com/sociomantic-tsunami/ocean/blob/v3.x.x/src/ocean/util/app/DaemonApp.d#L843) that the application needs to register it explicitly.

nemanja-boric-sociomantic:
I'm not sure this is the right approach. User's implementation of the `DaemonApp` does specify handled signals and `onSignal` and IMHO, they should match each other. Even if the `SIGTERM` is in the list implicitly, it is not explicitly, if you choose to override `onSignal` without explicitly calling `super.onSignal`.

Other issue would be if you switch the orders of the `signal` and `super`, you're in a problem:

```
super ();
this.optional_settings.signals = [SIGKILL]; // whoops
```

mihails-strasuns-sociomantic:
I think it was a mistake on my side to define https://github.com/sociomantic-tsunami/ocean/blob/v3.x.x/src/ocean/util/app/DaemonApp.d#L390-L391 like this.

It should be  `if (settings.signals.length == 0) settings.signals = [ SIGTERM ];` instead.

mihails-strasuns-sociomantic:
So the documentation is really correct here - one is supposed to define all signals manually if non-default handling is needed. And default sigterm handler is here for those apps who don't have any configuration of their own.

gkotian:
Alright then. I'll close this issue under the assumption that you guys will take care of updating DaemonApp's constructor in an upcoming release.

I'd also like to take this opportunity to thank you guys for your work on ocean. I work on several applications that use this stellar library. Good job!

PR #543, opened 2018-06-27
Do not enable SIGTERM unconditionally

mihails-strasuns-sociomantic:
Original logic was flawed - if derived application overrides
`onSignal` it may decide to not handle `SIGTERM` there. Thus it is
only necessary to add it to default signals if no custom ones are
defined.

See also https://github.com/sociomantic-tsunami/ocean/issues/542

### DIFF
--- a/relnotes/signal.migration.md
+++ b/relnotes/signal.migration.md
@@ -1,0 +1,10 @@
+### DaemonApp doesn't append SIGTERM to signal list
+
+Previously `DaemonApp` would append `SIGTERM` to handled signal
+list even if there is already an app-defined one. It was a bug - automatic
+handling of `SIGTERM` was only intended to happen if there is no custom
+signal handling defined in derived application.
+
+Starting with 5.0.0 `DaemonApp` will behave as originally intended - if your
+application was relying on old behaviour, make sure to add `SIGTERM` to signal
+list explicitly now.
--- a/src/ocean/util/app/DaemonApp.d
+++ b/src/ocean/util/app/DaemonApp.d
@@ -368,8 +368,10 @@ public abstract class DaemonApp : Application,
     {
         super(name, desc);
 
-        // DaemonApp always handles SIGTERM:
-        settings.signals ~= core.sys.posix.signal.SIGTERM;
+        // If derived app does not handle signals explicitly, default
+        // DaemonApp handler will be used which handles SIGTERM
+        if (settings.signals.length == 0)
+            settings.signals = [ core.sys.posix.signal.SIGTERM ];
 
         // Create and register arguments extension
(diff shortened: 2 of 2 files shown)


mihails-strasuns-sociomantic:
Not sure about the target branch - this is a bug fix but it is theoretically possible for an app to depend on old incorrect behavior (by not specifying `SIGTERM` but handling it). Delaying it to v5.0.0 feels a bit of an overkill though :(

gavin-norman-sociomantic:
I'm not quite clear what the plan is here.

As far as I understand the code before this PR, the intention is this:
* `SIGTERM` is always handled.
* The `SIGTERM` handler shuts down the scheduler / epoll.

Of course, the actual implementation of this is flawed, as the derived class might override `onSignal` and not call `super.onSignal`.

The current change in this PR is "the user is responsible for `SIGTERM` handling". Is this what we really want?

mihails-strasuns-sociomantic:
> The current change in this PR is "the user is responsible for SIGTERM handling". Is this what we really want?

Yes, this is what documentation was saying all the time. I don't think it was ever intention to always force `SIGTERM` handling even if app has logic of its own, at least for me intention was to ensure clean `SIGTERM` handling out of the box in case app just uses defaults.

mihails-strasuns-sociomantic:

gavin-norman-sociomantic:
Ok that makes sense. What I'm concerned about, though, is whether the rationale that you describe was clear to all users in the past.

>Yes, this is what documentation was saying all the time.

Could you point me to documentation where this is clearly and unambiguously stated? That should determine the destination branch for this PR, I feel (i.e. is this a bug fix or a breaking change).

mihails-strasuns-sociomantic:
This issue comes originally from @gkotian pointing to https://github.com/sociomantic-tsunami/ocean/blob/v3.x.x/src/ocean/util/app/DaemonApp.d#L843 and noticing that it does not conceptually match current implementation. Usage example is not unambiguous enough though.

mihails-strasuns-sociomantic:

gavin-norman-sociomantic:
As the doc isn't unambiguous, and there's no urgency to change this, I'd err on the side of caution and apply it in the next major.

nemanja-boric-sociomantic:
This is something that we should aim to make simpler in the new app framework, btw.

mihails-strasuns-sociomantic:
Changed base, added notes.

gavin-norman-sociomantic (approved):

gavin-norman-sociomantic:
Feel free to merge when green.

gavin-norman-sociomantic:
Haha, as soon as I submitted that comment, it went red :(

Issue #519, opened 2018-05-29
Create simplified app framework


gavin-norman-sociomantic:
The [current framework](https://github.com/sociomantic-tsunami/ocean/tree/v3.x.x/src/ocean/util/app), while flexible and extensible, has turned out to be massively overwrought for our needs. Adding new features invariably involves delving into the extensions logic, which is far from simple.

The framework was initially designed to be able to support many different types of application, each formed of a mix of the provided extensions. In practice, it turns out that we only really need ~2 different application types: a simple CLI app, and a full daemon app.

The flexibility of the app framework is thus not needed. It would be beneficial to develop a simpler alternative.

gavin-norman-sociomantic:
Suggested approach:
1. Create a new `DaemonApp` class. Not derived from anything. One abstract method `int mainAppLogic`.
2. For each extension used by the old daemon app:
  a. Extract the core functionality (into a separate struct / class / free functions -- as appropriate).
  b. Adapt the extension to use the new, extracted code, internally.
3. Add the extension's functionality to the new daemon app, using the new, extracted code.

gavin-norman-sociomantic:
Another idea would be to create `ServerDaemon` and `StreamProcessorDaemon`!

mihails-strasuns-sociomantic:
Btw while we are at it, I would love to see a dedicated `ocean.application` top-level package for new stuff and, in general, moving more of big functionality clusters top-level. That makes them more discover-able and easier to split into smaller sub-libraries if we ever want to.

Just a suggestion :)

gavin-norman-sociomantic:
Sounds like a good move. I agree that we have too much stuff crammed inside `ocean.util` :D

gavin-norman-sociomantic:
A thought on a simpler, intermediary step that might be worthwhile: now that the core logic of the extensions has been [separated out](https://github.com/sociomantic-tsunami/ocean/pull/660), we could replace the extensions in the app classes with using the separated code directly. (Without changing the API.)

gavin-norman-sociomantic:
Hm, looks like the "without changing the API" part would be complicated. The set of extensions is public.

PR #544, opened 2018-07-03
New daemon app base class

gavin-norman-sociomantic:


gavin-norman-sociomantic:
Very rough WIP on #519.

gavin-norman-sociomantic:
Rebased on v4.x.x, so as to use the results of https://github.com/sociomantic-tsunami/ocean/pull/534, from v4.2.0.

gavin-norman-sociomantic:
Updated with latest progress:
* Now includes almost all features of the old `DaemonApp` (completely untested still!).
* Moved everything into a new package `ocean.application`.

gavin-norman-sociomantic:
As this framework currently stands, the current "hook" points exist, where the user might want to override / add behaviour:
* Setting up extra args (override `initArgs`)
* Setting up extra unix socket commands (override `initUnixSocketCommands`)
* Special args validation (override `validateArgs`)
* Special args handling (override `handleArgs`)
* Config parsing (override `parseConfig`)
* Main app logic (override `mainAppLogic`)
* Signal handler (override `onSignal`)
* Stats timer handle (override `onStatsTimer`)

Many of these currently entail the nasty "override and call super" pattern. It'd be nice to avoid this, as far as possible. Suggestions welcome.

src/ocean/application/Daemon.d
```diff
@@ -0,0 +1,597 @@
...
+        this.unix_socket.commands.addHandler("reopen_files",
+            &this.reopenFilesCommand);
+    }
+
+    private void initSignals ( )
+    {
+        this.signals = new Signals(&this.onSignal);
+        this.signals.handle(this.settings.signals);
+        this.signals.ignore(this.settings.ignore_signals);
+    }
```
nemanja-boric-sociomantic: Some of these `init*` methods are private and some are protected.

gavin-norman-sociomantic: Yeah, there was some logic to this at some point, though it may not be consistent now. The idea was that derived classes will commonly need to override some things, whereas other things should not be overrideable.

gavin-norman-sociomantic:
Updated:
* Added hooks framework. (Suggested by @mihails-strasuns.) This does seem like a more manageable means of allowing derived classes (potentially hierarchies of them) to add behaviour at specific points.

gavin-norman-sociomantic:
Added doc in `Daemon.d`.

src/ocean/application/Daemon.d
```diff
@@ -0,0 +1,769 @@
...
+
+        /// Application version info. Mandatory.
+        Version.VersionInfo ver;
+
+        /// Configuration files to parse.
+        istring[] config_files = [ "etc/config.ini" ];
+
+        /// If true, errors when parsing configuration files are treated as
+        /// fatal -- the app will not start.
+        bool config_errors_fatal = true;
```
gavin-norman-sociomantic, line 106: Does anyone have a real use case for setting this to `false`? (This would be the equivalent to the old "loose config parsing".) (If there are no real use cases, I'll remove the option to change it.)

gautam-kotian-sociomantic: Some apps have the ability to reload their configurations on receipt of a signal (or a command via the UNIX socket). I don't know what would happen if an error was made in the config file and then the reload command was sent. But supporting "loose config parsing" to handle this situation is anyway not a proper solution. So I don't have any strong arguments for keeping this boolean variable.

gavin-norman-sociomantic: Ah right, good point. I'd be interested to hear how that works in your apps, as this is something that I was wondering about adding (at some point) as a general feature of the daemon base class.

gavin-norman-sociomantic: Looking at the config re-parsing code from one of these apps, it does the following on receipt of SIGHUP:
```D
this.config_ext.preRun(this, null);
```
This triggers the reload of all config files, and calls the app-defined behaviour in `processConfig` again.

gavin-norman-sociomantic: This would be a nice behaviour to support "out of the box".

gavin-norman-sociomantic:
Added config reload unix socket comment.

mihails-strasuns:
Tricky thing with config reloading is that it only reliably works if modified bits are "secondary" and can be trivially re-initialized. Trying to modify something like listen port number can easily explode depending on how app stores and initializes connection handler.

nemanja-boric-sociomantic:
Yes, perhaps we should enhance `ConfigFiller` to mark the fields as `re-initilisable`.

gavin-norman-sociomantic:
>Tricky thing with config reloading is that it only reliably works if modified bits are "secondary" and can be trivially re-initialized. Trying to modify something like listen port number can easily explode depending on how app stores and initializes connection handler.

Yeah, my plan was to just note this requirement in the documentation for now. We can work on a more robust approach down the line. (What Nemanja suggested sounds reasonable.)

gavin-norman-sociomantic:
Updated config reloading so that the only internal component that can be reconfigured is the logging system. (I've tested this and it works. Reconfiguring the stats logger does not currently work.)

gavin-norman-sociomantic:
Added components doc.

gavin-norman-sociomantic:
All testing with the dhtnode adaptation to `Daemon` has been successful (see https://github.com/sociomantic-tsunami/dhtnode/pull/143#issuecomment-406560037).

gavin-norman-sociomantic:
Updated and ready for more serious review:
* Squashed commits.
* Added release notes.
* Rebased on v4.x.x.

gavin-norman-sociomantic:
Open questions:
1. Sorting of signals by severity?
2. Include some kind of migration notes in the release notes?
3. Fixing D2 build errors ;)

mihails-strasuns:
FYI: I have not yet done any source review but will try to focus on it tomorrow.

src/ocean/application/Daemon.d
gavin-norman-sociomantic: Ok so I've added support for config reloading. A suggestion for `config_errors_fatal`: we say that this only applies at startup. When the config file is parsed in response to the unix socket command, parser errors are logged, but do not cause the application to exit.

mihails-strasuns:
D2 builds fail ;)

gavin-norman-sociomantic:
>D2 builds fail ;)

Indeed. See "open questions", [above](https://github.com/sociomantic-tsunami/ocean/pull/544#issuecomment-407439586).

relnotes/daemon.feature.md
```diff
@@ -0,0 +1,44 @@
...
+  this registry.
+* A unix socket command interface, including support for the following
+  built-in commands:
+    * `show_version`
+    * `show_build_info`
+    * `reopen_files`: reopens the specified files (must be registered with
+      the open files registry; see above)
+    * `reload_config`: re-parses the config files and reconfigures any
+      internal components that can be (this currently includes only the
+      logging system).
```
mihails-strasuns, line 39: AppStats via unix socket won't be there by default?

src/ocean/application/Daemon.d
```diff
@@ -0,0 +1,814 @@
...
+    public struct Settings
+    {
+        /// Application name. Mandatory.
+        istring name;
+
+        /// Application description string.
+        istring desc;
+
+        /// How the program is supposed to be invoked.
+        istring usage;
```
mihails-strasuns: Is not of much use for `DaemonApp`, is it?

src/ocean/application/Daemon.d
mihails-strasuns: I think most practical solution is to not allow loose parsing when reloading but instead ignore reloading command completely if there is any config error (and keep running with the old one). Having partially updated config because of typo can be scary.

src/ocean/application/Daemon.d
```diff
@@ -0,0 +1,814 @@
...
+    /// PID lock helper.
+    private PidLock pid;
+
+    /// Signal handlers.
+    private Signals signals;
+
+    /// Process stats tracker.
+    private CpuMemoryStats process_stats;
+
+    // TODO: global process monitor (see ocean.io.select.client.EpollProcess)
```
mihails-strasuns, line 154: Still todo?

src/ocean/application/Daemon.d
```diff
@@ -0,0 +1,814 @@
...
+    /// start-up procedure.
+    private struct Hooks
+    {
+        /// Hooks called just before args handling has occurred.
+        void delegate ( )[] pre_handle_args;
+
+        /// Hooks called during validation of args. If a validation error
+        /// occurs, a description of the error should be appended to the array
+        /// of strings passed to the delegate.
+        void delegate ( istring[] )[] validate_args;
```
mihails-strasuns: Why not just return error and append it on call site? Current API is asking for abuse :) Also it must be `ref istring[]` for appending to workl.

src/ocean/application/Daemon.d
```diff
@@ -0,0 +1,814 @@
...
+
+        /// Hooks called to process the parsed config files.
+        void delegate ( )[] process_config;
```
mihails-strasuns, line 168: No `pre_run` hooks?

src/ocean/application/Daemon.d
```diff
@@ -0,0 +1,814 @@
...
+        critical signals (like `SIGABRT` or `SIGSEGV`) where the application
+        generally shouldn't be allowed to proceed. For these cases, setup an
+        asynchronous signal handler using `sigaction` instead.
+
+        Params:
+            signals = list of signals that fired since the last event loop cycle
+
+    ***************************************************************************/
+
+    protected abstract void onSignal ( int[] signals );
```
mihails-strasuns, line 280: So no default signal handling at all?

mihails-strasuns (commented):
Looks promising. My main higher-level design question is if we indeed want to keep model of giant impl class with few user-facing to override. It feels that now it becomes especially confusing because most of API is not intended to be user-facing anymore thus combined docs can be misleading.

What do you thing about my earlier proposal to have a separate class/interface with just methods intended to be implemented in app code and passing instance implementing it to `DaemonApp` instead?

gavin-norman-sociomantic:
Updated (D2 fixes).

relnotes/daemon.feature.md
gavin-norman-sociomantic: No, this is only used by a few very old applications.

src/ocean/application/Daemon.d
gavin-norman-sociomantic: Ha, good point.

src/ocean/application/Daemon.d
gavin-norman-sociomantic: :+1:

src/ocean/application/Daemon.d
gavin-norman-sociomantic: I didn't find a use for them, in my adaptation of the dhtnode at least. Anything "pre-run" can equally easily be done first thing inside `run()`. Can you think of something specific that couldn't work that way?

src/ocean/application/Daemon.d
gavin-norman-sociomantic: Ah good question. This is something that slipped my mind. Thinking about it, though, maybe forcing the user to consider how they want to handle signals is the best way. The default in `DaemonApp` is essentially just one line (`theScheduler.shutdown();`), so I don't think it's especially important. (This could simply be suggested in the docs.) What do you think?

src/ocean/application/Daemon.d
gavin-norman-sociomantic: This is more of a "might be a nice thing to add in the future". Not intended to be addressed in this PR. (We could add an issue for this, when this PR is finalised.)

gavin-norman-sociomantic:
>What do you thing about my earlier proposal to have a separate class/interface with just methods intended to be implemented in app code and passing instance implementing it to DaemonApp instead?

I'll try that. A kind of `DaemonAppLogic` interface. Sounds like it could be nice.

gavin-norman-sociomantic:
Updated:
* Removed `Settings.usage`.
* Changed type of args validation delegates.

gavin-norman-sociomantic:
>What do you thing about my earlier proposal to have a separate class/interface with just methods intended to be implemented in app code and passing instance implementing it to DaemonApp instead?

Hmm, it's actually pretty inconvenient and adds loads of boilerplate. Two approaches I've looked at:
1. A Core class containing all private members, an API class containing all public/protected members. The API class owns an instance of the Core class. However, the Core class needs access to all the protected bits of the API class, so they need to be passed into all of the methods of Core. (And there are quite a lot of them.)
2. A Core class containing all protected/private members, an API class containing all public members. The API class owns an instance of the Core class. However, the API class needs to expose all the protected bits of the API class, so we need to add getters for them all. (And there are quite a lot of them.)

A mixin containing all the protected/private members would probably work, but that hides the protected fields from the API, which isn't ideal :/

mihails-strasuns:
Wasn't the idea that protected fields are not part of user-facing API anymore and won't be exposed? The way I expected it to work is there are few public bits (config, timers) and everything else is not accessible at all.

src/ocean/application/Daemon.d
Geod24: @gavin-norman-sociomantic : The app I maintained does. It was desirable with split config files, since you could want to test a new version which removes a config parameter used everywhere. Think of:

```
[LOG]
mode = append
file_size = 100_000
```

In your common config file, and both field being required. If you want to test, say, a new version of Ocean which removes `file_size`, you either have to specify the whole configuration in its own file and use that or rely on loose parsing. I choose to do the later (you can look at the commit / comment in the app in question).

However if someone relies on this, @mihails-strasuns ' solution means that they essentially can't reload their config. Doesn't sound too appealing.

gavin-norman-sociomantic:
As things currently stand, all protected fields are intended to be accessible. That includes:
```D
    protected Settings settings;
    protected Arguments args;
    protected ConfigParser config;
    protected StatsLog stats_log;
    protected OpenFiles files;
    protected UnixSocketCommands unix_socket;
    protected Timers timers;
```
(These are all things that implementing classes will want to access.)

mihails-strasuns-sociomantic:
TBH this is something I don't like a lot as it undermines much of intended simplification of new design. Now there are bunch of protected fields that shouldn't be used by apps and bunch of protected fields that should be used. Can we just make those public?

gavin-norman-sociomantic:
>Now there are bunch of protected fields that shouldn't be used by apps and bunch of protected fields that should be used. Can we just make those public?

Nope. All protected fields are intended to be used by apps. Anything that isn't intended to be used by derived classes is private (as you would expect).

mihails-strasuns-sociomantic:
That doesn't look like that to me at all when I read the code :X Note that I make strong distinction between "derived classes for more specific app bases in ocean" and "derived classes for app logic" and don't think those should be treated the same.

gavin-norman-sociomantic:
Could you give an example of a protected field that you think should not be accessible to derived classes (of either kind)?

gavin-norman-sociomantic:
I'm totally happy to rework this btw, I just want to understand your objections and get a clear picture of an alternative structure.

gavin-norman-sociomantic:
Idea from an old issue: Make help text behaviour overrideable. The current behaviour is to just call `Arguments.displayHelp()`. Applications may have specialised, run-time generated help text that they wish to display. (e.g. a swarm client tool, for example, could display the list of supported requests, which is generated at run-time.)

mihails-strasuns:
This is probably more relevant to new version of `CliApp` though (this is planned too, right?)

gavin-norman-sociomantic:
True. Yeah, I suppose it should be planned too ;)

src/ocean/application/Daemon.d
mihails-strasuns: I was thinking about it in context of https://github.com/sociomantic-tsunami/ocean/issues/500 - right now recommended way to initialize app is in `processConfig` override but it would be nice to support dedicated spot for that in a new system.

gavin-norman-sociomantic: I see. That sounds like a reasonable case. I'll add hooks just before the call to `runInTask`.

src/ocean/application/Daemon.d
matthias-wende-sociomantic: I think as it is initialized by default, the scheduler should be shut down by default as well. I would find it inconsistent to shutdown in instance by default that I haven't set up. Also I guess using the daemon base class should be as simple as possible to use especially for new users. Forcing the user to know about these details, ask someone or read docs contradicts that point. Nevertheless providing an application template might be an alternative approach.

gavin-norman-sociomantic: >I think as it is initialized by default, the scheduler should be shut down by default as well.

I agree with that. The difficulty is that there's no default signal handling, so it's not clear where scheduler shut down should occur.

mihails-strasuns:
What is  the current state of this @gavin-norman-sociomantic ?

gavin-norman-sociomantic:
Pending me, you, or Scott having some time to look at playing with improving it.

mihails-strasuns:
@gavin-norman-sociomantic I presume you don't plan to rush it for 4.4.0 right? :)

gavin-norman-sociomantic:
No, delaying until the next minor is fine.

mihails-strasuns:
@gavin-norman-sociomantic do you intend to work on this in any foreseeable future? :)

gavin-norman-sociomantic:
Basically no. I think it's important to continue with this topic, but I personally won't have any time in the near future.

Options:
* Close.
* Someone else take over.

gavin-norman-sociomantic:
Another option would be to merge the "Extract *" commits now and leave the design of the new app base class until later.

mihails-strasuns:
That sounds good. I doubt anyone can take this over until some new cores appear so merging at least some refactoring would be better than nothing.

gavin-norman-sociomantic:
Split into #660. Closing.

PR #548, opened 2018-07-11
Fix eraseStaticLines() if the terminal output is null

daniel-zullo:
If the terminal output is initialized to null in the AppStatus
constructor, then a program will segfault if the `eraseStaticLines()`
is called.

### DIFF
--- a/src/ocean/io/console/AppStatus.d
+++ b/src/ocean/io/console/AppStatus.d
@@ -480,6 +480,11 @@ public class AppStatus
 
     public void eraseStaticLines ( )
     {
+        if (this.terminal_output is null)
+        {
+            return;
+        }
+
         // Add +2: One for header and one for footer
         for (size_t i = 0; i < this.static_lines.length + 2; i++)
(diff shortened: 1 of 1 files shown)


mihails-strasuns-sociomantic:
Am I understanding correctly that the case for this is to use `AppStatus` for just log file output? Shouldn't "new" daemon apps use syslog for that?

(Fix is LGTM, making sure I understand the context)

daniel-zullo:
> Am I understanding correctly that the case for this is to use AppStatus for just log file output? 

IIUC the case for this is to connect to an output device, a unix socket in my use case.
@nemanja-boric-sociomantic you might want to confirm my assumption. Also please check/review this patch

nemanja-boric-sociomantic:
Yes, we're still supporting `AppStatus`, but over the unix domain socket. @daniel-zullo-sociomantic thank you for your PR!

nemanja-boric-sociomantic (approved):

PR #549, opened 2018-07-11
Fix multiple shutdown attempt corrupting state

mihails-strasuns:
`shutdown` is expected to be called only once, which will result in all
tasks being killed eventually. Calling it again does not serve any extra
purpose but can mess with internal task queue states.

### DIFF
--- a/src/ocean/task/Scheduler.d
+++ b/src/ocean/task/Scheduler.d
@@ -260,6 +260,10 @@ final class Scheduler : IScheduler
 
     public void shutdown ( )
     {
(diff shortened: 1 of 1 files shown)


nemanja-boric-sociomantic (approved):

gavin-norman-sociomantic (approved):

gavin-norman-sociomantic:
Was this related to the problems with the ring queue that we were talking about yesterday?

mihails-strasuns:
One of them, though seems like there are more.

Issue #289, opened 2017-10-06
Revisit assert/verify/enforce separation


mihails-strasuns-sociomantic:
https://github.com/sociomantic-tsunami/ocean/issues/212 resulted in a very straightforward migration where all `assert` places got changed to `verify`/`test` without any detailed consideration. This unblocks further D2 build deployment but in the long term we would need something finer tuned.

We should revisit all modules either one by one or by package and check what can be changed back to `assert` and what should really be an `enforce` with proper exception type. Eventually `-release` should be able to remove all non-essential checks.

mathias-lang-sociomantic:
>  This unblocks further D2 build deployment but in the long term we would need something finer tuned.

Does it ? For the dhtnode, don't we need to do the same in swarm and dhtproto ?

gavin-norman-sociomantic:
Yes, we need to do the same process in the other libraries. From ocean's side, though, things are complete.

mihails-strasuns-sociomantic:
Right, I only speak in context of ocean as it is ocean issue :)

leandro-lucarella-sociomantic:
Maybe a global issue in *backend* could help track the global progress (or a GitHub project).

mihails-strasuns-sociomantic:
That is just a bit too globa :). Ocean alone will take uncertain amount of time

PR #550, opened 2018-07-12
Fix broken shutdown if scheduling from event loop context

mihails-strasuns:
With this fix, any such tasks will be ignored allowing for clean exit.
There is a concern of losing app data but it will have to be addressed
in a separate changest, targetting minor or maybe even major branch.

### DIFF
--- a/src/ocean/task/Scheduler.d
+++ b/src/ocean/task/Scheduler.d
@@ -331,6 +331,11 @@ final class Scheduler : IScheduler
     {
         if (this.state == State.Shutdown)
         {
+            // Simply returning here would be generally sufficient to make sure
+            // no new tasks get added after shutdown. However, it is of some
+            // merit to try to kill everything as soon as possible thus
+            // scheduler kills the caller tasks on any attempt to schedule a new
+            // one.
             auto caller_task = Task.getThis();
             if (caller_task !is null)
                 caller_task.kill();
--- a/src/ocean/task/Scheduler.d
+++ b/src/ocean/task/Scheduler.d
@@ -334,6 +334,7 @@ final class Scheduler : IScheduler
             auto caller_task = Task.getThis();
             if (caller_task !is null)
                 caller_task.kill();
+            return;
         }
 
         try


src/ocean/task/Scheduler.d
```diff
@@ -334,6 +334,7 @@ final class Scheduler : IScheduler
             auto caller_task = Task.getThis();
             if (caller_task !is null)
                 caller_task.kill();
+            return;
```
gavin-norman-sociomantic, line 342: Unrelated to this PR, but it might be worth adding a comment as to why the calling task is killed if `schedule` is called during shutdown.

src/ocean/task/Scheduler.d
```diff
@@ -615,6 +615,8 @@ final class Scheduler : IScheduler
     public void processEvents ( )
     {
         auto task = Task.getThis();
+        assert(task !is null);
```
gavin-norman-sociomantic: Why not `verify`?

mihails-strasuns: It will crash anyway even if check is missing

src/ocean/task/Scheduler.d
mihails-strasuns: Will add

mihails-strasuns:
Updated with a comment

src/ocean/task/Scheduler.d
gavin-norman-sociomantic: Mm I don't get what you mean. Aren't these the possible cases in D2 builds?
* `assert`, `task is null`: crash.
* `assert`, `task !is null`: keeps running.
* `verify`, `task is null`: throws.
* `verify`, `task !is null`: keeps running.

mihails-strasuns: Few more options depending on `-release` but essentially yes. But don't forget that it crashes _right now_ if condition is false because there is no check at all. So this exactly matches assert purpose - turn crash / corruption into an easier debuggable crash.

daniel-zullo (commented):
LGTM

daniel-zullo:
Ping. Can we please label this PR as high priority? I need this patch to make a daemon application works properly.

src/ocean/task/Scheduler.d
gavin-norman-sociomantic: But didn't we go through and convert all such cases into `verify`s?

mihails-strasuns-sociomantic: We converted such _existing_ cases for _backwards compatibility_. That doesn't mean all them should really remain as such and definitely doesn't apply for new cases. See also https://github.com/sociomantic-tsunami/ocean/issues/289

mihails-strasuns-sociomantic:
I will make a release as soon as @gavin-norman-sociomantic gives a green light :)

src/ocean/task/Scheduler.d
gavin-norman-sociomantic: Maybe I'm missing something here. Looking at the [assert / verify doc](https://github.com/sociomantic-tsunami/sociomantic/blob/master/Code/D/assert-vs-enforce.md#what-kind-of-error-check-to-use):
>In practice, this means that anything checked by asserts must also be covered at some point in the code by verify, enforce, etc. In this way, the error conditions will still be checked in release builds, but non-release builds can benefit from additional error checks to aid debugging.

Does this apply in this situation? (i.e. is the condition `task !is null` being checked elsewhere?)

mihails-strasuns-sociomantic: Yes, "crash with segfault" is that extra check in this case.

gavin-norman-sociomantic: It sounds like you're saying our guidelines are incorrect.

>Checks for programmer error / basic sanity or safety (e.g. pointer checks, integer expressions) should be performed using ocean's verify

mihails-strasuns-sociomantic: Ugh, yes. This statement is definitely incorrect. Checks for programmer error should always be done with `assert` unless you can't afford it to be removed. So, for example, protecting cast + raw memory access has to be done with `verify` because if you miss it, app will end up with memory corruption and that is very bad. But protecting null pointers with `verify` is almost always a waste, because it will crash the app immediately anyway.

gavin-norman-sociomantic: Well... the whole point of inventing `verify` was to avoid apps crashing due to an `assert` firing in D2.

mihails-strasuns-sociomantic: If that was the case, we have wasted a lot of effort on something fundamentally flawed. You can't avoid crashes no matter how hard you try and must embrace them :) From my PoV goal of `verify` is to protect against memory corruptions and similar damage.

mihails-strasuns-sociomantic: To word it differently - the fact that right now dhtnode can hit an `assert` (now `verify`) condition and survive after is a bug in dhtnode that I totally expect to be fixed some day, not a feature to support.

gavin-norman-sociomantic: >the fact that right now dhtnode can hit an assert (now verify) condition and survive after is a bug in dhtnode that I totally expect to be fixed some day

This is the key. Using `assert`s to notify us of such bugs causes crashes. Using `verify`s to notify us of such bugs (potentially) avoids crashes and allows us to log the bug. From the POV of a dhtnode, the latter is _way_ more desirable than the former.

mihails-strasuns-sociomantic: I disagree. Consider this PR - before my change it would always just segfault on wrong usage, in both D1 and D2. There are no asserts, just plain segfault. The very fact that it was working like that in deployed live apps for a very long time is a direct proof that this is perfectly fine.

This is reality that can't be worked around - no matter what policy you take on `assert`, your app will crash eventually and bugs will have to be fixed. Effort to make it as harmless as possible is meaningful, effort to prevent it altogether is not.

gavin-norman-sociomantic: >The very fact that it was working like that in deployed live apps for a very long time is a direct proof that this is perfectly fine.

It's only fine because it never came up. If it had have come up, an `assert` here would be as bad as a plain segfault, in terms of direct impact to the application.

>Effort to make it as harmless as possible is meaningful, effort to prevent it altogether is not.

This is exactly what I'm suggesting. Throwing is generally less harmful than crashing. Both have equal value in terms of notifying maintainers of bugs.

gavin-norman-sociomantic (approved):

gavin-norman-sociomantic:
Green light.

PR #552, opened 2018-07-17
Fix startEventHandling verify condition

### LINKED ISSUE #528: `startEventHandling` requires epoll instance, it's not optional as it says
```
    public void startEventHandling ( EpollSelectDispatcher epoll )
    {
        verify(epoll !is null);
        verify(this.epoll is null,
            "Either pass epoll to the ctor or startEventHandling, not both");
```

The error message is wrong, because the user _must_ pass epoll to `startEventHandling` (because of the first verify).

mihails-strasuns:
Fixes #528


nemanja-boric-sociomantic (approved):
Doh!

PR #558, opened 2018-07-25
Bugfix: `FullQueryParamSet.parse` did not add remaining elements.

david-eckardt-sociomantic:
If the `query` string passed to `FullQueryParamSet.parse` (inherited from `QueryParamSet`) contained parameters with unknown keys, they were not added to the list of remaining elements.

This was because the super class did not call a potentially overridden method. (I made this mistake ages ago before @leandro-lucarella-sociomantic pointed it out to me.) A second commit changes all unnecessary `super` calls to `this` calls.

### DIFF
--- a/src/ocean/net/util/QueryParams.d
+++ b/src/ocean/net/util/QueryParams.d
@@ -255,7 +255,7 @@ class QueryParamSet: ParamSet
 
         foreach (key, val; query_params.set(query))
         {
-            super.set(key, val);
+            this.set(key, val);
         }
     }
 }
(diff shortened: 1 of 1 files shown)


src/ocean/net/util/QueryParams.d
```diff
@@ -255,7 +255,7 @@ class QueryParamSet: ParamSet
 
         foreach (key, val; query_params.set(query))
         {
-            super.set(key, val);
+            this.set(key, val);
```
gavin-norman-sociomantic, line 258: Nasty.

src/ocean/net/util/QueryParams.d
```diff
@@ -445,3 +445,77 @@ unittest
...
+
+    with (params.remaining_elements[1])
+    {
+        test!("==")(key, "tritt");
+        test!("is")(val, val.init);
+    }
+
+    bool[4] found;
+
+    foreach (key, val; params)
```
gavin-norman-sociomantic, line 474: (Probably unrelated to this PR.) Interesting API, that `opIn_r` doesn't report these extra values as being present, but `opApply` iterates over them.

gavin-norman-sociomantic:
Should be against v3.8.x (the oldest supported branch).

src/ocean/net/util/QueryParams.d
david-eckardt-sociomantic: Good point. IMHO they should not show up via `in` or `foreach`. It seems too risky to change it just for correctness, though, it could break some mission-critical server code. We can fix this if someone actually needs the correct behaviour.

david-eckardt-sociomantic:
Changed base to v3.8.x.

src/ocean/net/util/QueryParams.d
gavin-norman-sociomantic: :+1: 
Maybe something for ocean v5.x.x.

gavin-norman-sociomantic (approved):

src/ocean/net/util/QueryParams.d
gavin-norman-sociomantic: Issue for v5.0.0: https://github.com/sociomantic-tsunami/ocean/issues/559.

PR #598, opened 2018-07-27
Refactor public methods to avoid in contract

daniel-zullo:
`in` contracts containing asserts should only be used on private
methods which are called via a public API including verify/enforce
checks as they will be completely removed from the program when
compiled in `-release` mode.

All checks in `in` contracts have been converted to verify statements
in the method body.

PR #601, opened 2018-07-30
Few missed assert->verify conversions

mihails-strasuns:


daniel-zullo:
@mihails-strasuns the patch does pretty much what it is done in #598, is there any special reason to apply this patch instead?

mihails-strasuns:
I have totally missed that PR somehow, sorry! Will go with yours, it is better :)

daniel-zullo:
cool, thanks!

PR #607, opened 2018-08-06
Fix segfault in AppStatus.num_static_lines when terminal_output is null

don-clugston-sociomantic:
AppStatus.terminal_output could be null.
num_static_lines() needs the same check that displayStaticLines() has, to avoid a segfault.

This is important because when writing to a socket instead of the console, it is common for terminal_output to be null.

### DIFF
--- a/src/ocean/io/console/AppStatus.d
+++ b/src/ocean/io/console/AppStatus.d
@@ -530,7 +530,7 @@ public class AppStatus
     {
         this.resetStaticLines();
 
-        if ( this.is_redirected )
+        if ( this.is_redirected || this.terminal_output is null )
         {
             this.static_lines.length = size;
 


gavin-norman-sociomantic:
Could you rebase onto v4.2.x?

gavin-norman-sociomantic:
Unless the fix is urgent, leaving for @nemanja-boric-sociomantic to confirm.

don-clugston-sociomantic:
Rebased onto 4.2.x.   (This is high priority though, it prevents daemonisation of any app which uses a split display).

gavin-norman-sociomantic:
I'll review this then.

gavin-norman-sociomantic (approved):
LGTM

PR #608, opened 2018-08-06
Add helpers for auto-relinquishing shared resources used by a context

gavin-norman-sociomantic:
Straight outta swarm.

### DIFF
--- a/src/ocean/util/container/VoidBufferAsArrayOf.d
+++ b/src/ocean/util/container/VoidBufferAsArrayOf.d
@@ -1,0 +1,220 @@
+/*******************************************************************************
+
+    Struct template which wraps a void[] with an API allowing it to be safely
+    used as an array of another type.
+
+    It is, of course, possible to simply cast a void[] to another type of array
+    and use it directly. However, care must be taken when casting to an array
+    type with a different element size. Experience has shown that this is likely
+    to hit undefined behaviour. For example, casting the array then sizing it
+    has been observed to cause segfaults, e.g.:
+
+    ---
+        void[]* void_array; // acquired from somewhere
+
+        struct S { int i; hash_t h; }
+        auto s_array = cast(S[]*)void_array;
+        s_array.length = 23;
+    ---
+
+    The exact reason for the segfaults is not known, but this usage appears to
+    lead to corruption of internal GC data (possibly type metadata associated
+    with the array's pointer).
+
+    Sizing the array first, then casting is fine, e.g.:
+
+    ---
+        void[]* void_array; // acquired from somewhere
+
+        struct S { int i; hash_t h; }
+        (*void_array).length = 23 * S.sizeof;
+        auto s_array = cast(S[])*void_array;
+    ---
+
+    The helper VoidBufferAsArrayOf simplifies this procedure and removes the
+    risk of undefined behaviour by always handling the void[] as a void[]
+    internally.
+
+    Copyright:
+        Copyright (c) 2017-2018 dunnhumby Germany GmbH.
+        All rights reserved
+
+    License:
+        Boost Software License Version 1.0. See LICENSE.txt for details.
+
+*******************************************************************************/
+
+module ocean.util.container.VoidBufferAsArrayOf;
+
+import ocean.transition;
+
+/*******************************************************************************
+
+    Struct template which wraps a void[] with an API allowing it to be safely
+    used as an array of another type.
+
+    Params:
+        T = element type of array which the API mimics
+
+*******************************************************************************/
+
+public struct VoidBufferAsArrayOf ( T )
+{
+    // T == void is not only pointless but is also invalid: it's illegal to pass
+    // a void argument to a function (e.g. opCatAssign).
+    static assert(!is(T == void));
+
+    /// Pointer to the underlying void buffer. Must be set before use by struct
+    /// construction.
+    private void[]* buffer;
+
+    // The length of the buffer must always be an even multiple of T.sizeof.
+    invariant ( )
+    {
+        assert(this.buffer.length % T.sizeof == 0);
+    }
+
+    /***************************************************************************
+
+        Returns:
+            a slice of this array
+
+    ***************************************************************************/
+
+    public T[] array ( )
+    {
+        return cast(T[])(*this.buffer);
+    }
+
+    /***************************************************************************
+
+        Returns:
+            the number of T elements in the array
+
+    ***************************************************************************/
+
+    public size_t length ( )
+    {
+        return this.buffer.length / T.sizeof;
+    }
+
+    /***************************************************************************
+
+        Sets the length of the array.
+
+        Params:
+            len = new length of the array, in terms of the number of T elements
+
+    ***************************************************************************/
+
+    public void length ( size_t len )
+    {
+        this.buffer.length = len * T.sizeof;
+        enableStomping(*this.buffer);
+    }
+
+    /***************************************************************************
+
+        Appends an array of elements.
+
+        Note that mutable copies of appended elements are made internally, but
+        to access them from the outside, the constness of T applies.
+
+        but
+        are inaccessible from the outside.
+
+        Params:
+            arr = elements to append
+
+        Returns:
+            a slice of this array, now with the specified elements appended
+
+    ***************************************************************************/
+
+    public T[] opCatAssign ( in T[] arr )
+    {
+        return cast(T[])((*this.buffer) ~= cast(void[])arr);
+    }
+
+    /***************************************************************************
+
+        Appends an element.
+
+        Note that a mutable copy of the appended element is made internally, but
+        to access it from the outside, the constness of T applies.
+
+        Params:
+            element = element to append
+
+        Returns:
+            a slice of this array, now with the specified element appended
+
+    ***************************************************************************/
+
+    public T[] opCatAssign ( in T element )
+    {
+        return this.opCatAssign((&element)[0 .. 1]);
+    }
+}
+
+///
+unittest
+{
+    // Backing array.
+    void[] backing;
+
+    // Wrap the backing array for use as an S[].
+    struct S
+    {
+        ubyte b;
+        hash_t h;
+    }
+
+    auto s_array = VoidBufferAsArrayOf!(S)(&backing);
+
+    // Append some elements.
+    s_array ~= S();
+    s_array ~= [S(), S(), S()];
+
+    // Resize the array.
+    s_array.length = 2;
+
+    // Iterate over the elements.
+    foreach ( e; s_array.array() ) { }
+}
+
+version ( UnitTest )
+{
+    import ocean.core.Test;
+
+    align ( 1 ) struct S
+    {
+        ubyte b;
+        hash_t h;
+    }
+}
+
+unittest
+{
+    void[] backing;
+
+    auto s_array = VoidBufferAsArrayOf!(S)(&backing);
+
+    test!("==")(s_array.length, 0);
+    test!("==")(s_array.buffer.length, 0);
+
+    s_array ~= S(0, 0);
+    test!("==")(s_array.array(), [S(0, 0)]);
+    test(s_array.length == 1);
+    test(s_array.buffer.length == S.sizeof);
+
+    s_array ~= [S(1, 1), S(2, 2), S(3, 3)];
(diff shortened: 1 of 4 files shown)


src/ocean/util/container/VoidBufferAsArrayOf.d
```diff
@@ -0,0 +1,220 @@
+/*******************************************************************************
```
gavin-norman-sociomantic, line 1: Not sure if `ocean.util.container` is a good location for this. Other buffery things live there, so put there by default.

src/ocean/util/container/VoidBufferAsArrayOf.d
```diff
@@ -0,0 +1,220 @@
...
+
+*******************************************************************************/
+
+public struct VoidBufferAsArrayOf ( T )
+{
+    // T == void is not only pointless but is also invalid: it's illegal to pass
+    // a void argument to a function (e.g. opCatAssign).
+    static assert(!is(T == void));
+
+    /// Pointer to the unerlying void buffer. Must be set before use by struct
```
joseph-wakeling-sociomantic: `underlying` ;-)

src/ocean/util/container/pool/AcquiredResources.d
```diff
@@ -0,0 +1,813 @@
...
+        T = type of resource
+
+*******************************************************************************/
+
+public struct Acquired ( T )
+{
+    import ocean.util.container.VoidBufferAsArrayOf;
+
+    /// Type of a new resource.
+    static if ( is(typeof({T* t = new T;})) )
```
joseph-wakeling-sociomantic: Might be helpful to have a small comment to explain this `static if/else` choice ... ?  I presume it's meant to distinguish between cases where `T` is a value type and where `T` is a reference type?

gavin-norman-sociomantic:
Updated with Joe's suggestions.

gavin-norman-sociomantic:
I investigated adding a template to wrap a bunch of freelists and acquirers, based on a type tuple. However, it's complicated, as the `Acquired.acquire` method requires a `lazy Elem`, which would be awkward to specify. Maybe something for the future.

src/ocean/util/container/VoidBufferAsArrayOf.d
mihails-strasuns: Makes sense to me.

mihails-strasuns (commented):
This may need more simple examples in release notes. I can imagine that primary use case is to simply have bunch of different type buffers backed up by single `void[]` freelist. How does one achieve it?

gavin-norman-sociomantic:
:+1: Makes sense. I'll add one.

gavin-norman-sociomantic:
>I can imagine that primary use case is to simply have bunch of different type buffers backed up by single void[] freelist. How does one achieve it?

https://github.com/sociomantic-tsunami/ocean/pull/608/files#diff-60bf769b1fa6c58721c779b667ed7aa1R369 gives the basic idea of this. It's ~70 lines, though, so I don't think it makes sense to copy into the release notes. I've added a note about checking the usage examples in the module.

gavin-norman-sociomantic:
Also added more usage info to the module header.

mihails-strasuns:
Hm, it makes me wonder if this is of much use to apps in its current form. In 70 lines of code one can implement own wrapper for `FreeList` from scratch and I am not sure how many apps need resource management as convoluted as in swarm.

Don't mind merging this as-is but curious about your opinion - would it make sense to enhance this with something more simple or that would be better suited as a separate utility?

gavin-norman-sociomantic:
Yeah, I wonder as well. We should get some of the people who were keen on having this in ocean to look at the PR and see if it would help them.

gavin-norman-sociomantic:
(I suppose the actual definition of the shared resources class is only about 40 lines, in that example. The other lines are an example of using it in an execution context.)

>We should get some of the people who were keen on having this in ocean to look at the PR and see if it would help them.

@joseph-wakeling-sociomantic was one such person. Joe, the question is whether code like [this example](https://github.com/sociomantic-tsunami/ocean/blob/d0d49621b752c6198e421d62e5c7bf90eff5b0b9/src/ocean/util/container/pool/AcquiredResources.d#L392-L455) would be helpful to you, or whether it'd be easier to implement your own wrapper around `FreeList` in your app?

joseph-wakeling-sociomantic:
I've got to be honest and say this all looks overcomplicated and intrusive to me.  It looks like it works well for a case where (say) one is grabbing a bunch of temporary resources at the start of a `Task` context, or maybe where one has a resource supplier as part of a context object.  It's possibly not so useful for the use-cases I have in mind where the code is generally a class that encapsulates some calculations requiring temporary buffers under the hood.

What would probably be more useful for my case is a _thin_ wrapper around `FreeList` that allows for the option to get and recycle buffers with different element types.  Typical use case would be something like:

```D
class MyMagicClass
{
    // privately encapsulated supplier of buffers
    private MagicBufferSupplier buffer_pool;

    this () { this.buffer_pool = new MagicBufferSupplier; }

    // templated method, so this can be invoked for different data types
    // and Just Work with one underlying buffer pool
    void foo (T) (...)
    {
        auto buf = this.buffer_pool.get!(T)(...);
        scope (exit) this.buffer_pool.recycle(buf);

        // ... do stuff with the buffer
    }
}
```

... makes sense?

joseph-wakeling-sociomantic:
Of course, note that part of the reason behind the above is a wish to encapsulate the buffer supplier.  This makes sense where e.g. the code written is not as part of an app but as a standalone library component (which is a long-term design goal for the particular code I have in mind).  I imagine that for app components it would be more reasonable to have the resource supplier as a global app resource.

gavin-norman-sociomantic:
Yeah, what you have in mind is a bit different. Could probably be implemented on top of what's already in this PR, though.

joseph-wakeling-sociomantic:
Can you talk through a bit how you see that working?  It's not obvious to me how it could operate without exposing app or library functionality to a large chunk of global mutable state.  (But maybe you're not seeing that as a problem?)

gavin-norman-sociomantic:
Perhaps I'm missing something. There's nothing inherent in the code added by this PR that means that the shared resources must be global. You could, for example, instantiate a shared resources pool inside `MyMagicClass`, to use your example.

joseph-wakeling-sociomantic:
I guess so.  It just feels like the amount of boilerplate to do so is quite large (but maybe I'm missing something too).  I'd assumed that the typical use-case for it would be as one resource for an entire app.

gavin-norman-sociomantic:
Agreed. I just wanted to clarify that we're on the same page re: it not being intrinsically global.

So the main point of this PR is to provide building blocks for:
* A shared resources pool. (Intended to be owned at the app level, but that's not mandatory.)
* A tracker for resources acquired by a task (or other context), with the ability to automatically relinquish them back to the shared pool when the task ends.

Are either of those points applicable to your use case? (I'm getting the feeling not.)

joseph-wakeling-sociomantic:
> Are either of those points applicable to your use case?

Not really, no.  I think the use-case is much more of small encapsulated buffer pools for specific pieces of functionality, where relinquishing the resources is not tied to `Task` completion but (typically) to scope exit.

The one thing that _is_ needed for my use-case is the ability to have one pool which can serve multiple different types of buffer from the caller's point of view.

gavin-norman-sociomantic:
>where relinquishing the resources is not tied to Task completion but (typically) to scope exit.

The exact scope at which the resources are relinquished isn't important. (My mention of `Task`s was just an example. In fact, in *proto, we relinquish on `scope(exit)`, as you say.)

joseph-wakeling-sociomantic:
OK.  So sounds like the second point is a common area of interest.  From a usage point of view it's convenient if the code is not boilerplate-y, i.e. it doesn't require up-front declaration of the different supported types (cf. my pseudo-code example above).

gavin-norman-sociomantic:
Yep. I could add a simplistic buffer getter built on top of the existing code in this PR, say.

gavin-norman-sociomantic:
Usage would look like this:
```D
class MyMagicClass
{
    private MagicBufferSupplier buffer_pool;

    this () { this.buffer_pool = new MagicBufferSupplier; }

    // templated method, so this can be invoked for different data types
    // and Just Work with one underlying buffer pool
    void foo (T) (...)
    {
        scope buffers = this.buffer_pool.new AcquiredResources;

        // get one or more buffers in this context.
        // they will be automatically relinquished on scope(exit)
        buffers.get!(T)(...);

        // ... do stuff with the acquired buffers
    }
}
```

joseph-wakeling-sociomantic:
Seems reasonable.  A bit of a shame the boilerplate-y `scope buffers = ...` line is needed, but that's no big deal.

Will the individual buffers be recycled when the scope of the `AcquiredResources` instance ends, or when their own scope ends?  For the use-case I have there is some convenience in being able to recycle individual buffers at the end of their scope.

gavin-norman-sociomantic:
>Seems reasonable. A bit of a shame the boilerplate-y scope buffers = ... line is needed, but that's no big deal.

But you don't have to manually relinquish the buffers. So that line from your version is removed ;)

>Will the individual buffers be recycled when the scope of the AcquiredResources instance ends, or when their own scope ends?

When `AcquiredResources`' scope exits. If you want sub-scopes, you can add them manually with `{}`, as usual.

joseph-wakeling-sociomantic:
OK, all good then.  There might be some value in allowing some generic "get and recycle a buffer with element type `T`" functionality, versus a helper class like `AcquiredResources` that manages the scoping on your behalf -- but that's not a really big deal.  This seems like a nice addition to ocean.

gavin-norman-sociomantic:
Yeah, it sounds to me like what you really want, Joe, is a simple wrapper around a free-list that allows one to acquire and recycle arbitrarily typed buffers (without any scoped tracking or auto-relinquishing mechanism). This is a different thing, really.

joseph-wakeling-sociomantic:
Agreed.  And that is what I noted right at the beginning of this discussion ;-)
https://github.com/sociomantic-tsunami/ocean/pull/608#issuecomment-415359907

gavin-norman-sociomantic:
>We should get some of the people who were keen on having this in ocean to look at the PR and see if it would help them.

So it seems Joe isn't really the intended audience ;) We should find someone who is. (Or perhaps the whole idea of this PR is specific to swarm and rumours of people needing such support in ocean were ill-founded!)

joseph-wakeling-sociomantic:
Personally I think it's a useful feature for ocean, it's just not appropriate to the one particular use-case that sparked the discussion.

I can almost certainly find other contexts in which the use-case is valid.  And it's not like it's _invalid_ for my actual use-case; it's just that other tradeoffs mean that it may be preferable to approach that from a slightly different point of view.

gavin-norman-sociomantic:
I can confirm that two applications use exactly the pattern that this PR implements. (Albeit with very simplistic use cases, where only buffers of a single type are acquired.)

nemanja-boric-sociomantic:
There are also more advanced requirements - such as acquiring the buffers based on the anticipated grow size, and having a rudimentary support in the ocean is a good step forward towards implementing more advanced memory management patterns.

joseph-wakeling-sociomantic:
Yup. If you look at my use-case (which involves creating buffers to interact with C number-crunching libraries), a need is to be able to specify a _minimum_ length for the buffer.  In this way we avoid any need for costly `enableStomping`/`assumeSafeAppend` calls.

joseph-wakeling-sociomantic:
If we had those sort of extra helper features, it would make a lot more sense to use `AcquiredResources` etc.

mihails-strasuns:
@gavin-norman-sociomantic can you sum things up regarding what to do here? :)

gavin-norman-sociomantic:
Summary:
* The PR as presented is heavily used in swarm / *proto.
* Other real-time server apps in our system could use the PR as presented, but only its basic features (acquiring and relinquishing buffers from/to a pool).
* What Joe wants is something a bit different. (But could be implemented on top of what's in this PR.)

mihails-strasuns:
== let's merge and make another PR for Joe utils later?

gavin-norman-sociomantic:
I feel that would be reasonable.

joseph-wakeling-sociomantic:
I'm fine with that.

mihails-strasuns (approved):

gavin-norman-sociomantic:
@joseph-wakeling-sociomantic care to create an issue explaining what features you'd like?

joseph-wakeling-sociomantic:
Sure thing.  I'll follow up later today.

PR #611, opened 2018-08-13
Flush BufferedOutput on close

### LINKED ISSUE #578: Buffered output not flush()ed before close()
@leandro-lucarella-sociomantic :

This is a pretty nasty bug:

``` d
import ocean.io.Stdout;
import ocean.io.stream.Buffered;
import ocean.io.device.File;

void main()
{
    auto output = new BufferedOutput(new File("test.txt", File.WriteCreate));
    output("hello\n");
    version (Flush) output.flush();
    output.close();
}
```

When compiled without `-version=Flush` it produces an empty file! I don't see anything in the docs referring to this behaviour, and even if it is, is completely wrong and unintuitive!

mihails-strasuns:
Fixes #578

### DIFF
--- a/src/ocean/io/stream/Buffered.d
+++ b/src/ocean/io/stream/Buffered.d
@@ -1168,6 +1168,18 @@ public class BufferedOutput : OutputFilter, OutputBuffer
 
     /***************************************************************************
 
+        Flushes the buffer and closes the stream.
+
+    ***************************************************************************/
+
+    final override void close ( )
+    {
+        this.flush();
+        super.close();
+    }
+
+    /***************************************************************************
+
         Drain buffer content to the specific conduit.
 
         Write as much of the buffer that the associated conduit can consume.
--- a/integrationtest/buffered/main.d
+++ b/integrationtest/buffered/main.d
@@ -1,0 +1,25 @@
+module integrationtest.buffered.main;
+
+import ocean.io.stream.Buffered;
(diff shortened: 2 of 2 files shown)


src/ocean/io/stream/Buffered.d
```diff
@@ -1166,6 +1166,18 @@ public class BufferedOutput : OutputFilter, OutputBuffer
         return this;
     }
 
+    /***************************************************************************
+
+        Flushes the buffer and closes the stream.
+
+    ***************************************************************************/
+
+    final override void close ( )
```
gavin-norman-sociomantic, line 1175: Would be nice to add a unittest for this.

mihails-strasuns: That would be inherently flakey because you don't control buffering afaics.

gavin-norman-sociomantic: Hmm, is there some way that the buffer would be flushed magically behind the scenes?

I was thinking of something like:
1. Open file with buffered output.
2. Write something smaller than the buffer size.
3. Close the file.
4. Check that the file content on disk is correct. (i.e. the buffer was flushed.)

mihails-strasuns: Yeah, makes sense. Added.

gavin-norman-sociomantic (approved):

PR #612, opened 2018-08-14
[WIP] Add convenience method to schedule a delegate

gavin-norman-sociomantic:


gavin-norman-sociomantic:
Does something like this seem reasonable? Just trying to simplify scheduler usage in scripts / test code.

mihails-strasuns:
I don't think it is a good idea. Right now it roughly same amount of code using anonymous class feature, for example https://github.com/sociomantic-tsunami/ocean/blob/e53ac93fbf3bfa9b2dceec1a2b6dc4a0ec7f78b2/src/ocean/util/app/ext/TaskExt.d#L145-L153

Once you allow raw delegates, you have to worry both for:

1) Not making a closure by an accident
2) Ensuring that delegate context does not outlive the task execution (which can happen at arbitrary time)

Thus I am afraid providing such shortcut encourages one to not think about important memory concern.

mihails-strasuns:
In simple scripts this should be simple enough:
```D
theScheduler.schedule(new class Task {
    override void run ( ) {
        // would-be delegate body here
    }
});
```

gavin-norman-sociomantic:
Yeah, I see what you mean. Although being able to schedule a delegate is only slightly less code than making an anonymous class, it also has the advantage to the user of not having to know what the `Task` class requires.

mihails-strasuns:
For me benefit of anonymous class is that it contains explicit `new` thus no one can possibly miss GC allocation by writing this (or worse, corrupted stack pointer).

> it also has the advantage to the user of not having to know what the Task class requires

Do we actually want that? :)

gavin-norman-sociomantic:
I was thinking about this some more, and I think maybe actually what we need in simple scripts is something even more dumb:
```D
void main ( )
{
  Scheduler.runInTask({
    // do something
  }); // automatically initialises the scheduler, if not already done
}
```

mihails-strasuns:
Initialization has to stay explicit to actually allocate singleton object (also there are no sane defaults for scripts imho). But we can do something like `theScheduler.run({})` which adds one task and starts event loop. Probably even two of them - also `theScheduler.runAndQuit({})` to shutdown scheduler after root task finishes.

gavin-norman-sociomantic:
>Initialization has to stay explicit to actually allocate singleton object (also there are no sane defaults for scripts imho).

Huh, I always just use `initScheduler(Scheduler.Configuration.init);` Seems to work fine.

>But we can do something like theScheduler.run({}) which adds one task and starts event loop. Probably even two of them - also theScheduler.runAndQuit({}) to shutdown scheduler after root task finishes.

Right, exactly, I meant to say that it should run the event loop too.

nemanja-boric-sociomantic:
> Do we actually want that? :)

I think this is good enough for writing the simple scripts that would be using swarm, for example. Writing this in an application code should be a no-go, but for the simple one-run scripts this is acceptable IMHO.

gavin-norman-sociomantic:
Exactly. Being able to write code like this would be ideal:
```D
void main ( )
{
  Scheduler.runAndExit({
    Dht.connect(10_000); // Task-blocking connect & handshake
    Dht.put("channel", 23, "value"); // Task-blocking Put request
  });
}
```

gavin-norman-sociomantic:
>Initialization has to stay explicit to actually allocate singleton object (also there are no sane defaults for scripts imho).

Interested in your reasoning for this statement, @mihails-strasuns . I always just use `initScheduler(Scheduler.Configuration.init);` in scripts, and this seems to work fine.

gavin-norman-sociomantic:
Replaced by #659.

PR #593, opened 2018-08-15
`ocean.text.Arguments.getInt()` does not extract specified int type

mihails-strasuns:
@joseph-wakeling-sociomantic : 

The `getInt` method requests a template parameter for the int type, but that type is not used when attempting to extract the integer value from the command line argument: instead, a `long` is always extracted.

Ocean's `toInteger` method should probably be used internally instead.

### DIFF
--- a/src/ocean/text/Arguments.d
+++ b/src/ocean/text/Arguments.d
@@ -611,7 +611,7 @@ import ocean.io.stream.Format : FormatOutput;
 import ocean.math.Math;
 import ocean.text.Util;
 import ocean.text.convert.Formatter;
-import ocean.text.convert.Integer_tango;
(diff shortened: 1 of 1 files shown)


mihails-strasuns:
Good catch, I didn't think about unsigned values.

mihails-strasuns:
Fix attached

joseph-wakeling-sociomantic:
Commit title should probably mention `ulong` not `uint` ... ?

Is it worth adding a unittest to validate that `getInt` now works with large `ulong` values?

mihails-strasuns:
Updated

joseph-wakeling-sociomantic (approved):
LGTM, thanks!

PR #388, opened 2018-08-15
Validation of overridden config value 

mihails-strasuns:
It is possible to override a config value which doesn't exist in the config file, e.g. 

`./myapp -O non.existent=value`

Is this intended behavior or should it perhaps be handled in [ConfigExt.validateArgs()](https://github.com/sociomantic-tsunami/ocean/blob/53aa6a8970366c4d7d23e362b026540825bc6488/src/ocean/util/app/ext/ConfigExt.d#L241)?

### DIFF
--- a/src/ocean/util/app/ext/ConfigExt.d
+++ b/src/ocean/util/app/ext/ConfigExt.d
@@ -419,6 +419,9 @@ class ConfigExt : IApplicationExtension, IArgumentsExtExtension
         if (key.length == 0)
             return "Empty key for config option override: " ~ opt;
 
+        if (!this.config.exists(category, key))
+            return "Attempt to override non-existent config entry";
+
         return null;
     }
 }
@@ -440,6 +443,7 @@ version (UnitTest)
 unittest
 {
     auto ext = new ConfigExt;
+    ext.config.parseString("[cat]\nkey = oops");
 
     // Errors are compared only with startsWith(), not the whole error
     void testParser ( istring opt, istring exp_cat, istring exp_key,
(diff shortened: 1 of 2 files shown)


mihails-strasuns:
Fix attached. Usual concern of bug fix vs breaking change though.

andreas-bok-sociomantic:
> Merging #388 into v4.2.x will increase coverage by <.01%.
The diff coverage is 100%.

Impressive :)

src/ocean/util/app/ext/ConfigExt.d
```diff
@@ -419,6 +419,9 @@ class ConfigExt : IApplicationExtension, IArgumentsExtExtension
         if (key.length == 0)
             return "Empty key for config option override: " ~ opt;
 
+        if (!this.config.exists(category, key))
+            return "Attempt to override non-existent config entry";
```
Geod24, line 423: Does this also error if loose parsing is enabled ? IMO it should not, but I don't see any code to check for it.

mihails-strasuns: Yes, it will. Based on our last discussion with @gavin-norman-sociomantic & co intent is to get rid of loose parsing completely, so it is hardly a problem on its own.

mihails-strasuns:
Rebased to v5.x.x

mihails-strasuns:
Ping @gavin-norman-sociomantic @nemanja-boric-sociomantic

src/ocean/util/app/ext/ConfigExt.d
gavin-norman-sociomantic: My feeling is that arg overriding of config values is a separate question to how the config parser works (i.e. strict or "loose").

gavin-norman-sociomantic (approved):
I'll leave 24 hours for discussion on the loose parsing point. If no further comments arise, I'll merge.

PR #616, opened 2018-08-17
Add ebtree based TreeMap data structure

nemanja-boric-sociomantic:
This adds a TreeMap, which is an ebtree based map with malloc-allocated
nodes, +with a very low overhead for the entire map.

### DIFF
--- a/src/ocean/util/container/map/TreeMap.d
+++ b/src/ocean/util/container/map/TreeMap.d
@@ -1,0 +1,485 @@
+/*******************************************************************************
+
+    EBtree based map.
+
+    Compared to a standard HashMap, a tree-based map is useful in situations
+    where the number of elements in the map is hard to predict. The overhead
+    of an ebtree is very low compared to a hash map, as the overhead of a hash
+    map is the whole bucket set array, while for an ebtree is only a little
+    struct. Additionally, any tree modification is permitted during the
+    iteration.
+
+    The map internally allocates internal nodes which contain elements
+    using `malloc`, so they are invisible to the GC.
+
+    Copyright: Copyright (c) 2016-2018 dunnhumby Germany GmbH. All rights reserved
+
+    License:
+        Boost Software License Version 1.0. See LICENSE.txt for details.
+
+*******************************************************************************/
+
+module ocean.util.container.map.TreeMap;
+
+import ocean.util.container.ebtree.c.eb64tree;
+import ocean.util.container.ebtree.c.ebtree;
+
+/*******************************************************************************
+
+    Params:
+        T = user value type to store
+
+*******************************************************************************/
+
+struct TreeMap ( T )
+{
+    import ocean.transition;
+
+    /***************************************************************************
+
+        Internal node structure.
+
+    ***************************************************************************/
+
+    private struct Node
+    {
+        /***********************************************************************
+
+            Internal ebtree node
+
+        ***********************************************************************/
+
+        eb64_node node;
+
+        /**********************************************************************
+
+            User's value
+
+        **********************************************************************/
+
+        T value;
+    }
+
+    /***************************************************************************
+
+        The `malloc` allocator.
+
+    ***************************************************************************/
+
+    private static MallocAllocator allocator;
+
+    /***************************************************************************
+
+        The ebtree root.
+
+    ***************************************************************************/
+
+    private eb_root root = empty_unique_ebtroot;
+
+    /***************************************************************************
+
+        Returns:
+            true if the map is empty or false if it contains elements.
+
+    ***************************************************************************/
+
+    public bool is_empty ( ) // const
+    {
+        return this.root.is_empty();
+    }
+
+    /***************************************************************************
+
+        Reinitialises the internal tree struct.
+
+        This method should be called only if it is known that the tree is empty,
+        otherwise pointers to malloc-allocated nodes are lost so they cannot be
+        deallocated.
+
+    ***************************************************************************/
+
+    public void reinit ( )
+    {
+        this.root = eb_root.init;
+        this.root.unique = true;
+    }
+
+    /***************************************************************************
+
+        Adds a new node to the map for key or obtains the existing if already in
+        the map.
+
+        Params:
+            key   = the key to add a new node for or obtain the existing one
+            added = outputs true a new node was added or false if key was found
+                    in the map so that the existing node is returned
+
+        Returns:
+            the node associated with key. If added is true then it was newly
+            created and added, otherwise it is the already existing node in the
+            tree.
+
+    ***************************************************************************/
+
+    public T* put ( ulong key, out bool added )
+    {
+        return &(this.put_(key, added).value);
+    }
+
+    /***********************************************************************
+
+        Looks up the value for key in the map.
+
+        Params:
+            key = the key to look up node for
+
+        Returns:
+            the value associated with key or null if not found.
+
+    ***********************************************************************/
+
+    private T* opIn_r ( ulong key )
+    {
+        return &((cast(Node*)eb64_lookup(&this.root, key)).value);
+    }
+
+    /***********************************************************************
+
+        Obtains the first value in the map.
+
+        Returns:
+            the first value or `null` if the map is empty.
+
+    ***********************************************************************/
+
+    public T* first ()
+    {
+        return this.getBoundary!(true)();
+    }
+
+    /***********************************************************************
+
+        Obtains the last value in the map.
+
+        Returns:
+            the last value or `null` if the map is empty.
+
+    ***********************************************************************/
+
+    public T* last ()
+    {
+        return this.getBoundary!(false)();
+    }
+
+    /***********************************************************************
+
+        foreach iterator over nodes in the tree. Any tree modification is
+        permitted during iteration.
+
+    ***********************************************************************/
+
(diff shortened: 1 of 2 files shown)


src/ocean/util/container/TreeMap.d
```diff
@@ -0,0 +1,599 @@
...
+    By default the map uses plain `eb64_node` nodes.
+    To attach values to an ebtree node, create a struct, say, `Node`, whose
+    first field is of type `eb64_node`, followed by any number of other fields,
+    and pass this struct as the `TreeMap` template parameter.
+      - Do not manipulate the first `eb64_node` field (or the plain node if not
+        passing an own struct). Reading may be useful as it contains the key.
+      - Instances of this struct handed out by `TreeMap` are malloc-allocated so
+        they are invisible to the GC.
+
+    We are using an ebtree with malloc-allocated nodes, rather than a hash map,
```
gavin-norman-sociomantic: Suggest rewording this to:
"
Compared to a standard `HashMap`, a tree-based map is useful in situations where the number of elements in the map is hard to predict. The overhead of an ebtree is very low compared to a hash map, as the overhead of a hash map is the whole bucket set array, while for an ebtree is only a little struct. 
"

src/ocean/util/container/TreeMap.d
```diff
@@ -0,0 +1,599 @@
...
+    while for an ebtree is only a little struct.
+
+    Copyright: Copyright (c) 2016-2018 dunnhumby Germany GmbH. All rights reserved
+
+    License:
+        Boost Software License Version 1.0. See LICENSE.txt for details.
+
+*******************************************************************************/
+
+module ocean.util.container.TreeMap;
```
gavin-norman-sociomantic: I wonder about putting it in `util.container.map`.

src/ocean/util/container/TreeMap.d
```diff
@@ -0,0 +1,599 @@
...
+
+/******************************************************************************/
```
gavin-norman-sociomantic: Could get rid of this weird line, while we're at it ;)

src/ocean/util/container/TreeMap.d
```diff
@@ -0,0 +1,599 @@
...
+
+import ocean.util.container.ebtree.c.eb64tree;
+import ocean.util.container.ebtree.c.ebtree;
+
+/*******************************************************************************
+
+    Template params:
```
gavin-norman-sociomantic: `Params:`

src/ocean/util/container/TreeMap.d
```diff
@@ -0,0 +1,599 @@
...
+
+            bool added;
+            MyClass obj  = map.put(12345, added, new MyClass);
+            MyClass obj2 = map.getUserElement(12345);
+            map.remove(obj);
+        ---
+
+    ***************************************************************************/
+
+    static if (is(typeof(Node.user_element_with_treemap_backlink) UserElement))
```
gavin-norman-sociomantic: Suggest removing this alternative API. It's confusing and doesn't seem to really add much.

src/ocean/util/container/TreeMap.d
```diff
@@ -0,0 +1,599 @@
...
+        public static void remove ( ref Node node )
+        {
+            static if (is(Node == eb64_node))
+                eb64_delete(&node);
+            else
+                eb64_delete(&node.tupleof[0]);
+
+            allocator.deallocate(&node);
+        }
+    }
```
gavin-norman-sociomantic: Suggest also adding `public bool remove ( ulong key )`.

gavin-norman-sociomantic (changes requested):

nemanja-boric-sociomantic:
Updated with all the points above addressed.

src/ocean/util/container/map/TreeMap.d
```diff
@@ -0,0 +1,352 @@
...
+        else
+            alias eb64_last eb64_function;
+
+        return cast(Node*)eb64_function(&this.root);
+    }
+
+    /***********************************************************************
+
+        foreach iterator over nodes in the tree. Any tree modification is
+        permitted during iteration.
```
gavin-norman-sociomantic, line 177: This is actually a big advantage of this container, too. Worth mentioning in the relnotes / module header.

src/ocean/util/container/map/TreeMap.d
```diff
@@ -0,0 +1,352 @@
...
+            }
+            else
+            {
+                free(node);
+            }
+        }
+    }
+}
+
+/*******************************************************************************
```
gavin-norman-sociomantic: Annoying ocean comment: a usage example and unittest would be nice! :grimacing:

nemanja-boric-sociomantic: Sigh... ;-)

nemanja-boric-sociomantic: I think we cut out the wrong API. As this stands, this map is just a map of `id`s, since AFAICS `eb64_node` (to which we get the pointer from `Put`) can't hold any user defined value. We need the "other" API that supports any arbitrary data type, as long as they have `eb_node` as a first element.

nemanja-boric-sociomantic: This having said, I'm not convinced anymore this is a good fit for Ocean :-/.

nemanja-boric-sociomantic:
Updated with the API based around user defined type and few extras (instead of `getBoundary` we now have `first` and `last` methods, `opApply` gives both key and the value, etc). Example unittest is added as well.

src/ocean/util/container/map/TreeMap.d
```diff
@@ -0,0 +1,488 @@
...
+
+    EBtree based map.
+
+    By default the map uses plain `eb64_node` nodes.
```
gavin-norman-sociomantic: The module doc needs to be updated.

src/ocean/util/container/map/TreeMap.d
```diff
@@ -0,0 +1,488 @@
...
+    }
+
+    /***********************************************************************
+
+        foreach iterator over nodes in the tree. Any tree modification is
+        permitted during iteration.
+
+    ***********************************************************************/
+
+    public int opApply ( int delegate ( ref ulong key, ref T* node ) dg )
```
gavin-norman-sociomantic: Suggest changing `node` -> `value`.

src/ocean/util/container/map/TreeMap.d
```diff
@@ -0,0 +1,488 @@
...
+
+        return stop;
+    }
+
+
+    /***********************************************************************
+
+        Removes node from the map and deallocates it.
+
+        DO NOT USE node from this point!
```
gavin-norman-sociomantic: I guess this comment no longer applies.

src/ocean/util/container/map/TreeMap.d
```diff
@@ -0,0 +1,488 @@
...
+            eb_node = next;
+        }
+
+        return stop;
+    }
+
+
+    /***********************************************************************
+
+        Removes node from the map and deallocates it.
```
gavin-norman-sociomantic: Needs updating.

src/ocean/util/container/map/TreeMap.d
```diff
@@ -0,0 +1,488 @@
...
+    }
+}
+
+version (UnitTest)
+{
+    import ocean.core.Test;
+}
+
+///
+unittest {
```
gavin-norman-sociomantic: Unconventional style.

gavin-norman-sociomantic (commented):
Yeah, this looks much better now!

src/ocean/util/container/map/TreeMap.d
nemanja-boric-sociomantic: Thanks, I've missed this one.

src/ocean/util/container/map/TreeMap.d
nemanja-boric-sociomantic: Yeah. It should be "DO NOT USE any previous pointers to the accompanying value from this point!" or something similar.

nemanja-boric-sociomantic:
Updated with all comments addressed.

src/ocean/util/container/map/TreeMap.d
```diff
@@ -0,0 +1,484 @@
+/*******************************************************************************
+
+    EBtree based map.
+
+    Compared to a standard HashMap, a tree-based map is useful in situations
+    where the number of elements in the map is hard to predict. The overhead
+    of an ebtree is very low compared to a hash map, as the overhead of a hash
+    map is the whole bucket set array, while for an ebtree is only a little
+    struct.
+
```
gavin-norman-sociomantic: Should also mention the modification / iteration thing here.

nemanja-boric-sociomantic: Sorry I could swear I added it here. Perhaps it was lost in refactoring.

gavin-norman-sociomantic: You added it to the release notes.

nemanja-boric-sociomantic:
Updated.

src/ocean/util/container/map/TreeMap.d
nemanja-boric-sociomantic: You are right, I've checked `git reflog` and it was never committed.

gavin-norman-sociomantic (approved):

PR #617, opened 2018-08-17
Add statistics counter for generating a time histogram

nemanja-boric-sociomantic:
A timing statistics counter for any sort of transaction that takes a
microsecond to a second to complete. Collects the following statistical
information:

- a logarithmic time histogram with bins from ≥1µs to <1s, three bins per
  power of ten in a  stepping of 1 .. 2 .. 5 .., plus one bin for each <1µs
  and ≥1s,

- the total number of transactions and the aggregated total completion time.

### DIFF
--- a/src/ocean/math/TimeHistogram.d
+++ b/src/ocean/math/TimeHistogram.d
@@ -1,0 +1,414 @@
+/*******************************************************************************
+
+    A timing statistics counter for any sort of transaction that takes a
+    microsecond to a second to complete. Collects the following statistical
+    information:
+    - a logarithmic time histogram with bins from ≥1µs to <1s, three bins per
+      power of ten in a  stepping of 1 .. 2 .. 5 .., plus one bin for each <1µs
+      and ≥1s,
+    - the total number of transactions and the aggregated total completion time.
+
+    To reset all counters to zero use `TimeHistogram.init`.
+
+    copyright: Copyright (c) 2018 dunnhumby Germany GmbH. All rights reserved.
+
+    License:
+        Boost Software License Version 1.0. See LICENSE.txt for details.
+
+*******************************************************************************/
+
+module ocean.math.TimeHistogram;
+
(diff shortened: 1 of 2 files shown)


src/ocean/util/TimeHistogram.d
```diff
@@ -0,0 +1,414 @@
...
+    To reset all counters to zero use `TimeHistogram.init`.
+
+    copyright: Copyright (c) 2018 dunnhumby Germany GmbH. All rights reserved.
+
+    License:
+        Boost Software License Version 1.0. See LICENSE.txt for details.
+
+*******************************************************************************/
+
+module ocean.util.TimeHistogram;
```
gavin-norman-sociomantic: Should probably live in https://github.com/sociomantic-tsunami/ocean/tree/v4.x.x/src/ocean/math, along with `BinaryHistogram`?

nemanja-boric-sociomantic: Moved.

src/ocean/math/TimeHistogram.d
```diff
@@ -0,0 +1,414 @@
...
+
+        Params:
+            us = the transaction completion time in microseconds
+
+        Returns:
+            us
+
+    ***************************************************************************/
+
+    ulong countMicros ( ulong us )
```
gavin-norman-sociomantic: `public`.

gavin-norman-sociomantic (commented):
Apart from one tiny comment, LGTM.

nemanja-boric-sociomantic:
I've added `public` also to `total_time_micros`, `count` and `bins`. AFAIR these are also supposed to be used from the outside.

gavin-norman-sociomantic:
Yeah, this struct just trusts its user to not mess with its internals. I think that's fine.

gavin-norman-sociomantic (approved):

PR #615, opened 2018-08-17
Add thread workers support to ocean

nemanja-boric-sociomantic:
This adds support for calling user delegates in a task-blocking fashion which will be then executed in separate threads. This makes it possible to perform blocking system calls (reading from disk or, for example, using `curl_easy_*` API) as a task-blocking operations.

For usage example, see `integrationtests/asyncio/main.d`.

PR #618, opened 2018-08-17
Add ebtree TreeQueue data structure.

nemanja-boric-sociomantic:
`TreeQueue`, the data structure based on the ebtree is added where the unique
items can be pushed, queried and popped in the order of pushing.

### DIFF
--- a/relnotes/treequeue.feature.md
+++ b/relnotes/treequeue.feature.md
@@ -1,0 +1,6 @@
+### Add ebtree TreeQueue data structure.
+
+* `ocean.util.container.TreeQueue`
+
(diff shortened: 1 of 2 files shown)


gavin-norman-sociomantic:
Verbal discussion: we agreed that this data structure is too specific to be put into ocean. It's required by the current implementation of the thread pool (#615), but it should be easy to adapt that to use ocean's standard `LinkedListQueue`.

gavin-norman-sociomantic:
(Blocked for now. We can close if the plan above works out.)

nemanja-boric-sociomantic:
I'll just close this, since AsyncIO PR is already updated to use different queue. We can reopen it if necessary.

PR #621, opened 2018-08-24
[REG 4.2.6] Arguments used to allow empty/non-existent integer argument

mihails-strasuns:
Regression from https://github.com/sociomantic-tsunami/ocean/commit/c141b3c967d88e3b88e60d5151a42efe45358f3a#diff-59a9e0a12c22f0a8f0e7160c47f92956

### DIFF
--- a/src/ocean/text/Arguments.d
+++ b/src/ocean/text/Arguments.d
@@ -1278,6 +1278,7 @@ public class Arguments
 
         Returns:
             integer value of the parameter assigned to the argument
+            0 if value is missing or not a valid integer
 
     ***************************************************************************/
 
@@ -1293,8 +1294,10 @@ public class Arguments
         }
 
         T num;
-        enforce(toInteger(value, num));
-        return num;
+        if (toInteger(value, num))
+            return num;
+        else
+            return 0;
     }
 
 
@@ -2663,6 +2666,10 @@ unittest
     args("num").params(1);
     test(args.parse("--num 18446744073709551615"));
     test(args.getInt!(ulong)("num") == ulong.max);
+
+    args = new Arguments;
+    args("num").params(1);
+    test(args.getInt!(ulong)("num") == 0);
 }
 
 // Test for D2 'static immutable'


gavin-norman-sociomantic (approved):

src/ocean/text/Arguments.d
```diff
@@ -1293,8 +1293,10 @@ public class Arguments
         }
 
         T num;
-        enforce(toInteger(value, num));
-        return num;
+        if (toInteger(value, num))
```
gavin-norman-sociomantic, line 1297: Actually, probably also worth mentioning this behaviour in the method doc. (The lack of doc is probably partly what caused this regression in the first place.)

mihails-strasuns: Updated

gavin-norman-sociomantic (approved):

gavin-norman-sociomantic:
LGTM. Feel free to auto-merge when the tests pass.

src/ocean/text/Arguments.d
```diff
@@ -1278,6 +1278,7 @@ public class Arguments
 
         Returns:
             integer value of the parameter assigned to the argument
+            0 if value is missing or not a valid integer
```
joseph-wakeling-sociomantic, line 1281: Obviously we need to fix the regression for now, but is this a desirable behaviour?  It looks like this makes it impossible to tell the difference between `--arg 0` and `--arg somethinginvalid`.

Maybe we should look at updating this behaviour in the next major release?

gavin-norman-sociomantic: Indeed. It is pretty odd behaviour. Currently, you need to do this:
```D
if ( arg.assigned.length > 0 )
{
  arg.getInt(...);
}
else
  // undefined
```

nemanja-boric-sociomantic: Looking at the offending turtle's usage [1]:

```
 auto progress = args.getInt!(long)("progress");
```
I think this should be written as:

```
auto progress = args.exists("progress")? args.getInt!(long)("progress") : 0;
```

if it doesn't care about it, or it should have specified the default value.

[1] https://github.com/sociomantic-tsunami/turtle/blob/a0da9cc4269d3d83c730c2295dc7f79f7964a15c/src/turtle/runner/Runner.d#L692

PR #622, opened 2018-08-27
Introduce SslClientConnection

don-clugston-sociomantic:
This is a minimal binding to OpenSSL, and also introduces an asynchronous Task-based implementation of an SSL client. I am not really sure about the right approach to error handling. Here, I've created a reusable `SslException` class based on `IOException`, but I have not used it when intiializing the SSL library.

However, this code does all that is necessary to use OpenSSL asynchronously. It was quite tricky to get it right, since some of the behaviour is only documented in very obscure places. Some of the documentation which I have created for the OpenSSL binding, is more complete than the documentation on the OpenSSL website.

### DIFF
--- a/relnotes/sslclient.feature.md
+++ b/relnotes/sslclient.feature.md
@@ -1,0 +1,7 @@
+### New bindings for OpenSSL and new SslClientConnection class
+
+* `ocean.net.ssl.SslClientConnection`
+
+  This new class provides an asynchronous Task-based implementation of an SSL
+  client. It provides `connect`, `read`, and `write` primitives, which are the
+  basis for higher-level protoocols like HTTPS.
--- a/src/ocean/net/ssl/SslClientConnection.d
+++ b/src/ocean/net/ssl/SslClientConnection.d
@@ -1,0 +1,584 @@
+/******************************************************************************
+
+    Task-based implementation of an SSL connection
+
+    copyright: Copyright (c) 2018 dunnhumby Germany GmbH. All rights reserved
+
+
+    Usage example: See documented unittest of X.
+
+*******************************************************************************/
+
+module ocean.net.ssl.SslClientConnection;
+
+import ocean.net.ssl.openssl.OpenSsl;
+import ocean.text.util.StringC;
+import ocean.transition;
+
+
+
+/***************************************************************************
+
+    Global context used for all SSL connections
+
+***************************************************************************/
+
+private SSL_CTX * globalSslContext;
+
+
+
+/***************************************************************************
+
+    Class representing a single SSL client connection.
+    Can only be called from inside a Task.
+
+***************************************************************************/
+
+public class SslClientConnection
+{
+    private import core.sys.posix.netdb : AF_UNSPEC, SOCK_STREAM;
+
+    static import ocean.core.ExceptionDefinitions;
+
+    private import ocean.io.select.protocol.task.TaskSelectClient;
+
+    private import ocean.sys.socket.AddrInfo;
+
+    private import ocean.sys.socket.model.ISocket;
+
+    private import ocean.sys.Epoll: epoll_event_t;
+
+    private alias epoll_event_t.Event Event;
+
+
+    /***************************************************************************
+
+        Exception class thrown on errors.
+
+    ***************************************************************************/
+
+    public static class SslException :
+        ocean.core.ExceptionDefinitions.IOException
+    {
+        import ocean.core.Exception: ReusableExceptionImplementation;
+
+        mixin ReusableExceptionImplementation!() ReusableImpl;
+
+        /*******************************************************************
+
+            Sets the exception instance.
+
+            Params:
+                file_path = path of the file
+                func_name = name of the method that failed
+                msg = message description of the error
+                file = file where exception is thrown
+                line = line where exception is thrown
+
+        *******************************************************************/
+
+        public typeof(this) set ( cstring host_path,
+                istring func_name,
+                cstring msg,
+                istring file = __FILE__, long line = __LINE__)
+        {
+            this.error_num = error_num;
+            this.func_name = func_name;
+
+            this.ReusableImpl.set(this.func_name, file, line)
+                .fmtAppend(": {} {} on {}", this.func_name, msg, host_path);
+
+            return this;
+        }
+    }
+
+
+    /***************************************************************************
+
+        A minimal implementation of ISocket
+
+    ***************************************************************************/
+
+    private static class SimpleSocket : ISocket
+    {
+        public this ()
+        {
+            super (addrinfo.sizeof);
+        }
+
+        public override void formatInfo ( ref char[] buf, bool io_error )
+        {
+
+        }
+    }
+
+    /***************************************************************************
+
+        The socket which is used for the connection
+
+    ***************************************************************************/
+
+    private SimpleSocket socket;
+
+
+    /***************************************************************************
+
+        The SSL connection
+
+    ***************************************************************************/
+
+    private SSL * sslHandle;
+
+
+    /***************************************************************************
+
+        SelectClient used for blocking the calling Task
+
+    ***************************************************************************/
+
+    private TaskSelectClient select_client;
+
+
+    /***************************************************************************
+
+        Instance of AddrInfor used for resolving host names
+
+    ***************************************************************************/
+
+    private AddrInfo addr_info;
+
+
+    /***********************************************************************
+
+        Reusable exception instance
+
+    ***********************************************************************/
+
+    private SslException exception;
+
+
+    /***************************************************************************
+
+        Constructor
+
+    ***************************************************************************/
+
+    public this ( )
+    {
+        this.addr_info = new AddrInfo();
+
+        this.socket = new SimpleSocket();
+
+        this.select_client = new TaskSelectClient(this.socket,
+            &this.socket.error);
+
+        this.exception = new SslException;
+    }
+
+
+    /***************************************************************************
+
+        Create an SSL connection. Blocks the calling task until the handshake
+        is complete.
+
+        Params:
+            host_name = the name of the host
+            host_port = the port to use (eg "443" for an HTTPS connection)
+
+        Throws:
+            IOException if the connection failed
+
+    ***************************************************************************/
+
+    public void connect ( cstring host_name, cstring host_port )
+    {
+        // Resolve the host
+
+        // We should be able to use AddrInfo.getIp but cannot because it doesn't
+        // support AF_UNSPEC.
+
+        auto addr_err = this.addr_info.get(host_name, host_port, AF_UNSPEC,
+            SOCK_STREAM, 0);
+
+        if ( addr_err != 0 )
+        {
+            this.ssl_error(host_name, "connect", gai_strerror(addr_err));
+        }
+
+        this.socket.socket(addr_info.info().ai_family,
+            addr_info.info().ai_socktype | SocketFlags.SOCK_NONBLOCK,
+            addr_info.info().ai_protocol);
+
+        if ( !socket.connect(addr_info.info().ai_addr) )
+        {
+            this.error(host_name, "socket.connect", "Failed to connect");
+        }
+
+        this.sslHandle = SSL_new(globalSslContext);
+
+        if ( !this.sslHandle )
+        {
+            this.error(host_name, "connect", "sslNew failed");
+        }
+
+        if (!SSL_set_fd(this.sslHandle, this.socket.fileHandle))
+        {
+            this.error(host_name, "connect", "sslSetFd failed");
+        }
+
+        // Set it into "connect" state, ie it is a client, not a server.
+        // (For a server, SSL_set_accept_state would be called instead).
+
+        SSL_set_connect_state(this.sslHandle);
+
+        ERR_clear_error();
+
+        int r;
+
+        while ( (r = SSL_do_handshake(this.sslHandle)) != 1 )
+        {
+            int err = SSL_get_error(this.sslHandle, r);
+
+            if ( err ==  SSL_ERROR_WANT_WRITE )
+            {
+                this.select_client.ioWait(Event.EPOLLOUT);
+            }
+            else if (err == SSL_ERROR_WANT_READ)
+            {
+                this.select_client.ioWait(Event.EPOLLIN);
+            }
+            else
+            {
+                this.error(host_name, "connect", "Handshake failed");
+            }
+        }
+    }
+
+
+    /***************************************************************************
+
+        Write a string to the SSL connection
+
+        The calling task is blocked until the write has completed.
+
+        Params:
+            request = the string to send
+
+        Throws:
+            IOException if the write failed
+
+    ***************************************************************************/
+
+    public void write ( cstring request )
+    {
+        int w;
+
+        while ( (w = SSL_write(this.sslHandle, request.ptr,
+            cast(int)request.length)) <= 0 )
+        {
+            auto err = SSL_get_error(this.sslHandle, w);
+
+            // At any time, a renegotiation is possible, so a call to SSL_write
+            // can also cause read operations.
+
+            if ( err ==  SSL_ERROR_WANT_WRITE )
+            {
+                this.select_client.ioWait(Event.EPOLLOUT);
+            }
+            else if (err == SSL_ERROR_WANT_READ)
+            {
+                this.select_client.ioWait(Event.EPOLLIN);
+            }
+            else
+            {
+                // The write failed
+
+                this.ssl_error("", "write", ERR_reason_error_string(err));
+            }
+        }
+
+        if ( w != request.length )
+        {
+            this.error("", "write", "Mismatch in number of bytes written");
+
+        }
+    }
+
+
+    /***************************************************************************
+
+        Reads a string from the SSL connection
+
+        The calling task is blocked until the read has completed.
+
+        Params:
+            buffer = array to store the string
+
+        Returns:
+            a slice into buffer of the bytes which were read.
+
+        Throws:
+            IOException if the read failed
+
+    ***************************************************************************/
+
+    public mstring read ( mstring buffer )
+    {
+        this.select_client.ioWait(Event.EPOLLIN);
+
+        ERR_clear_error();
+
+        int r;
+
+        while ( (r = SSL_read(this.sslHandle, buffer.ptr,
+            cast(int)buffer.length)) <= 0 )
+        {
+            auto err = SSL_get_error(this.sslHandle, r);
+
+            // At any time, a renegotiation is possible, so a call to SSL_read
+            // can also cause write operations.
+
+            if ( err ==  SSL_ERROR_WANT_WRITE )
+            {
+                this.select_client.ioWait(Event.EPOLLOUT);
+            }
+            else if (err == SSL_ERROR_WANT_READ)
+            {
+                this.select_client.ioWait(Event.EPOLLIN);
+            }
+            else
+            {
+                // The read failed
+
+                this.ssl_error("", "read", ERR_reason_error_string(err));
+            }
+
+        }
+        return buffer[0 .. r];
+    }
+
+
+    /***************************************************************************
+
+        Validate the X509 certificate.
+
+        The relevant documents are RFC 5280 and RFC 6125.
+
+        Params:
+            host_name = the name of the host
+
+        Throws:
+            IOException if validation fails
+
+    ***************************************************************************/
+
+    public void validateCertificate (cstring host_name)
+    {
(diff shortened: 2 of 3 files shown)


src/ocean/net/ssl/openssl/OpenSsl.d
```diff
@@ -0,0 +1,564 @@
+/******************************************************************************
+
+    C binding to OpenSSL
+
+    copyright: Copyright (c) 2018 sociomantic labs. All rights reserved
```
gavin-norman-sociomantic: `dunnhumby Germany GmbH` ;)

src/ocean/net/ssl/SslClientConnection.d
```diff
@@ -0,0 +1,586 @@
+/******************************************************************************
+
+    Task-based implementation of an SSL connection
+
+    copyright: Copyright (c) 2018 sociomantic labs. All rights reserved
```
gavin-norman-sociomantic: `dunnhumby Germany GmbH` ;)

src/ocean/net/ssl/SslClientConnection.d
```diff
@@ -0,0 +1,586 @@
...
+
+    Example:
+
+    Here is the fundamental code for an HTTPS client.
+    A proper HTTP client would need to parse the HTTP response header to
+    determine how many bytes should be read; this example will always trigger
+    an error after the final bytes are read.
+
+    ---
+    void run ()
```
gavin-norman-sociomantic: Could you make this a documented unittest? It needn't actually run the code in the unittest, but would be nice to check that it compiles.

src/ocean/net/ssl/SslClientConnection.d
```diff
@@ -0,0 +1,586 @@
...
+
+
+
+/***************************************************************************
+
+    Global context used for all SSL connections
+
+***************************************************************************/
+
+private SSL_CTX *globalSslContext;
```
gavin-norman-sociomantic: C-style `*` placement.

src/ocean/net/ssl/SslClientConnection.d
```diff
@@ -0,0 +1,586 @@
...
+
+
+
+/***************************************************************************
+
+    Class representing a single Https client connection.
```
gavin-norman-sociomantic: Https -> SSL.

src/ocean/net/ssl/SslClientConnection.d
```diff
@@ -0,0 +1,586 @@
...
+    private SSL * sslHandle;
+
+
+    /***************************************************************************
+
+        SelectClient used for blocking the calling Task
+
+    ***************************************************************************/
+
+    private TaskSelectClient select_client;
```
gavin-norman-sociomantic, line 139: Initial, high-level design question: you're wrapping a socket and a task select client. Another approach would be to accept those in the ctor (or the I/O methods). Pros / cons?

gavin-norman-sociomantic (commented):
First question and a few minor points.

src/ocean/net/ssl/SslClientConnection.d
don-clugston-sociomantic: I think that this class really needs to own the socket. It is stored inside `sslHandle`, and if the app played around with the socket, all kinds of weird things would happen.

It is true that the socket is only directly required in the `connect` function, where the `sslHandle` is created. But the `select_client` is required everywhere, and the three members are coupled together in a complicated way.

don-clugston-sociomantic:
Repushed with the first basic fixes made. I have made the example compilable in a unittest, though it is not actually executed.

src/ocean/net/ssl/SslClientConnection.d
```diff
@@ -0,0 +1,622 @@
...
+
+        public typeof(this) set ( cstring host_path,
+                istring func_name,
+                cstring msg,
+                istring file = __FILE__, long line = __LINE__)
+        {
+            this.error_num = error_num;
+            this.func_name = func_name;
+
+            this.ReusableImpl.set(this.func_name, file, line);
```
mihails-strasuns: Reusable now supports formatter:

```D
this.ReusableImpl.set(this.func_name, file, line)
    .fmtAppend(": {} {} on {}", this.func_name, msg, host_path);
```

src/ocean/net/ssl/SslClientConnection.d
```diff
@@ -0,0 +1,622 @@
...
+    ***************************************************************************/
+
+    private static class SimpleSocket : ISocket
+    {
+        public this ()
+        {
+            super (addrinfo.sizeof);
+        }
+
+        public override void formatInfo ( ref char[] buf, bool io_error )
```
mihails-strasuns, line 109: Needs a comment why having empty impl is OK here

mihails-strasuns (approved):
Can you please add a short app that uses this to connect to some SSL host and put it into `examples` folder? Setting full integration test would be even better but that is much more effort of course.

mihails-strasuns (changes requested):
(wrong review status :P)

mihails-strasuns:
Also:

- missing release notes
- will break existing projects at link step because of new dependency - can't remember what process we agreed on to handle this

src/ocean/net/ssl/SslClientConnection.d
don-clugston-sociomantic: I can't answer that. I don't know why that function is abstract in the first place. I shouldn't have to declare this class at all.

don-clugston-sociomantic:
I don't know how to do tests for this code, since it requires a server to connect to. How do we normally deal with that?

src/ocean/net/ssl/SslClientConnection.d
don-clugston-sociomantic: That function is not used by `ISocket`. I do not know why it exists.

mihails-strasuns:
By running both server and client code as part of test binary, see https://github.com/sociomantic-tsunami/ocean/blob/v4.x.x/integrationtest/httpserver/main.d for example.

Not sure if we have SSL-able server impl in ocean though :angel: 

But as I have mentioned, just having example that connects to something like `sociomantic.com` is fine too for a start.

src/ocean/net/ssl/SslClientConnection.d
mihails-strasuns: @gavin-norman-sociomantic @david-eckardt-sociomantic ?

nemanja-boric-sociomantic:
Yeah, since this is an example and not the integrationtest, that should be fine.

src/ocean/net/ssl/SslClientConnection.d
nemanja-boric-sociomantic: As @mihails-strasuns said, this is better to go to `examples`.

nemanja-boric-sociomantic: Actually for the code as short as here, the unittest is also convenient. Just wrap it inside the function so it doesn't execute it:

```
unittest
{
     void connectAndRead ()
     { /* code */ }
}
```

src/ocean/net/ssl/SslClientConnection.d
nemanja-boric-sociomantic: It's not used by `ISocket` but it's used by the implementations of the socket - it's like `toString` for various sockets, IIRC. I'll try to find examples.

nemanja-boric-sociomantic: https://github.com/sociomantic-tsunami/ocean/blob/3ee545df3c2e0aec2e9396ad4e91d7f5f700073e/src/ocean/sys/socket/UnixSocket.d#L161

usage:

https://github.com/sociomantic-tsunami/ocean/blob/3ee545df3c2e0aec2e9396ad4e91d7f5f700073e/src/ocean/net/server/connection/IConnectionHandler.d#L369

(and that's called from https://github.com/sociomantic-tsunami/ocean/blob/3ee545df3c2e0aec2e9396ad4e91d7f5f700073e/src/ocean/net/server/SelectListener.d#L561)

nemanja-boric-sociomantic: So you should provide simple human readable string here for logging purposes.

src/ocean/net/ssl/SslClientConnection.d
```diff
@@ -0,0 +1,622 @@
...
+            {
+                this.select_client.ioWait(Event.EPOLLIN);
+            }
+            else
+            {
+                // The write failed
+
+                this.ssl_error("", "write", ERR_reason_error_string(err));
+            }
+        }
```
nemanja-boric-sociomantic, line 298: It looks that this code for handing the epoll integration is same in all cases. Can it be moved to a single method?

src/ocean/net/ssl/SslClientConnection.d
nemanja-boric-sociomantic: I see you already did it. Perhaps remove this code here and point people to the documented unittest at the end of the file to avoid stale code?

src/ocean/net/ssl/SslClientConnection.d
```diff
@@ -0,0 +1,622 @@
...
+
+    // Technically, this isn't a fatal error, but many subsequent operations
+    // will fail
+
+    return res;
+}
+
+
+
+unittest
```
nemanja-boric-sociomantic, line 546: Add `///` to generate documentation from this unittest.

src/ocean/net/ssl/SslClientConnection.d
don-clugston-sociomantic: Can't there be a default implementation? It's not of any interest for this module. It seems crazy to have to provide a derived class, just for debugging!

nemanja-boric-sociomantic: Yeah, the default implementation could be:

```
        sformat(buf, "fd={}, ioerr={}",
            this.fileHandle, io_error);
```

src/ocean/net/ssl/openssl/OpenSsl.d
```diff
@@ -0,0 +1,564 @@
...
+
+        Retrieve the function pointers for SSLv3, or v2 if v3 is unavailable
+
+        Returns;
+            A struct with all of the function pointers pointing to the SSL v3
+            functions, falling back to v2 if v3 is unavailable
+
+    ***************************************************************************/
+
+    public ssl_method_st * SSLv23_method();
```
david-eckardt-sociomantic: Please attach the `*` to the pointer base type.

src/ocean/net/ssl/SslClientConnection.d
```diff
@@ -0,0 +1,622 @@
...
+
+        ERR_clear_error();
+
+        int r;
+
+        while ( (r = SSL_do_handshake(this.sslHandle)) != 1 )
+        {
+            int err = SSL_get_error(this.sslHandle, r);
+
+            if ( err ==  SSL_ERROR_WANT_WRITE )
```
david-eckardt-sociomantic, line 242: Suggest `switch (SSL_get_error(this.sslHandle, r))`.

src/ocean/net/ssl/SslClientConnection.d
```diff
@@ -0,0 +1,622 @@
...
+        if ( cert )
+        {
+            X509_free(cert); // Free the certificate immediately
+        }
+
+        if (!cert)
+        {
+            this.error(host_name, "SSL_get_peer_certificate",
+                "No certificate was presented");
+        }
```
david-eckardt-sociomantic, line 388: Suggest
```d
if (X509* cert = SSL_get_peer_certificate(this.sslHandle))
{
    X509_free(cert); // Free the certificate immediately
}
else
{
    this.error(host_name, "SSL_get_peer_certificate",
        "No certificate was presented");
}
```
This protects from accidentally using `cert` if it is `null` or after it has been deallocated.

src/ocean/net/ssl/SslClientConnection.d
```diff
@@ -0,0 +1,622 @@
...
+    {
+        // Success
+
+        return 0;
+    }
+
+    // Technically, this isn't a fatal error, but many subsequent operations
+    // will fail
+
+    return res;
```
david-eckardt-sociomantic: The return code is the same for `res == 1` (success) and `res == 0`. Shouldn't `res == 0` be an error? The caller won't be able to tell.

src/ocean/net/ssl/SslClientConnection.d
```diff
@@ -0,0 +1,622 @@
...
+            IOException if the write failed
+
+    ***************************************************************************/
+
+    public void write ( cstring request )
+    {
+        int w;
+
+        while ( (w = SSL_write(this.sslHandle, cast(void *)request.ptr,
+            cast(int)request.length)) <= 0 )
```
david-eckardt-sociomantic, line 277: Casting `const` away from `request.ptr` deserves a note.

mihails-strasuns: No need to cast at all, just declare `SSL_write` as taking const argument.

src/ocean/net/ssl/SslClientConnection.d
don-clugston-sociomantic: I'm not sure if the code is fundamentally the same, or if it just looks the same by accident.
If the socket is blocking, the return values of `SSL_write` are different to `SSL_read`, and `SSL_do_handshake` is completely different, and even when non-blocking, the error codes other than `SSL_ERROR_WANT_READ` and `SSL_ERROR_WANT_WRITE`, are different for reads, writes, and handshakes. It might be necessary to treat them differently in the future.

It was also a non-obvious feature of the code. Many examples which I saw, didn't realize that you need to do this for writes, as well as for reads. The fact that you might need to do a write while you are doing a read, is highly counter-intuitive. Whereas it's totally expected for handshaking.

Since this pattern only appears in 3 places (and no other SSL calls behave this way), I don't think it's possible to reduce code size by attempting to extract it.

nemanja-boric-sociomantic: > the error codes other than SSL_ERROR_WANT_READ and SSL_ERROR_WANT_WRITE, are different for reads, writes, and handshakes

OK, this is a good point, and I agree good point for keeping it as is.

nemanja-boric-sociomantic: (it was not visible from the code I've seen).

src/ocean/net/ssl/SslClientConnection.d
don-clugston-sociomantic: Good catch. The code at this point was badly wrong.

src/ocean/net/ssl/SslClientConnection.d
```diff
@@ -0,0 +1,618 @@
...
+    }
+}
+
+
+
+/*******************************************************************************
+
+    Initializes SSL and creates a global SSL_CTX object
+
+    This function must be called before an SSL clients can be created
```
don-clugston-sociomantic: Note to self, this should be "any SSL clients"

src/ocean/net/ssl/SslClientConnection.d
david-eckardt-sociomantic: What you could do is to add
```d
private void doSslIo ( lazy int io, void delegate(int err) handle_error )
{
    while (true)
    {
        int w = io;
        if (w > 0)
            break;

        auto err = SSL_get_error(this.sslHandle, w);
        // At any time, a renegotiation is possible, so a call to SSL_write
        // can also cause read operations.
        switch (err)
        {
            case SSL_ERROR_WANT_WRITE:
                this.select_client.ioWait(Event.EPOLLOUT);
                break;
            case SSL_ERROR_WANT_READ:
                this.select_client.ioWait(Event.EPOLLIN);
                break;
            default:
                handle_error(err);
        }
    }
}
```
and call it in `connect`, `read` and `write`.

don-clugston-sociomantic: The interesting thing that Nemanja has noticed, is that the `ioWait` calls are actually the only place that relies on epoll, or tasks. So, if those `select_client.ioWait` calls were abstracted, it would be possible to use this code in the existing swarm implementation, for example.
This gives two competing refactorings for the same piece of code.

don-clugston-sociomantic: BTW a really nasty feature of the code is that although all three functions return `SSL_ERROR_WANT_WRITE` if they want to write, the norm (non-error condition) return value of each function has a totally different meaning for the three functions.

mihails-strasuns:

don-clugston-sociomantic:
Repushed with the simple changes addressed. I'm not sure about the others.
Because of some other work I'm doing, I need additional functions from OpenSSL, and it is higher priority. Should I add that additional binding to this PR? Or keep it separate? Or split the openSSL binding into a separate PR?

nemanja-boric-sociomantic:
I say move all OpenSSL bindings introduced here + your additional ones into a separate PR and since I think that's very non-controversial thing it can be merged really soon.

don-clugston-sociomantic:
The OpenSSL binding has been expanded, and moved to a new PR at
https://github.com/sociomantic-tsunami/ocean/pull/632

don-clugston-sociomantic:
I have rebased this onto https://github.com/sociomantic-tsunami/ocean/pull/632, which should be merged first.

mihails-strasuns:
Needs rebase now

don-clugston-sociomantic:
Rebased onto the latest v4.x.x

relnotes/sslclient.feature.md
```diff
@@ -0,0 +1,12 @@
+### New bindings for OpenSSL and new SslClientConnection class
```
gavin-norman-sociomantic, line 1: 1. Any reason to not squash this with the commit that adds the new class?
2. The SSL bindings were added in another PR, right? (I guess the release notes were forgotten. Just checking that we don't end up with doubles.)

src/ocean/net/ssl/SslClientConnection.d
gavin-norman-sociomantic: Suggest removing this doc here and just saying "Usage example: See documented unittest of X.".

src/ocean/net/ssl/SslClientConnection.d
gavin-norman-sociomantic: Maybe make a PR for this?

src/ocean/net/ssl/SslClientConnection.d
```diff
@@ -0,0 +1,612 @@
...
+
+        Params:
+            host_name = the name of the host
+
+        Throws:
+            IOException if validation fails
+
+    ***************************************************************************/
+
+    public void validateCertificate (cstring host_name)
```
gavin-norman-sociomantic, line 375: Is there any situation where you'd want to do this not at the same time as connecting?

gavin-norman-sociomantic (commented):
A few small comment.

src/ocean/net/ssl/SslClientConnection.d
don-clugston-sociomantic: Technically, it is an optional step.
There are cases where you want to allow a connection to be made, even when the certificate has expired or is otherwise invalid. Curl provides this via the `-k` or `--insecure` command-line option. Some of our apps are forced to use this for some client servers. You would only do this in cases where a secure connection isn't actually required.

relnotes/sslclient.feature.md
don-clugston-sociomantic: 1. It's not done yet because things can change. 2 is an example of that.

mihails-strasuns:
Looks like last @gavin-norman-sociomantic comments are still not resolved

don-clugston-sociomantic:
Repushed with documentation updated, and rebased onto 4.4.x. This removes the 'SSL binding' part of this PR, which was merged long ago.

If there are still issues with this, it might be worth splitting the first commit into a tiny PR of its own, and maybe also making a PR to make `ISocket` non-abstract, as Gavin suggested. Just so that we can close as much of this PR as possible.

mihails-strasuns (approved):
LGTM, up to @gavin-norman-sociomantic for final merge

don-clugston-sociomantic:
I have squashed the release notes into the 'SSlClientConnection' commit.

src/ocean/net/ssl/SslClientConnection.d
gavin-norman-sociomantic: Does `/***` generate a documented unittest as well? The [language spec](https://dlang.org/spec/unittest.html#documented-unittests) is unclear on what exactly is required.

mihails-strasuns: It should.

gavin-norman-sociomantic (approved):

PR #625, opened 2018-08-29
Separate lifetime for missing records for `ExpiringCache`

### LINKED ISSUE #539: Enhance expiring cache with per-record lifetime
The current expiring cache implementation has a lifetime value that is set in the constructor (https://github.com/sociomantic-tsunami/ocean/blob/v3.x.x/src/ocean/util/container/cache/ExpiringCache.d#L84) making it impossible to store different records for different lifetimes in the same cache. To do this currently would require multiple caches which complicates application logic.

It would be nice if different lifetimes for records in the same cache were possible so, for example, records expected to change frequently have a short lifetime and more stable records have a longer lifetime.

mihails-strasuns:
Fixes #539

### DIFF
--- a/relnotes/expiring.feature.md
+++ b/relnotes/expiring.feature.md
@@ -1,0 +1,12 @@
+### Separate lifetime for empty records of ExpiringCache
+
+`ocean.util.container.cache.ExpiringCache`
+
+Now supports extra optional constructor argument - lifetime used for empty
+cached records. If not specified, regular lifetime value will be used, same as
+before.
+
+ExpiringCache considers a cached record with array length 0 as being "empty".
+
+This feature is intended for applications caching failures as empty values so
+that shorter expiration can be used for those.
--- a/src/ocean/util/container/cache/ExpiringCache.d
+++ b/src/ocean/util/container/cache/ExpiringCache.d
@@ -6,6 +6,10 @@
     removing  elements where the difference between the current time and the
     createtime is greater than that lifetime value.
 
+    The cache can be configured to apply different lifetimes to empty and
+    non-empty values (where a cached record with array length 0 is considered
+    "empty"). See constructor arguments.
+
     Copyright:
         Copyright (c) 2009-2016 dunnhumby Germany GmbH.
         All rights reserved.
@@ -54,12 +58,24 @@ class ExpiringCache ( size_t ValueSize = 0 ) : Cache!(ValueSize, true),
 
     /***************************************************************************
 
-        Life time for all items in seconds; may be changed at any time.
+        Life time for non-empty items in seconds; may be changed at any time.
         This value must be at least 1.
 
     ***************************************************************************/
 
     public time_t lifetime;
+
+    /***************************************************************************
+
+        Life time for empty items in seconds; may be changed at any time.
+        This value must be at least 1. Can be same as `lifetime`.
+
+        ExpiringCache considers a cached record with array length 0 as being
+        "empty".
+
+    ***************************************************************************/
+
+    public time_t empty_lifetime;
 
     /***************************************************************************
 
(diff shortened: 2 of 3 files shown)


relnotes/expiring.feature.md
```diff
@@ -0,0 +1,12 @@
+### Separate lifetime for empty records of ExpiringCache
+
+`ocean.util.container.cache.ExpiringCache`
+
+Now supports extra optional constructor argument - lifetime used for empty
+cached records. If not specified, regular lifetime value will be used, same as
+before.
+
+By empty record `ExpiringCache` considers one with internaly array length of 0.
```
gavin-norman-sociomantic: Wording:
"
`ExpiringCache` considers a cached record with array length 0 as being "empty".
"

src/ocean/util/container/cache/ExpiringCache.d
```diff
@@ -54,13 +54,22 @@ class ExpiringCache ( size_t ValueSize = 0 ) : Cache!(ValueSize, true),
```
gavin-norman-sociomantic, line 58: Suggest adding this to the module doc:
"
The cache can be configured to apply different lifetimes to empty and non-empty values (where a cached record with array length 0 is considered "empty"). See constructor arguments.
"

src/ocean/util/container/cache/ExpiringCache.d
```diff
@@ -54,13 +54,22 @@ class ExpiringCache ( size_t ValueSize = 0 ) : Cache!(ValueSize, true),
...
+        Life time for non-empty items in seconds; may be changed at any time.
         This value must be at least 1.
 
     ***************************************************************************/
 
     public time_t lifetime;
 
+    /***************************************************************************
+
+        Life time for empty items in seconds; may be changed at any time.
```
gavin-norman-sociomantic, line 70: Suggest defining what empty means here too.

gavin-norman-sociomantic (commented):
LGTM. Couple of doc suggestions.

mihails-strasuns:
Updated:
- fixed wording
- workaround to fix `-release` build

src/ocean/util/container/cache/ExpiringCache.d
```diff
@@ -304,9 +323,6 @@ class ExpiringCache ( size_t ValueSize = 0 ) : Cache!(ValueSize, true),
     }
     body
     {
-        verify (this.lifetime > 0,
```
ben-palmer-sociomantic, line 307: Removal of this check could be in its own commit.

src/ocean/util/container/cache/model/Value.d
```diff
@@ -130,6 +130,18 @@ template Value ( size_t ValueSize )
...
             }
+
+            /*******************************************************************
+
+                Returns:
+                    length of underlying array
+
+            *******************************************************************/
+
+            public size_t length ( )
```
ben-palmer-sociomantic, line 141: Another option would be to have `empty` instead of `length` IIUC, of less general use though.

ben-palmer-sociomantic (commented):
Just a couple of comments, the changes LGTM.

src/ocean/util/container/cache/ExpiringCache.d
mihails-strasuns: It is not removed, it is moved several lines below to check both lifetimes depending on item state.

ben-palmer-sociomantic: Oh sorry, I thought you removed it because it was already checked in the constructor. Ignore my comment then.

src/ocean/util/container/cache/model/Value.d
mihails-strasuns: Yeah I think `length` is better because it is a strict superset of functionality :)

PR #628, opened 2018-09-04
Rename ClassIterator -> ConfigIterator and support struct types.

scott-gibson-sociomantic:
Existing unit tests suggests that this was originally meant to be supported but not correctly done.
This updates it to be able to iterate over struct types.

### DIFF
--- a/src/ocean/util/config/ConfigFiller.d
+++ b/src/ocean/util/config/ConfigFiller.d
@@ -842,9 +842,12 @@ private bool hasField ( T ) ( T reference, cstring field )
     return was_found;
 }
 
+deprecated("Use ConfigIterator instead.")
+alias ConfigIterator ClassIterator;
+
 /*******************************************************************************
 
-    Class Iterator. Iterates over variables of a category
+    Config Iterator. Iterates over variables of a category
 
     Params:
         T = type of the class to iterate upon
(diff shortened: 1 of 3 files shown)


src/ocean/util/config/ConfigFiller.d
```diff
@@ -842,9 +842,12 @@ private bool hasField ( T ) ( T reference, cstring field )
     return was_found;
 }
 
+deprecated("Use ConfigIterator instead.")
```
gavin-norman-sociomantic, line 845: :police_car:

gavin-norman-sociomantic (commented):
This PR gives me deja vu. I thought we already did this?

scott-gibson-sociomantic:
Previously the ConfigFiller class was named ClassFiller and only worked on class types. This was changed to support structs and renamed to ConfigFiller, however the internal ClassIterator was never updated to support structs as well.

scott-gibson-sociomantic:
I added a deprecation message for the rename and a feature message for the struct support.

gavin-norman-sociomantic:
>Previously the ConfigFiller class was named ClassFiller and only worked on class types. This was changed to support structs and renamed to ConfigFiller, however the internal ClassIterator was never updated to support structs as well.

Ahh gotcha.

gavin-norman-sociomantic (approved):

PR #629, opened 2018-09-12
Add stats log reader

bogdan-szabo-sociomantic:


src/ocean/util/log/Stats.d
```diff
@@ -650,3 +656,369 @@ unittest
...
+            line = a line stat
+
+        Returns:
+            A StatsLine
+
+    **************************************************************************/
+
+    static public StatsLine opCall (cstring line)
+    {
+        static Buffer!(char) buffer_line;
```
nemanja-boric-sociomantic: Do you need this static variable? It feels like you could do all this just by slicing the `line`?

src/ocean/util/log/Stats.d
```diff
@@ -650,3 +656,369 @@ unittest
...
 }
+
+
+/**************************************************************************
+
+    Struct that parses a stats line
+
+**************************************************************************/
+
+public struct StatsLine
```
nemanja-boric-sociomantic: `StatsLine` could just hold line as `cstring` and just slice the `line` passed in `opCall`. That would mean that the lifetime of `StatsLine` is valid as long as `line` passed to `opCall` is valid, but I think that is fine. AFAICS, you don't need to pass the `StatsLine` around, and if you want you could provide `dup` method which would return you a copy of `StatsLine` which would manage its own memory.

That way I think you've removed all allocations here.

src/ocean/util/log/Stats.d
```diff
@@ -650,3 +656,369 @@ unittest
...
+        auto date_end_position = find(buffer_line, " ");
+        stats_line.date = buffer_line[0 .. date_end_position];
+
+        auto time_begin_position = date_end_position + 1;
+        auto time_end_position = find(buffer_line[time_begin_position .. len], " ") + time_begin_position;
+        stats_line.time = buffer_line[time_begin_position .. time_end_position];
+
+        stats_line.line = buffer_line[time_end_position .. len];
+
+        return stats_line;
```
nemanja-boric-sociomantic: So, in this method you always create new `StatsLine` and pass it to the caller. Creating `StatsLine` involves creating `Buffer` which `StatsLine` doesn't reuse in any way, and since it's private nobody else can do reuse it, so having `Buffer` there makes a little point (except to prevent copying StatsLines).

I think we should you could/should do it in a way where `StatsLine` is just a _view_ on the line.

bogdan-szabo-sociomantic:
@nemanja-boric-sociomantic ready for another look

src/ocean/util/log/Stats.d
```diff
@@ -650,3 +656,402 @@ unittest
...
+
+    try
+    {
+        data = new Array("".dup);
+        reader = new StatsLogReader(data);
+        reader.last();
+    }
+    catch(Exception e)
+    {
+        test!("==")(e.message, "The stats are empty");
```
mihails-strasuns: Why not simply `testThrown`?

src/ocean/util/log/Stats.d
```diff
@@ -650,3 +656,402 @@ unittest
...
+
+    /// It should not extract a missing key
+    try
+    {
+        line["missing_key"];
+    }
+    catch(Exception e)
+    {
+        test!("==")(e.message, "missing_key is not present in the stats");
+        has_exception = true;
```
mihails-strasuns: ditto `testThrown`

mihails-strasuns (approved):
LGTM apart from few nitpicks. One proposal to consider is to move it new `ocean.util.log.StatsReader` module, I personally like small independent modules more.

src/ocean/util/log/Stats.d
bogdan-szabo-sociomantic: because `testThrown` can not check the error message. I updated the code to use `testThrown` anyway... It looks like there are no tests that check for the error messages

bogdan-szabo-sociomantic:
@mihails-strasuns @nemanja-boric-sociomantic  ready for another look. I moved the code to a new module and refactored the tests to use the testThrown function

src/ocean/util/log/StatsReader.d
```diff
@@ -0,0 +1,393 @@
...
+    Currently, `StatsLogReader` parses an `InputStream` to `StatsLine`. The lines
+    can be iterated using a foreach. Sometimes only the last stat line is
+    important. The `StatsLogReader.last` method will return only that line.
+
+    The `StatsLine` struct parses one stat line.
+
+    Refer to the class' description for information about their actual usage.
+
+    Copyright:
+        Copyright (c) 2009-2018 Sociomantic Labs GmbH.
```
nemanja-boric-sociomantic:
```
        Copyright (C) 2018 dunnhumby Germany GmbH. All rights reserved
```

src/ocean/util/log/StatsReader.d
```diff
@@ -0,0 +1,393 @@
+/*******************************************************************************
+
+    Classes that can be used to parse the graphite stats.
+
+    Currently, `StatsLogReader` parses an `InputStream` to `StatsLine`. The lines
```
nemanja-boric-sociomantic: No need for `Currently,` here, just say `Parses an InputStream...`

nemanja-boric-sociomantic (approved):

nemanja-boric-sociomantic:
Thanks!

src/ocean/util/log/Stats.d
mihails-strasuns: Yeah we don't normally care about exact error message and it is not worth extra lines of boilerplate. If specific exception kind is crucial, it should be denoted by creating dedicated exception class and doing `testThrown!(MyExceptionClass)` instead.

PR #630, opened 2018-09-14
Allow for dynamic terminal width inside AppStatus

nemanja-boric-sociomantic:
AppStatus was assuming that main terminal exists and that
it can be queried for its width. However this is not true
when starting programs as daemons from systemd - the main terminal
doesn't exist (stdout may or may be not redirected), so the
width is zero. However, user may want to attach to the unix socket
and get the output from the AppStatus.

AppStatus is changed so it a) accept assumed terminal width at
construction which, b) can be overridden by passing the width
to the command (`app_status 120` for example).

### DIFF
--- a/src/ocean/io/console/AppStatus.d
+++ b/src/ocean/io/console/AppStatus.d
@@ -71,6 +71,7 @@ import ocean.time.model.IMicrosecondsClock;
 import ocean.time.MicrosecondsClock;
 import ocean.transition;
 import ocean.util.container.AppendBuffer;
+import ocean.text.convert.Integer;
 
 import ocean.util.log.Event;
 import ocean.util.log.InsertConsole;
@@ -119,6 +120,15 @@ public class AppStatus
     ***************************************************************************/
 
     protected alias .TerminalOutput TerminalOutput;
+
+
+    /**************************************************************************
+
+        Number of columns in Terminal
+
+    ***************************************************************************/
+
+    private int terminal_columns;
 
 
     /***************************************************************************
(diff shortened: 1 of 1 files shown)


don-clugston-sociomantic:
Why not make the default width of 80 columns, rather than 0?

nemanja-boric-sociomantic:
It's a bit confusing and I don't have a good answer. The default terminal width is whatever reported by the terminal. In case that's 0, we can make this value ` = 80` and that would override the zero. But what if the width of the terminal is 120, and this makes it drop to 80? I think its clearer if this value overrides the defaults, no matter if they are zero or not.

Alternatively, this could be renamed to `fallback_terminal_width` and set to 80.

mihails-strasuns (approved):

src/ocean/io/console/AppStatus.d
```diff
@@ -445,6 +471,13 @@ public class AppStatus
         this.connectOutput(unix_socket_file,
             unix_terminal_output);
 
+        if (command.length > 0)
+        {
+            int columns;
+            toInteger(command[0], columns);
```
daniel-zullo: Optional: `this.terminal_columns` can be directly used here instead.

src/ocean/io/console/AppStatus.d
```diff
@@ -359,6 +374,16 @@ public class AppStatus
         this.stream = stream;
         this.terminal_output = terminal_output;
         this.old_terminal_size = Terminal.rows;
+
+        if (terminal_columns)
```
daniel-zullo: you might want to explicitly compare `terminal_columns > 0`

don-clugston-sociomantic:
> It's a bit confusing and I don't have a good answer. The default terminal width is whatever reported by the terminal. In case that's 0, we can make this value = 80 and that would override the zero. But what if the width of the terminal is 120, and this makes it drop to 80? I think its clearer if this value overrides the defaults, no matter if they are zero or not.

The point is that the terminal width should never be zero under any circumstances. A zero width terminal is nonsense, it's not a terminal. I guess Linux allows it to avoid a singularity in the API but we're high enough level of abstraction that we can remove it.

nemanja-boric-sociomantic:
Updated based on both @daniel-zullo-sociomantic and @don-clugston-sociomantic 's suggestions.

don-clugston-sociomantic:
LGTM

PR #631, opened 2018-09-18
Automatically resume tasks after exception

mihails-strasuns:
Replaces previous ad-hoc solution which scaled terribly with growing
amount of task kinds. Biggest problem of previous solution was that
rethrown exception could be consumed by another caller task instead of
getting to `theScheduler.exception_handler` and thus
`continueAfterThrow` was not called.

### DIFF
--- a/relnotes/continuetask.deprecation.md
+++ b/relnotes/continuetask.deprecation.md
@@ -1,0 +1,10 @@
+### Usage of `Task.continueAfterThrow` is not needed anymore
+
+`ocean.task.Task.continueAfterThrow` method has been deprecated and is now
+implemented as no-op. Calls to this method can simply be removed; the required
+cleanup behaviour is now handled automatically by the scheduler. (A task that
+rethrows an unhandled exception to the caller will now be automatically resumed
+by the scheduler for cleanup, using the delayedResume functionality.)
+
+In rare cases when tasks are used without scheduler, caller must resume task
+directly instead of calling `continueAfterThrow`.
--- a/src/ocean/task/Task.d
+++ b/src/ocean/task/Task.d
@@ -32,6 +32,7 @@ import ocean.core.Buffer;
 import ocean.core.Verify;
 import ocean.io.model.ISuspendable;
 import ocean.task.internal.TaskExtensionMixins;
+import ocean.task.IScheduler;
 
 version (UnitTest)
 {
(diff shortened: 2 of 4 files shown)


mihails-strasuns:
FYI @don-clugston-sociomantic - I think this should fix app getting into corrupted state after consuming exception like you had recently.

src/ocean/task/Task.d
```diff
@@ -638,6 +630,12 @@ unittest
 {
     // test exception forwarding semantics
 
+
+    // if scheduler was initialized before it will have one pending task from
+    // thrown exception - drop it before proceeding:
```
don-clugston-sociomantic: This sentence is hard to understand, because the "before" is ambiguous. I think it should be more like:

```d
\\ If the scheduler was initialized before, it will have one pending task from
\\ the thrown exception. Drop it before proceeding.
```

mihails-strasuns: Updated

relnotes/continuetask.deprecation.md
```diff
@@ -0,0 +1,9 @@
+### Usage of `Task.continueAfterThrow` is not needed anymore
+
+`ocean.task.Task.continueAfterThrow` method has been deprecated and is now
+implemented as no-op. Instead a task which rethrows unhandled exception to the
+caller will be automatically resumed by scheduler for cleanup using
+`delayedResume` functionality.
```
nemanja-boric-sociomantic: I don't quite get this sentence: "Instead a task which rethrows unhandled exception to the caller will be automatically resumed by scheduler for cleanup using `delayedResume` functionality"

nemanja-boric-sociomantic (approved):
Other than small question in the release note, LGTM

relnotes/continuetask.deprecation.md
nemanja-boric-sociomantic: OK, I got what you want to say. I think it's just a missing comma issue :-).

"Instead, a task which rethrows ... will be automatically resumed ... "

gavin-norman-sociomantic: Suggestion:
"
`ocean.task.Task.continueAfterThrow` method has been deprecated and is now implemented as no-op. Calls to this method can simply be removed; the required cleanup behaviour is now handled automatically by the scheduler. (A task that rethrows an unhandled exception to the caller will now be automatically resumed by the scheduler for cleanup, using the `delayedResume` functionality.)
"

mihails-strasuns:
Updated with @gavin-norman-sociomantic wording

PR #632, opened 2018-09-19
Introduce minimal bindings to OpenSSL

don-clugston-sociomantic:
These are the minimal bindings required to implement an asynchronous SSL client, and to implement RSA message signing.

### DIFF
--- a/Build.mak
+++ b/Build.mak
@@ -37,4 +37,5 @@ $O/test-httpserver: override LDFLAGS += -lglib-2.0
 
 # Link unittests to all used libraries
 $O/%unittests: override LDFLAGS += -lglib-2.0 -lpcre -lxml2 -lxslt -lebtree \
-		-lreadline -lhistory -llzo2 -lbz2 -lz -ldl -lgcrypt -lgpg-error -lrt
+		-lreadline -lhistory -llzo2 -lbz2 -lz -ldl -lgcrypt -lgpg-error -lrt \
+		-lssl -lcrypto
--- a/src/ocean/net/ssl/openssl/OpenSsl.d
+++ b/src/ocean/net/ssl/openssl/OpenSsl.d
@@ -1,0 +1,816 @@
+/******************************************************************************
+
+    C binding to OpenSSL v1.0.x
+
+    The OpenSSL library is very large. This binding includes only a tiny
+    fraction of the available functions.
+
+    copyright: Copyright (c) 2018 dunnhumby Germany GmbH. All rights reserved
+
+*******************************************************************************/
+
+module ocean.net.ssl.openssl.OpenSsl;
+
+
+import core.stdc.config : c_ulong, c_long;
+import ocean.transition;
+
+
+extern (C):
+
+/*******************************************************************************
+
+    Opaque struct which contains function pointers for SSLv2 or
+    SSLv3/TLSv1 functions. This is roughly equivalent to a virtual function
+    table, but implemented in plain C.
+
+*******************************************************************************/
+
+public struct ssl_method_st;
+
+
+/*******************************************************************************
+
+    Opaque struct which implements an SSL connection
+
+*******************************************************************************/
+
+public struct SSL;
+
+
+/*******************************************************************************
+
+    Opaque struct which implements an SSL connection context
+
+*******************************************************************************/
+
+public struct SSL_CTX;
+
+
+/*******************************************************************************
+
+    Opaque struct which implements an X509 certificate
+
+*******************************************************************************/
+
+public struct X509;
+
+
+/*******************************************************************************
+
+    The context used wile verifying an X509 certificate
+
+*******************************************************************************/
+
+public struct X509_STORE_CTX;
+
+
+/*******************************************************************************
+
+    Opaque struct which holds RSA encryption parameters
+
+*******************************************************************************/
+
+public struct RSA;
+
+
+/*******************************************************************************
+
+    Opaque struct which holds a message digest
+
+*******************************************************************************/
+
+public struct EVP_MD;
+
+
+/*******************************************************************************
+
+    Opaque struct which holds a message digest context
+
+*******************************************************************************/
+
+public struct EVP_MD_CTX;
+
+
+/*******************************************************************************
+
+    Opaque struct which holds a private key
+
+*******************************************************************************/
+
+public struct EVP_PKEY;
+
+
+/*******************************************************************************
+
+    Opaque struct which holds a private key context
+
+*******************************************************************************/
+
+public struct EVP_PKEY_CTX;
+
+
+/*******************************************************************************
+
+    Opaque struct which holds a digest encryption engine
+
+*******************************************************************************/
+
+public struct ENGINE;
+
+
+/*******************************************************************************
+
+    Retrieve the function pointers for SSLv3, or v2 if v3 is unavailable
+
+    Returns;
+        A struct with all of the function pointers pointing to the SSL v3
+        functions, falling back to v2 if v3 is unavailable
+
+*******************************************************************************/
+
+public ssl_method_st* SSLv23_method ();
+
+
+/*******************************************************************************
+
+    Initialize the SSL library by registering algorithms
+
+    Returns:
+        Always returns 1.
+
+*******************************************************************************/
+
+public int SSL_library_init ();
+
+
+/*******************************************************************************
+
+    Registers the error strings for all libcrypto and libssl function
+
+*******************************************************************************/
+
+public void  SSL_load_error_strings ();
+
+
+/*******************************************************************************
+
+    Obtains a human-readable error message
+
+    Params:
+        e = the error code
+
+    Returns:
+        A human-readable string representing the error code.
+
+*******************************************************************************/
+
+public Const!(char *) ERR_reason_error_string (c_ulong e);
+
+
+/*******************************************************************************
+
+    Empties the current thread's error queue.
+
+*******************************************************************************/
+
+public void ERR_clear_error ();
+
+
+/*******************************************************************************
+
+    Returns the earliest error code from the thread's error queue and
+    removes the entry. This function can be called repeatedly until there
+    are no more error codes to return.
+
+*******************************************************************************/
+
+public c_ulong ERR_get_error ();
+
+
+/*******************************************************************************
+
+    Obtains the result code for a TLS/SSL I/O operation.
+
+    Gets a result code for a preceding call to SSL_connect, SSL_accept,
+    SSL_do_handshake, SSL_read, SSL_read_ex, SSL_peek, SSL_peek_ex,
+    SSL_write, or SSL_writex.
+
+    The OpenSSL interface is rather brittle.
+    The current thread's error queue must be empty before the call is made,
+    and no other OpenSSL calls should be made before calling SSL_get_error.
+
+    Params:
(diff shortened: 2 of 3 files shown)


Build.mak
```diff
@@ -37,4 +37,5 @@ $O/test-httpserver: override LDFLAGS += -lglib-2.0
 
 # Link unittests to all used libraries
 $O/%unittests: override LDFLAGS += -lglib-2.0 -lpcre -lxml2 -lxslt -lebtree \
-		-lreadline -lhistory -llzo2 -lbz2 -lz -ldl -lgcrypt -lgpg-error -lrt
+		-lreadline -lhistory -llzo2 -lbz2 -lz -ldl -lgcrypt -lgpg-error -lrt \
+		-lssl -lcrypto
```
nemanja-boric-sociomantic, line 41: Oh god, is this a breaking change or not :-/.

nemanja-boric-sociomantic: If worse comes to worse, we could mark it as `+breaking`.

The reason is that ocean users do compile ocean unittests (but they don't run them) in their unittest target IIRC and I wonder if this will make them update their dependencies.

Thinking again, I don't think it will be an issue, as I never used `-lhistory` in my Build.maks

don-clugston-sociomantic: Yeah. Well actually there are no unittests, so it isn't technically required. But I think it is bad form to include the bindings without documenting what linker flags are required.

src/ocean/net/ssl/openssl/OpenSsl.d
```diff
@@ -0,0 +1,833 @@
...
+        Params:
+            ssl = the SSL object which obtained the peer certificate
+
+        Returns:
+            X509_V_OK if the verification succeeded or no peer certificate was
+            presented, ot an error code if the verification failed.
+
+    ***************************************************************************/
+
+    public c_long SSL_get_verify_result (/*const*/ SSL* ssl);
```
nemanja-boric-sociomantic: What's the reason for this not being `Const!(SSL)*` and as you have `Const!(char*) CAfile` in other method that accepts parameter as const?

don-clugston-sociomantic: I have used `Const` whenever it's ultimately a string, so that it works with string literals. In the case where they are opaque structs it doesn't really make any difference if it is const or not, I think.
Do we have a policy about constness of C bindings?

Build.mak
Geod24: @nemanja-boric-sociomantic : IIRC Makd only includes the imported modules. As no part of Ocean depends on this, it should be fine. But feel free to test :)

src/ocean/net/ssl/openssl/OpenSsl.d
```diff
@@ -0,0 +1,833 @@
+/******************************************************************************
+
+    C binding to OpenSSL
+
+    copyright: Copyright (c) 2018 dunnhumby Germany GmbH. All rights reserved
```
Geod24: This should go last, otherwise the next paragraph is part of this section.

src/ocean/net/ssl/openssl/OpenSsl.d
```diff
@@ -0,0 +1,833 @@
...
+*******************************************************************************/
+
+module ocean.net.ssl.openssl.OpenSsl;
+
+
+import core.stdc.config : c_ulong, c_long;
+import ocean.transition;
+
+
+extern (C)
```
Geod24: `extern(C):` to remove an useless level of indentation ?

src/ocean/net/ssl/openssl/OpenSsl.d
```diff
@@ -0,0 +1,833 @@
...
+
+        Retrieve the function pointers for SSLv3, or v2 if v3 is unavailable
+
+        Returns;
+            A struct with all of the function pointers pointing to the SSL v3
+            functions, falling back to v2 if v3 is unavailable
+
+    ***************************************************************************/
+
+    public ssl_method_st* SSLv23_method();
```
Geod24: Nit: Space

src/ocean/net/ssl/openssl/OpenSsl.d
```diff
@@ -0,0 +1,833 @@
...
+    /***************************************************************************
+
+        Initialize the SSL library by registering algorithms
+
+        Returns:
+            Always returns 1.
+
+    ***************************************************************************/
+
+    public int SSL_library_init ();
```
Geod24: Which OpenSSL version are you binding to ?
This one is deprecated in v1.1.0 (which has a couple of breaking changes) but v1.1.0 is only available from Bionic (18.04), Xenial ships 1.0.2

src/ocean/net/ssl/openssl/OpenSsl.d
```diff
@@ -0,0 +1,833 @@
...
+
+        Params:
+            e = the error code
+
+        Returns:
+            A human-readable string representing the error code.
+
+    ***************************************************************************/
+
+    public Const!(char *) ERR_reason_error_string (c_ulong e);
```
Geod24: Nit: `Const!(char)*`. But usually we (used to?) go with `Const!(char)*`.
One of the advantage is that it makes `auto` slightly more usable, but not a big deal either way.

src/ocean/net/ssl/openssl/OpenSsl.d
```diff
@@ -0,0 +1,833 @@
...
+        Empties the current thread's error queue.
+
+    ***************************************************************************/
+
+    public void ERR_clear_error ();
+
+
+    /***************************************************************************
+
+        Returns the earliest error code from the thread's error queue and
```
Geod24:
```
Returns:
...
```

src/ocean/net/ssl/openssl/OpenSsl.d
```diff
@@ -0,0 +1,833 @@
...
+        removes the entry. This function can be called repeatedly until there
+        are no more error codes to return.
```
Geod24: What does it returns then ?

src/ocean/net/ssl/openssl/OpenSsl.d
```diff
@@ -0,0 +1,833 @@
...
+
+    Params:
+        key = the bytes to decode
+
+    Returns:
+        A decoded structure, or NULL if an error occurs.
+
+*******************************************************************************/
+
+public RSA * decodeRSAPrivateKey (Const!(void)[] key)
```
Geod24: Nit: `RSA*`

src/ocean/net/ssl/openssl/OpenSsl.d
```diff
@@ -0,0 +1,833 @@
...
+{
+    // If this isn't null, memory corruption will happen
+
+    RSA * rsa = null;
+
+    // This variable gets corrupted by the call
```
Geod24, line 811: What do you mean corrupted ? Does it expects an lvalue because it zeroes it ?

src/ocean/net/ssl/openssl/OpenSsl.d
```diff
@@ -0,0 +1,833 @@
...
+
+/*******************************************************************************
+
+    Decodes a PKCS#1 RSAPrivateKey structure, creating an RSA object
+
+    Params:
+        key = the bytes to decode
+
+    Returns:
+        A decoded structure, or NULL if an error occurs.
```
Geod24: Nit: `null`

src/ocean/net/ssl/openssl/OpenSsl.d
```diff
@@ -0,0 +1,833 @@
...
+            1 if any cipher could be selected, 0 on complete failure
+
+    ***************************************************************************/
+
+    public int SSL_set_cipher_list (SSL* ssl, Const!(char*) str);
+
+
+    /***************************************************************************
+
+        Configure the available TLS v1.3 ciphersuites for an ssl object
```
Geod24: Wow, bleeding edge. Then you must be binding to v1.1.1, right ?

src/ocean/net/ssl/openssl/OpenSsl.d
Geod24: > Do we have a policy about constness of C bindings?

Explicitly ? No AFAIK. But most (all?) binding I've seen just sticked to what C does.
If it's a pointer to constant data, then it has a `const`.

Geod24 (commented):
Would be nice to mention the version used / expected.

src/ocean/net/ssl/openssl/OpenSsl.d
don-clugston-sociomantic: v1.0.x. We need to support Xenial.

src/ocean/net/ssl/openssl/OpenSsl.d
don-clugston-sociomantic: I mean it is set to a nonsense value by the function call. It's obscene.

src/ocean/net/ssl/openssl/OpenSsl.d
don-clugston-sociomantic: Well, there are things at the end of the file which are NOT `extern(C)`.

src/ocean/net/ssl/openssl/OpenSsl.d
don-clugston-sociomantic: Ugh. You're right.  Actually I never call either of these functions, I only ever used the default cipher list, so they don't need to be included at all.
I'll remove it from the bindings, since there is nothing else here which supports v1.1.x.

don-clugston-sociomantic:
Repushed incorporating suggestions from our dearly departed.

src/ocean/net/ssl/openssl/OpenSsl.d
mihails-strasuns: You can explicitly mark them as `extern(D)` to override default

src/ocean/net/ssl/openssl/OpenSsl.d
mihails-strasuns: We need to support both :/
This will need compile-time version flag to chose between 1.1 and 1.0

mihails-strasuns (commented):
LGTM but there is an open question about buionaic/xenial 1.0/1.1

src/ocean/net/ssl/openssl/OpenSsl.d
don-clugston-sociomantic: The call still works in Bionic, though.

don-clugston-sociomantic: In Bionic, calling this function is no longer necessary, the library performs most initialization automatically. It is still OK to call it, and that preserves backwards compatibility.

don-clugston-sociomantic:
At this stage we don't need any functionality which is SSL version-specific. This code has been tested on both xenial and bionic.

don-clugston-sociomantic:
Repushed with `extern(C):` added to remove useless indentation.

mihails-strasuns (approved):
LGTM from my side

src/ocean/net/ssl/openssl/OpenSsl.d
Geod24: Somehow reminds me of [this](http://lkml.iu.edu/hypermail/linux/kernel/1711.2/01701.html)

Geod24 (approved):

PR #635, opened 2018-09-25
Move assumeUnique helper function to ocean.core.TypeConvert

joseph-wakeling-sociomantic:
This helper function is not merely transitional: it will continue to be useful in D2-only applications.

A selective `public import` has been added to `ocean.transition` so as to avoid any breaking change for downstream apps and libraries, but the change should allow projects moving to D2 only to have a migration path away from `ocean.transition` while continuing to use `assumeUnique`.

### DIFF
--- a/src/ocean/core/TypeConvert.d
+++ b/src/ocean/core/TypeConvert.d
@@ -17,10 +17,55 @@ module ocean.core.TypeConvert;
 
 
 
-import ocean.transition;
 import ocean.meta.traits.Indirections;
+import ocean.meta.types.Qualifiers;
+import ocean.meta.types.Typedef;
 
 version ( UnitTest ) import ocean.core.Test;
+
+
+/*******************************************************************************
+
+    Trivial wrapper for a cast from any array to an array of immutable elements
+    (e.g. from any string to an immutable string), to make code more readable.
+    Its use is only legal if no one else has a reference to the contents of the
+    `input` array.  Cf. `std.exception.assumeUnique` in Phobos.
+
+    Params:
+        input = slice whose contents to cast to immutable; this reference
+            will be nullified to prevent any further access from this
+            mutable handle
+
+    Returns:
+        slice of immutable elements corresponding to the same segment of
+        memory referred to by `input`
+
+    Note:
+        D1 does not allow overloading on rvalue vs lvalue, nor does it have
+        anything similar to D2's `auto ref` feature.  At the same time, to
+        match Phobos semantics we need to nullify the slice that gets cast
+        to immutable.  Because of this our `assumeUnique` accepts only
+        rvalues: use temporary local variables to assign lvalues if any
+        need to be used with `assumeUnique`.
+
+
+*******************************************************************************/
+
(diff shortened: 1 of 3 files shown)


src/ocean/core/TypeConvert.d
```diff
@@ -22,6 +22,50 @@ import ocean.meta.traits.Indirections;
...
+    input = null;
+    return cast(Immut!(T)[]) tmp;
+}
+
+unittest
+{
+    auto s1 = "aaa".dup;
+    auto s2 = assumeUnique(s1);
+    test!("==")(s2, "aaa");
+    test!("is")(s1, null);
```
mihails-strasuns: `null` -> `mstring.init` to fix D1 build

joseph-wakeling-sociomantic:
@mihails-strasuns fixed.  I've also replaced the `ocean.transition` import in `ocean.core.TypeConvert` to avoid any possibility of circular dependencies:

```diff
diff --git a/src/ocean/core/TypeConvert.d b/src/ocean/core/TypeConvert.d
index 046d0cfe..96070f56 100644
--- a/src/ocean/core/TypeConvert.d
+++ b/src/ocean/core/TypeConvert.d
@@ -17,7 +17,8 @@ module ocean.core.TypeConvert;
 
 
 
-import ocean.transition;
+import ocean.meta.types.Qualifiers;
+import ocean.meta.types.Typedef;
 import ocean.meta.traits.Indirections;
 
 version ( UnitTest ) import ocean.core.Test;
@@ -62,7 +63,7 @@ unittest
     auto s1 = "aaa".dup;
     auto s2 = assumeUnique(s1);
     test!("==")(s2, "aaa");
-    test!("is")(s1, null);
+    test!("is")(s1, mstring.init);
 }
```

joseph-wakeling-sociomantic:
The codecov report is a dirty rotten liar.  We have exactly the same tests!

joseph-wakeling-sociomantic:
Tweaked slightly to retain alphabetical order of imports:

```diff
diff --git a/src/ocean/core/TypeConvert.d b/src/ocean/core/TypeConvert.d
index 96070f56..ddc87c03 100644
--- a/src/ocean/core/TypeConvert.d
+++ b/src/ocean/core/TypeConvert.d
@@ -17,9 +17,9 @@ module ocean.core.TypeConvert;
 
 
 
+import ocean.meta.traits.Indirections;
 import ocean.meta.types.Qualifiers;
 import ocean.meta.types.Typedef;
-import ocean.meta.traits.Indirections;
 
 version ( UnitTest ) import ocean.core.Test;
```

joseph-wakeling-sociomantic:
Just as a minor implementation note: my first draft tried to use `d1to2fix_inject` to make this `auto ref` for D2 code (on the grounds that we might as well make it easy to support lvalues for D2-only apps/libs even if we can't do it for D1 code).

That failed because `d1to2fix_inject` does not support injecting `auto` (I assume for good and deliberate reason), but I tested the validity of `auto ref` by manually editing the D2 code, and things worked.  So this would conceivably be a way to bring the function's behaviour closer to Phobos if there is willingness to support it.

PR #636, opened 2018-09-26
Add D2-only lvalue support to assumeUnique

joseph-wakeling-sociomantic:
While we cannot do anything for D1 code, this patch adds a new D2-only `assumeUnique` overload which accepts the input array by value, which will allow D2 programs to call `assumeUnique` with lvalues.

This is a shameless copy of what is done with the phobos `assumeUnique` implementation and should behave identically.  A `Credits:` section has been added to the function documentation to reflect this.


mihails-strasuns (approved):

joseph-wakeling-sociomantic:
@mihails-strasuns I'm happy to follow up with support for the other phobos `assumeUnique` overload if that would be nice to have.

PR #637, opened 2018-09-26
Stop printing at \0

andrew-stevenson-sociomantic:
address_pathnul is a buffer containing a C string

### DIFF
--- a/src/ocean/net/server/unix/UnixListener.d
+++ b/src/ocean/net/server/unix/UnixListener.d
@@ -221,12 +221,12 @@ public class UnixSocketListener ( CommandHandlerType ) : SelectListener!(
             char[0x100] buf;
             auto msgnul = strerror_r(errno, buf.ptr, buf.length);
             log.error("Unable to delete socket \"{}\": {}",
-                      this.address_pathnul[0 .. $ - 1],
+                      this.address_pathnul[0 .. strlen(this.address_pathnul.ptr)],
                       msgnul[0 .. strlen(msgnul)]);
         }
         else
         {
-            log.info("Deleted socket \"{}\"", this.address_pathnul[0 .. $ - 1]);
+            log.info("Deleted socket \"{}\"", this.address_pathnul[0 .. strlen(this.address_pathnul.ptr)]);
         }
     }
 }


src/ocean/net/server/unix/UnixListener.d
```diff
@@ -226,7 +227,7 @@ public class UnixSocketListener ( CommandHandlerType ) : SelectListener!(
         }
         else
         {
-            log.info("Deleted socket \"{}\"", this.address_pathnul[0 .. $ - 1]);
```
matthias-wende-sociomantic, line 229: I can't see what is wrong with the old code. Can you explain this please? Also I think it would be good to apply the same change to line 224.

nemanja-boric-sociomantic: The `address_pathnul` is a static array of size `sockaddr_un.sun_path.length`. So if it stores a string of three elements, and assuming that `sockaddr_un.sun_path.length` is 12 elements, the old code will print the first three elements, then nul and finally garbage after it (7 characters, because it will not print the last element).

nemanja-boric-sociomantic: > Also I think it would be good to apply the same change to line 224.

:+1:

andrew-stevenson-sociomantic: `address_pathnul` is a D char array - i.e a struct with at least `length` and `ptr`. https://github.com/sociomantic-tsunami/ocean/blob/bb525d17a230c87e4cf0a7443c3a473266dcb5d9/src/ocean/net/server/unix/UnixListener.d#L134
Its a static array so its length will always be the same (just over 100 bytes on most platforms IIRC).

This array is filled with the path to the socket and a nul byte as we are passing it to a C API where that is the representation of strings.
https://github.com/sociomantic-tsunami/ocean/blob/bb525d17a230c87e4cf0a7443c3a473266dcb5d9/src/ocean/net/server/unix/UnixListener.d#L165
If we imagine our path was, say 20 bytes long and `address_pathnul` is 100 bytes long then we end up with 20 bytes of path, one byte of nul, and 79 bytes of unitialised memory (though as this is D they will be 0xFF IIRC).

At line 229 we print `address_pathnul[0 .. $ - 1]`.
https://github.com/sociomantic-tsunami/ocean/blob/6ed6af575862ffbd61812a6b8fe78ca40ab60078/src/ocean/net/server/unix/UnixListener.d#L229
That's bytes 0 to the length of `address_pathnul` - 2. Using our example above then we print bytes 0 to 98 inclusive which will be 20 bytes of our path, one nul byte, and 77 bytes of 0xFFs. What we really want to do is print from byte 0 until the first nul byte. There are many ways to achieve this and this PR does one of them - by creating a new struct where `ptr` points to `address_pathnul.ptr` but length is set to `strlen(address_pathnul.ptr)`.

You are right that line 224 looks buggy as well - I just didn't hit that one.
https://github.com/sociomantic-tsunami/ocean/blob/bb525d17a230c87e4cf0a7443c3a473266dcb5d9/src/ocean/net/server/unix/UnixListener.d#L225

nemanja-boric-sociomantic: Sigh, there goes my poor explanation :'(

> though as this is D they will be 0xFF IIRC).

that is correct.

matthias-wende-sociomantic: Thanks, for that comprehensive explanation :).

andrew-stevenson-sociomantic: and now I look at 224 I see the same problem has already been tackled a different way in this file so I will do it the same to remain consistent.

andrew-stevenson-sociomantic:
Updated to fix the similar bug spotted by @matthias-wende-sociomantic and to use `strlen` directly as per the pre-existing code.

nemanja-boric-sociomantic:
I've changed the base of this PR to include it into a patch release, could you just rebase on v4.2.x and force push?

src/ocean/net/server/unix/UnixListener.d
```diff
@@ -19,6 +19,7 @@ module ocean.net.server.unix.UnixListener;
 import ocean.net.server.unix.UnixConnectionHandler;
 import ocean.net.server.SelectListener;
 import ocean.io.select.EpollSelectDispatcher;
+import ocean.text.util.StringC;
```
daniel-zullo: I guess this import is no longer needed (but you might need to import `ocean.stdc.string` to get `strlen()`)

andrew-stevenson-sociomantic: Probably right about not needing the import any more. `strlen` was already in use in this file (which is why I switched techniques) so someone else must have already imported it for me :-)

andrew-stevenson-sociomantic:
Rebased

nemanja-boric-sociomantic:
Thanks!

nemanja-boric-sociomantic (approved):

nemanja-boric-sociomantic (approved):

PR #638, opened 2018-10-04
Ignore empty lines while parsing stats files

bogdan-szabo-sociomantic:
I just figured out that the stats files always end with an empty end line and it will crash the parser. This pr addresses this issue. 

related to  #629

### DIFF
--- a/src/ocean/util/log/StatsReader.d
+++ b/src/ocean/util/log/StatsReader.d
@@ -83,6 +83,7 @@ public struct StatsLine
 
     static public StatsLine opCall (cstring line)
     {
+        enforce(line.length > 0, "Can not parse an empty line");
         StatsLine stats_line;
         auto len = line.length;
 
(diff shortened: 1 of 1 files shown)


src/ocean/util/log/StatsReader.d
```diff
@@ -252,6 +253,11 @@ public class StatsLogReader
 
         foreach (line; this.lines)
         {
+            if (line.length == 0)
```
nemanja-boric-sociomantic: It seems that it might be better to wrap `this.lines` into a small struct which provides `opApply` and just ignores empty lines, then to repeat this over and over?

bogdan-szabo-sociomantic:
@nemanja-boric-sociomantic done

src/ocean/util/log/StatsReader.d
```diff
@@ -215,13 +216,56 @@ unittest
 
 public class StatsLogReader
 {
+    /**************************************************************************
+
+        Struct used to iterate through valid stat lines
```
nemanja-boric-sociomantic: I suggest we extend documentation to say that it's silently ignoring empy stat lines, and that no other validity check (such as if the stat line is well formed) is performed, as "valid stat lines" is not really true - you can have non-empty but broken stats line.

bogdan-szabo-sociomantic: done

nemanja-boric-sociomantic (approved):

nemanja-boric-sociomantic:
Thanks!

PR #640, opened 2018-10-08
Add suspendables to throttler in consistent state 

matthias-wende-sociomantic:
This change ensures that the suspend state of an ISuspendable instance
when added to the throttler is consistent with the throttler state.

This patch somehow also works around the the following problem.

If a RequestSuspender instance is reused by a new Request it'll set the suspend_requested flag to false.

https://github.com/sociomantic-tsunami/dhtproto/blob/v14.x.x/src/dhtproto/client/legacy/internal/request/model/IBulkGetRequest.d#L133
https://github.com/sociomantic-tsunami/swarm/blob/v5.x.x/src/swarm/client/request/helper/RequestSuspender.d#L148

When adding this instance to the throttlers list of suspendables it will be already present and therefore neither appended to the list of suspendables nor suspended.

https://github.com/sociomantic-tsunami/ocean/blob/v4.x.x/src/ocean/io/model/ISuspendableThrottler.d#L77

Now if at the same time the application (i.e. the applications throttler) is in suspended state, the request will be started in a non suspended state and will not get suspended at least until the throttler resumes and suspends again.

https://github.com/sociomantic-tsunami/ocean/blob/v4.x.x/src/ocean/io/model/ISuspendableThrottler.d#L139

**Note** that that doesn't address the issue that a suspendable is kept in the list of suspendables when a request finishes (compere also the comment below). The possibility for clients to unregister suspendables from the Throttller when a request finishes will be addressed later.

### DIFF
--- a/src/ocean/io/model/ISuspendableThrottler.d
+++ b/src/ocean/io/model/ISuspendableThrottler.d
@@ -91,6 +91,63 @@ abstract public class ISuspendableThrottler
         }
     }
 
+    version (UnitTest){
+        import ocean.core.Test;
+    }
+    unittest
+    {
+        class SuspendableThrottler : ISuspendableThrottler
+        {
+            override protected bool suspend ( )
+            {
+                return false;
+            }
+
+            override protected bool resume ( )
+            {
+                return false;
+            }
+        }
+
+        class Suspendable : ISuspendable
+        {
+            bool suspended_ = false;
+
+            public void suspend ( )
+            {
+                this.suspended_ = true;
+            }
+
+            public void resume ( )
+            {
+                this.suspended_ = false;
+            }
+
+            public bool suspended ( )
+            {
+                return suspended_;
+            }
+        }
+
+        auto suspendable_throttler = new SuspendableThrottler;
+        auto suspendable = new Suspendable;
+
+        // add a suspended ISuspendable to a non suspended throttler
+        suspendable.suspend();
+        suspendable_throttler.addSuspendable(suspendable);
+
+        test!("==")(suspendable.suspended(), false,
+            "Suspended suspendable not resumed.");
+
+        suspendable_throttler.suspendAll();
+
+        // add a resumed ISuspendable to a suspended throttler
+        suspendable.resume();
+        suspendable_throttler.addSuspendable(suspendable);
+
+        test!("==")(suspendable.suspended(), true,
+            "Resumed suspendable not suspended.");
+    }
 
     /***************************************************************************
 
--- a/src/ocean/io/model/ISuspendableThrottler.d
+++ b/src/ocean/io/model/ISuspendableThrottler.d
@@ -65,7 +65,9 @@ abstract public class ISuspendableThrottler
     /***************************************************************************
 
         Adds a suspendable to the list of suspendables which are to be
-        throttled. If it is already in the list, nothing happens.
+        throttled if it's not already in there.
+        Also ensures that the state of the added suspendable is consistent
+        with the state of the throttler.
 
         Params:
             s = suspendable
@@ -77,10 +79,15 @@ abstract public class ISuspendableThrottler
         if ( !this.suspendables.contains(s) )
         {
             this.suspendables ~= s;
-            if (this.suspended_)
-            {
-                s.suspend();
-            }
+        }
+
+        if (this.suspended_ && !s.suspended())
+        {
+            s.suspend();
+        }
+        else if (!this.suspended_ && s.suspended())
+        {
+            s.resume();
         }
     }
 


src/ocean/io/model/ISuspendableThrottler.d
```diff
@@ -77,10 +77,11 @@ abstract public class ISuspendableThrottler
...
             this.suspendables ~= s;
-            if (this.suspended_)
-            {
-                s.suspend();
-            }
+        }
+
+        if (this.suspended_ && !s.suspended())
+        {
+            s.suspend();
```
daniel-zullo, line 86: should it be suspended if it isn't added to the `suspendables`? Anyway if there is any reason to do so then the documentation of the method should be updated.

daniel-zullo:
If this is a bug fix then it would be good to have/add a unit test

src/ocean/io/model/ISuspendableThrottler.d
matthias-wende-sociomantic: Yes that's the whole point. Recycled instances are in the list and they should be suspended.

matthias-wende-sociomantic: And good point. I updated the doc accordingly ;).

matthias-wende-sociomantic:
> If this is a bug fix then it would be good to have/add a unit test

Yeah, thanks for pointing that out. I've added one for this particular case.

gavin-norman-sociomantic:
Hmm wasn't the idea originally to add a method allowing the user to remove a suspendable from the throttler?

matthias-wende-sociomantic:
Yes, that was the old plan before this new one :).
I favour this solution for several reasons.
- It fixes the bug as well.
- It is much simpler (i.e. no changes in swarm and dhtproto).
- It fixes the bug without changing clients.

At the same time the original idea introduced some new regression to the client and I have ~no~ intension to track this down ~if this patch works~ _later_.

gavin-norman-sociomantic:
You describe this situation:
1. A suspendable is registered with the throttler.
2. The suspendable ceases to be valid (e.g. the request it represented finished) and is recycled into a pool of suspendables. It is not unregistered from the throttler.
3. The suspendable is acquired from the pool and reinitialised.
4. The suspendable is registered with the throttler. As it is already registered, some special behaviour is required.

What I'm concerned about is this: what happens if, after step 2, the throttler tries to suspend? It would be calling a method of the recycled suspendable, which would do...? The fact that invalid suspendables can (must, currently) stay registered with the throttler seems like a bug to me, not something that we should patch up to make work.

What do you think?

gavin-norman-sociomantic:
Note that a recycled invalid suspendable is only one case. What if the invalid suspendable has been garbage collected, say?

matthias-wende-sociomantic:
Yes that's a valid point a should be addressed as well. My description of this patch was somehow misleading. I've updated the PR and commit descriptions to what this patch is actually about.

src/ocean/io/model/ISuspendableThrottler.d
```diff
@@ -91,6 +91,64 @@ abstract public class ISuspendableThrottler
...
+        suspendable.suspend();
+        suspendable_throttler.addSuspendable(suspendable);
+
+        test!("==")(suspendable.suspended(), false, "Unit test failed, " ~
+            "suspended suspendable not resumed.");
+
+        suspendable_throttler.suspendAll();
+
+        // add a resumed ISuspendable to a suspended throttler
+        // note that the instance is "recycled"
```
gavin-norman-sociomantic: This one comment probably isn't needed any more.

src/ocean/io/model/ISuspendableThrottler.d
```diff
@@ -91,6 +91,64 @@ abstract public class ISuspendableThrottler
...
+        }
+
+        auto suspendable_throttler = new SuspendableThrottler;
+        auto suspendable = new Suspendable;
+
+        // add a suspended ISuspendable to a non suspended throttler
+        suspendable.suspend();
+        suspendable_throttler.addSuspendable(suspendable);
+
+        test!("==")(suspendable.suspended(), false, "Unit test failed, " ~
```
gavin-norman-sociomantic: Generally, I don't think you need to write "Unit test failed". It's clear from the context.

src/ocean/io/model/ISuspendableThrottler.d
```diff
@@ -91,6 +91,64 @@ abstract public class ISuspendableThrottler
...
+            "suspended suspendable not resumed.");
+
+        suspendable_throttler.suspendAll();
+
+        // add a resumed ISuspendable to a suspended throttler
+        // note that the instance is "recycled"
+        suspendable.resume();
+        suspendable_throttler.addSuspendable(suspendable);
+
+        test!("==")(suspendable.suspended(), true, "Unit test failed, " ~
```
gavin-norman-sociomantic: Generally, I don't think you need to write "Unit test failed". It's clear from the context.

gavin-norman-sociomantic (commented):
Yep, the new explanation makes much more sense, outside of the context of the bug you were investigating.

gavin-norman-sociomantic:
It seems like this should be a patch release.

matthias-wende-sociomantic:
Updated, base branch changed + rebased on top of new branch, accidentally deleted branch, recovered the branch and finally reopened the PR :).

gavin-norman-sociomantic (approved):

src/ocean/io/model/ISuspendableThrottler.d
```diff
@@ -77,13 +79,75 @@ abstract public class ISuspendableThrottler
...
+        }
+
+        if (this.suspended_ && !s.suspended())
+        {
+            s.suspend();
+        }
+        else if (!this.suspended_ && s.suspended())
+        {
+            s.resume();
+        }
```
daniel-zullo, line 91: Why make it simple if we could make it complicated?

```D
        if (this.suspended_ != s.suspended())
        {
            if (this.suspended_)
                s.suspend();
            else
                s.resume();
        }
```

src/ocean/io/model/ISuspendableThrottler.d
```diff
@@ -77,13 +79,75 @@ abstract public class ISuspendableThrottler
...
+    }
+
+    version (UnitTest){
```
daniel-zullo, line 94: the format/style is wrong

src/ocean/io/model/ISuspendableThrottler.d
```diff
@@ -77,13 +79,75 @@ abstract public class ISuspendableThrottler
...
+            }
+
+            public void resume ( )
+            {
+                this.suspended_ = false;
+            }
+
+            public bool suspended ( )
+            {
+                return suspended_;
```
daniel-zullo, line 128: missing `this.`

src/ocean/io/model/ISuspendableThrottler.d
gavin-norman-sociomantic: This PR has been merged, but that's a good point.

matthias-wende-sociomantic: Yep, that's looks indeed better.

Issue #643, opened 2018-10-11
Handle tasks spawned by a task pool in SpecializedPools


daniel-zullo:
The `SpecializedPools`  is not used if the task is spawned/started from another object rather than a plain task, for instance, spawned by a TaskPool. 
`SpecializedPools` works fine if a plain task is directly scheduled.

```D
unittest
{
    const NUM_START_CALLS = 100;

    static class SubTask : Task
    {
        static int value;

        public void copyArguments ()
        {
        }

        override public void run ()
        {
            // only works using the specialized pool
            long[5_000] big_array;

            .wait(20); // so that the task gets suspended

            SubTask.value++;
        }
    }

    class PoolTask : Task
    {
        TaskPool!(SubTask) my_task_pool;

        public this ()
        {
            this.my_task_pool = new TaskPool!(SubTask);
        }

        public void run ()
        {
            for (int i = 0; i < NUM_START_CALLS; i++)
                this.my_task_pool.start();

            this.my_task_pool.awaitRunningTasks();
        }
    }

    class PlainTask : Task
    {
        SubTask[NUM_START_CALLS] my_task_array;

        public this ()
        {
            foreach (ref task; this.my_task_array)
                task = new SubTask();
        }

        public void run ()
        {
            foreach (ref task; this.my_task_array)
            {
                theScheduler.schedule(task);
            }
        }
    }

    SchedulerConfiguration config;
    config.task_queue_limit = 200;
    config.worker_fiber_limit = 10;
    config.suspended_task_limit = 20;
    config.worker_fiber_stack_size = 32_768;

    config.specialized_pools = [
        config.PoolDescription(SubTask.classinfo.name, 65_536)
    ];

    initScheduler(config);
    theScheduler.schedule(new PlainTask()); // Works
    theScheduler.schedule(new PoolTask());  // Fails: segfault
    theScheduler.eventLoop();

    test!("==")(SubTask.value, NUM_START_CALLS * 2);
}
```

daniel-zullo:
This bug/issue is solved by updating the method `findPool()` to: 

```diff
diff --git a/src/ocean/task/internal/SpecializedPools.d b/src/ocean/task/internal/SpecializedPools.d
index 2dbdd71..12c3c08 100644
--- a/src/ocean/task/internal/SpecializedPools.d
+++ b/src/ocean/task/internal/SpecializedPools.d
@@ -14,10 +14,12 @@
 
 module ocean.task.internal.SpecializedPools;
[...]

Issue #461, opened 2018-02-26
Deprecate usage of task as `ISuspendable`


mihails-strasuns-sociomantic:
Currently ocean implemeted `ISuspendable` interface for tasks and uses that in some examples with suspendable throttler. It is not correct though - throttler expects to be able resume and suspend from any context and with tasks it is only legal to suspend from within a task and resume externally.

mihails-strasuns-sociomantic:
Sadly, one can't deprecate implementing an interface - thus this will have to wait for v5 to be directly removed.

mihails-strasuns:
Moving to `4.4.0+breaking` because of how much of a negative impact this issue brings.

PR #645, opened 2018-10-12
Remove `OwnedTask` from task pools

mihails-strasuns:
It was added as a trick to recycle tasks before termination hook
existed. Now it has become liability.

### DIFF
--- a/relnotes/taskpool.feature.md
+++ b/relnotes/taskpool.feature.md
@@ -1,0 +1,15 @@
+### Change of `StreamProcessor`, `TaskPool` and `ThrottledTaskPool` internals
+
+Also affects `StreamProcessor` as it uses `ThrottledTaskPool` internally.
+
+Both `TaskPool` and `ThrottledTaskPool` stopped defining internal private
+`OwnedTask` to wrap user-supplied type in. Instead, regular termination hook
+infrastructure is used.
+
+Both also don't try to catch and rethrow exceptions arising from a task. This is
+not needed because recycling is already done by the termination hook system.
+
+Overall, only expected observable impact is fixed interaction between
+specialized worker fiber pools and `TaskPool`. However, because of how different
+internal logic is now, it is highly recommended to carefully test applications
+when upgrading.
--- a/src/ocean/task/TaskPool.d
+++ b/src/ocean/task/TaskPool.d
@@ -75,32 +75,6 @@ class TaskPool ( TaskT : Task ) : ObjectPool!(Task)
 
     public alias TaskT TaskType;
 
-    /***************************************************************************
-
-        Internal helper class derived from user-supplied task type that is used
-        to track reference to outer pool for recycling.
-
-        Params:
-            TaskT = user-supplied task type
-
-    ***************************************************************************/
-
-    protected class OwnedTask : TaskT
-    {
-        /***********************************************************************
-
-            Recycles the task by first calling its own recycle method to
-            reset it to a clean state, and returns it to the pool of
-            available tasks instances afterwards.
-
-        ***********************************************************************/
-
-        override public void recycle ( )
-        {
-            super.recycle();
-            this.outer.recycle(this);
-        }
-    }
 
     /***************************************************************************
 
@@ -122,28 +96,28 @@ class TaskPool ( TaskT : Task ) : ObjectPool!(Task)
         if (this.num_busy() >= this.limit())
             return false;
 
-        auto task = cast(TaskT) this.get(new OwnedTask);
+        auto task = cast(TaskT) this.get(new TaskT);
         assert (task !is null);
 
-        try
-        {
-            task.copyArguments(args);
-            theScheduler.schedule(task);
-        }
-        catch (TaskKillException e)
-        {
-            // don't try recycling task upon TaskKillException as this is not
-            // normal code flow and it may have already been recycled by
-            // finishing on its own
-            throw e;
-        }
-        catch (Exception e)
-        {
-            this.recycle(task);
-            throw e;
-        }
-
+        task.copyArguments(args);
+        this.startImpl(task);
         return true;
+    }
+
+    /***************************************************************************
+
+        Common part of start implementation reused by derivatives. Split into
+        separate method to ensure that recycling hook won't be omitted.
+
+        Params:
+            task = already setup task to run and recyle
+
+    ***************************************************************************/
+
+    protected void startImpl ( Task task )
+    {
+        task.terminationHook(&this.taskTerminationHook);
+        theScheduler.schedule(task);
     }
 
     /***************************************************************************
@@ -227,29 +201,26 @@ class TaskPool ( TaskT : Task ) : ObjectPool!(Task)
             if (this.num_busy() >= this.limit())
                 return false;
 
-            auto task = cast(TaskT) this.get(new OwnedTask);
+            auto task = cast(TaskT) this.get(new TaskT);
             assert (task !is null);
 
-            try
-            {
-                task.deserialize(serialized);
-                theScheduler.schedule(task);
-            }
-            catch (TaskKillException e)
-            {
-                // don't try recycling task upon TaskKillException as this is not
-                // normal code flow and it may have already been recycled by
-                // finishing on its own
-                throw e;
-            }
-            catch (Exception e)
-            {
-                this.recycle(task);
-                throw e;
-            }
+            task.deserialize(serialized);
+            this.startImpl(task);
 
             return true;
         }
+    }
+
+    /***************************************************************************
+
+        Used to recycle pool tasks when they finish
+
+    ***************************************************************************/
+
+    private void taskTerminationHook ( )
+    {
+        auto task = Task.getThis();
+        this.recycle(task);
     }
 }
 
--- a/src/ocean/task/ThrottledTaskPool_test.d
+++ b/src/ocean/task/ThrottledTaskPool_test.d
@@ -43,16 +43,24 @@ class ProcessingTask : Task
     int x;
 
     static int total;
+    static size_t recycle_count;
 
     void copyArguments ( int x )
     {
         this.x = x;
     }
 
+    override public void recycle ( )
+    {
+        recycle_count++;
+    }
+
     override void run ( )
     {
-        .wait(100);
+        test!("==")(theScheduler.getStats.worker_fiber_busy, 0);
+
         ++total;
+        .wait(100);
 
         if (this.x == 1000)
             theScheduler.shutdown();
@@ -81,7 +89,7 @@ class Generator : ISuspendable
 
     void start ( )
     {
-        this.timer.set(0, 10, 0, 10);
+        this.timer.set(0, 1, 0, 1);
         this.resume();
     }
 
@@ -111,8 +119,14 @@ class Generator : ISuspendable
 unittest
 {
     SchedulerConfiguration config;
-    config.worker_fiber_limit = 10;
-    config.task_queue_limit = 30;
+    with (config)
+    {
+        worker_fiber_limit = 10;
+        task_queue_limit = 30;
+        specialized_pools = [
+            PoolDescription(ProcessingTask.classinfo.name, 10240)
+        ];
+    }
     initScheduler(config);
 
     auto generator = new Generator;
@@ -120,9 +134,11 @@ unittest
     generator.start();
     theScheduler.eventLoop();
 
+    test!("==")(ProcessingTask.recycle_count, ProcessingTask.total);
+    test!("==")(generator.pool.num_busy(), 0);
+
     // exact number of tasks that will be processed before the shutdown
     // may vary but it must always be at most 1000 + task_queue_limit
     test!(">=")(ProcessingTask.total, 1000);
     test!("<=")(ProcessingTask.total, 1000 + config.task_queue_limit);
-
 }
--- a/src/ocean/task/ThrottledTaskPool_test.d
+++ b/src/ocean/task/ThrottledTaskPool_test.d
@@ -24,34 +24,12 @@ import ocean.task.util.Timer;
 
 import ocean.io.model.ISuspendable;
 import ocean.io.Stdout;
+import ocean.io.select.client.TimerEvent;
+
 import ocean.core.Test;
 import ocean.core.Verify;
 
 static import core.thread;
-
-/*******************************************************************************
-
-    "stream" or "generator" class, one which keeps producing new records for
-    processing at throttled rate
-
-*******************************************************************************/
-
-class Generator : Task
-{
-    bool delegate(int) process_dg;
-
-    this ( typeof(this.process_dg) dg )
-    {
-        this.process_dg = dg;
-    }
-
-    override void run ( )
-    {
-        int i;
-        while (++i)
-            verify(process_dg(i));
-    }
-}
 
 /*******************************************************************************
 
@@ -81,6 +59,55 @@ class ProcessingTask : Task
     }
 }
 
+/*******************************************************************************
+
+    "stream" or "generator" class, one which keeps producing new records for
+    processing at throttled rate
+
+*******************************************************************************/
+
+class Generator : ISuspendable
+{
+    TimerEvent timer;
+    ThrottledTaskPool!(ProcessingTask) pool;
+    int counter;
+
+    this ( )
+    {
+        this.timer = new TimerEvent(&this.generate);
+        this.pool = new ThrottledTaskPool!(ProcessingTask)(10, 0);
+        this.pool.throttler.addSuspendable(this);
+    }
+
+    void start ( )
+    {
+        this.timer.set(0, 10, 0, 10);
+        this.resume();
+    }
+
+    override void resume ( )
+    {
+        theScheduler.epoll.register(this.timer);
+    }
+
+    override void suspend ( )
+    {
+        theScheduler.epoll.unregister(this.timer);
+    }
+
+    override bool suspended ( )
+    {
+        return this.timer.is_registered;
+    }
+
+    bool generate ( )
+    {
+        this.pool.start(this.counter);
+        ++this.counter;
+        return true;
+    }
+}
+
 unittest
 {
     SchedulerConfiguration config;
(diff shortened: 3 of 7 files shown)


mihails-strasuns:
@don-clugston-sociomantic @daniel-zullo-sociomantic can you please test this with your application use cases?

daniel-zullo:
The unit tests in https://github.com/sociomantic-tsunami/ocean/issues/643#issue-369011733 are similar to some of our use cases and you might want to add them (if you haven't done it yet)

src/ocean/task/util/StreamProcessor_test.d
```diff
@@ -21,75 +21,106 @@ import ocean.task.util.Timer;
...
     SchedulerConfiguration config;
     config.worker_fiber_limit = 10;
     config.task_queue_limit = 30;
-    initScheduler(config);
-
-    auto throttler_config = ThrottlerConfig(10, 1);
-    auto stream_processor = new StreamProcessor!(ProcessingTask)(throttler_config);
-    auto generator = new Generator(&stream_processor.process);
-    stream_processor.addStream(generator);
+    initScheduler(config, epoll);
```
daniel-zullo: I don't think it is needed to define and pass `epoll` to the scheduler as the scheduler would instantiate it and then `theScheduler.epoll()` can be passed to the Generator. But I might be missing something

don-clugston-sociomantic (commented):
I have tested this, it definitely fixes the problem and works in my use case.

src/ocean/task/util/StreamProcessor_test.d
mihails-strasuns: Correct, it is simply a finger memory from me.

daniel-zullo:
I am afraid the current patch does not work with my apps. I am investigating and I'll report back once I find it out

daniel-zullo:
One problem is calling `awaitRunningTasks()` from a `TaskPool` freezes the application.

```
[ocean.task.internal.FiberPoolWithQueue] Recycling worker fiber <0X00007FF014893080>
[ocean.task.util.Event] Resuming task 0X00007FF016C78300 by trigger of event 0X00007FF013044C00
[ocean.task.Task] <0X00007FF016C78300> has been resumed by main thread or event loop
[ocean.task.Task] <0X00007FF016C78300> termination (end of main function)
[ocean.task.Task] Calling 2 termination_hooks for task <0X00007FF016C78300>
[ocean.task.Task] Recycling task <0X00007FF016C78300>
[ocean.task.internal.FiberPoolWithQueue] Recycling worker fiber <0X00007FF014896A00>
[ocean.task.util.Event] Resuming task 0X00007FF016C78100 by trigger of event 0X00007FF00E03CC00
[ocean.task.Task] <0X00007FF016C78100> has been resumed by main thread or event loop
[ocean.task.Task] <0X00007FF016C78100> termination (end of main function)
[ocean.task.Task] Calling 2 termination_hooks for task <0X00007FF016C78100>
[ocean.task.Task] Recycling task <0X00007FF016C78100>
[ocean.task.internal.FiberPoolWithQueue] Recycling worker fiber <0X00007FF0149B5B00>
[ocean.task.util.Event] Resuming task 0X00007FF016C79E00 by trigger of event 0X00007FF00EBB6C00
[ocean.task.Task] <0X00007FF016C79E00> has been resumed by main thread or event loop
[ocean.task.Task] <0X00007FF016C79E00> termination (end of main function)
[ocean.task.Task] Calling 2 termination_hooks for task <0X00007FF016C79E00>
[ocean.task.Task] Recycling task <0X00007FF016C79E00>
[ocean.task.internal.FiberPoolWithQueue] Recycling worker fiber <0X00007FF01498EF00>
[ocean.task.util.Event] Resuming task 0X00007FF016C79900 by trigger of event 0X00007FF0124E4C00
[ocean.task.Task] <0X00007FF016C79900> has been resumed by main thread or event loop
[ocean.task.Task] <0X00007FF016C79900> termination (end of main function)
[ocean.task.Task] Calling 2 termination_hooks for task <0X00007FF016C79900>
[ocean.task.Task] Recycling task <0X00007FF016C79900>
[ocean.task.internal.FiberPoolWithQueue] Recycling worker fiber <0X00007FF0148C1B00>
```

Calling 2 `temination_hooks` does not seem to be the right thing.

daniel-zullo:
```D
unittest
{
    const NUM_RECORDS = 10_000;

    static class Generator : Task
    {
        void delegate(int) process_dg;

        void copyArguments (typeof(this.process_dg) dg)
        {
            this.process_dg = dg;
        }

        override void run ()
        {
            for (int i = 0; i < NUM_RECORDS; i++)
            {
                this.process_dg(i);
                .wait(10);
            }
        }
    }

    static class ProcessingTask : Task
    {
        int counter;

        static int num_processed_tasks;

        override void run ()
        {
            auto mod = counter % 6;
            switch (mod)
            {
                case 0:
                    .wait(mod * 10000);
                    break;

                case 1:
                    .wait(mod * 100);
                    break;

                case 2:
                    .wait(mod * 30000);
                    break;

                case 3:
                    .wait(mod * 1_000);
                    break;

                case 4:
                    .wait(mod * 100);
                    break;

                case 5:
                    .wait(mod * 5000);
                    break;

                default:
                    assert(false);
            }

            num_processed_tasks++;
        }

        void copyArguments (int counter)
        {
            this.counter = counter;
        }
    }

    static class MainTask : Task
    {
        override void run ()
        {
            // suspend once because `await` safeguards against being run
            // before the scheduler starts
            theScheduler.processEvents();

            auto stream_processor = new StreamProcessor!(ProcessingTask)(
                ThrottlerConfig.init);

            auto generator = new Generator();
            generator.copyArguments(&stream_processor.process);
            stream_processor.addStream(generator);

            theScheduler.await(generator);
        }
    }

    SchedulerConfiguration config;
    config.task_queue_limit = 1_000;
    config.worker_fiber_limit = 100;
    config.suspended_task_limit = 200;

    initScheduler(config);

    test!("==")(ProcessingTask.num_processed_tasks, 0);

    theScheduler.schedule(new MainTask);

    theScheduler.eventLoop();

    test!("==")(ProcessingTask.num_processed_tasks, NUM_RECORDS);
}
```


The unit test above fails with this patch too.
```
ocean.task.ThrottledTaskPool] Throttler.resume(200) : used = 202, total = 1000, pool.busy = 3672, pool.limit = 18446744073709551615
[ocean.task.Task] Recycling task <0X00007F4A08C22600>
[ocean.task.internal.FiberPoolWithQueue] Reusing worker fiber <0X00007F4A08796E80> to run scheduled task <0X00007F4A08C22380>
[ocean.task.Task] <0X00007F4A08C22380> start of main function
[ocean.task.util.Timer] Suspending task <0X00007F4A08C22380> for 60000 microseconds
[ocean.task.Task] <0X00007F4A08C22380> is suspending itself
[ocean.task.util.Timer] Resuming task <0X00007F4A08BE9600> by timer event
[ocean.task.Task] <0X00007F4A08BE9600> has been resumed by main thread or event loop
[ocean.task.Task] <0X00007F4A08BE9600> termination (end of main function)
[ocean.task.Task] Calling 1 termination_hooks for task <0X00007F4A08BE9600>
[ocean.task.ThrottledTaskPool] Throttler.resume(200) : used = 201, total = 1000, pool.busy = 3672, pool.limit = 18446744073709551615
[ocean.task.Task] Recycling task <0X00007F4A08BE9600>
[ocean.task.internal.FiberPoolWithQueue] Reusing worker fiber <0X00007F4A08727A80> to run scheduled task <0X00007F4A08C24F00>
[ocean.task.Task] <0X00007F4A08C24F00> start of main function
[ocean.task.util.Timer] Suspending task <0X00007F4A08C24F00> for 3000 microseconds
[ocean.task.Task] <0X00007F4A08C24F00> is suspending itself
[ocean.task.util.Timer] Resuming task <0X00007F4A08B8C380> by timer event
[ocean.task.Task] <0X00007F4A08B8C380> has been resumed by main thread or event loop
[ocean.task.Task] <0X00007F4A08B8C380> termination (end of main function)
[ocean.task.Task] Calling 1 termination_hooks for task <0X00007F4A08B8C380>
[ocean.task.ThrottledTaskPool] Throttler.resume(200) : used = 200, total = 1000, pool.busy = 3672, pool.limit = 18446744073709551615
Unhandled exception in epoll/scheduler
	Use `theScheduler.resume(task)` to resume a task from termination hook of another task (./src/ocean/task/Task.d:338)
Aborted (core dumped)
```

mihails-strasuns:
Thanks for the report! I will add this as test and investigate.

mihails-strasuns:
@daniel-zullo-sociomantic finally got to look into this and looks like your sample is not intended to work by design. Check last lines explaining the crash:

```
Unhandled exception in epoll/scheduler
	Use `theScheduler.resume(task)` to resume a task from termination hook of another task (./src/ocean/task/Task.d:338)
Aborted (core dumped)
```

This happens because you are using `Task` as `ISuspendable` and that is a historical mistake that will be removed in v5.0.0 (https://github.com/sociomantic-tsunami/ocean/issues/461) - and is not guaranteed to work with current code base either. If I replace it with epoll-based generator, your sample works for me.

daniel-zullo:
right, my unit test example was just an extension of the stream processor test done in https://github.com/sociomantic-tsunami/ocean/blob/v4.x.x/src/ocean/task/util/StreamProcessor_test.d#L28-L99 so it'd mean that example is wrong too.

Anyway my app has another problem, it freezes with this patch as I mentioned in a comment above. I see if can get more info about it or try to reproduce the problem in a unit test

mihails-strasuns:
That example is changed as part of this PR - it was indeed wrong :)

src/ocean/task/TaskPool.d
```diff
@@ -193,7 +153,7 @@ class TaskPool ( TaskT : Task ) : ObjectPool!(Task)
 
             ++count;
 
-            task.terminationHook({
+            task.terminationHookUnique({
```
daniel-zullo: @mihails-strasuns 
There is a problem in one application where the task (main task) gets suspended below (`if (count > 0)  current_task.suspend()`) and the termination hook added here is never executed. 

This is pretty much the scenario I have: 
```
- Scheduler stats when `awaitRunningTasks()`  is called:
stats.task_queue_busy: 0
stats.suspended_queue_busy: 0
stats.worker_fiber_busy: 1    (usually fiber busy is greater than 1, but I found this case interesting to debug)

- the variable `count` in this method goes greater than zero

- [ocean.task.Task] <0X00007F30FB70A580> is suspending itself (main/current task)

- the task <0X00007F30FB70A580> is never resumed
```

src/ocean/task/TaskPool.d
```diff
@@ -227,30 +187,28 @@ class TaskPool ( TaskT : Task ) : ObjectPool!(Task)
...
+    /***************************************************************************
+
+        Used to recycle pool tasks when they finish
+
+    ***************************************************************************/
+
+    private void taskTerminationHook ( )
+    {
+        auto task = Task.getThis();
+        this.recycle(task);
```
daniel-zullo, line 223: Should the task be first recycled by calling its own recycle method as it was done for `OwnedTask`?

src/ocean/task/TaskPool.d
daniel-zullo: the number of items in `tasks_iterator` is always equal to `count` so it seems the check `if (!this.isBusy(this.toItem(task)))`  does not kick in

src/ocean/task/TaskPool.d
mihails-strasuns: Good catch!

mihails-strasuns: Actually no, this is a wrong lead - `recycle` method will be called anyway after termination hook because of https://github.com/sociomantic-tsunami/ocean/blob/v4.x.x/src/ocean/task/Task.d#L493

But it does show that this PR changes order of recycling calls - not sure yet if it is of any relevance

src/ocean/task/TaskPool.d
mihails-strasuns: That rings a bell in combination with observation in another comment - maybe there is an issue about order of termination hooks that I have missed. Trying to craft a synthetic test case now.

nemanja-boric-sociomantic:
Daniel Zullo:

> there is a conflict between a patch in ocean #645 and the implementation of the `Controllers` in swarm. The `Suspendable` implements  `ISuspendable` but it always throws in `suspended()`

Mihails:

> NB: this is not directly related to #645. This issues makes it impossible to use controller as a stream in `StreamProcessor` in general

nemanja-boric-sociomantic:
This is not related to 645, but to #640 

cc @matthias-wende-sociomantic

src/ocean/task/ThrottledTaskPool_test.d
```diff
@@ -24,75 +24,104 @@ import ocean.task.util.Timer;
 
 import ocean.io.model.ISuspendable;
 import ocean.io.Stdout;
+import ocean.io.select.client.TimerEvent;
+import ocean.io.select.EpollSelectDispatcher;
```
daniel-zullo: This import might not be needed in `Stop using tasks as ISuspendable in tests` commit. Please double check

src/ocean/task/ThrottledTaskPool_test.d
```diff
@@ -24,75 +24,104 @@ import ocean.task.util.Timer;
...
-class ProcessingTask : Task
+class Generator : ISuspendable
 {
-    int x;
+    TimerEvent timer;
+    ThrottledTaskPool!(ProcessingTask) pool;
+    int counter;
 
-    static int total;
+    this ( EpollSelectDispatcher epoll )
```
daniel-zullo: It seems the input `epoll` parameter is not used so it could be removed.

src/ocean/task/ThrottledTaskPool_test.d
```diff
@@ -24,75 +24,104 @@ import ocean.task.util.Timer;
...
+
+    bool generate ( )
+    {
+        this.pool.start(this.counter);
+        ++this.counter;
+        return true;
     }
 }
 
+
```
daniel-zullo: adding this line might not be related

src/ocean/task/ThrottledTaskPool_test.d
```diff
@@ -24,75 +24,104 @@ import ocean.task.util.Timer;
...
 unittest
 {
+    auto epoll = new EpollSelectDispatcher;
```
daniel-zullo: this is not needed

src/ocean/task/ThrottledTaskPool_test.d
```diff
@@ -24,75 +24,104 @@ import ocean.task.util.Timer;
...
+
     SchedulerConfiguration config;
     config.worker_fiber_limit = 10;
     config.task_queue_limit = 30;
-    initScheduler(config);
+    initScheduler(config, epoll);
```
daniel-zullo: `theScheduler.epoll()` instead

src/ocean/task/util/StreamProcessor_test.d
```diff
@@ -21,59 +21,88 @@ import ocean.task.util.Timer;
 
 import ocean.io.model.ISuspendable;
 import ocean.io.Stdout;
-import ocean.core.Test;
+import ocean.io.select.client.TimerEvent;
+import ocean.io.select.EpollSelectDispatcher;
```
daniel-zullo: ditto

relnotes/taskpool.feature.md
```diff
@@ -0,0 +1,13 @@
+### Change of `TaskPool` and `StreamProcessor` internals
+
+Both `TaskPool` and `StreamProcessor` stopped defining internal private
```
daniel-zullo: I think the `StreamProcessor` defined it indirectly maybe I want to mention here the `ThrottledTaskPool` instead

daniel-zullo (commented):
I only have minor comments to be discussed and LGTM

mihails-strasuns:
Updated per @daniel-zullo-sociomantic comments

src/ocean/task/TaskPool.d
```diff
@@ -122,30 +96,30 @@ class TaskPool ( TaskT : Task ) : ObjectPool!(Task)
...
-        }
-
+        task.copyArguments(args);
+        this.startImpl(task);
         return true;
     }
 
+    /***************************************************************************
+
+        Common part of start implemntation reused by derivatives. Split into
```
daniel-zullo: typo in `implemntation`

daniel-zullo (approved):
a cosmetic comment and LGTM

relnotes/taskpool.feature.md
```diff
@@ -0,0 +1,15 @@
+### Change of `TaskPool` and `ThrottledTaskPool` internals
+
+Also affects `StreamProcessor` as it uses `ThrottledTaskPool` internally.
+
+Both `TaskPool` and `ThrottledTaskPool` stopped defining internal private
+`OwnedTask` to wrap user-supplied type in. Instead, regular termination hook
+infrastructure is used.
+
+Both also don't try to catch and rethrow exceptions arising from a task - that
+is already done by termination hook system.
```
matthias-wende-sociomantic: This sounds like catching the exceptions is done by the termination hook system.
Do I understand right, that what you actually mean is: try to catch the exception to call `this.recycle(task)` is not needed because the recycle call is done by the (new) termination hook?

mihails-strasuns: Yup

mihails-strasuns: Reworded

matthias-wende-sociomantic (commented):
Apart from the failing unit test it lgtm.

relnotes/taskpool.feature.md
```diff
@@ -0,0 +1,15 @@
+### Change of `TaskPool` and `ThrottledTaskPool` internals
+
+Also affects `StreamProcessor` as it uses `ThrottledTaskPool` internally.
```
leandro-lucarella-sociomantic, line 3: Maybe add also `StreamProcessor` as an affected class/module in the title?

leandro-lucarella-sociomantic (commented):
I have no idea about the impact of this and only went very quickly through the code, so just one small documentation comment.

PR #647, opened 2018-10-18
Fix equality expression in ocean/core/Traits

RazvanN7:
This is currently blocking [1]. Short explanation: is expression currently has a bug where if a type is checked for equality with another type, the alias this is considered an exact match. The modified code took advantage of this bug.

[1] https://github.com/dlang/dmd/pull/8839

### DIFF
--- a/src/ocean/core/Traits.d
+++ b/src/ocean/core/Traits.d
@@ -2019,7 +2019,7 @@ template ParameterTupleOf( Fn )
         alias Tuple!(Params) ParameterTupleOf;
     else static if( is( Fn Params == delegate ) )
         alias Tuple!(ParameterTupleOf!(Params)) ParameterTupleOf;
-    else static if( is( Fn Params == Params* ) )
+    else static if( is( Fn Params : Params* ) )
         alias Tuple!(ParameterTupleOf!(Params)) ParameterTupleOf;
     else
         static assert( false, "Argument has no parameters." );


nemanja-boric-sociomantic:
Yup, that will work - travis will now compile both D1 and D2 code.

RazvanN7:
Sorry, I just didn't thought about the conversion when I made the PR. Thanks for the tip

src/ocean/core/Traits.d
```diff
@@ -2019,7 +2019,7 @@ template ParameterTupleOf( Fn )
         alias Tuple!(Params) ParameterTupleOf;
     else static if( is( Fn Params == delegate ) )
```
Geod24, line 2020: The change should probably be applied to the other ones as well.

RazvanN7: I tried, but `is (Fn Params : delegate)` issues a compilation error.

RazvanN7: Maybe nothing is implicitly convertible to a `function` or `delegate` ?

Geod24: Interestingly, it's a parsing error. But the grammar allows it: https://dlang.org/spec/expression.html#is_expression

RazvanN7: Yes, I know. Puzzling.

mihails-strasuns: Shouldn't `alias this` of `delegate` member field match this? This looks like different version of the very same bug your linked DMD PR tries to fix.

Not a problem from our side but you may want to check it while you are at it.

mihails-strasuns (approved):
Thanks!

I intend to make next ocean minor release next week - as soon as it is out, dlang CI should pick that tag for testing.

PR #649, opened 2018-10-24
Make validateTime and validateDate public

andreas-bok-sociomantic:
Expose validateTime and validateDate in
ocean.text.convert.DateTime.

The methods are helpful for simply validation
of strings from e.g. config files which are
dates/time values.

### DIFF
--- a/src/ocean/text/convert/DateTime.d
+++ b/src/ocean/text/convert/DateTime.d
@@ -304,7 +304,7 @@ private bool validateCharacters ( cstring str, cstring extra = "" )
 
 *******************************************************************************/
 
-private bool validateDate ( uint day, uint month, uint year )
+public bool validateDate ( uint day, uint month, uint year )
 {
     if ( year < 1900 )
     {
@@ -337,7 +337,7 @@ private bool validateDate ( uint day, uint month, uint year )
 
 *******************************************************************************/
 
-private bool validateTime ( int hour, int minute, int second )
+public bool validateTime ( int hour, int minute, int second )
 {
     if ( hour < 0 || hour > 23 )
     {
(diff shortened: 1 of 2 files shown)


src/ocean/text/convert/DateTime.d
```diff
@@ -304,7 +304,7 @@ private bool validateCharacters ( cstring str, cstring extra = "" )
 
 *******************************************************************************/
 
-private bool validateDate ( uint day, uint month, uint year )
+validate bool validateDate ( uint day, uint month, uint year )
```
nemanja-boric-sociomantic: `validate` ;-)

nemanja-boric-sociomantic:
I'm skeptical if this is the right place for these methods to live. Also missing release notes.

andreas-bok-sociomantic:
> I'm skeptical if this is the right place for these methods to live. 

Sure. Didn't consider other places since `convert.DateTime` are their current kiez. However since it anyway would go into a minor release they could go somewhere more appropriate. 

Also missing release notes. 

:+1:

andreas-bok-sociomantic:
Updated based on comments from @nemanja-boric-sociomantic

Previous head was 4989cff3, new **HEAD** at 7e945aa8.

<details>
<summary>Diff between heads are:</summary>

```diff

 src/ocean/text/convert/DateTime.d | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/src/ocean/text/convert/DateTime.d b/src/ocean/text/convert/DateTime.d
index 0f865c5e..7b5f4333 100644
--- a/src/ocean/text/convert/DateTime.d
+++ b/src/ocean/text/convert/DateTime.d
@@ -304,7 +304,7 @@ private bool validateCharacters ( cstring str, cstring extra = "" )
 
 *******************************************************************************/
 
-validate bool validateDate ( uint day, uint month, uint year )
+public bool validateDate ( uint day, uint month, uint year )
 {
     if ( year < 1900 )
     {

```
</details>

mihails-strasuns:
Still missing release notes

andreas-bok-sociomantic:
> Still missing release notes

But is the location of the util methods in `ocean.text.convert.DateTime`, cf. the comment by @nemanja-boric-sociomantic:  

> I'm skeptical if this is the right place for these methods to live.

mihails-strasuns:
`ocean.time` would be more appropriate but I don't want to add new module for just a few functions. I'd suggest to keep it here for now - if there will be more time utilities of such kind, those can be deprecated and moved later.

andreas-bok-sociomantic:
Added release notes

mihails-strasuns (approved):

PR #650, opened 2018-10-24
Make closures CI script fail for parts of pipeline

mihails-strasuns:


nemanja-boric-sociomantic (approved):

joseph-wakeling-sociomantic:
Unfortunately I'm not sure this works, because the `!` will override a non-zero exit status caused by a `make` failure as well as by `grep` not finding a `closure` in the output.  (This was tried with dmxnet.)

To test, try replacing `fasttest` with a non-existent make target (e.g. `fastte`).  The closure check will pass even though the build should have failed.

Issue #652, opened 2018-10-26
Support deep-copying empty Contiguous instances


joseph-wakeling-sociomantic:
The `ocean.util.serialize.contiguous.Util.copy` function throws a `SanityException` when the `src` parameter is an empty `Contiguous` instance (i.e. `src.ptr is null`), because the underlying `Deserializer.deserialize` call requires data to be present.

However, it seems perfectly reasonable that `copy` should handle empty `Contiguous` instances, in which case the `dst` parameter should also wind up empty.

PR #653, opened 2018-10-26
Support copying empty Contiguous instances

joseph-wakeling-sociomantic:
The original implementation of `ocean.serialize.contiguous.Util.copy` will throw a `SanityException` if `src` contains no data, because the `Deserializer.deserialize` method requires that underlying struct data is present.

This patch fixes the issue by ensuring `copy` checks `src.ptr` and then if it is `null` resets `dst` instead of deep-copying to it.

Note that we need to temporarily cast away `const`-ness of `src` to work around the fact that the `ptr` method does not currently support it.  We do not modify any underlying data, so this should not break the promise of the `copy` function signature.

Fixes https://github.com/sociomantic-tsunami/ocean/issues/652.

### DIFF
--- a/src/ocean/util/serialize/contiguous/Util.d
+++ b/src/ocean/util/serialize/contiguous/Util.d
@@ -43,7 +43,19 @@ import ocean.core.Test;
 
 public Contiguous!(S) copy(S) ( in Contiguous!(S) src, ref Contiguous!(S) dst )
 {
-    Deserializer.deserialize(src.data, dst);
+    // We have to cast away `const` here since the `ptr`
+    // method does not support it.  We are not modifying
+    // anything so we do not break the `in` promise of
+    // the function signature.
+    if ((cast(Contiguous!(S)) src).ptr is null)
+    {
+        dst.reset();
+    }
+    else
+    {
+        Deserializer.deserialize(src.data, dst);
+    }
+
     return dst;
 }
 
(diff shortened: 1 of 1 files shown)


joseph-wakeling-sociomantic:
Whoops, I forgot to target v4.2.x first time round.  Fixed for merge purposes, but CI seems to be embuggerated.  Can we get Travis to deal with this?  (Should be no big deal in any case since the patch is trivial and tests based on v4.x.x should be as valid as with v4.2.x.)

nemanja-boric-sociomantic:
> Whoops, I forgot to target v4.2.x first time round. Fixed for merge purposes, but CI seems to be embuggerated. Can we get Travis to deal with this? (Should be no big deal in any case since the patch is trivial and tests based on v4.x.x should be as valid as with v4.2.x.)

Amend and force push should make Travis happy again.

joseph-wakeling-sociomantic:
Done.  I've taken the opportunity to tweak the commit message slightly.

matthias-wende-sociomantic (approved):
Sounds totally reasonable and lgtm. Still I'll keep merging the PR to someone with deeper ocean experience.

joseph-wakeling-sociomantic:
@mihails-strasuns is probably the one who should give final OK.

mihails-strasuns (approved):

Issue #648, opened 2018-10-23
Closure-allocation CI script hardcodes assumption of D1 => D2 conversion


joseph-wakeling-sociomantic:
The `ci/closures.sh` script used to check for closure allocations makes a hardcoded assumption that the code being checked is D1 and therefore needs to be converted to D2 before the check is made, using `make d2conv`.

However, this does not apply when used with a `+d2` release, or a dependent codebase that is D2-only.  The script should not run `make d2conv` in this case.

(Example use-case is with dmxnet: https://github.com/sociomantic-tsunami/dmxnet/pull/89 where we would like to have the closure allocation check, but can't use this while converting to D2-only.  We can work around simply by just copying the script and deleting the conversion, but it'd be nice to have one simple solution for this.)

nemanja-boric-sociomantic:
I don't think it's the idea that you use `ci/` directory from ocean in your project, though. You should have a copy of the script, or the script should be a part of beaver, because we're that doesn't fall into semver guarantees (we usually change it as a part of minor and sometimes even patch release).

joseph-wakeling-sociomantic:
Hmmm well.  It _is_ being used that way in some projects (e.g. swarm), although I agree it's probably not really the intended use-case.

I do think it's a problem though that it can't be run on `+d2` auto-converted releases.

nemanja-boric-sociomantic:
Nope, swarm has it's own copy:

https://github.com/sociomantic-tsunami/swarm/blob/v5.x.x/.travis.yml#L55
https://github.com/sociomantic-tsunami/swarm/blob/v5.x.x/ci/closures.sh

nemanja-boric-sociomantic:
However, you're right that something is wrong with the` +d2` builds:

https://travis-ci.org/sociomantic-tsunami/ocean/builds/442572615

joseph-wakeling-sociomantic:
> Nope, swarm has it's own copy:

Well, duh.  I'm sure there _is_ at least one project that uses ocean's script directly, though.

joseph-wakeling-sociomantic:
For D2-only projects, though, I'm overthinking it: it should suffice to just do the `make fasttest | grep "closure"` directly (cf. the updated dmxnet PR), assuming that `DFLAGS=-vgc` is set in `env`.

mihails-strasuns:
Should be fixed now for ocean

PR #654, opened 2018-10-26
Ensure closures script fails if make fails

joseph-wakeling-sociomantic:
The intent of the script here, and the use of `set -o pipeline`, is that the script should fail if either `make fasttest` fails or `grep` finds a matching line; in other words, if the exit status of `make` is non-zero or that of `grep` is zero.

`-o pipeline` means that non-zero exit statuses from `make` propagate to the exit status of `grep`, but the leading `!` means this is inverted to result in a (falsely) passing test.

Applying the `!` only to the output status of `grep` itself would fix this, but leaves a second problem: `grep` may itself fail, in which case it will report exit status 2.

This patch therefore rewrites the check to use `test` to ensure that the exit status of `grep` is 1.

This should mean that the exit status of the pipeline will be non-zero in any of the following circumstances:

  * `make` fails
  * `grep` fails
  * `grep` finds at least one line containing `closure`

... while if `make` passes and `grep` does not find `closure`, we will get the zero exit status wanted to pass the test.

Note that this may mean that the closure script will now fail for `+d2` auto-converted releases, since the script will apply double conversion to the source code and `make` will hence fail.

### DIFF
--- a/ci/closures.sh
+++ b/ci/closures.sh
@@ -1,12 +1,17 @@
 #!/bin/bash
-set -xe
-set -o pipefail
+set -x
 
 # Enables printing of all potential GC allocation sources to stdout
 export DFLAGS=-vgc
 
 # Prepare sources
-beaver dlang make d2conv
+beaver dlang make d2conv || exit 1
 
-# Ensure that there are no lines about closure allocations in the output
-! beaver dlang make fasttest 2>&1 | grep "closure"
+# Run tests and write compiler output to temporary file
+compiler_output=`mktemp`
+beaver dlang make fasttest 2>&1 > $compiler_output || exit 1
+
+# Ensure there are no lines about closure allocations in the output.
+# Note explicit check for `grep` exit status 1, i.e. no lines found.
+grep -e "closure" $compiler_output
+test $? -eq 1 || exit 1


joseph-wakeling-sociomantic:
@nemanja-boric-sociomantic @mihails-strasuns FYI.

joseph-wakeling-sociomantic:
Interesting that the new version failed.  I'm going to push a series of `squash!` patches just to check that we are not accidentally getting failure from some other source.

joseph-wakeling-sociomantic:
I'm guessing that the reason is that the `grep` command generates a non-zero code, and this is enough to count as failure given `set -x`.

joseph-wakeling-sociomantic:
I've tried a small rewrite using manual error checks instead of `set -e`.  The first version is designed to fail if D2 conversion fails, just to ensure that the checks are all working right.

joseph-wakeling-sociomantic:
I've squashed things down and edited the in-script comments.  This is now ready for final review.

joseph-wakeling-sociomantic:
However, I would like to have some thoughts on this point related to https://github.com/sociomantic-tsunami/ocean/issues/648:

> Note that this update may mean that the closure script will now fail for
> `+d2` auto-converted releases: the hardcoded `make d2conv` means that it
> will run `d1to2fix` on already-converted code, and thus `make fasttest`
> will fail.

It might be an idea to address that issue first before taking this in.

joseph-wakeling-sociomantic:
@leandro-lucarella-sociomantic @mihails-strasuns it's probably a good idea for you to take a glance at this one.

ci/closures.sh
```diff
@@ -1,12 +1,22 @@
...
-set -xe
+set -x
 set -o pipefail
 
 # Enables printing of all potential GC allocation sources to stdout
 export DFLAGS=-vgc
 
 # Prepare sources
-beaver dlang make d2conv
+beaver dlang make d2conv; test $? -eq 0 || exit 2
```
mihails-strasuns:
```suggestion
beaver dlang make d2conv || exit 1
```

ci/closures.sh
```diff
@@ -1,12 +1,22 @@
...
+# This means tests will fail if any of these conditions are met:
+#
+#   * `make` fails (exit status 2, propagated by `set -o pipefail`)
+#   * `grep` fails (exit status 2)
+#   * `grep` finds `closure` (exit status 0)
+#
+# Note that `make` can in principle exit with status 1 too, but only
+# if the `-q` option is used, which should never be the case for our
+# builds, so we can assume exit status 1 only comes from `grep`.
+beaver dlang make fasttest 2>&1 | (grep -e "closure"; test $? -eq 1)
```
mihails-strasuns: I'd rather use something like named pipe and multi-line scirpt. For example (untested):

```suggestion
mkfifo output
beaver dlang make fasttest 2>&1 > output || exit 1
grep "closure" output
test $? -eq 1 || exit 1
```

joseph-wakeling-sociomantic: Out of curiosity, since you wrote the original script: why is the `2>&1` necessary/wanted?  I ask because the question came up with a team member in another context.  I assume it's in order to ensure that the test works regardless of whether the compiler outputs to `stdout` or `stderr`, but thought I'd ask.

joseph-wakeling-sociomantic:
<details><summary>Updated in line with @mihails-strasuns' suggestions.</summary>

```diff
diff --git a/ci/closures.sh b/ci/closures.sh
index c3bcec26..679fcebd 100755
--- a/ci/closures.sh
+++ b/ci/closures.sh
@@ -1,22 +1,17 @@
 #!/bin/bash
 set -x
-set -o pipefail
 
 # Enables printing of all potential GC allocation sources to stdout
 export DFLAGS=-vgc
 
 # Prepare sources
-beaver dlang make d2conv; test $? -eq 0 || exit 2
+beaver dlang make d2conv || exit 1
+
+# Write compiler output to a named pipe
+mkfifo compiler_output || exit 1
+beaver dlang make fasttest 2>&1 > compiler_output || exit 1
 
 # Ensure there are no lines about closure allocations in the output.
 # Note explicit check for `grep` exit status 1, i.e. no lines found.
-# This means tests will fail if any of these conditions are met:
-#
-#   * `make` fails (exit status 2, propagated by `set -o pipefail`)
-#   * `grep` fails (exit status 2)
-#   * `grep` finds `closure` (exit status 0)
-#
-# Note that `make` can in principle exit with status 1 too, but only
-# if the `-q` option is used, which should never be the case for our
-# builds, so we can assume exit status 1 only comes from `grep`.
-beaver dlang make fasttest 2>&1 | (grep -e "closure"; test $? -eq 1)
+grep -e "closure" compiler_output
+test $? -eq 1 || exit 1
```
</details>

I'll follow up with some `fixup!` patches just to validate that the script really works as intended.

joseph-wakeling-sociomantic:
BTW @mihails-strasuns this hangs on the `mkfifo` command on my system (with or without the `|| exit 1`).

joseph-wakeling-sociomantic:
I suspect that the `|| exit 1` on the end of the `make fasttest` line is causing the whole write-to-fifo process to hang, as it won't complete until a reader handles the output.

joseph-wakeling-sociomantic:
Given the apparent complications of this alternative approach, is it worth pursuing?  Or better to revert back to the working form I had originally?

mihails-strasuns:
Apologies for a late response. I did some local tests and I think this simple replacement using tempfile is how I would prefer it to look like:

```bash
#!/bin/bash
set -x

# Enables printing of all potential GC allocation sources to stdout
export DFLAGS=-vgc

# Prepare sources
beaver dlang make d2conv || exit 1

# Grab compilation output
output=`mktemp`
beaver make fasttest 2>&1 > $output || exit 1

# Ensure that there are no lines about closure allocations in the output
grep "closure" $output
test $? == 1 || exit 1
```

Will that work for you?

mihails-strasuns:
Ping @joseph-wakeling-sociomantic

joseph-wakeling-sociomantic:
Hah, you know the tempfile approach was one of the first I ever tried for this?  I think I dropped it on the basis of feedback about the possibilities of the `pipefail` setting.

mihails-strasuns:
I had some vague memories that there was an attempt but can't remember why it was discarded :)

joseph-wakeling-sociomantic:
Here you go.  If it works then I'll push a few follow-up `fixup` just to validate that the different points of failure work as expected.

joseph-wakeling-sociomantic:
The last of the current fixup patches should restore us to the correct (and passing) behaviour.  I'll then squash so we can do final review.

joseph-wakeling-sociomantic:
Squashed.  @mihails-strasuns I would be interested in your feedback on potential issues with the `make d2conv` part: I think this version of the script will cause failures in builds of `+d2` tags.  So we may need to further tweak the script, or find some other way to avoid this outcome.

joseph-wakeling-sociomantic:
If it's not clear, the anticipated failure is:

* Travis tries to build a `+d2` tag

* all the regular test stages pass

* the closure allocation check is launched

* `make d2conv` is run on already-converted code

* `make fasttest` fails because of doubly-converted code (`scope scope someDg` seems to be the usual thing the compiler fails on)

ci/closures.sh
```diff
@@ -1,12 +1,17 @@
...
 export DFLAGS=-vgc
 
 # Prepare sources
-beaver dlang make d2conv
+beaver dlang make d2conv || exit 1
 
-# Ensure that there are no lines about closure allocations in the output
-! beaver dlang make fasttest 2>&1 | grep "closure"
+# Run tests and write compiler output to temporary file
+compiler_output=`mktemp`
```
joseph-wakeling-sociomantic, line 11: Minor note: we could use something like `mktemp --tmpdir closure-check.XXXXXXXX` to generate a temp file whose name clearly associates it with this test.

mihails-strasuns:
Test are disabled for `+d2` tags -https://github.com/sociomantic-tsunami/ocean/blob/v4.x.x/.travis.yml#L30

ci/closures.sh
mihails-strasuns: What for? It is printed to stdout for inspection and we don't preserve those tmp files.

mihails-strasuns (approved):

PR #656, opened 2018-10-30
Workaround optimizer bug

stefan-koch-sociomantic:
This commit works around a bug in dmd where it'll
pack two distinct values into one register,
which does not work with the current x87 codegen

### DIFF
--- a/src/ocean/math/IEEE.d
+++ b/src/ocean/math/IEEE.d
@@ -1718,21 +1718,27 @@ real NaN(ulong payload)
 ulong getNaNPayload(real x)
 {
     verify(!!isNaN(x));
+    // x_ptr is needed to create a separate alias to x
+    // which the optimizer cannot see through
+    // this will prevent an optimization which
+    // will cause an ice in newer dmd versions
+    auto x_ptr = &x;
+
     static if (real.mant_dig == 53) {
-        ulong m = *cast(ulong *)(&x);
+        ulong m = *cast(ulong *)(x_ptr);
         // Make it look like an 80-bit significand.
         // Skip exponent, and quiet bit
         m &= 0x0007_FFFF_FFFF_FFFF;
         m <<= 10;
     } else static if (real.mant_dig==113) { // quadruple
         version(LittleEndian) {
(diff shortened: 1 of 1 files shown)


tiyash-basu-sociomantic:
Upon trying it out, this workaround does avert the aforementioned dmd bug. Thanks @stefan-koch-sociomantic!

Geod24:
Is there an issue number for this?

stefan-koch-sociomantic:
@Geod24 not yet.

The dmd bug is introduced in https://github.com/dlang/dmd/pull/6192

stefan-koch-sociomantic:
@Geod24 now there is:
https://issues.dlang.org/show_bug.cgi?id=19343

don-clugston-sociomantic:
I'm not sure about the meaning of the comment in the code.
```d
   // x_ptr is needed to create a seperate alias to x
    // which the optimizer cannot see through
```
Does it need to be in that exact form? Can we get rid of one or more of the `size_t`s?
It'd be much nicer if it was just
`auto  x_ptr = &x;`
but I presume that doesn't actually work.

src/ocean/math/IEEE.d
```diff
@@ -1718,21 +1718,27 @@ real NaN(ulong payload)
 ulong getNaNPayload(real x)
 {
     verify(!!isNaN(x));
+    // x_ptr is needed to create a seperate alias to x
```
iain-buclaw-sociomantic: Spelling: separate

src/ocean/math/IEEE.d
```diff
@@ -1718,21 +1718,27 @@ real NaN(ulong payload)
...
+    // which the optimizer cannot see through
+    // this will prevent an optimisation which
```
iain-buclaw-sociomantic: You use optimize on one line, then optimise on the next.

iain-buclaw-sociomantic:
Can't you just use union access?

stefan-koch-sociomantic:
@don-clugston-sociomantic actually that does work.
Amended.

matthias-wende-sociomantic (approved):
:+1:

don-clugston-sociomantic (approved):
LGTM

PR #658, opened 2018-11-04
Workaround task pool regression from v4.4.0

mihails-strasuns:
FYI @jens-mueller-sociomantic @bogdan-szabo-sociomantic

### DIFF
--- a/src/ocean/task/TaskPool.d
+++ b/src/ocean/task/TaskPool.d
@@ -29,6 +29,8 @@ import ocean.task.Task;
 import ocean.task.IScheduler;
 
 import ocean.core.Enforce;
+import ocean.core.Buffer;
+import ocean.core.array.Mutation; /* : insertShift */;
 import ocean.meta.types.Function /* : ParametersOf */;
 import ocean.meta.traits.Aggregates /* : hasMember */;
 import ocean.meta.AliasSeq;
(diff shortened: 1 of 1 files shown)


bogdan-szabo-sociomantic:
I am not sure if I understand what is `task.tupleof`... but it looks right. Maybe a test would be helpful...

src/ocean/task/TaskPool.d
```diff
@@ -116,7 +118,15 @@ class TaskPool ( TaskT : Task ) : ObjectPool!(Task)
 
     protected void startImpl ( Task task )
     {
-        task.terminationHook(&this.taskTerminationHook);
+        // HACK: ocean scheduler disallows any context switching while handling
+        // termination hooks, but this is currently violated by swarm neo
+        // `resume` implementation for `ISuspendable`. Until swarm is fixed,
+        // this dirty hack ensures recycling always happens last among
+        // termination hooks
+        static assert (is(typeof(task.tupleof[1]) == Buffer!(void delegate())));
```
bogdan-szabo-sociomantic, line 131: please add a message to this assert

mihails-strasuns:
> I am not sure if I understand what is task.tupleof... but it looks right. Maybe a test would be helpful...

It is a trick to ignore `private` protection for a field. `tupleof` always ignores protection attributes so if you know exactly order of fields it is possible to access them directly. In this case I didn't want to make termination hooks `public` for a short-term hack thus this approach.

I will expand the comment.

PR #659, opened 2018-11-14
Implement utility for task usage in scripts

mihails-strasuns:
Fixes #612

### DIFF
--- a/relnotes/quickrun.feature.md
+++ b/relnotes/quickrun.feature.md
@@ -1,0 +1,23 @@
+### Helper to quickly spawn a task for scripting
+
+* `ocean.task.util.QuickRun`
+
+Utility intended for script-like usage:
+
+```D
+  int main ( )
+  {
+    initScheduler(SchedulerConfiguration.init);
+
+    return quickRun({
+      // use task-blocking functions:
+      Dht.connect(10_000);
+      Dht.put("channel", 23, "value");
+      return 0;
+    });
+  }
+```
+
+It turns delegate argument into a task and runs it through scheduler. Starts the
+event loop automatically and will finish only when all tasks finish and no
+registered events remain (or `theScheduler.shutdown` is called).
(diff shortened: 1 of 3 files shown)


src/ocean/task/util/QuickRun.d
```diff
@@ -0,0 +1,81 @@
...
+    verify(task.finished());
+    return task.result;
+}
+
+///
+unittest
+{
+    int main ( )
+    {
+        // initScheduler(SchedulerConfiguration.init);
```
gavin-norman-sociomantic, line 54: What does this line being commented out imply?

relnotes/quickrun.feature.md
```diff
@@ -0,0 +1,23 @@
+### Helper to quickly spawn a task for scripting
+
+* `ocean.task.util.QuickRun`
+
+Utility intended for script-like usage:
+
+```D
+  int main ( )
+  {
+    initScheduler(SchedulerConfiguration.init);
```
gavin-norman-sociomantic, line 10: I would still argue for silently doing this as a default if the user doesn't do it, but it's not a big deal if you still disagree.

src/ocean/task/util/QuickRun.d
mihails-strasuns: I don't want to import `ocean.task.Scheduler` here (because import cycles) but wanted to leave some example. Not sure how to handle this :/

mihails-strasuns: I don't want to import `ocean.task.Scheduler` here (because import cycles) but wanted to leave some example. Not sure how to handle this :/

relnotes/quickrun.feature.md
mihails-strasuns: Yeah I haven't changed my opinion on topic ;)

src/ocean/task/util/QuickRun.d
gavin-norman-sociomantic: I see. Maybe:
```D
// Make sure the scheduler is initialised first, e.g.:
// initScheduler(SchedulerConfiguration.init);
```

mihails-strasuns:
Updated:
- comment as suggested by @gavin-norman-sociomantic

gavin-norman-sociomantic (approved):

PR #661, opened 2018-11-16
GC stats

bogdan-szabo-sociomantic:
This is a class that adds GC stats. We started to reuse this code a lot in our apps and I was thinking to add it to ocean. What do you think?

### DIFF
--- a/relnotes/gcstats.feature.md
+++ b/relnotes/gcstats.feature.md
@@ -1,0 +1,10 @@
+### Garbage Collector stats
+
+* `ocean.application.components.GCStats`,
+
+  `GCStats` can be used with dmd-transitional in order to get stats about the
+  garbage collector. It will collect:
+    - `gc_run_duration` The number of microseconds the garbage collector ran
+        for during the last collection cycle.
+    - `gc_run_percentage` The percentage of time that was spent by the
+        garbage collector during the last collection cycle.
--- a/relnotes/gcstats.feature.md
+++ b/relnotes/gcstats.feature.md
@@ -8,3 +8,8 @@
         for during the last collection cycle.
     - `gc_run_percentage` The percentage of time that was spent by the
         garbage collector during the last collection cycle.
+
+* `ocean.util.app.DaemonApp`
+
+  `DaemonApp` class now contains the `reportGCStats()` method which will add
+  the GC stats to the stats log file.
--- a/src/ocean/application/components/GCStats.d
+++ b/src/ocean/application/components/GCStats.d
@@ -1,0 +1,175 @@
+/*******************************************************************************
+
+    Contains API to obtain garbage collector stats.
+
+    Copyright:
+        Copyright (c) 2009-2018 dunnhumby Germany GmbH.
+        All rights reserved.
+
+    License:
+        Boost Software License Version 1.0. See LICENSE_BOOST.txt for details.
+        Alternatively, this file may be distributed under the terms of the Tango
+        3-Clause BSD License (see LICENSE_BSD.txt for details).
+
+*******************************************************************************/
+
+module ocean.application.components.GCStats;
+
+import core.memory;
+
+import ocean.core.Traits : hasMember, ParameterTupleOf;
+import ocean.time.MicrosecondsClock;
+
+/*******************************************************************************
+
+    Class used for getting used GC stats for the current process. This will
+    work only with dmd-transitional.
+
+*******************************************************************************/
+
+public class GCStats
+{
+    static if ( hasMember!(GC, "monitor") )
+    {
+        /// The type of the gcEnd delegate
+        private alias ParameterTupleOf!(GC.monitor)[1] MonitorEndParams;
+
+        /// Integers used by the gcEnd delegate
+        private alias ParameterTupleOf!(MonitorEndParams)[0] MonitorInt;
+    }
+    else
+    {
+        /// Default type for integers used by monitor
+        private alias long MonitorInt;
+    }
+
+    /**********************************************************************
+
+        Structure representing recorded GC stats.
+
+    **********************************************************************/
+
+    public struct Stats
+    {
+        /**********************************************************************
+
+            The number of microseconds the garbage collector ran for in the
+            last stats collection.
+
+        **********************************************************************/
+
+        public size_t gc_run_duration;
+
+        /**********************************************************************
+
+            The percentage of time that was spent by the garbage collector in
+            the last stats collection.
+
+        **********************************************************************/
+
+        public float gc_run_percentage = 0;
+    }
+
+    /// ditto
+    private Stats stats;
+
+    /// Timestamp for when the last garbage collection started
+    private ulong gc_start_us;
+
+    /**********************************************************************
+
+        Timestamp for when the last stats were collected used to
+        calculate the percentage of the time spent in GC between stats
+        cycles.
+
+    **********************************************************************/
+
+    private ulong last_collected_timestamp;
+
+    /**********************************************************************
+
+        Reset statistics.
+
+    **********************************************************************/
+
+    private void reset ( )
+    {
+        this.last_collected_timestamp = MicrosecondsClock.now_us();
+        this.stats.gc_run_duration = 0;
+    }
+
+    /**********************************************************************
+
+        Start stats reporting
+
+    **********************************************************************/
+
+    public void start ( )
+    {
+        static if ( hasMember!(GC, "monitor") )
+        {
+            GC.monitor(&this.gcBegin, &this.gcEnd);
+        }
+    }
+
+    /**********************************************************************
+
+        Stop stats reporting
+
+    **********************************************************************/
+
+    public void stop ( )
+    {
+        static if ( hasMember!(GC, "monitor") )
+        {
+            GC.monitor(null, null);
+        }
+    }
+
+    /***************************************************************************
+
+        Get's the GC stats for the current process.
+
+        Returns:
+            Stats instance recording the current stats.
+
+    ***************************************************************************/
+
+    public Stats collect ()
+    {
+        auto result = this.stats;
+        this.reset();
+
+        return result;
+    }
+
+    /**********************************************************************
+
+        Called each time the GC starts a collection
+
+    **********************************************************************/
+
+    private void gcBegin ( )
+    {
+        this.gc_start_us = MicrosecondsClock.now_us();
+    }
+
+    /***************************************************************************
+
+        Called when the GC completes a collection.
+
+        Params:
+            freed = the number of bytes freed overall
+            pagebytes = the number of bytes freed within full pages.
+
+    ***************************************************************************/
+
+    private void gcEnd ( MonitorInt freed, MonitorInt pagebytes )
+    {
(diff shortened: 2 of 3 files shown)


gavin-norman-sociomantic:
Seems like a good addition :+1: 

Two comments:
1. I wonder whether `ocean.sys.Stats` is a good fit. The other code in that module is dealing with OS-level stats, whereas the new code deals with GC = D runtime-level stats. A better location doesn't come to mind immediately, though. @nemanja-boric-sociomantic @mihails-strasuns any thoughts?
2. There should probably be a method to write these stats into the `DaemonApp`'s stats log. ([Something like this.](https://github.com/sociomantic-tsunami/ocean/blob/fc67da7e1769468a6744dbaaf109e03aaf72a49d/src/ocean/util/app/DaemonApp.d#L650-L653))

src/ocean/sys/Stats.d
```diff
@@ -236,3 +240,197 @@ public class CpuMemoryStats
...
+
+        **********************************************************************/
+
+        private alias ParameterTupleOf!(MonitorEndParams)[0] MonitorInt;
+    }
+    else
+    {
+        /**********************************************************************
+
+            Default type for intgers used by monitor
```
nemanja-boric-sociomantic: typo `intgers`

src/ocean/sys/Stats.d
```diff
@@ -236,3 +240,197 @@ public class CpuMemoryStats
...
+        Structure representing recorded GC stats.
+
+    **********************************************************************/
+
+    public struct Stats
+    {
+        /**********************************************************************
+
+            The number of microseconds for how much the garbage collector
+            was run.
```
nemanja-boric-sociomantic: This should say "The number of microseconds for how much the garbage collector was run in total." to make it more clear.

src/ocean/sys/Stats.d
```diff
@@ -236,3 +240,197 @@ public class CpuMemoryStats
...
+
+            The number of bytes freed overall
+
+        **********************************************************************/
+
+        int freed;
+
+        /**********************************************************************
+
+            The number of bytes freed within full pages
```
nemanja-boric-sociomantic: This member could have a better explanation for the users (without them knowing about the GC internals)

bogdan-szabo-sociomantic:
I added the stats to the DaemonApp class. I am not sure where to move the GCStats class. Any ideas?

src/ocean/sys/Stats.d
```diff
@@ -236,3 +240,197 @@ public class CpuMemoryStats
...
+
+        Timestamp for when the last garbage collection started
+
+    **********************************************************************/
+
+    private ulong gc_start_us;
+
+    /**********************************************************************
+
+        Timestamp for when the last stats were written to hdd
```
nemanja-boric-sociomantic: "Timestamp for when the last stats were collected" - there's nothing here implying the stats will go to the hard drive.

src/ocean/sys/Stats.d
```diff
@@ -236,3 +240,197 @@ public class CpuMemoryStats
...
+        }
+    }
+
+    /**********************************************************************
+
+        Stop stats reporting
+
+    **********************************************************************/
+
+    public void stop ( )
```
nemanja-boric-sociomantic: this method looks the same as `start`?

src/ocean/sys/Stats.d
nemanja-boric-sociomantic: I also don't get if this one is right. Doesn't this pretty much say "timestamp when the stats collecting was started or when it was reset"?

mihails-strasuns:
I  think it fits new https://github.com/sociomantic-tsunami/ocean/tree/v4.x.x/src/ocean/application/components

src/ocean/sys/Stats.d
nemanja-boric-sociomantic: (that being said, I doubt this is useful)

nemanja-boric-sociomantic: If you insist, you could make it `last_collected_timestamp` and set it in `collect`

nemanja-boric-sociomantic: (still, I would just remove this member, it doesn't bring anything IMHO and makes the stats more confusing).

bogdan-szabo-sociomantic: I updated the comment and the name. I did this pr in a hurry, so I forgot to use it. :P

It was needed to calculate the percentage time for how long the gc was run.

src/ocean/sys/Stats.d
bogdan-szabo-sociomantic: I'm not an expert in GC internals... any suggestions for how I can reword this?

mihails-strasuns: It is actually implementation defined. In dmd-transitional both `free` values will be the same - https://github.com/sociomantic-tsunami/dmd-transitional/blob/v2.078.3.sX/patches/druntime/0004-Add-temproary-GC.monitor-implementation.patch#L127

Thinking about this more, I'd suggest to only use `GC.monitor` only for timing stats. Additional GC stats can optionally be retrieved by manual `GC.stats` call and this will be more future-proof.

src/ocean/sys/Stats.d
nemanja-boric-sociomantic: How do you calculate the percentage of the time based on this? I think what you need is the time while the program is running, right?

nemanja-boric-sociomantic:
OK, sorry I messed up with my comments:

> This should say "The number of microseconds for how much the garbage collector was run in total." to make it more clear.

this is actually wrong. This is "the number of the microseconds the GC run since the last stats collection".

And now the percentage makes sense :-).

relnotes/gcstats.feature.md
```diff
@@ -0,0 +1,12 @@
+### Allow changing of the output streams for AppStatus
```
gavin-norman-sociomantic: Mmm... is this copy pasta?

relnotes/gcstats.feature.md
```diff
@@ -0,0 +1,12 @@
...
+
+* `ocean.application.components.GCStats`,
+
+  `GCStats` can be used with dmd-transitional in order to get stats about the
+  garbage collector. It will collect:
+    - `gc_run_duration` The number of microseconds for how much the
```
gavin-norman-sociomantic: "The number of microseconds the garbage collector ran for" sounds more natural.

gavin-norman-sociomantic: Is this the number of microsecs that it ran for during a certain period?

src/ocean/application/components/GCStats.d
```diff
@@ -0,0 +1,216 @@
...
+    Class used for getting used GC stats for the current process. This will
+    work only with dmd-transitional.
+
+*******************************************************************************/
+
+public class GCStats
+{
+    static if ( hasMember!(GC, "monitor") )
+    {
+        /**********************************************************************
```
gavin-norman-sociomantic: General point: feel free to use `///` comments for field documentation in ocean. (But no need to change this PR now.)

src/ocean/application/components/GCStats.d
```diff
@@ -0,0 +1,216 @@
...
+        Start stats reporting
+
+    **********************************************************************/
+
+    public void start ( )
+    {
+        static if ( hasMember!(GC, "monitor") )
+        {
+            GC.monitor(&this.gcBegin, &this.gcEnd);
+        }
```
gavin-norman-sociomantic, line 112: Should we output a warning or something if `hasMember!(GC, "monitor")` is false?

relnotes/gcstats.feature.md
```diff
@@ -10,3 +10,6 @@
             garbage collector.
     - `freed` The number of bytes freed overall
     - `pagebytes` The number of bytes freed within full pages
+
+  `DaemonApp` contains now `reportGCStats` method which will report the GC
```
gavin-norman-sociomantic: You need to mention the module that is affected, like in line 3.

gavin-norman-sociomantic (commented):
Looks good apart from the small comments...

relnotes/gcstats.feature.md
bogdan-szabo-sociomantic: It is for the time that the stats are collected

src/ocean/application/components/GCStats.d
bogdan-szabo-sociomantic: it is my pleasure to get rid of this large comments

src/ocean/application/components/GCStats.d
bogdan-szabo-sociomantic: The warning is useful only if the class is used. Should I use a `Stderr.formatln`?

src/ocean/sys/Stats.d
bogdan-szabo-sociomantic: the duration is computed here https://github.com/sociomantic-tsunami/ocean/pull/661/files#diff-8e40a07cedca7d7b1b28be980b65f3faR211 and is for each collection cycle

src/ocean/sys/Stats.d
bogdan-szabo-sociomantic: done

src/ocean/sys/Stats.d
nemanja-boric-sociomantic: Yup, see my later comment: https://github.com/sociomantic-tsunami/ocean/pull/661#issuecomment-442044627

nemanja-boric-sociomantic: Maybe just put `(used to calculate the percentage of the time spent in GC between stats cycles)`

bogdan-szabo-sociomantic: I updated the comments and the release notes

bogdan-szabo-sociomantic:
ping @gavin-norman-sociomantic @nemanja-boric-sociomantic

src/ocean/application/components/GCStats.d
gavin-norman-sociomantic: Hmm, it's an interesting scenario. The small danger is that an app is happily calling `collect` and logging the GC stats, but in reality no stats are actually being collected, due to the `static if` here not being compiled in. That's only possible in the case of someone using the wrong DMD version, though, so is not a major risk. I'd say your approach of just ignoring this case seems fine.

src/ocean/util/app/DaemonApp.d
```diff
@@ -652,6 +666,18 @@ public abstract class DaemonApp : Application,
...
     }
 
+    /***************************************************************************
+
+        Collects GC stats and reports them to stats log. Should be
+        called periodically (inside onStatsTimer).
+
+    ***************************************************************************/
+
+    protected void reportGCStats ( )
```
gavin-norman-sociomantic, line 676: This new method should be mentioned in the release notes. (Didn't you have this before?)

bogdan-szabo-sociomantic: I think I lost the change during a rebase... I added it back

gavin-norman-sociomantic (approved):

gavin-norman-sociomantic:
Thanks!

PR #662, opened 2018-11-16
TaskSuspender utility

mihails-strasuns:
Based on idea by @daniel-zullo-sociomantic

### DIFF
--- a/src/ocean/task/Scheduler.d
+++ b/src/ocean/task/Scheduler.d
@@ -656,12 +656,10 @@ final class Scheduler : IScheduler
         {
             Task task;
             verify(this.suspended_tasks.pop(task));
-            if (this.state == State.Shutdown)
-                task.kill();
-            else
-                this.resumeTask(task);
+            this.resumeTask(task);
 
-            // if calling last task resulted in a shutdown
+            // if calling last task resulted in a shutdown there will be no
+            // suspended tasks left (shutdown clears them)
             if (this.state == State.Shutdown)
                 break;
         }
--- a/src/ocean/task/Scheduler.d
+++ b/src/ocean/task/Scheduler.d
@@ -660,6 +660,10 @@ final class Scheduler : IScheduler
                 task.kill();
             else
                 this.resumeTask(task);
+
+            // if calling last task resulted in a shutdown
+            if (this.state == State.Shutdown)
+                break;
         }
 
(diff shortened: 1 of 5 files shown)


src/ocean/task/Scheduler.d
```diff
@@ -660,6 +660,10 @@ final class Scheduler : IScheduler
                 task.kill();
             else
                 this.resumeTask(task);
+
+            // if calling last task resulted in a shutdown
+            if (this.state == State.Shutdown)
+                break;
```
daniel-zullo, line 664: would it be possible to add the `break` statement right after `task.kill()`? or could `this.resumeTask(task)` change the state of `this.state`?

src/ocean/task/util/TaskSuspender.d
```diff
@@ -0,0 +1,87 @@
...
+
+    /***************************************************************************
+
+        Implements resuming as delayed resuming
+
+    ***************************************************************************/
+
+    override public void resume ( )
+    {
+        theScheduler.delayedResume(task);
```
daniel-zullo: you might want to prepend `this.`  to task

daniel-zullo (commented):
I only have a cosmetic comment and a question and LGTM.
I'll test this patch in one application and report back. Thanks a lot!

src/ocean/task/Scheduler.d
mihails-strasuns: Yep, it is for the latter - tesumed task calls shutdown internally and scheduler gets into shutdown state on the middle of processing suspend queue.

daniel-zullo: right, then alternatively we could have

```D
        if (this.state != State.Shutdown)
            this.resumeTask(task);
                   
        // if calling last task resulted in a shutdown
        if (this.state == State.Shutdown)
        {
            task.kill();
            break;
        }
```
But I'm not sure that makes it better. It's up to you...

daniel-zullo:
I've run a couple of tests and went all good.

src/ocean/task/Scheduler.d
mihails-strasuns: I have realized that original check is actually not needed at all - calling `shutdown` will kill all suspended tasks anyway. Updated to remove it completely.

PR #663, opened 2018-11-29
prepare for precise garbage collection

rainers:
Bucket should use void[] to ensure the data is scanned by a precise GC

This is necessary to let https://github.com/dlang/druntime/pull/1603 pass tests with precise collection enabled.

Maybe other ubyte[] data, e.g. VoidBufferImpl or MemoryDevice, are also used with data with indirections, but I'm just poking into the dark.

### DIFF
--- a/src/ocean/util/container/map/model/Bucket.d
+++ b/src/ocean/util/container/map/model/Bucket.d
@@ -80,9 +80,20 @@ public struct Bucket ( size_t V, K = hash_t )
 
          **********************************************************************/
 
-        public alias void[V] Val;
+        public alias ubyte[V] Val;
 
-        public Val val;
+        union
+        {
+            public Val val;
+
+            /**********************************************************************
+
+            Mark this memory as potentially containing pointers and align size of data
+            to next multiple of (void*).sizeof.
+
+            **********************************************************************/
+            private void*[(V + (void*).sizeof - 1) / (void*).sizeof] _void;
+        }
 
         /**********************************************************************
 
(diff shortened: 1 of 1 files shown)


mihails-strasuns:
@gavin-norman-sociomantic and @david-eckardt-sociomantic should know better about this specific code but usually in ocean `ubyte[]` is used exactly with intention to _not_ be scanned by GC.

rainers:
> `ubyte[]` is used exactly with intention to _not_ be scanned by GC.

The comment right above suggests otherwise: "Is extremely important that this "val" is at the top of the struct, otherwise the struct can get into alignment issues that can cause the GC to miss the pointers in the data."

Unfortunately void[V] doesn't work in D1 (complains about "no initalizer"), but void*[] will need more changes.

I also noticed that the `_padding` looks suspicious: V=7 will add 7 more bytes of padding, not 1 as I would expect.

mihails-strasuns:
Yeah that looks suspicious - but I'd like to hear from original authors of the code first :)

david-eckardt-sociomantic:
> usually in ocean `ubyte[]` is used exactly with intention to not be scanned by GC.

This is about buffer allocation. `new ubyte[n]` sets the GC “no-scan” flag for the allocated buffer while `new void[n]` doesn’t. But this is not the case here; this issue is about a struct field of fixed-size array type instead. We always assumed there is no precise GC, and the `void[n]` fixed-size array type is an error in D1, so we used `ubyte[n]` assuming it is scanned by the GC.

rainers:
Thanks, just in time for https://github.com/dlang/druntime/pull/2418. I guess it will be some time until this gets into a tagged version that can be used by tests on buildkite?

mihails-strasuns:
Hard to say, there is at least one other change that definitely needs to get into this minor release, don't know if anything else comes. Few weeks should be a reasonable estimate.

mihails-strasuns:
@rainers released

PR #664, opened 2018-12-10
Add select cycle callbacks to eventLoop

matthias-wende-sociomantic:
Generalizes the idea of a single select cycle hook
to an array of hooks.

### DIFF
--- a/src/ocean/io/select/EpollSelectDispatcher.d
+++ b/src/ocean/io/select/EpollSelectDispatcher.d
@@ -54,6 +54,8 @@ import ocean.core.array.Search;
 
 import ocean.util.container.AppendBuffer;
 
+import ocean.util.container.queue.DynamicQueue;
+
 import ocean.time.timeout.model.ITimeoutManager;
 
 import ocean.sys.Epoll;
@@ -89,6 +91,14 @@ public class EpollSelectDispatcher : IEpollSelectDispatcherInfo
      **************************************************************************/
 
     alias ISelectClient.Event Event;
+
+    /***************************************************************************
+
+        Delegate to be called each time select cycle finished.
+
+     **************************************************************************/
+
+    alias void delegate () SelectCycleCallback;
 
     /***************************************************************************
 
@@ -144,6 +154,15 @@ public class EpollSelectDispatcher : IEpollSelectDispatcherInfo
     ***************************************************************************/
 
     private epoll_event_t[] selected_set;
+
+    /***************************************************************************
+
+        Queue of delegates to be called (and dismissed) after the
+        epoll select cycle finishes.
+
+    ***************************************************************************/
+
+    private DynamicQueue!(SelectCycleCallback) select_cycle_callbacks;
 
     /***************************************************************************
 
@@ -269,6 +288,9 @@ public class EpollSelectDispatcher : IEpollSelectDispatcherInfo
         this.handle = this.timeout_enabled?
             new TimeoutSelectedKeysHandler(&this.unregister, this.e, this.timeout_manager, max_events) :
             new SelectedKeysHandler(&this.unregister, this.e);
+
+        this.select_cycle_callbacks = new DynamicQueue!(SelectCycleCallback);
+        this.select_cycle_callbacks.auto_shrink = false;
     }
 
     /***************************************************************************
@@ -298,6 +320,25 @@ public class EpollSelectDispatcher : IEpollSelectDispatcherInfo
         {
             close();
         }
+    }
+
+    /***************************************************************************
+
+        Adds new callback to the queue of delegates to be called after the
+        current select cycle. It will only be kept for a single cycle.
+
+        Params:
+            cb = callback to call once after the current cycle ends. Must return
+                `true` to block event loop from terminating in the current
+                cycle. Returning `false` is primarily intended for internal
+                epoll/scheduler facilities and unlikely to be of use in an
+                application code.
+
+    ***************************************************************************/
+
+    public void onCycleEnd ( SelectCycleCallback cb )
+    {
+        this.select_cycle_callbacks.push(cb);
     }
 
     /***************************************************************************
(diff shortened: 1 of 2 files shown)


mihails-strasuns:
I had something a bit different in mind. Essentially replacing https://github.com/sociomantic-tsunami/ocean/blob/v4.x.x/src/ocean/task/Scheduler.d#L162 with `FixedRingQueue!(void delegate()) cycle_hooks`  and turning resuming of a suspended task into simply one kind of possible cycle hooks (+ deprecating old facility). That can be done on purely scheduler level too, without affecting `EpollSelectDispatcher`.

I have rough implementation in the works right now, will submit by tomorrow.

matthias-wende-sociomantic:
Why do you want to have the cycle_hooks on scheduler level and not in the `EpollSelectDispatcher`? It doesn't make any semantic difference, right? But I would like to have it in the dispatcher since then it can be used by potentially different users.

src/ocean/io/select/EpollSelectDispatcher.d
```diff
@@ -300,6 +317,36 @@ public class EpollSelectDispatcher : IEpollSelectDispatcherInfo
...
+
+        Registers a cycle hook that will be executed when the select cycle
+        finished.
+
+        Params:
+            hook = delegate to be called after the select cycle finished
+
+    ***************************************************************************/
+
+    public void cycleHook (bool delegate() hook)
```
david-eckardt-sociomantic: Either use the type alias here i.e.
```d
public void cycleHook (selectCycleHook hook)
```
or don't use a type alias at all (my personal preference).

mihails-strasuns:
> Why do you want to have the cycle_hooks on scheduler level and not in the EpollSelectDispatcher?

This is a tough call. Key thing to keep in mind is that this is intended to replace `suspended_tasks` facility and right now there are some places where the scheduler makes use of knowledge about this queue state (for example, shutdown process). I am not sure if it is strictly necessary to keep it this way but it was my first intuitive call to keep things as similar as possible to the status quo.

Also historically we kept modifications to `EpollSelectDispatcher` to bare minimum building stuff on top instead (primarily out of feat to break some weird legacy stuff) but I guess it is less of a concern now.

I don't have a hard opinion either way.

src/ocean/io/select/EpollSelectDispatcher.d
```diff
@@ -740,10 +787,17 @@ public class EpollSelectDispatcher : IEpollSelectDispatcherInfo
         bool caller_work_pending = false;
 
         if (select_cycle_hook !is null)
-            caller_work_pending = select_cycle_hook();
+            select_cycle_hooks ~= select_cycle_hook;
+
+        foreach (hook; select_cycle_hooks)
+        {
+            caller_work_pending |= hook();
```
david-eckardt-sociomantic: `hook()` may call `cycleHook` and/or `removeCycleHook`. This leads to complications which need to be addressed:
1. `cycleHook` increases the length of the `select_cycle_hooks` array. This can cause buffer relocation, which would turn the array iterating over into a dangling slice. Using a `for` loop instead should be safe AFAICS.
2. `removeCycleHook` changes the order of the elements in `select_cycle_hooks`, it moves the last to the position of the removed element. If this happens during this iteration and the removed hook has already been called, the previously last hook in the array will be skipped. I can think of two ways to prevent this from happening:
    - Do not move array elements in `removeCycleHook`, instead set the deleted element to `null`.
    - Use a linked list rather than an array for `select_cycle_hooks`.

mihails-strasuns: Yeah this is one thing that should be addressed by https://github.com/sociomantic-tsunami/ocean/pull/666

gavin-norman-sociomantic:
I'm more on the side of adding this to epoll, rather than the scheduler. As Matthias said, there are other use cases for epoll cycle hooks.

src/ocean/io/select/EpollSelectDispatcher.d
```diff
@@ -740,10 +756,17 @@ public class EpollSelectDispatcher : IEpollSelectDispatcherInfo
         bool caller_work_pending = false;
 
         if (select_cycle_hook !is null)
-            caller_work_pending = select_cycle_hook();
+            select_cycle_hooks ~= select_cycle_hook;
```
gavin-norman-sociomantic: Missing `this.`s.

src/ocean/io/select/EpollSelectDispatcher.d
```diff
@@ -740,10 +787,17 @@ public class EpollSelectDispatcher : IEpollSelectDispatcherInfo
...
+
+        foreach (hook; select_cycle_hooks)
+        {
+            caller_work_pending |= hook();
+        }
 
         while ( this.registered_clients.length && !this.shutdown_triggered )
         {
+            caller_work_pending = false;
```
matthias-wende-sociomantic, line 790: This has to be set to false after the select call or otherwise select will always block.

matthias-wende-sociomantic:
Release notes are added and rebased on latest release. This is ready for review now.

src/ocean/io/select/EpollSelectDispatcher.d
```diff
@@ -32,6 +32,7 @@ module ocean.io.select.EpollSelectDispatcher;
 import ocean.transition;
 
 import ocean.core.Verify;
+import ocean.core.array.Mutation : moveToEnd;
```
mihails-strasuns: Not used anymore

src/ocean/io/select/EpollSelectDispatcher.d
```diff
@@ -90,6 +93,14 @@ public class EpollSelectDispatcher : IEpollSelectDispatcherInfo
 
     alias ISelectClient.Event Event;
 
+    /***************************************************************************
+
+        Delegate to be called each time select cycle finished.
+
+     **************************************************************************/
+
+    alias void delegate () selectCycleCallback;
```
mihails-strasuns: It is a type, thus CamelCase per our style guidelines.

mihails-strasuns (approved):
Just a few nitpicks, LGTM

PR #665, opened 2018-12-10
EmailSender: Fix the envelope sender address

daniel-zullo:
Set the envelope sender address as it was empty and
this is where delivery problems are sent to.
Merely setting `From` in the header will not be
sufficient and bounces will return to the wrong place.

Also if `From` in the header does not match the envelope
sender, the e-mail could be considered spoofing and gets
rejected.

### DIFF
--- a/src/ocean/net/email/EmailSender.d
+++ b/src/ocean/net/email/EmailSender.d
@@ -57,7 +57,7 @@ class EmailSender
 
     public this ( )
     {
-        this.process = new Process("sendmail -t", null);
+        this.process = new Process();
     }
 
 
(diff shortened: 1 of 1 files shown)


daniel-zullo:
@don-clugston-sociomantic FYI and can you please have a look at it?

src/ocean/net/email/EmailSender.d
```diff
@@ -162,7 +162,8 @@ class EmailSender
         {
             try
             {
-                execute;
+                setArgs("sendmail", "-t", "-f", sender).execute();
```
mateusz-gozdek-sociomantic, line 165: I'm not sure if From should always be the same as Sender address. Also, perhaps less intrusive way would be to make setting sender optional?

daniel-zullo: Good point. But I think that would be a new feature, this bug-fix is mainly to prevent the envelope sender is empty

don-clugston-sociomantic:
LGTM. I like how simple and direct this fix is.

mihails-strasuns (approved):
Makes sense to me.

daniel-zullo:
Thanks!

PR #667, opened 2018-12-11
Few codecov config tweaks

mihails-strasuns:


mihails-strasuns:
Going to ask codecov support about multi-comment spam.

PR #668, opened 2018-12-11
Adapt scheduler to use epoll cycle callbacks

mihails-strasuns:
@matthias-wende-sociomantic I was not able to finish this to the state that passes tests today but this should explain what I am trying to do. Lot of stuff needs to be changed and deprecation process is pain, but I do like how it is fleshing out design-wise.

### DIFF
--- a/src/ocean/core/TypeConvert.d
+++ b/src/ocean/core/TypeConvert.d
@@ -19,6 +19,7 @@ module ocean.core.TypeConvert;
 
 import ocean.meta.traits.Indirections;
 import ocean.meta.types.Qualifiers;
+import ocean.meta.types.Function;
 import ocean.meta.types.Typedef;
 
 version ( UnitTest ) import ocean.core.Test;
@@ -323,14 +324,14 @@ unittest
             called
 
     Returns:
-        forged delegate that can be passed to any API expecting regular `void
-        delegate()`
+        forged delegate that can be passed to any API expecting regular `T
+        delegate()` where T is the return type of F
 
 *******************************************************************************/
 
-void delegate() toContextDg ( alias F ) ( void* context )
+ReturnTypeOf!(F) delegate() toContextDg ( alias F ) ( void* context )
 {
-    static assert (is(typeof(&F) == void function (void*)));
+    static assert (is(typeof({ F((void*).init); })));
 
     // This code makes use of two facts:
     //    1) The D ABI allows aggregate methods to be converted to delegates,
@@ -341,16 +342,16 @@ void delegate() toContextDg ( alias F ) ( void* context )
 
     static struct Fake
     {
-        void method ( )
+        ReturnTypeOf!(F) method ( )
         {
             void* context = cast(void*) this;
 
             // do real work via provided F function:
-            F(context);
+            return F(context);
         }
     }
 
-    void delegate() dg = &Fake.init.method;
+    auto dg = &Fake.init.method;
     dg.ptr = context;
     return dg;
 }
--- a/src/ocean/task/Scheduler.d
+++ b/src/ocean/task/Scheduler.d
@@ -26,6 +26,7 @@ import core.thread;
 import ocean.transition;
 import ocean.core.Enforce;
 import ocean.core.Verify;
+import ocean.core.TypeConvert;
 import ocean.io.select.EpollSelectDispatcher;
 import ocean.util.container.queue.FixedRingQueue;
 import ocean.meta.traits.Indirections;
@@ -81,20 +82,20 @@ final class Scheduler : IScheduler
 
     /***************************************************************************
 
-        Thrown when task is being added for later resuming via
-        Scheduler.processEvents but matching queue is full
-
-    ***************************************************************************/
-
-    private SuspendQueueFullException suspend_queue_full_e;
-
-    /***************************************************************************
-
         Worker fiber pool, allows reusing fiber memory to run different tasks
 
     ***************************************************************************/
 
     private FiberPoolWithQueue fiber_pool;
+
+    /***************************************************************************
+
+        Indicates if this.selectCycleHook is already registered for the next
+        epoll cycle.
+
+    ***************************************************************************/
+
+    bool select_cycle_hook_registered;
 
     /***************************************************************************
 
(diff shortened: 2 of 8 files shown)


src/ocean/io/select/EpollSelectDispatcher.d
```diff
@@ -802,6 +803,58 @@ public class EpollSelectDispatcher : IEpollSelectDispatcherInfo
...
+                    throw e;
+                }
+            }
+
+            // only call items already in the queue, this enables patterns
+            // like "callback that does something and adds itself again for the
+            // next cycle"
+
+            for (int i; i < this.select_cycle_callbacks.length(); ++i)
+            {
```
matthias-wende-sociomantic: When the next callback might add itself to the queue this loop is infinite, right?
Assigning `this.select_cycle_callbacks.length()` to a variable and compare against this would solve this problem.

gavin-norman-sociomantic (commented):
Aside from the slack discussion re: deprecating `eventLoop`, the basic concept LGTM.

src/ocean/io/select/EpollSelectDispatcher.d
matthias-wende-sociomantic: OK, I was to slow :). Just realized that I didn't send my review yesterday and now it's obsolete :).

mihails-strasuns: You caught the mistake, that is what matters ;)

src/ocean/io/select/EpollSelectDispatcher.d
```diff
@@ -737,16 +769,20 @@ public class EpollSelectDispatcher : IEpollSelectDispatcherInfo
 
         this.unhandled_exception_hook = unhandled_exception_hook;
 
-        bool caller_work_pending = false;
-
         if (select_cycle_hook !is null)
-            caller_work_pending = select_cycle_hook();
+            this.onCycleEnd(select_cycle_hook);
```
matthias-wende-sociomantic: This should run more then once :).

mihails-strasuns:
Tweaked design based on discussion with @matthias-wende-sociomantic in Slack. Should be review-ready no, apart from missing release notes.

matthias-wende-sociomantic (commented):
Design wise I'm happy with the changes :). I'm a bit confused about the order of the commits though and by that I mean introducing the `select_cycle_callbacks` with bool return value first adapting everything to work with it and then changing the semantics to `void` callbacks at the end.

I'd rather prefer to introduce the finalized `select_cycle_callbacks` first and then use it for changing the `delayed_resume`.

One suggestion to get a better structure could be:
- Adapt my existing PR (#664) to introduce the new `select_cycle_callbacks`
- Use this PR for `delayed_resume` and change the `select_cycle_hook` to become part of the callbacks
(these 2 would be exactly the state of this PR)
- Use the `select_cycle_callbacks` for throttler resume termination hook.
(this is the reason why this change was proposed at the beginning)

Apart from structuring I haven't looked deep into code details yet. I'll do it later (i.e. when I have more time).

mihails-strasuns:
Order of commits is definitely not final - I left it this way for review because it follows thought flow and I felt that it is easier to review this way.

You can pick relevant bits into #664 (adding release notes and stuff 👼) and I will rebase after it is merged.

mihails-strasuns:
Rebased, added release notes, squashed related commits. Also tweaked implementation a bit to avoid duplicate cycle callbacks in scheduler.

After this, the only important remaining bit is to change throttledtaskpool to use cycle callbacks too - will do that in a separate PR.

src/ocean/core/TypeConvert.d
```diff
@@ -328,9 +329,9 @@ unittest
 
 *******************************************************************************/
 
-void delegate() toContextDg ( alias F ) ( void* context )
+ReturnTypeOf!(F) delegate() toContextDg ( alias F ) ( void* context )
```
gavin-norman-sociomantic, line 332: The Returns doc should be updated.

src/ocean/task/Scheduler.d
```diff
@@ -113,6 +114,15 @@ final class Scheduler : IScheduler
...
     private SpecializedPools specialized_pools;
 
+    /***************************************************************************
+
+        Tracks how many tasks are currently pending resume via epoll
+        cycle callbacks.
+
+    ***************************************************************************/
+
+    size_t cycle_pending_task_count;
```
gavin-norman-sociomantic: Should this be private?

src/ocean/task/Scheduler.d
```diff
@@ -678,32 +673,10 @@ final class Scheduler : IScheduler
...
         }
 
-        return this.suspended_tasks.length > 0 || queued;
-    }
-
-    /***************************************************************************
-
-        Helper method which combines recurring pattern of resuming some task
-        and handling potential exceptions.
+        if (queued || this.fiber_pool.limit() == this.fiber_pool.num_busy())
```
gavin-norman-sociomantic: I don't get the use of `queued` here. Could you add a comment explaining?

src/ocean/task/Scheduler.d
```diff
@@ -678,32 +673,10 @@ final class Scheduler : IScheduler
...
+            this.epoll.onCycleEnd(&this.selectCycleHook);
```
gavin-norman-sociomantic, line 674: Should ` this.select_cycle_hook_registered` not be set to true, in this case? (The following line sets it to false.)

src/ocean/task/Scheduler.d
mihails-strasuns: It was wrong artifact of old design. Updated to match current approach.

src/ocean/task/Scheduler.d
mihails-strasuns: Was missing `else` ;)

gavin-norman-sociomantic: Sneaky ;)

gavin-norman-sociomantic (approved):
With the last tweaks, LGTM. I'll leave it for @matthias-wende-sociomantic to give the final approval, though, as he has a deeper knowledge of the current issues.

matthias-wende-sociomantic (approved):
I hereby give my final approval! :grin:

gavin-norman-sociomantic:
Excellent work chaps!

PR #669, opened 2018-12-21
Adapt `removeShift` such that it support 0 element removal

matthias-wende-sociomantic:


matthias-wende-sociomantic:
@mihails-strasuns I created a new pr for this, as I think it is a bug fix and should be independent from the other PR's.

mihails-strasuns (approved):

mihails-strasuns:
How about turning it into point release in that case? :)

PR #672, opened 2019-01-22
Fix runtime assertion

mihails-strasuns:
Looks like calling `Fiber.reset` immediately puts it into `HOLD` state, thus it is not legal to reset fiber twice even if it is never started in between.


gavin-norman-sociomantic:
For historical records (and reviewability), it'd be nice for the commit message to explain what's happening.

mihails-strasuns:
Sure, done.

nemanja-boric-sociomantic (approved):

gavin-norman-sociomantic (approved):

PR #680, opened 2019-02-26
Resolve infinite loop on a Cache clear

### LINKED ISSUE #679: Infinite loop on ICache.clear and corrupted state
ICache.clear produce an infinite loop in some situations.

ICache is a base class for a bunch of different caches.
ICache has 2 containers inside TimeToIndex and KeyToNode. 
KeyToNode is HashMap of TimeToIndex.Node* with custom allocator.
The HashMap use liked lists in each Bucket. 
Loops in Bucket's liked lists cause infinite traverse time of that lists.

Except for infinite loop on clear this bug could be a cause of a corrupted state of the hash map. 
When we have the same linked list node several times in one or many linked lists then we lose data on each hashmap write operation. The allocator can return pointer on linked list node which isn't free and contains some data. When we put new data in this node we erase data which was there before. 
The allocator can return pointer on some node several times and as a result, this node could be present several times in some list or even several times in different linked lists.

Here's a unit test to reproduce this issue

```
unittest
{
    KeyToNode am_related = new KeyToNode(100000);
    am_related.clear();
     am_related.put(9207674216414740734);
    am_related.put(8595442437537477107);
    am_related.clear();
    am_related.put(8595442437537477107);
    am_related.clear();
    am_related.put(9207674216414740734);
    am_related.put(8595442437537477106);
    am_related.put(8595442437537477108);
    am_related.put(8595442437537477110);
    am_related.put(8595442437537477112);
    am_related.clear();
    assert(false, "isn't reacheble");
}
```

Aleksandr-Mamonov-Dunnhumby:
BucketElementFreelist store free/dealocated Elements in a linked list.
on new item request it use item from linked list if it isn't empty.
if linked list is empty ArrayAllocatedFreeBucketElements ask
GenericArrayPool for a new item.

If ArrayAllocatedFreeBucketElements.clear calls GenericArrayPool.clear
then Pool marks all Elements as free. Meanwhile BucketList puts same
Elements in linked list. When BucketList will has use all Elements in a
linked list it will ask Pool for new items and a Pull will return just used
Elements (Elements which are using right now).

Without Pool.clear call Pool returns each Element once and then it reused
by BucketList.
Fix #679

### DIFF
--- a/src/ocean/util/container/cache/model/containers/KeyToNode.d
+++ b/src/ocean/util/container/cache/model/containers/KeyToNode.d
@@ -69,6 +69,8 @@ class KeyToNode: HashMap!(TimeToIndex.Node*)
     /***************************************************************************
 
         Bucket elements allocator.
+        The elements are allocated only during container construction,
+        and never freed.
 
     ***************************************************************************/
 
@@ -85,22 +87,37 @@ class KeyToNode: HashMap!(TimeToIndex.Node*)
 
     public this ( size_t n )
     {
-        super(this.allocator = new ArrayAllocatedFreeBucketElements(n), n);
+        this.allocator = new ArrayAllocatedFreeBucketElements(n);
+        super(this.allocator, n);
     }
+}
 
-    /***************************************************************************
+unittest
+{
+    KeyToNode map = new KeyToNode(5);
+    (*map.put(1)) = new TimeToIndex.Node();
+    (*map.put(2)) = new TimeToIndex.Node();
+    (*map.put(3)) = new TimeToIndex.Node();
+    (*map.put(4)) = new TimeToIndex.Node();
+    (*map.put(5)) = new TimeToIndex.Node();
+    map.clear();
+    map.clear();
+    map.clear();
+}
 
-        Removes all elements from all buckets.
-
-        Returns:
-            this instance
-
-     **************************************************************************/
-
-    public override typeof(this) clear ( )
-    {
-        super.clear();
-        this.allocator.pool.clear();
-        return this;
-    }
+unittest
+{
+    KeyToNode map = new KeyToNode(100000);
+    map.clear();
+    map.put(9207674216414740734);
+    map.put(8595442437537477107);
+    map.clear();
+    map.put(8595442437537477107);
+    map.clear();
+    map.put(9207674216414740734);
+    map.put(8595442437537477106);
+    map.put(8595442437537477108);
+    map.put(8595442437537477110);
+    map.put(8595442437537477112);
+    map.clear();
 }


src/ocean/util/container/map/model/BucketElementFreeList.d
```diff
@@ -436,5 +266,186 @@ abstract class IBucketElementFreeList: IAllocator
...
+        Params:
+            predicate = delegate which define node searching criteria
+
+        Returns:
+            return first node match to predicate
+            either
+            return null if there is no node which match to predicate
+
+     **************************************************************************/
+    public ListNode* find( bool delegate(ListNode*) predicate)
```
stefan-koch-sociomantic: why do you need this.
Linked list search is freaking (if not pointing into a contiguous memory-block).
it should not expose a find primitive.

stefan-koch-sociomantic:
at least find is private now, good.

mihails-strasuns:
I am not very familiar with container code, would prefer @gavin-norman-sociomantic to have a look.

matthias-wende-sociomantic:
I had a first quick look. Can you please explain in more detail why it is necessary to change `BucketElementFreeList` and the relation to this bugix.

I also wonder if removing the `IBucketElementFreeList` interface is really necessary in this PR, not that I'm against it, but less changes would make it simpler to review.

matthias-wende-sociomantic (approved):
Approved! It makes sense after the explanation you gave to me. Still I would wait until @gavin-norman-sociomantic gives his final approval.

src/ocean/util/container/cache/model/containers/KeyToNode.d
```diff
@@ -85,22 +85,37 @@ class KeyToNode: HashMap!(TimeToIndex.Node*)
 
     public this ( size_t n )
     {
-        super(this.allocator = new ArrayAllocatedFreeBucketElements(n), n);
+        this.allocator = new ArrayAllocatedFreeBucketElements(n);
+        super(this.allocator, n);
```
don-clugston-sociomantic, line 91: This seems to be a refactoring, nothing to do with the fix.

don-clugston-sociomantic:
This was a bit hard to understand.

From looking at this code, the bug appears to be visible from the doc comments at the start of the module:
` Limits the number of available mappings to a fixed value and preallocates all bucket elements in an array buffer.`

When the container is cleared, the bucket elements should remain preallocated. So it shouldn't clear them.

I suggest changing the doc comments for the allocator member to be something like:
`
        Bucket elements allocator.
        The elements are allocated only during container construction, and never freed.
`

src/ocean/util/container/cache/model/containers/KeyToNode.d
Aleksandr-Mamonov-Dunnhumby: yes, it's a refactoring

matthias-wende-sociomantic:
Agreed, this is hard to understand. I checked again with @Aleksandr-Mamonov-Dunnhumby and now feel confident to merge. Thanks everyone!

don-clugston-sociomantic:
LGTM

PR #681, opened 2019-02-26
Conversion to D2-only for v5.0.0

mihails-strasuns:
Includes automatic conversion from d1to2fix and removal of `version(D_Version2)` blocks.
More cleanup will follow in separate PRs after - there are plenty of things to improve but I am not sure how many will be finished by the time I leave.


src/ocean/core/SmartEnum.d
```diff
@@ -639,11 +629,11 @@ private template ShortestName ( T ... )
...
-        enum_id ~ " length = " ~ CTFE.toString(T.length) ~ "; " ~
-        enum_id ~ " min = " ~ CTFE.toString(MinValue!(T)) ~ "; " ~
-        enum_id ~ " max = " ~ CTFE.toString(MaxValue!(T)) ~ "; " ~
-        enum_id ~ " min_descr_length = " ~ CTFE.toString(ShortestName!(T)) ~ "; " ~
-        enum_id ~ " max_descr_length = " ~ CTFE.toString(LongestName!(T)) ~ "; ";
+        "enum length = " ~ CTFE.toString(T.length) ~ "; " ~
+        "enum min = " ~ CTFE.toString(MinValue!(T)) ~ "; " ~
+        "enum max = " ~ CTFE.toString(MaxValue!(T)) ~ "; " ~
+        "enum min_descr_length = " ~ CTFE.toString(ShortestName!(T)) ~ "; " ~
+        "enum max_descr_length = " ~ CTFE.toString(LongestName!(T)) ~ "; ";
```
leandro-lucarella-sociomantic, line 636: This looks unrelated...

leandro-lucarella-sociomantic (commented):
I just gave it a quick look.

I'm really scared now that this is really happening...

src/ocean/core/SmartEnum.d
mihails-strasuns: It is related to a change earlier in this module - `enum_id` was a constant defined based on D version. Now it can be removed and only possible value used directly.

matthias-wende-sociomantic (approved):
Approved apart from the "Remvoe" spelling mistake (in one of the commit messages).

Although given that many of the changes have a fairly automatic nature or maybe even because of that, it might be worth waiting for a second review this time.

src/ocean/application/components/PidLock.d
```diff
@@ -161,14 +161,14 @@ public struct PidLock
         }
 
         // Clear the any previous contents of the file
-        this.enforcePosix!(ftruncate)(this.lock_fd, 0);
+        (&this).enforcePosix!(ftruncate)((&this).lock_fd, 0);
```
Geod24: Personally I would suggest not adding those in the first place. They are just noise that needs to be removed afterwards, except for a very few of them which can be reviewed case-by-case.

mihails-strasuns: That was my first approach (with tweaked d1to2fix) but after trying it in practice I got scared of regressions because cases like `return this` can potentially keep compiling but with a changed meaning. It feels safer to cleanup it in chunks that actually can be reasonably reviewed.

mihails-strasuns:
- Fixed commit title typo

stefan-koch-sociomantic (approved):
this is the type of PR which can only be done by a blindfold merge.
I give my blindfolded approval.

mihails-strasuns:
Will wait for acknowledgement from @gavin-norman-sociomantic and merge after.

mihails-strasuns:
Updated:
- remove lot of redundant `&this` with some sed

PR #684, opened 2019-03-11
Add a cli arguments parser for date intervals

bogdan-szabo-sociomantic:


relnotes/dateintervalargs.feature.md
```diff
@@ -0,0 +1,7 @@
+### Add a cli arguments parser for date intervals
+
+`ocean.text.arguments.DateIntervalArgs`
+
+DateIntervalArgs is an utility forr the apps that get a time interval in the
```
scott-gibson-sociomantic: For*

src/ocean/text/arguments/DateIntervalArgs.d
```diff
@@ -0,0 +1,429 @@
...
+    public long begin ( )
+    {
+        if ( this.cli_begin == 0 && this.cli_end == 0 )
+        {
+            return time(null) - this.default_interval;
+        }
+
+        if ( this.cli_begin == 0 )
+        {
+            return time(null);
```
scott-gibson-sociomantic: Hmm i'm not sure if having this value change every time it is called is a good idea. Perhaps it should be set to whatever time "processArgs" was called.

src/ocean/text/arguments/DateIntervalArgs.d
```diff
@@ -0,0 +1,429 @@
...
+
+        return this.cli_begin;
+    }
+
+    /***************************************************************************
+
+        Get the parsed end-date value
+
+        Returns:
+            The interval end as a timestamp
```
scott-gibson-sociomantic: Good idea to specify unix timestamp here.

src/ocean/text/arguments/DateIntervalArgs.d
```diff
@@ -0,0 +1,429 @@
...
+    {
+        this.default_interval = default_interval;
+    }
+
+    /***************************************************************************
+
+        Get the parsed begin-date value
+
+        Returns:
+            The interval begin as a timestamp
```
scott-gibson-sociomantic: Good idea to specify unix timestamp here.

src/ocean/text/arguments/DateIntervalArgs.d
```diff
@@ -0,0 +1,429 @@
...
+
+    /// The interval that should be used when dates are not set by the user
+    private long default_interval;
+
+    /**************************************************************************
+
+        Constructor
+
+        Params:
+            default_interval = the default interval iin seconds when the
```
scott-gibson-sociomantic: iin -> in.

src/ocean/text/arguments/DateIntervalArgs.d
```diff
@@ -0,0 +1,429 @@
...
+public class DateIntervalArgs
+{
+    /// The begin and end of the run, if given on the command line
+    private long cli_begin;
+
+    /// ditto
+    private long cli_end;
+
+    /// The interval that should be used when dates are not set by the user
+    private long default_interval;
```
scott-gibson-sociomantic: Default interval in what? Years? Seconds? Minutes? I assume seconds so i'd rename to default_interval_sec.

src/ocean/text/arguments/DateIntervalArgs.d
```diff
@@ -0,0 +1,429 @@
...
+        Returns:
+            The interval end as a timestamp
+
+    ***************************************************************************/
+
+    public long end ( )
+    {
+        if ( this.cli_end == 0 )
+        {
+            return time(null);
```
scott-gibson-sociomantic: Same goes for this, I believe it should be based on when processArgs was called, otherwise it's an ever changing value.

src/ocean/text/arguments/DateIntervalArgs.d
```diff
@@ -0,0 +1,429 @@
...
+    long tempValue;
+
+    char unit = value[value.length - 1];
+
+    if ( !toLong(value[0..$-1], tempValue) )
+    {
+        throw new Exception("The provided time interval has an invalid value.");
+    }
+
+    switch ( unit )
```
scott-gibson-sociomantic: this looks like it'd be easy to support seconds. Any reason not too?

scott-gibson-sociomantic (changes requested):
A few comments but in general looks useful.

src/ocean/text/arguments/DateIntervalArgs.d
bogdan-szabo-sociomantic: no reason. I just added it.

relnotes/dateintervalargs.feature.md
```diff
@@ -0,0 +1,7 @@
+### Add a cli arguments parser for date intervals
+
+`ocean.text.arguments.DateIntervalArgs`
+
+DateIntervalArgs is an utility for the apps that get a time interval for the
```
mihails-strasuns: Needs example.

src/ocean/text/arguments/DateIntervalArgs.d
```diff
@@ -0,0 +1,436 @@
...
+
+    copyright: Copyright (c) 2018 dunnhumby Germany GmbH. All rights reserved
+
+*******************************************************************************/
+
+module ocean.text.arguments.DateIntervalArgs;
+
+import core.stdc.time;
+
+import ocean.util.app.model.IApplication;
```
mihails-strasuns: Why this import is needed?

src/ocean/text/arguments/DateIntervalArgs.d
```diff
@@ -0,0 +1,436 @@
...
+import ocean.text.convert.Integer;
+import ocean.transition;
+
+version (UnitTest)
+{
+    import Test = ocean.core.Test;
+}
+
+/// Class that handles the begin-date and end-date CLI arguments
+public class DateIntervalArgs
```
mihails-strasuns: It doesn't look like this needs to be a class, struct with a few static methods seems suitable too (and thus should be preferred as more memory-friendly solution).

src/ocean/text/arguments/DateIntervalArgs.d
```diff
@@ -0,0 +1,436 @@
...
+        Template Params:
+            required = True if the fields are mandatory
+
+        Params:
+            app = The application
+            args = The arguments to add to
+
+    ***************************************************************************/
+
+    public void setupArgs ( bool required ) ( Arguments args )
```
mihails-strasuns: I don't see why `required` is a template argument here.

src/ocean/text/arguments/DateIntervalArgs.d
```diff
@@ -0,0 +1,436 @@
...
+    /***************************************************************************
+
+        Process the begin-date and end-date arguments
+
+        Params:
+            args = The arguments to process
+
+    ***************************************************************************/
+
+    public void processArgs ( Arguments args )
```
mihails-strasuns: How about making this a static method which returns `DateInterval` struct instead?

src/ocean/text/arguments/DateIntervalArgs.d
```diff
@@ -0,0 +1,436 @@
...
+        }
+
+        long result;
+        DateConversion dummy_conv;
+        if ( timeToUnixTime(value, result, dummy_conv) )
+        {
+            result += date_time;
+        }
+
+        return result;
```
mihails-strasuns: This will silently do nothing in case neither of conversions work. You probably want to throw some exception.

src/ocean/text/arguments/DateIntervalArgs.d
```diff
@@ -0,0 +1,436 @@
...
+            this.cli_begin = time(null);
+        }
+
+        if ( this.cli_end == 0 )
+        {
+            this.cli_end = time(null);
+        }
+    }
+
+    private long getlong ( cstring value, long date_time )
```
mihails-strasuns: Method documentation

mihails-strasuns (changes requested):
I am very sorry for a delayed review, this has somehow slipped past my notifications.

src/ocean/text/arguments/DateIntervalArgs.d
bogdan-szabo-sociomantic: I guess there is no reason to have a struct with static methods... they can be some functions

bogdan-szabo-sociomantic:
@scott-gibson-sociomantic @mihails-strasuns ready for another look

src/ocean/text/arguments/DateIntervalArgs.d
mihails-strasuns: I assumed you may want to keep it for namespacing purpose (those method names are rather widespread) but if it is not a concern, sure, free functions will do just fine too.

mihails-strasuns (approved):
CI is red now but design-wise this looks good to me. There is a natural direction for enhancement for cases when one wants to have multiple time intervals arguments but I think it can wait until there is a real use case.

bogdan-szabo-sociomantic:
great! I fixed the D2 problems

mihails-strasuns:
Thank you for taking time to contribute to ocean :)

@scott-gibson-sociomantic any last remarks from your side?

scott-gibson-sociomantic (approved):
No more changes from me, i'm happy with the improvements made.

PR #687, opened 2019-03-22
Drop wchar/dchar support from ocean.text.Util

mihails-strasuns:
This PR can also be used as form of textbook example of applying `inout` to our codebase and how to workaround some of its limitations.

### DIFF
--- a/src/ocean/text/Search.d
+++ b/src/ocean/text/Search.d
@@ -17,7 +17,7 @@
 
 module ocean.text.Search;
 
-import ocean.transition;
+import ocean.meta.types.Qualifiers : mstring, cstring;
 
 import Util = ocean.text.Util;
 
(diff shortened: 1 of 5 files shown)


src/ocean/text/Search.d
```diff
@@ -553,7 +552,6 @@ public struct SearchFruct(T)
 
     private void reset ()
     {
-        auto what = cast(char[]) this.what;
         if (fore)
             for (ptrdiff_t i=0; i < what.length; ++i)
```
matthias-wende-sociomantic: You may want to add `this` to `what`, right?

matthias-wende-sociomantic (commented):
I think there are some missing this's ;). Other than that I approve.

mihails-strasuns:
`this` updated

PR #688, opened 2019-03-25
Ignore `remaining_elements` in `FullQueryParamSet`

### LINKED ISSUE #559: Adapt `QueryParamSet.opApply` to not include `remaining_elements`
`opIn_r` doesn't report these extra values as being present, but `opApply` iterates over them. This is a pretty inconsistent API.

See discussion at https://github.com/sociomantic-tsunami/ocean/pull/558/files#r205364031.

mihails-strasuns:
Fixes #559

### DIFF
--- a/relnotes/paramset.breaking.md
+++ b/relnotes/paramset.breaking.md
@@ -1,0 +1,9 @@
+### Ignore remaining elements in `FullQueryParamSet.opApply`
+
+Now `opApply` behavior in `FullQueryParamSet` is consistent with other methods
+such as `opIn` and will ignore remaining (not specified in constructor)
+elements.
+
+This is a silent breaking change in runtime behavior - please review your app
+usage of `FullQueryParamSet` carefully and switch to plain `QueryParamSet` if
+necessary.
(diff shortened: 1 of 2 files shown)


matthias-wende-sociomantic (commented):
Wouldn't it be better to implement the `opIn` to get the consistency instead of removing the `opApply`? That probably depends on the general usage of that class and might be a question for @gavin-norman-sociomantic.

mihails-strasuns:
AFAIU `opIn` is already present and works like `opApply` after this change. See the discussion linked from the issue.

matthias-wende-sociomantic:
> AFAIU opIn is already present and works like opApply after this change.

Yes that's true. What I mean is that due to this patch it will change the semantics of the `opApply`. Instead it would also be possible to change how the `opIn` works by overriding the `opIn` from the super class.

mihails-strasuns:
Right, but why would you want that? To me excluding unregistered elements makes more sense considering this is a derived class with sole purpose of limiting set of expected parameters.

mihails-strasuns:
Replaced with https://github.com/sociomantic-tsunami/ocean/pull/696

PR #689, opened 2019-03-25
Start using package.d feature

### LINKED ISSUE #686: Switch from `package_.d` to `package.d` in v5.x.x
Now that ocean v5.x.x branch is D2-only, it is possible to replace ad-hoc `package_.d` modules with `package.d` ones natively supported by compiler.

mihails-strasuns:
Fixes #686


mihails-strasuns:
Trivial, self-merging.

PR #691, opened 2019-04-02
"yesterday" and single value support improvements for TimeIntervalArgs.

scott-gibson-sociomantic:
This PR adds the following features
1. specifying "yesterday" will use the unix timestamp of the previous day
2. specifying a single parameter now supported
    - With timestamp (including 'yesterday' it will be 24 hours range after given time.
    - With a duration provided it will be a range of X in the past to the current time.

it also adds an overload for the 'time' method to unsure the unit tests are using a stable value.

---

I've added a second commit to standardize how the interval is defined. It is to never include the final
parsed second. For example, `"-t 1h" with now:1000 (start:940, end:999)`. This was done after an internal poll and discussion.

### DIFF
--- a/src/ocean/text/arguments/TimeIntervalArgs.d
+++ b/src/ocean/text/arguments/TimeIntervalArgs.d
@@ -2,6 +2,42 @@
 
     Handler for time interval CLI arguments
 
+    One Argument
+    ------------
+
+    Date:
+        Gives the range for that day.
+        "-t 2019-04-01" (start: 2019-04-01 00:00:00, end: 2019-04-01 23:59:59 )
+
+    TimeStamp
+        Gives a range from the provided timestamp plus one day.
+        "-t 1554132392" (start: 1554132392, end: 1554218791)
+
+    Timerange
+        Gives X amount in the past NOT including the current second.
+        "-t 1m" with now:1000 (start:940, end:999)
+
+    Two Arguments
+    -------------
+
+    Dates:
+        Start is parsed inclusively, end is parsed exclusively
+        eg. -t "2019-04-01 2019-04-02"
+            (start: 2019-04-01 00:00:00)
+            (  end: 2019-04-02 23:59:59)
+
+    TimeStamp
+        Start is treated as is, end is NOT included in final range.
+        eg. -t 100 200 = (begin: 100, end: 199)
+
+    Timerange
+        End range includes the starting second.
+        eg. -t 100 1m = (begin: 100, end: 159)
+        eg. -t 1m 100 = (begin: 60, end: 99)
+
+    Intended Range Usage
+        for ( long i = range.begin; i <= range.end; ++i )
+
     copyright: Copyright (c) 2018 dunnhumby Germany GmbH. All rights reserved
 
 *******************************************************************************/
@@ -56,6 +92,10 @@ public void setupTimeIntervalArgs ( Arguments args, bool required )
             "an integer for unix timestamps\n\t" ~
             "a time duration `{int}m`. Supported units are [m]inutes, [h]ours and [d]ays\n\t" ~
             "an iso1806 date (YYYY-MM-DD)");
+    args("time-interval-exclude")
+        .help("If not set, then a range of '-t 2019-04-01 2019-04-02' will \n" ~
+              "include the end date's data. eg. (end: 2019-04-02 23:59:59) \n" ~
+              "If set, end date data will not be included, eg.(end: 2019-04-01 23:59:59)");
 
     if ( required )
     {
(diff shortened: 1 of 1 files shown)


scott-gibson-sociomantic:
One question that perhaps needs answering is if if the current second should be included in the 'end' of a range when calling with a single time interval. Eg. if i said create a range of 1d in the past, should it include the current second because when we provide a range of "2019-01-01 to 2019-02-01" we don't include 2019-02-01 00:00:00 in the range.

Edit:
Currently it is including the 'now' second, but i think it should be changed for consistency sake.

mihails-strasuns:
@scott-gibson-sociomantic I'd ask that question to actual users of this utility :) Neither seems more "correct" to me, it is simply a matter of practical convenience.

bogdan-szabo-sociomantic:
I think it's up to the app to use a closed or opened time interval. So probably there is enough to do `interval.end = interval.begin + SECONDS_IN_DAY;` here: https://github.com/sociomantic-tsunami/ocean/pull/691/files#diff-417c9e96a89af2c59b9e11a8237611adR117

otherwise LGTM

src/ocean/text/arguments/TimeIntervalArgs.d
```diff
@@ -2,6 +2,39 @@
...
+        Gives the range for that day.
+        "-t 2019-04-01" (start: 2019-04-01 00:00:00, end: 2019-04-01 23:59:59 )
+
+    TimeStamp
+        Gives a range from the provided timestamp plus one day.
+        "-t 1554132392" (start: 1554132392, end: 1554218791)
+
+    Timerange
+        Gives X amount in the past NOT including the current second.
+        "-t 1h" with now:1000 (start:940, end:999)
```
bogdan-szabo-sociomantic: with this case, `1h` = `59m59s`

I know that this is equivalent for `-t 3600s`, but this will just run for the last `3599` seconds, which might not make much sense :smile: 

I also wonder how this should work:
"-t 1s" with now:1000 (start:999, end:999)?

also, what `-t 0h` should do?

src/ocean/text/arguments/TimeIntervalArgs.d
```diff
@@ -48,12 +82,17 @@ public void setupTimeIntervalArgs ( Arguments args, bool required )
...
-        .params(2)
-        .help("Specify a time interval. It get 2 values, where they can be:\n\t" ~
+        .params(1, 2)
+        .help("Specify a time interval. It 1 or 2 values, where they can be:\n\t" ~
             "'now' an alias for the current timestamp\n\t" ~
+            "'yesterday' an alias for yesterdays date\n\t" ~
             "an integer for unix timestamps\n\t" ~
             "a time duration `{int}m`. Supported units are [m]inutes, [h]ours and [d]ays\n\t" ~
             "an iso1806 date (YYYY-MM-DD)");
+    args("time-interval-exclude")
```
bogdan-szabo-sociomantic, line 95: this argument might be too much ...

src/ocean/text/arguments/TimeIntervalArgs.d
```diff
@@ -76,10 +115,10 @@ public void setupTimeIntervalArgs ( Arguments args, bool required )
 
 public istring validateTimeIntervalArgs ( Arguments args )
 {
-    if ( args["time-interval"].assigned.length != 2 &&
-        args["time-interval"].assigned.length != 0 )
+    if ( args["time-interval"].assigned.length < 1 ||
```
bogdan-szabo-sociomantic: this might be more readable:
```
 auto num_args = args["time-interval"].assigned.length;
 if ( num_args < 1 || num_args > 2 )
```

bogdan-szabo-sociomantic (approved):
Just a few notes, but it LGTM.

src/ocean/text/arguments/TimeIntervalArgs.d
scott-gibson-sociomantic: The '1h" was a mistake, it should say `1m`.
-t 3600s is still 3600 seconds worth of data, because it is intended that the ranges' end is to be used in the query eg.
```d
for ( long i = range.begin; i <= range.end; ++i )
```

`-t 1s` would mean you get the previous second of data. So you'd get the data from just 999.

-t 0h i believe should be an error (if it's not already)

src/ocean/text/arguments/TimeIntervalArgs.d
scott-gibson-sociomantic: Added

src/ocean/text/arguments/TimeIntervalArgs.d
scott-gibson-sociomantic: Added a commit that treats 0 as an error.

bogdan-szabo-sociomantic:
can we merge this one? please?

PR #692, opened 2019-04-02
Restore "Add suspendables to throttler in consistent state"

### LINKED ISSUE #657: Re-add reverted throttler consistency check
https://github.com/sociomantic-tsunami/ocean/pull/640 was released but then reverted, as it introduced a breaking change. (Applications that use a throttler in conjunction with a swarm neo suspendable request encountered runtime errors. See https://github.com/sociomantic-tsunami/ocean/pull/651.)

mihails-strasuns:
Fixes #657

### DIFF
--- a/src/ocean/io/model/ISuspendableThrottler.d
+++ b/src/ocean/io/model/ISuspendableThrottler.d
@@ -65,7 +65,9 @@ abstract public class ISuspendableThrottler
     /***************************************************************************
 
         Adds a suspendable to the list of suspendables which are to be
-        throttled. If it is already in the list, nothing happens.
+        throttled if it's not already in there.
+        Also ensures that the state of the added suspendable is consistent
+        with the state of the throttler.
 
         Params:
             s = suspendable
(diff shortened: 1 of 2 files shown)


src/ocean/io/model/ISuspendableThrottler.d
```diff
@@ -74,16 +76,76 @@ abstract public class ISuspendableThrottler
...
+        suspendable.suspend();
+        suspendable_throttler.addSuspendable(suspendable);
+
+        test!("==")(suspendable.suspended(), false,
+            "Suspended suspendable not resumed.");
+
+        suspendable_throttler.suspendAll();
+
+        // add a resumed ISuspendable to a suspended throttler
+        suspendable.resume();
```
matthias-wende-sociomantic, line 143: Actually this line seems superfluous.

matthias-wende-sociomantic: Ignore my comment. Maybe it's worth having it. It seems easier to understand the test when you set the state explicitly at the beginning :).

matthias-wende-sociomantic (approved):

src/ocean/io/model/ISuspendableThrottler.d
mihails-strasuns: You wrote this originally, I have only copied it :D

PR #695, opened 2019-04-03
Replace hex string literal in Tiger.d with ordinary string literal

JinShil:
Hex string literals are deprecated in D.  We are trying to remove them from the language (https://github.com/dlang/dmd/pull/9549).  Since Ocean is part of D's test suite, we need this hex string remedied.


mihails-strasuns:
@JinShil it is already fixed in v5.x.x branch, and not intended to be addressed in v4.x.x

Current plan is to release v5.0.0 this week.

PR #699, opened 2019-04-24
 Add a client for sending stats to prometheus

tiyash-basu-sociomantic:
Prometheus sends HTTP GET requests to an application at `/metrics` endpoint to pull stats from it. This PR adds a listener for these requests, and a handler to response with stats to prometheus.

Also accompanying here are classes that interface between the handler and a client application, such that the handler is able to invoke callbacks that collect stats from the client application upon receiving requests from prometheus.

### DIFF
--- a/src/ocean/util/prometheus/collector/Collector.d
+++ b/src/ocean/util/prometheus/collector/Collector.d
@@ -1,0 +1,327 @@
+/*******************************************************************************
+
+    Contains methods that collect stats from primitive data members of structs
+    or classes to respond to Prometheus queries with.
+
+    The metric designs specified by Prometheus
+    (`https://prometheus.io/docs/concepts/metric_types/`) have not been
+    implemented yet. So, as of now, this collector can process only primitive
+    data-type members from a struct or a class. However, the current Prometheus
+    stat collection framework could be enhanced to support metrics with a very
+    little effort.
+
+    In Prometheus' data model, the stats that are measured are called Metrics,
+    and the dimensions along which stats are measured are called Labels.
+
+    Metrics can be any measurable value, e.g., CPU or memory consumption.
+
+    Labels resemble key-value pairs, where the key is referred to as a label's
+    name, and the value as a label's value. A label name would refer to the name
+    of a dimension across which we want to measure stats. Correspondingly, a
+    label value would refer to a point along the said dimension. A stat can have
+    more than one label, if it is intended to be measured across multiple
+    dimensions.
+
+    Stats with labels look like the following example
+    `
+    promhttp_metric_handler_requests_total{code="200"} 3
+    promhttp_metric_handler_requests_total{code="500"} 0
+    promhttp_metric_handler_requests_total{code="503"} 0
+    `
+    Here `promhttp_metric_handler_requests_total` is the metric, `code` is
+    the label name, `"200"`, `"500"` and `"503"` are the label values, and `3`,
+    `0` and `0` are the respective metric values.
+
+    On the data visualization side of Prometheus, fetching stats using queries
+    is analogous to calling functions. A metric name is analogous to a function
+    name, and a label is analogous to a function parameter. Continuing with the
+    above example, the following query will return `3`:
+    `
+    promhttp_metric_handler_requests_total {code="200"}
+    `
+
+    (For detailed information about Prometheus-specific terminology, e.g.,
+    Metrics and Labels, please refer to
+    `https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels`.)
+
+    This module contains a class which can be used to collect metrics as well as
+    labels from composite-type values, and format them in a way acceptable for
+    Prometheus.
+
+    Copyright:
+        Copyright (c) 2019 dunnhumby Germany GmbH.
+        All rights reserved
+
+    License:
+        Boost Software License Version 1.0. See LICENSE.txt for details.
+
+*******************************************************************************/
+
+module ocean.util.prometheus.collector.Collector;
+
+/*******************************************************************************
+
+    This class provides methods to collect metrics with no label, one label, or
+    multiple labels, using overloaded definitions of the `collect` method.
+
+    Once the desired stats have been collected, the accumulated response
+    string can be collected using the `getCollection` method.
+
+    Additionally, the `reset` method can be used to reset stat
+    collection, when the desired stats have been collected from a Collector
+    instance.
+
+*******************************************************************************/
+
+public class Collector
+{
+    import core.stdc.time;
+    import Traits = ocean.core.Traits;
+    import ocean.math.IEEE;
+    import ocean.text.convert.Formatter;
+    import ocean.transition;
+    import StatFormatter = ocean.util.prometheus.collector.StatFormatter;
+
+    /// A buffer used for storing collected stats. Is cleared when the `reset`
+    /// method is called.
+    private mstring collect_buf;
+
+    /***************************************************************************
+
+        Returns:
+            The stats collected since the last call to the `reset` method, in a
+            textual respresentation that can be readily added to a response
+            message body.
+            The specifications of the format of the collected stats can be found
+            at `https://prometheus.io/docs/instrumenting/exposition_formats/`.
+
+    ***************************************************************************/
+
+    public cstring getCollection ( )
+    {
+        return this.collect_buf;
+    }
+
+    /// Reset the length of the stat collection buffer to 0.
+    public void reset ( )
+    {
+        this.collect_buf.length = 0;
+        enableStomping(this.collect_buf);
+    }
+
+    /***************************************************************************
+
+        Collect stats from the data members of a struct or a class and prepare
+        them to be fetched upon the next call to `getCollection`. The
+        specifications of the format of the collected stats can be found at
+        `https://prometheus.io/docs/instrumenting/exposition_formats/`.
+
+        Params:
+            ValuesT = The struct or class type to fetch the stat names from.
+            values  = The struct or class to fetch stat values from.
+
+    ***************************************************************************/
+
+    public void collect ( ValuesT ) ( ValuesT values )
+    {
+        static assert (is(ValuesT == struct) || is(ValuesT == class),
+            "'values' parameter must be a struct or a class.");
+
+        StatFormatter.formatStats(values, this.collect_buf);
+    }
+
+    /***************************************************************************
+
+        Collect stats from the data members of a struct or a class, annotate
+        them with a given label name and value, and prepare them to be fetched
+        upon the next call to `getCollection`.
+        The specifications of the format of the collected stats can be found
+        at `https://prometheus.io/docs/instrumenting/exposition_formats/`.
+
+        Params:
+            LabelName = The name of the label to annotate the stats with.
+            ValuesT   = The struct or class type to fetch the stat names from.
+            LabelT    = The type of the label's value.
+            values    = The struct or class to fetch stat values from.
+            label_val = The label value to annotate the stats with.
+
+    ***************************************************************************/
+
+    public void collect ( istring LabelName, ValuesT, LabelT ) (
+        ValuesT values, LabelT label_val )
+    {
+        static assert (is(ValuesT == struct) || is(ValuesT == class),
+            "'values' parameter must be a struct or a class.");
+
+        StatFormatter.formatStats!(LabelName)(values, label_val,
+            this.collect_buf);
+    }
+
+    /***************************************************************************
+
+        Collect stats from the data members of a struct or a class, annotate
+        them with labels from the data members of another struct or class, and
+        prepare them to be fetched upon the next call to `getCollection`.
+        The specifications of the format of the collected stats can be found
+        at `https://prometheus.io/docs/instrumenting/exposition_formats/`.
+
+        Params:
+            ValuesT = The struct or class type to fetch the stat names from.
+            LabelsT = The struct or class type to fetch the label names from.
+            values  = The struct or class to fetch stat values from.
+            labels  = The struct or class holding the label values to annotate
+                      the stats with.
+
+    ***************************************************************************/
+
+    public void collect ( ValuesT, LabelsT ) ( ValuesT values, LabelsT labels )
+    {
+        static assert (is(ValuesT == struct) || is(ValuesT == class),
+            "'values' parameter must be a struct or a class.");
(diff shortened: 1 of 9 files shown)


src/ocean/util/prometheus/server/PrometheusListener.d
```diff
@@ -0,0 +1,161 @@
...
+import ocean.util.prometheus.server.PrometheusHandler;
+
+/*******************************************************************************
+
+    A listener for prometheus' pull requests. Listens to the `/metrics` endpoint
+    of the given server address, and invokes the handler for incoming requests.
+
+    Derives from `SelectListener` with it's handler specified as
+    `PrometheusHandler`.
+
```
gavin-norman-sociomantic: Could you document the template params?

src/ocean/util/prometheus/server/PrometheusListener.d
```diff
@@ -0,0 +1,161 @@
+/*******************************************************************************
+
+    A listener for prometheus requests to pull stats.
+
+    Once instantiated with the set of callbacks that collect stats for given
+    stats and labels, the callbacks will be called for every incoming prometheus
+    request, and the accumulated stats will be appended to the response message.
+
+    Example:
```
gavin-norman-sociomantic: Would this work as a documented unittest?

src/ocean/util/prometheus/server/PrometheusHandler.d
```diff
@@ -0,0 +1,155 @@
...
+/*******************************************************************************
+
+    An HTTP response handler for prometheus' requests, used in
+    `PrometheusListener`
+
+    Responds to GET requests on;y..
+
+*******************************************************************************/
+
+public class PrometheusHandler : HttpConnectionHandler
```
gavin-norman-sociomantic, line 22: Ah, shame we don't have an `HttpTaskConnectionHandler` yet, but never mind.

src/ocean/util/prometheus/server/PrometheusHandler.d
```diff
@@ -0,0 +1,155 @@
...
+module ocean.util.prometheus.server.PrometheusHandler;
+
+import ocean.net.http.HttpConnectionHandler;
+
+/*******************************************************************************
+
+    An HTTP response handler for prometheus' requests, used in
+    `PrometheusListener`
+
+    Responds to GET requests on;y..
```
gavin-norman-sociomantic: Typo.

src/ocean/util/prometheus/server/PrometheusHandler.d
```diff
@@ -0,0 +1,155 @@
...
+            HTTP status code
+
+    ***************************************************************************/
+
+    override protected HttpResponseCode handleRequest (
+        out cstring response_msg_body )
+    {
+        if (this.request.uri_string() != "/metrics")
+        {
+            PrometheusHandler.log.info("Received unexpected request uri: {}",
```
gavin-norman-sociomantic: Warning?

src/ocean/util/prometheus/server/PrometheusHandler.d
```diff
@@ -0,0 +1,155 @@
...
+        }
+
+        try
+        {
+            response_msg_body = this.collector_registry.collect();
+            return HttpResponseCode.OK;
+        }
+        catch (Exception ex)
+        {
+            err_buf.length = 0;
```
gavin-norman-sociomantic, line 109: Also log the exception?

src/ocean/util/prometheus/collector/CollectorRegistry.d
```diff
@@ -0,0 +1,209 @@
...
+        All rights reserved
+
+    License:
+        Boost Software License Version 1.0. See LICENSE.txt for details.
+
+*******************************************************************************/
+
+module ocean.util.prometheus.collector.CollectorRegistry;
+
+/*******************************************************************************
```
gavin-norman-sociomantic: You can use `/// ditto` here.

src/ocean/util/prometheus/collector/CollectorRegistry.d
```diff
@@ -0,0 +1,209 @@
...
+        this.collector_callbacks ~= collector;
+    }
+
+    /***************************************************************************
+
+        Collects stats by calling all delegates specified for stat collection
+        and returns the accumulated result.
+
+        Returns:
+            The stats accumulated since last time this method was called.
```
gavin-norman-sociomantic, line 74: Worth noting what format these are in? (e.g. mention the module that does the rendering / link to online doc about the format?)

src/ocean/util/prometheus/collector/CollectorRegistry.d
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    ///
+    public void collectDg2 ( Collector collector )
+    {
+        collector.collect(test_stats, test_labels);
+    }
+}
+
+/// Test collection from a single delegate
```
gavin-norman-sociomantic, line 154: Are these intended as tests or usage examples? (They look like tests to me, as I don't suppose using `toDg` is recommended usage.) Any unittest commented with `///` will be treated as a usage example, so you might want to just use `//` for these.

src/ocean/util/prometheus/collector/Collector.d
```diff
@@ -0,0 +1,332 @@
+/*******************************************************************************
+
+    Contains methods that collect stats from primitive data members of structs
+    or classes to respond to prometheus queries with.
+
+    The metric designs specified by prometheus
+    (https://prometheus.io/docs/concepts/metric_types/) have not been
+    implemented yet. So, as of now, this collector can can process only
+    primitive data-type members from a struct or a class. However, the current
+    prometheus stat collection framework can support metrics with very little
```
gavin-norman-sociomantic: "can support" -> "could be enhanced to support" ?

src/ocean/util/prometheus/collector/Collector.d
```diff
@@ -0,0 +1,332 @@
...
+    1.  For collecting stats from a struct or class:
+        `void collect ( ValuesT ) ( ValuesT values )`
+
+        Example:
+        ```
+        CpuMemoryStats system_stats = new CpuMemoryStats();
+
+        void collect ( Collector collector )
+        {
+            collector.collect(this.system_stats.collect());
```
gavin-norman-sociomantic: Maybe add a comment: `CpuMemoryStats.collect() returns a struct whose fields we want to collect via Prometheus.`

src/ocean/util/prometheus/collector/Collector.d
```diff
@@ -0,0 +1,332 @@
...
+        }
+        ```
+
+    2.  For collecting stats from a struct or class, and with one label:
+        `void collect ( istring LabelName, ValuesT, LabelT ) (
```
gavin-norman-sociomantic: You need to explain how the label works. It's not clear to me, as someone who's not read anything about Prometheus before.

src/ocean/util/prometheus/collector/Collector.d
```diff
@@ -0,0 +1,332 @@
...
+        ```
+        CpuMemoryStats system_stats = new CpuMemoryStats();
+
+        void collect ( Collector collector )
+        {
+            collector.collect!("pid")(this.system_stats.collect(), getpid());
+        }
+        ```
+
+    3.  For collecting stats from a struct or class, along with a set of labels
```
gavin-norman-sociomantic: Ditto about the labels.

src/ocean/util/prometheus/collector/Collector.d
```diff
@@ -0,0 +1,332 @@
...
+    The metric designs specified by prometheus
+    (https://prometheus.io/docs/concepts/metric_types/) have not been
+    implemented yet. So, as of now, this collector can can process only
+    primitive data-type members from a struct or a class. However, the current
+    prometheus stat collection framework can support metrics with very little
+    effort.
+
+    This class provides three methods for stat collection:
+
+    1.  For collecting stats from a struct or class:
```
gavin-norman-sociomantic: Could these examples be documented unittests? (Verbal explanations here, code examples as documented unittests?)

src/ocean/util/prometheus/collector/Collector.d
```diff
@@ -0,0 +1,332 @@
...
+    import ocean.math.IEEE;
+    import ocean.text.convert.Formatter;
+    import ocean.transition;
+    import ocean.util.prometheus.collector.StatFormatter;
+
+    /// A buffer used for storing collected stats. Is cleared when the `reset`
+    /// method is called.
+    private mstring collect_buf;
+
+    /// Default constructor.
```
gavin-norman-sociomantic: I don't think there's any need to include this.

src/ocean/util/prometheus/collector/Collector.d
```diff
@@ -0,0 +1,332 @@
...
+    public this ( ) { }
+
+    /***************************************************************************
+
+        Returns:
+            The stats collected until the last call to the `reset` method, in a
```
gavin-norman-sociomantic: "until" -> "since".

src/ocean/util/prometheus/collector/Collector.d
```diff
@@ -0,0 +1,332 @@
...
+            textual respresentation that can be readily added to a response
+            message body.
```
gavin-norman-sociomantic, line 94: Would also be nice to link something that describes the format here.

src/ocean/util/prometheus/collector/Collector.d
```diff
@@ -0,0 +1,332 @@
...
+
+    struct Labels
+    {
+        hash_t id;
+        cstring job;
+        float perf;
+    }
+}
+
+/// Test collecting stats having initial values, but without any label.
```
gavin-norman-sociomantic: Looks like these are tests, not usage examples, so should have `//` instead of `///`.

src/ocean/util/prometheus/collector/StatFormatter.d
```diff
@@ -0,0 +1,384 @@
...
+import ocean.net.http.TaskHttpConnectionHandler;
+
+/*******************************************************************************
+
+    Contains methods that format primitive data members from structs or classes
+    into strings that can be exported as responses to prometheus queries.
+
+*******************************************************************************/
+
+public final abstract class StatFormatter
```
gavin-norman-sociomantic: More idiomatic would be to not use a class and instead declare your functions at module scope. If anyone wants to wrap them in a "namespace", they can do so with a named import.

gavin-norman-sociomantic (commented):
Some minor comments, but generally this is looking really good. Nicely structured. Good work!

The biggest thing, for me, is the lack of description about some of the Prometheus terminology, e.g. the labels in the `format` functions.

src/ocean/util/prometheus/server/PrometheusHandler.d
tiyash-basu-sociomantic: We do have a `ocean.net.http.TaskHttpConnectionHandler`, which makes using the task system (or at least the global `theScheduler` instance) mandatory. I thought using `HttpConnectionHandler` would be slightly less restrictive, while offering the same set of functionalities.

Do you suggest that we use `TaskHttpConnectionHandler` here?

src/ocean/util/prometheus/server/PrometheusHandler.d
tiyash-basu-sociomantic: I will update the message being logged here to be more specific.

Prometheus only needs the `/metrics` endpoint for pulling stats, and hence every other endpoint results in a `NotImplemented` response here. However, this is a general HTTP handler listening at a port, so anyone can send it a GET request at any endpoint. So, in that sense, perhaps it is not necessarily a warning?

src/ocean/util/prometheus/collector/Collector.d
tiyash-basu-sociomantic: The examples there were using `CpuMemoryStats.collect()` method, which would return dynamic values, which may make it a bit dodgy for unit tests. A similar example now exists as a documented unit test in `PrometheusListener`.

This module already had some unit tests (the ones you mentioned about [here](https://github.com/sociomantic-tsunami/ocean/pull/699#discussion_r278489075)). Now only some act as documentation/examples (the rest are now only test-cases) which roughly cover the use-cases of in this class.

src/ocean/util/prometheus/collector/CollectorRegistry.d
tiyash-basu-sociomantic: I removed usages of `toDg` from these tests (it looks like it was needed before because the calbacks were module-level functions). If it's alright, I have kept the tests commented with `///` to make them serve as examples as well.

tiyash-basu-sociomantic:
Thanks for the review!

I have now added some additional documentation about the Prometheus-specific terminology used here. I have also included links to relevant pages on Prometheus' website as references.

Ready for another look.

src/ocean/util/prometheus/server/PrometheusHandler.d
gavin-norman-sociomantic: Ha! I was looking in the wrong place for it ;)

Yeah, on the whole I agree that compatibility is probably more important here.

src/ocean/util/prometheus/collector/Collector.d
gavin-norman-sociomantic: The reason for avoiding example code in comments is that it's not being compiled, so is very vulnerable to getting out-dated.

tiyash-basu-sociomantic: Right, I agree with you. I also noticed, that although I removed the examples, I forgot to remove the method signatures. :)

Edit: Removed the method signatures as well.

tiyash-basu-sociomantic:
Created a `.keep` file in the `relnotes` folder, and added a release note for the prometheus integration feature.

src/ocean/util/prometheus/server/PrometheusListener.d
gavin-norman-sociomantic: Oh wait, sorry, I misread this. I thought that these were params of `PrometheusListener`, whereas they're actually params used to instantiate `SelectListener`. Probably better to remove the comments then. Sorry!

gavin-norman-sociomantic (commented):
New doc greatly improves comprehensibility :+1:

tiyash-basu-sociomantic:
I have removed the documentation as you pointed out.

Thanks for the review!

PR #700, opened 2019-04-30
Reset all fields in the gc stats

bogdan-szabo-sociomantic:
@scott-gibson-sociomantic noticed that not all fields are updated after a reset. This is a fix for that.

### DIFF
--- a/src/ocean/application/components/GCStats.d
+++ b/src/ocean/application/components/GCStats.d
@@ -95,7 +95,7 @@ public class GCStats
     private void reset ( )
     {
         this.last_collected_timestamp = MicrosecondsClock.now_us();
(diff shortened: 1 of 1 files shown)


gavin-norman-sociomantic:
Is this a bug fix? (If so, it should be applied to v4.5.x -- the oldest supported minor branch.)

matthias-wende-sociomantic:
Can you please add more context, why and where this is actually needed.

ben-palmer-sociomantic:
> Is this a bug fix?

Agree this doesn't look like a bug fix to me.

ben-palmer-sociomantic:
Any progress on this? Or I will close it as it seems to be inactive.

PR #701, opened 2019-04-30
 Explicitely handle `void*` types in `ReduceType`

tiyash-basu-sociomantic:
Dereferencing in typeof(*(void*).init) causes the compilation error
```
Error: expression *null is void and has no value
```

indicating that `void*` cannot be implicitly reduced. So, it must be handled explicitly.

### DIFF
--- a/src/ocean/meta/types/ReduceType.d
+++ b/src/ocean/meta/types/ReduceType.d
@@ -212,3 +212,82 @@ private struct ReduceTypeImpl ( Reducer )
         return result;
     }
 }
+
+
+version (UnitTest)
+{
+    import ocean.meta.types.Qualifiers;
+
+    private struct TestAggregate
+    {
+        int x;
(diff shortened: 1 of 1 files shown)


gavin-norman-sociomantic:
@tiyash-basu-sociomantic if this is a patch, it should be based on v5.0.x.

gavin-norman-sociomantic (commented):
LGTM.

Just needs:
- Correct branch.
- FIXME comment with link to issue.

matthias-wende-sociomantic (approved):

matthias-wende-sociomantic:
I'm happy to merge as soon as Gavin's comments are addressed .

tiyash-basu-sociomantic:
Rebased on correct branch, and added a FIXME comment linking to issue https://github.com/sociomantic-tsunami/ocean/issues/704.

PR #702, opened 2019-04-30
[WIP] Test d1 changes from ocean4 in the CI

matthias-wende-sociomantic:


matthias-wende-sociomantic:
not needed any longer

PR #707, opened 2019-05-08
Add message when the gc monitor is not available

bogdan-szabo-sociomantic:
Right now there is no way to check if your binary can use this feature. With this feature, you will be notified if the feature is available or not.

### DIFF
--- a/src/ocean/application/components/GCStats.d
+++ b/src/ocean/application/components/GCStats.d
@@ -39,6 +39,9 @@ public class GCStats
     }
     else
     {
+        pragma(msg, "There will be no gc stats! You need to use dmd transitional:");
+        pragma(msg, "https://github.com/sociomantic-tsunami/dmd-transitional");
+
         /// Default type for integers used by monitor
         private alias long MonitorInt;
     }


src/ocean/application/components/GCStats.d
```diff
@@ -39,6 +39,8 @@ public class GCStats
     }
     else
     {
+        pragma(msg, "There will be no gc stats! You need to use dmd transitional.");
```
gavin-norman-sociomantic: Could you add a link to where someone can find info about dmd-transitional?

bogdan-szabo-sociomantic: hmm... like this one? https://github.com/sociomantic-tsunami/dmd-transitional

gavin-norman-sociomantic: Yep.

bogdan-szabo-sociomantic:
I'm not sure how relevant it is, but since dmd 2.085, there is a feature that allows you to query the gc stats.

https://dlang.org/changelog/2.085.0.html#core-memory-gc-profile-stats

gavin-norman-sociomantic:
Aha, interesting. I wonder if anyone here knows if that addition was a PR of ours that was merged, or is compatible with the stats in dmd-transitional, or is something completely different. @leandro-lucarella-sociomantic @don-clugston-sociomantic @stefan-koch-sociomantic any idea?

mihails-strasuns:
@gavin-norman-sociomantic completely different - it simply exposes the same data you could previously get via `--DRT-opt`, there is no monitor replacement.

leandro-lucarella-sociomantic:
Exactly, and I think you still have to run the app with some `--DRT` option to enable profiling. I don't remember either how it impacts performance.

src/ocean/application/components/GCStats.d
```diff
@@ -39,6 +39,9 @@ public class GCStats
     }
     else
     {
+        pragma(msg, "There will be no gc stats! You need to use dmd transitional:");
+        pragma(msg, "https://github.com/sociomantic-tsunami/dmd-transitional");
+
```
leandro-lucarella-sociomantic, line 44: This whole module requires `GC.monitor` to make sense right? So this will only trigger a message for people really depending on `GC.monitor`?

bogdan-szabo-sociomantic: yes

bogdan-szabo-sociomantic:
I don't know what and how it is implemented, but the api https://dlang.org/phobos/core_memory.html#.GC.ProfileStats has exactly what we need

leandro-lucarella-sociomantic:
I think the message should also mention `dmd1` or this should be applied only to `v5.x.x`.

leandro-lucarella-sociomantic:
OK, I just checked and it is very weird. Most stats are collected regardless of the presence of the profile GC runtime flag, but not all of them. Weirdly enough, all timing stats (which I would assume are expensive as timing information is usually gotten via syscalls) is always available, but `maxPoolMemory` (which only involves branching once and one MUL) is not. This is actually the only stat that is not being collected without the runtime flag, so I guess this is just a bug.

leandro-lucarella-sociomantic:
Ah, actually that's not an issue, because the `GC.ProfileStats` only retrieve a very limited subset of the stats collected, which makes it even weirder that so many unused stats are collected... But then having the stats from phobos/the new GC api should be doable from v2.085+, we could add support to ocean for that.

ben-palmer-sociomantic:
LGTM, any more comments from anyone?

leandro-lucarella-sociomantic (approved):

leandro-lucarella-sociomantic:
Related: https://github.com/sociomantic-tsunami/ocean/issues/711

ben-palmer-sociomantic (approved):

PR #708, opened 2019-05-17
Rewrite outdated usage documentation to compilable example

johannes-riecken-dunnhumby:
The example usage in the documentation did not give correct results,
because json.print() now returns a string and takes different
parameters. Using compilable examples makes sure that usage examples can
still compile.

### DIFF
--- a/src/ocean/text/json/Json.d
+++ b/src/ocean/text/json/Json.d
@@ -980,7 +980,7 @@ class Json(T) : JsonParser!(T)
                     if (type is typeid(Composite))
                         v.set (va_arg!(Composite)(args));
                     else
-                    if (type is typeid(T[]))
+                    if (type is typeid(istring))
                         v.set (va_arg!(T[])(args));
                     else
                     if (type is typeid(void*))
--- a/src/ocean/text/json/Json.d
+++ b/src/ocean/text/json/Json.d
@@ -31,60 +31,61 @@ import ocean.text.json.JsonParser;
 
 import Float = ocean.text.convert.Float;
 
-version(UnitTest) import ocean.core.Test;
+version (UnitTest)
+{
+    import ocean.core.Test;
+    import ocean.io.Stdout;
+}
+
+///
+unittest
+{
+
+    // Typical usage is as follows:
+
+    auto json = new Json!(char);
+    json.parse(`{"t": true, "n":null, "array":["world", [4, 5]]}`);
+
+
+    // Convert back to text format:
+
+    Stdout.formatln("{}", json.value.print());
+
+
+    // Constructing json within your code leverages a handful of factories
+    // within a document instance. This example creates a document from an
+    // array of values:
+
+    with (json)
+        value = array (true, false, null, "text");
+    test!("==")(json.value.print(), `[true, false, null, "text"]`);
+
+
+    // Setting the document to contain a simple object instead:
+
+    with (json)
+        value = object (pair("a", value(10)));
+    test!("==")(json.value.print(), `{"a":10}`);
+
+
+    // Objects may be constructed with multiple attribute pairs like so:
+
+    with (json)
+        value = object (pair("a", value(10)), pair("b", value(true)));
+    test!("==")(json.value.print(), `{"a":10,"b":true}`);
+
+
+    // Substitute arrays, or other objects as values where appropriate:
+
+    with (json)
+        value = object (pair("a", array(10, true, object(pair("b")))));
+    test!("==")(json.value.print(), `{"a":[10, true, {"b":null}]}`);
+}
+
 
 /*******************************************************************************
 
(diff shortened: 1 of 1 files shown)


leandro-lucarella-sociomantic:
Super weird that the diff coverage of an **unittest** is only 6.25%, and that the total coverage increased (even if by almost nothing). On top of that codecov is failing to load for me so I can't check what's going on. Besides that, LGTM.

src/ocean/text/json/Json.d
```diff
@@ -31,60 +31,64 @@ import ocean.text.json.JsonParser;
...
+version (UnitTest)
+{
+    import ocean.core.Test;
+    import ocean.io.Stdout;
+}
+
+///
+unittest
+{
+    void example ( )
```
ben-palmer-sociomantic: I am not sure how exactly comments as unit tests work but do you need to actually run `example()`?

src/ocean/text/json/Json.d
```diff
@@ -31,60 +31,64 @@ import ocean.text.json.JsonParser;
...
+        /*
+            Typical usage is as follows:
+        */
+        auto json = new Json!(char);
+        json.parse(`{"t": true, "n":null, "array":["world", [4, 5]]}`);
+
+        /*
+            Convert back to text format:
+        */
+        Stdout.formatln("{}", json.value.print());
```
ben-palmer-sociomantic: Just want to check we need this output or can we use `test`?

ben-palmer-sociomantic (commented):
Couple of minor questions.

ben-palmer-sociomantic:
Is there any way to kick the codecov to run again?

src/ocean/text/json/Json.d
gavin-norman-sociomantic: Nope, usage examples (`///` followed by `unittest`) don't need to be run, they just need to be compiled.

leandro-lucarella-sociomantic:
I guess by force-rebasing and force-pushing.

johannes-riecken-dunnhumby:
> Super weird that the diff coverage of an **unittest** is only 6.25%, and that the total coverage increased (even if by almost nothing). On top of that codecov is failing to load for me so I can't check what's going on. Besides that, LGTM.

I assume that the way the code coverage tool works is that it sees how much of outside code is exercised within an assertion so that the `test!...` statements increases the code coverage of the definition of the json module's value function and print function and that's it? That would explain the results.

src/ocean/text/json/Json.d
johannes-riecken-dunnhumby: I guess it's a matter of taste, since the `example()`s are not run as unit tests any way. I've put in the `test` statements to show without running what the expected output should be and the `Stdout.formatln` statements to make it easy to adjust things and view the outputs in one place on the console when pasting the `example()` contents into a standalone main module for exploration.

ben-palmer-sociomantic:
> I guess by force-rebasing and force-pushing.

@johannes-riecken-dunnhumby could you give that a go and see if it makes this green?

ben-palmer-sociomantic:
Actually it seems to fail in D2. Can you test that locally?

johannes-riecken-dunnhumby:
> Actually it seems to fail in D2. Can you test that locally?

Yes, the json module doesn't work correctly in D2. I think I'll just allocate one hour daily for the three days left of this week on fixing this as it's a nice debugging practise with a medium-size relatively self-contained code, and if I still can't get it working, then I'll just close the PR. I can't just put the unit test in something like `version (D1)`, because then the code coverage would be too low when building for D2.

ben-palmer-sociomantic:
Changes look ok to me. I would just clean up the changes so you have:
1 commit with the D2 fixes
1 commit with the unit test changes

johannes-riecken-dunnhumby:
> Changes look ok to me. I would just clean up the changes so you have:
> 1 commit with the D2 fixes
> 1 commit with the unit test changes

Cool, done!

src/ocean/text/json/Json.d
```diff
@@ -31,60 +31,61 @@ import ocean.text.json.JsonParser;
...
-    Constructing json within your code leverages a handful of factories
-    within a document instance. This example creates a document from an
-    array of values:
-    ---
+///
+unittest
+{
+    /*
+        Typical usage is as follows:
+    */
```
ben-palmer-sociomantic: I think our normal style for both single and multi-line inline documentation is `//`. Same goes for the other comments.

ben-palmer-sociomantic (commented):
One more comment on the documentation style.

src/ocean/text/json/Json.d
johannes-riecken-dunnhumby: There you go!

ben-palmer-sociomantic:
LGTM

PR #710, opened 2019-05-20
Fix unnecessary memory allocation

daniel-zullo:
Otherwise a non collectable buffer will be allocated
leading to a memory leak.

### DIFF
--- a/src/ocean/io/device/File.d
+++ b/src/ocean/io/device/File.d
@@ -455,13 +455,10 @@ class File : Device, Device.Seek, Device.Truncate
 
                 // allocate enough space for the entire file
                 auto len = cast(size_t) file.length;
-                if (dst.length < len){
-                    if (dst is null){ // avoid setting the noscan attribute, one should maybe change the return type
-                        dst=new ubyte[](len);
-                    } else {
-                        dst.length = len;
-                        enableStomping(dst);
-                    }
+                if (dst.length < len)
+                {
+                    dst.length = len;
+                    enableStomping(dst);
                 }
 
                 //read the content


ben-palmer-sociomantic:
LGTM

matthias-wende-sociomantic (approved):

Geod24:
> Otherwise a non collectable buffer will be allocated leading to a memory leak.

Lacking context here, and curious about this construct. Why does it leads to a memory leak ? Since the buffer is `null` ?
I guess the original NOSCAN concern was for 32 bits and is much less of a concern nowadays (although it might affect GC mark phase).

stefan-koch-sociomantic:
I'll be honest I don't exactly know why this lead to leak. But it surely did ;)

don-clugston-sociomantic:
So the original purpose of the code was to ensure that the GC doesn't look for pointers inside the file contents.
But does this PR fix the leak, actually? Initially I misread this, I thought it was an `if (!dest.length)` so it would do an unnecessary allocation if the array had been shrunk. But since it explicitly checks for a null pointer, it should only ever be called once. So I think it should make no difference.

stefan-koch-sociomantic:
in our usecase the buffer was only used once.
and although it has gone out of scope it still wasn't collected.

daniel-zullo:
> Lacking context here, and curious about this construct. Why does it leads to a memory leak ? Since the buffer is null ?

Actually the leak was in the application side due to the buffer was used in a delegate.

stefan-koch-sociomantic:
Still this code is leads to puzzlers.
Actully the fix would be to `ubyte[]` everywhere in the file "api".

daniel-zullo:
> So the original purpose of the code was to ensure that the GC doesn't look for pointers inside the file contents.
> But does this PR fix the leak, actually? Initially I misread this, I thought it was an `if (!dest.length)` so it would do an unnecessary allocation if the array had been shrunk. But since it explicitly checks for a null pointer, it should only ever be called once. So I think it should make no difference.

That's also right. I think allocating a new buffer is misleading but that was not the intention of this PR. (Yesterday I wrongly thought this was the cause that the GC was not collecting the buffer)

matthias-wende-sociomantic:
Why is this closed? Even if it doesn't fix a bug it still seems to simplify the code or am I missing something?

stefan-koch-sociomantic:
it might cause additonal gc issues to appear when merged.
Then again, it might not.

ben-palmer-sociomantic:
I agree that this is useful code simplification. Should probably target a different branch though.

PR #713, opened 2019-06-04
Update deprecation message for `isStringType` 

tiyash-basu-sociomantic:
~This template has been deprecated in `ocean.core.Traits`, and the
[deprecation message](https://github.com/sociomantic-tsunami/ocean/blob/v5.x.x/src/ocean/core/Traits.d#L538) points to the template added here, which was missing
until now.~

 The deprecation message should redirect to `ocean.meta.traits.Arrays.isUTF8StringType` instead of `ocean.meta.traits.Basic.isStringType`

### DIFF
--- a/src/ocean/core/Traits.d
+++ b/src/ocean/core/Traits.d
@@ -535,7 +535,7 @@ public template TemplateInstanceArgs (alias Template, Type : Template!(TA), TA..
     public alias TA TemplateInstanceArgs;
 }
 
-deprecated("Use ocean.meta.traits.Basic.isStringType")
+deprecated("Use ocean.meta.traits.Arrays.isUTF8StringType")
 template isStringType( T )
 {
     static immutable bool isStringType = is( T : char[] )  ||


src/ocean/meta/traits/Basic.d
```diff
@@ -92,6 +92,46 @@ unittest
...
+
+    Params:
+        T = type to check
+
+    Returns:
+        `true` if `T` is a string type, false otherwise.
+
+*******************************************************************************/
+
+public template isStringType( T )
```
ben-palmer-sociomantic: Should we have a space between `isStringType` and `(`?

ben-palmer-sociomantic (commented):
One very minor formatting fix then LGTM.

src/ocean/meta/traits/Basic.d
```diff
@@ -92,6 +92,46 @@ unittest
...
+{
+    static immutable bool isStringType =
+        is( T : char[] )
+        || is( T : dchar[] )
+        || is( T : wchar[] )
+        || is( T : istring )
+        || is( T : cstring )
+        || is( T : mstring );
```
Geod24: Weeeell there's an interesting question here. Is a `Typedef` (or anything else that implicitly converts to a string) a string type ? If it is, you can just go with:
```
is(T : const(char)[]) || is(T : const(wchar)[]) || is(T : const(dchar)[])
```
Note however that we removed non-UTF8 support from many places in Ocean so you might just be fine reducing it to one condition.

But if `Typedef` should not match `isStringType`, then you need a long list of `is(T == ...)`.

src/ocean/meta/traits/Basic.d
```diff
@@ -92,6 +92,46 @@ unittest
...
+    static assert ( isStringType!(const(char[])) );
+    static assert ( isStringType!(immutable(char)[]) );
+    static assert ( isStringType!(immutable(char[])) );
+
+    static assert ( isStringType!(dchar[]) );
+    static assert ( isStringType!(wchar[]) );
+
+    static assert ( isStringType!(istring) );
+    static assert ( isStringType!(cstring) );
+    static assert ( isStringType!(mstring) );
```
Geod24: There is no test for `!isStringType`

tiyash-basu-sociomantic:
Well, it looks like there's already a `ocean.meta.traits.Arrays.isUTF8StringType`, so there is no need of implementing this here again.
The deprecation message needs to be fixed, however. I'll do it here.

Edit: On a second thought, maybe for now it's better to keep `ocean.meta.traits.Basic.isStringType` in this PR, until the following question is answered.

@mihails-strasuns @gavin-norman-sociomantic @leandro-lucarella-sociomantic Is the deprecation message [here](https://github.com/sociomantic-tsunami/ocean/blob/v5.x.x/src/ocean/core/Traits.d#L538) supposed to redirect to `ocean.meta.traits.Arrays.isUTF8StringType`, or should `ocean.meta.traits.Basic.isStringType` still be implemented?

src/ocean/meta/traits/Basic.d
tiyash-basu-sociomantic: Looking at other templates that were put in `ocean.meta.traits.Basic` from `ocean.core.Traits`, it seems that the functionality of a given template in `ocean.meta.traits.Basic` should be the same as that in `ocean.core.Traits`. So, I do not think the implementation of the template here should try to consider `Typedef`s (or other implicitly convertible types) as `string`s. What do you suggest?

ben-palmer-sociomantic: It could create problems for people porting from the deprecated code if the new code behaves differently I guess.

Geod24: >  it seems that the functionality of a given template in ocean.meta.traits.Basic should be the same as that in ocean.core.Traits. So, I do not think the implementation of the template here should try to consider Typedefs (or other implicitly convertible types) as strings. What do you suggest?

Well... The implementation you copied does support `Typedef`. Try `isStringType` on a `Typedef!string`. The question was basically: Should it be `is(T == v)` or `is(T : v)`.

mihails-strasuns:
It is a bug in the deprecation message, intention was to direct towards `ocean.meta.traits.Arrays.isUTF8StringType`. We made a decision some time ago to start dropping support of `dchar` and `wchar` strings from ocean so that would be a (tiny) part of it.

src/ocean/meta/traits/Basic.d
tiyash-basu-sociomantic: You're right, I missed that.
Anyway, as Mihails mentioned](https://github.com/sociomantic-tsunami/ocean/pull/713#issuecomment-498819829), I'll delete this template and update the deprecation message instead.

tiyash-basu-sociomantic:
Thanks for the response! Updated to fix the deprecation message.

Geod24 (approved):
PR title needs fixing, otherwise LGTM

ben-palmer-sociomantic:
I think you should be targeting the v4.x.x branch. We could also actually target v4.5.x as a bug fix if you really wanted but might be easier not to.

tiyash-basu-sociomantic:
> I think you should be targeting the v4.x.x branch. We could also actually target v4.5.x as a bug fix if you really wanted but might be easier not to.

It should target v5.0.x, since the deprecation message is absent in the v4.5.x branch. It was wrong to target v5.x.x anyway. Thanks for pointing out!

ben-palmer-sociomantic:
> It should target v5.0.x, since the deprecation message is absent in the v4.5.x branch

Ah right missed that. Cool.

tiyash-basu-sociomantic:
Changed target branch to v5.0.x and rebased.

matthias-wende-sociomantic (approved):

ben-palmer-sociomantic:
LGTM

PR #714, opened 2019-06-18
Make getIeeeFlags() naked

### LINKED ISSUE #496: D2 ocean fails to build if ldmd2 is used as the compiler
Current D2 ocean builds with recent dmd releases (e.g. 2.078.x, 2.079.x).  However, if `ldmd2` (the DMD-alike LDC frontend) is used, then unittests fail with the following message:

```
./src/ocean/math/IEEE.d(293): Error: unimplemented return type IeeeFlags for implicit abi return
```

The line in question is an `asm` block, which suggests that ocean's `asm` sections will need adaptation or re-implementation if they are to be supported by LDC.

This is a real shame, as we have some use-cases where LDC's much better/faster number-crunching would be a real benefit.

david-eckardt-frequenz:
Change the return type from a struct with one `int` member to plain `int`.

This function is implemented in assembly language. The function body doesn’t have a return statement so the compiler needs to know how to return the result implicitly. DMD seems to know what to do to implicitly return a struct with one `int` member while LDC doesn’t. LDC can be helped by returning `int` instead.

Fixes #496. That issue has been closed as “wontfix”, but IMHO this is not so much to support LDC but rather to correct sloppy code which DMD accepts and LDC doesn’t. This seems to be the only remaining obstacle for compiling ocean with LDC.

### DIFF
--- a/src/ocean/math/IEEE.d
+++ b/src/ocean/math/IEEE.d
@@ -281,22 +281,26 @@ private:
         {
             asm
             {
+                 naked;
                  fstsw AX;
                  // NOTE: If compiler supports SSE2, need to OR the result with
                  // the SSE2 status register.
                  // Clear all irrelevant bits
                  and EAX, 0x03D;
+                 ret;
             }
         }
         else version(D_InlineAsm_X86_64)
         {
             asm
             {
+                 naked;
                  fstsw AX;
                  // NOTE: If compiler supports SSE2, need to OR the result with
                  // the SSE2 status register.
                  // Clear all irrelevant bits
                  and RAX, 0x03D;
+                 ret;
             }
         } else {
            /*   SPARC:


gavin-norman-sociomantic:
@don-clugston-sociomantic do you know anything about this module?

gavin-norman-sociomantic:
@david-eckardt-frequenz :wave:

ben-palmer-sociomantic:
LGTM but I agree I think @don-clugston-sociomantic probably knows the most about this.

don-clugston-sociomantic:
Yes, I wrote that code a very very long time ago. At that time there were some limitations (for example, struct constructors did not exist). 

If you attempt to compile with this change in D1, you get:
``ocean/math/IEEE.d(341): Error: cannot implicitly convert expression (getIeeeFlags()) of type IeeeFlags to int``
But now that we're dropping D1 support, we can do this. How ironic.
LGTM

gavin-norman-sociomantic:
Blocked pending fixing up the faulty v5.1.0 tag. See #715.

ben-palmer-sociomantic:
So is this unblocked now?

gavin-norman-sociomantic:
Yes.

joseph-wakeling-frequenz:
One of the LDC devs told me today that it fails because LDC does not know how to deal with implicit return values from inline ASM -- but that if the ASM block is marked as `naked` LDC will trust the user to do the right thing.

So, an alternative fix might be to just mark the inline ASM block as `naked`.

JohanEngelen:
> This function is implemented in assembly language. The function body doesn’t have a return statement so the compiler needs to know how to return the result implicitly. DMD seems to know what to do to implicitly return a struct with one `int` member while LDC doesn’t. LDC can be helped by returning `int` instead.

Indeed, LDC only knows how to do asm implicit returns for a few basic types; the inline asm part of LDC does not know the ABI in general (it only knows it for x86 for a few basic types). If you mark the `asm` as `naked`, then LDC is OK with not knowing the return-ABI and trusts the user fully and in this case that should work just as well.

Edit: I highly recommend the GCC-style inline assembly expressions of LDC for writing asm pieces: https://wiki.dlang.org/LDC_inline_assembly_expressions

matthias-wende-sociomantic:
I think using `naked` seems to be the cleaner solution. What do you think @david-eckardt-frequenz?

Btw. Herzlichen Glückwunsch zum 1. Ozean PR :smiley:!

david-eckardt-frequenz:
`naked` would be cleaner, but that would require extending the assembly routine with the prologue and epilogue required by the ABI, which is currently added by the compiler: At least a `ret` instruction needs to be added, maybe register or stack cleanup as well.
This is probably as simple as adding the `ret`, but I am not an expert here. @don-clugston-sociomantic, would this work (code comments removed):
```d
static int getIeeeFlags()
{
    version(D_InlineAsm_X86)
    {
        asm
        {
             naked;
             fstsw AX;
             and EAX, 0x03D;
             ret;
        }
    }
    else version(D_InlineAsm_X86_64)
    {
        asm
        {
             naked;
             fstsw AX;
             and RAX, 0x03D;
             ret;
        }
    }
    else
    {
        static assert(0, "Not yet supported");
    }
}
```

don-clugston-sociomantic:
Yup, that should work. `naked` is really complicated if there are parameters, but in this case there are none, so a `ret` is all you need.

matthias-wende-sociomantic:
I'm also happy with changing the return type. Your decision ;).

david-eckardt-frequenz:
Updated.

src/ocean/math/IEEE.d
```diff
@@ -275,28 +275,32 @@ private:
         }
     }
 private:
-    static IeeeFlags getIeeeFlags()
+    static int getIeeeFlags()
```
joseph-wakeling-frequenz: Shouldn't we be able to replace `int` with `IeeeFlags` again now that we have `naked` and `ret` in the asm?

src/ocean/math/IEEE.d
```diff
@@ -338,7 +342,7 @@ public:
 }
 
 /// Return a snapshot of the current state of the floating-point status flags.
-IeeeFlags ieeeFlags() { return IeeeFlags.getIeeeFlags(); }
+IeeeFlags ieeeFlags() { return IeeeFlags(IeeeFlags.getIeeeFlags()); }
```
joseph-wakeling-frequenz: & the change to this line should also be possible to drop now that `naked` and `ret` are used ... ?

david-eckardt-frequenz:
Updated, returning `IeeeFlags`.

matthias-wende-sociomantic (approved):

joseph-wakeling-frequenz:
This got merged into v5.0.x but there is no point release, and it hasn't been merged into v5.x.x either.  Any chance of a new 5.0.x release and corresponding 5.1.y release ... ? :-)

PR #717, opened 2019-07-31
Support extended arguments to task pool deserialization.

scott-gibson-sociomantic:
The main driver behind this change is to support passing runtime-dependant information
(such as DHT, DMQ instances) into the deserialized tasks as previously the only way this
could be done was to create static member variables which isn't ideal.

I've also included a small refactoring of the ThrottledTaskPool to simplify the code as the overriden
start/restore are no longer needed and can be done in an overriden startImpl

release notes coming _soon_.

### DIFF
--- a/src/ocean/task/TaskPool.d
+++ b/src/ocean/task/TaskPool.d
@@ -180,17 +180,17 @@ class TaskPool ( TaskT : Task ) : ObjectPool!(Task)
             current_task.suspend();
     }
 
-    static if (hasMethod!(TaskT, "deserialize", void delegate(void[])))
+    static if (__traits(hasMember, TaskT, "deserialize"))
     {
         /***********************************************************************
 
             Starts a task in the same manner as `start` but instead calls the
-            `deserialize()` method on the derived task with a serialized buffer
-            of the state. This is to support dumping and loading tasks from disk.
+            `deserialize()` method on the derived task with arguments supported.
+            This is to support dumping and loading tasks from disk.
 
             Params:
-                serialized = Buffer containing serialized data for restoring
-                             the internal state of a task.
+                args = Arguments matching the function arguments of the
+                       'deserialize()' function of the task type.
 
             Returns:
                 'false' if new task can't be started because pool limit is reached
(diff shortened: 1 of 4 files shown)


src/ocean/task/TaskPool.d
```diff
@@ -180,33 +180,33 @@ class TaskPool ( TaskT : Task ) : ObjectPool!(Task)
...
+                       'deserialize()' function of the task type.
 
             Returns:
                 'false' if new task can't be started because pool limit is reached
                 for now, 'true' otherwise
 
         ***********************************************************************/
 
-        public bool restore ( void[] serialized )
+        public bool restore ( Args ... ) ( Args args )
```
Geod24: That looks wrong. This method is `override`n in a derived class, but it's a templated method. I'm surprised the compiler does not complain.

E.g. the following code triggers an assertion failure:
```d
class Foo
{
    void restore (Args...) (Args a) { assert(0); }
}

class Bar : Foo
{
    override void restore (Args...) (Args a) {}
}

void main ()
{
    scope bar = new Bar();
    boom(bar);
}

void boom (scope Foo f) { f.restore(); }
```

scott-gibson-sociomantic: My original reasoning for changing this to Args... instead of using something like ParametersOf!(TaskT.deserialize) was to support multiple types of function signatures. (IIRC using ParametersOf! means only the first declaration is used. Perhaps the proposed change causes more problems that i didn't anticipate. I removed the overwritten "restore" from the ThrottledTaskPool and i could set this as final but that's technically a api change then no?

Geod24: Yes it's a breaking change. You can actually generate an overload set matching another overload set, but it's a PITA. Want the code ?

matthias-wende-sociomantic:
What about adding a new templated restore method with a different name, that is final and can then be used to implement the original (virtual) restore method.

scott-gibson-sociomantic:
I've modified this to instead of using templated arguments to use ParametersOf!(TaskT.deserialize) as well as adding extra templated methods of TaskPoolSerializer.load in case someone inherited it. I believe this makes it have no breaking changes.

scott-gibson-sociomantic:
Apologies some changes (TaskPool.d) was missing in my last push. Updated

matthias-wende-sociomantic (approved):

PR #719, opened 2019-08-08
add 'isManifestConstant' function

bogdan-szabo-sociomantic:
The 'isManifestConstant' function checks if the type of a member inside a
struct or class is a D manifest constant.

### DIFF
--- a/src/ocean/meta/traits/Basic.d
+++ b/src/ocean/meta/traits/Basic.d
@@ -660,3 +660,63 @@ unittest
 
     static assert(isTypedef!(RealNum) == TypedefKind.Struct);
 }
+
+/******************************************************************************
+
+    Check if the type of a member inside a struct or class is a D manifest
+    constant
+
+    Params:
+        T = The struct that will be checked
+        name = The struct member that will be checked
+
+******************************************************************************/
+
+bool isManifestConstant ( T, string name ) ( )
+{
+    mixin(`return is(typeof(T.init.`~name~`)) && !is(typeof(&T.init.`~name~`));`);
+}
+
+unittest
+{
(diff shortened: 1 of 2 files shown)


src/ocean/meta/traits/Basic.d
```diff
@@ -660,3 +660,38 @@ unittest
 
     static assert(isTypedef!(RealNum) == TypedefKind.Struct);
 }
+
+/******************************************************************************
+
+    Check if the type of a member inside a struct or class is a D manifest
```
gavin-norman-sociomantic, line 666: Why inside a struct or class? Would `bool isManifestConstant ( T )` not be more general?

bogdan-szabo-sociomantic: I added another version of this function

gavin-norman-sociomantic: Is the class or struct member version necessary? Can't you just do `isManifestConstant!(MyStruct.some_field)`?

bogdan-szabo-sociomantic: I guess... I removed the prev version and I will update my original code

gavin-norman-sociomantic: Please ping when this PR is updated.

bogdan-szabo-sociomantic: It looks like I need the both functions. The following test fails using `static assert(!isManifestConstant!(Test.init.CustomType));` when `assert(!isManifestConstant!(Test, "CustomType"));` works

ben-palmer-sociomantic (approved):
LGTM

PR #720, opened 2019-08-08
Add 'snakeToPascal' function which converts a snake_case name to a PascalCase

bogdan-szabo-sociomantic:


src/ocean/meta/codegen/CTFE.d
```diff
@@ -119,3 +120,52 @@ unittest
     assert (stripQualifiedPrefix("aabb") == "aabb");
     static assert (stripQualifiedPrefix("aabb") == "aabb");
 }
+
+
+/******************************************************************************
+
+    Converts a snake_case name to a PascalCase.
+    This should be used only at CTFE since is not memory efficient.
```
ben-palmer-sociomantic: Can you describe what an example use case is here?

src/ocean/meta/codegen/CTFE.d
```diff
@@ -119,3 +120,52 @@ unittest
...
+        name = the event name
+
+    Returns:
+        The name in camel case
+
+******************************************************************************/
+
+string snakeToPascal ( string name )
+{
+    string result;
```
ben-palmer-sociomantic, line 142: You could add a static assert that this is CTFE (I think there is something like `_ctfe` or something you can check)

relnotes/snaketopascal.feature.md
```diff
@@ -0,0 +1,5 @@
+### snakeToPascal
```
ben-palmer-sociomantic: This should be a catchy title/description of the change.

ben-palmer-sociomantic (commented):
Made a couple of comments.

relnotes/snaketopascal.feature.md
gavin-norman-sociomantic: Ha, I've never heard of Pascal case before. Looks like the correct term though :+1:

gavin-norman-sociomantic (approved):

relnotes/snaketopascal.feature.md
vasileios-kyrillidis-sociomantic: Cause it's actually called UpperCamelCase. Pascal just happened to use it and now ppl call it this way. So now PascalCase is supposed to be a subset of CamelCase OR plainly UpperCamelCase.

gavin-norman-sociomantic: Right, UpperCamelCase is the term I've heard before. As long as it's a standard term, I think either is fine though.

src/ocean/meta/codegen/CTFE.d
bogdan-szabo-sociomantic: why you would forbid usage of this function at runtime?

src/ocean/meta/codegen/CTFE.d
bogdan-szabo-sociomantic: I'm not sure if I can.

src/ocean/meta/codegen/CTFE.d
ben-palmer-sociomantic: I'm just going off your comment in the documentation that says it should only be used at compile time.

src/ocean/meta/codegen/CTFE.d
liza-churkyn-sociomantic: If your code is following a certain coding standard, and you're also using mixin to generate some of your code, and you want the result of your mixin to also comply with the coding standard, without forcing the client code (which calls your mixin) to pass params to your mixin that follow your coding standard. I think that's the original use case of this case.

src/ocean/meta/codegen/CTFE.d
bogdan-szabo-sociomantic: yeah... but this is why I wrote should and not must... if someone wants to use this at run time, they should assume the gc allocations

matthias-wende-sociomantic: Maybe instead of documenting it should be used only at compile time you could indicate that it is allocating memory.

bogdan-szabo-sociomantic: @matthias-wende-sociomantic i am not sure that I understand that phrase. How do you indicate that you allocate memory in ocean?

gavin-norman-sociomantic: He means that instead of saying "This should be used only at CTFE since is not memory efficient.", you could say something like "This function is designed for use at compile time and is thus not careful about avoiding allocations.".

bogdan-szabo-sociomantic: ah cool. I updated the comment.

src/ocean/meta/codegen/CTFE.d
```diff
@@ -119,3 +120,54 @@ unittest
...
+
+/******************************************************************************
+
+    Converts a snake_case name to a PascalCase.
+
+    This function is designed for use at compile time and is thus not careful
+    about avoiding allocations.
+
+    Params:
+        name = the event name
```
gavin-norman-sociomantic: Event?

relnotes/snaketopascal.feature.md
```diff
@@ -0,0 +1,6 @@
+### Clear And Unbiased Facts About `snakeToPascal` (Without All the Hype)
```
gavin-norman-sociomantic: How about "Add `snakeToPascal` string case conversion function"? ;)

ben-palmer-sociomantic (commented):
I think it would be good still to include a use case in the documentation for the function (e.g. https://github.com/sociomantic-tsunami/ocean/pull/720#discussion_r312388544 ) but otherwise LGTM. Not sure how we get the travis tests passing again though.

ben-palmer-sociomantic:
Maybe adding one runtime test of the function will satisfy codecov.

bogdan-szabo-sociomantic:
@ben-palmer-sociomantic i updated the test to be run at runtime

ben-palmer-sociomantic:
LGTM, @liza-churkyn-sociomantic any other comments?

PR #723, opened 2019-08-09
Added HttpServer class as a simple alternative to start a web server

bogdan-szabo-sociomantic:


src/ocean/net/http/HttpServer.d
```diff
@@ -0,0 +1,232 @@
...
+    /**************************************************************************
+
+        Constructor
+
+        Params:
+            finalizer = finalizer callback of the select listener
+
+    **************************************************************************/
+
+    public this ( scope FinalizeDg finalizer, HttpHandler[] handlers )
```
Geod24: You really don't want to use `scope` here

src/ocean/net/http/HttpServer.d
```diff
@@ -0,0 +1,232 @@
...
+        return HttpResponseCode.OK;
+    }
+
+    auto config = HttpConfig("0.0.0.0", 8080);
+    auto server = new HttpServer([ &iDoNothing, &iDoSomething ]);
+
+    // The following line registers the server with the dispatcher, which in
+    // effect makes it operational, subject to the latter's scheduling.
+    // (Commented out here to not start a server instance in a unit test.)
+    // server.start(config);
```
Geod24: Just use a nested function which is never called:
```d
unittest
{
    void main ()
    {
        /// Magic
    }
}
```

src/ocean/net/http/HttpServer.d
```diff
@@ -0,0 +1,232 @@
...
+
+            return HttpResponseCode.BadRequest;
+        }
+
+        return HttpResponseCode.NotFound;
+    }
+
+    /**************************************************************************
+
+        Tells the request message body length.
```
Geod24, line 184: Missing newline to separate the doc

src/ocean/net/http/HttpServer.d
```diff
@@ -0,0 +1,232 @@
...
+    {
+        size_t size;
+
+        try
+        {
+            this.request.getUnsigned("Content-Length", size);
+        }
+        catch(Exception e)
+        {
+            log.error("{}", e.toString);
```
Geod24: `toString` allocates and walk the stack, hence it's a potential DoS vector. Usually we use `message` for this.

src/ocean/net/http/HttpServer.d
```diff
@@ -0,0 +1,232 @@
...
+            {
+                auto status = handler(this.request, response);
+
+                if ( status != HttpResponseCode.NotFound )
+                {
+                    return status;
+                }
+            }
+        }
+        catch ( Exception e )
```
Geod24, line 171: Your spacing

src/ocean/net/http/HttpServer.d
```diff
@@ -0,0 +1,232 @@
...
+
+    override protected size_t request_msg_body_length ( )
+    {
+        size_t size;
+
+        try
+        {
+            this.request.getUnsigned("Content-Length", size);
+        }
+        catch(Exception e)
```
Geod24: confuses me

src/ocean/net/http/HttpServer.d
```diff
@@ -0,0 +1,232 @@
...
+            [ HttpMethod.Get, HttpMethod.Put, HttpMethod.Post,
+              HttpMethod.Delete, HttpMethod.Options ]);
+        this.handlers = handlers;
+
+        this.log = Log.lookup("ocean.net.http.HttpServer.HttpRequestHandler");
+    }
+
+    /**************************************************************************
+
+        Method that handles the HTTP requests, it will call each
```
Geod24: I got a feeling something's missing here

src/ocean/net/http/HttpServer.d
```diff
@@ -0,0 +1,232 @@
...
+unittest
+{
+    HttpResponseCode iDoNothing( HttpRequest, out const(char)[] )
+    {
+        return HttpResponseCode.NotFound;
+    }
+
+    HttpResponseCode iDoSomething( HttpRequest, out const(char)[] response )
+    {
+        response = "I did something".dup;
```
Geod24: You don't need a `dup` here

Geod24:
So this iterates over all the handlers, in sequence ?
Anyone with a non-trivial API will want to have routes, and those are not supported as far as I can tell, and need to be handled manually, which the example unittest does not demonstrate how to do.

src/ocean/net/http/HttpServer.d
bogdan-szabo-sociomantic: weird... it's used like this everywhere in the released version. It compiles without it, so I removed `scope`

https://github.com/sociomantic-tsunami/ocean/blob/v5.1.0/integrationtest/selectlistener/UnixServer.d#L102

https://github.com/sociomantic-tsunami/ocean/blob/v5.1.0/src/ocean/net/http/HttpConnectionHandler.d#L125

https://github.com/sociomantic-tsunami/ocean/blob/v5.1.0/src/ocean/net/server/connection/IConnectionHandler.d#L161

https://github.com/sociomantic-tsunami/ocean/blob/v5.1.0/src/ocean/util/prometheus/server/PrometheusHandler.d#L66

and so on....

src/ocean/net/http/HttpServer.d
bogdan-szabo-sociomantic: I tried to do something similar to this:

https://github.com/sociomantic-tsunami/ocean/blob/v5.1.0/src/ocean/util/prometheus/server/PrometheusListener.d#L138

which looked good enough to be merged. Thanks for the tip!

bogdan-szabo-sociomantic:
@Geod24  thanks for the review! I did not try to write an url router here. I wrote  a web server that can handle multiple handlers. Right now the current api allows only one handler, and we can't easily use the prometheus stats with some other http handler on the same port.

bogdan-szabo-sociomantic:
@Geod24 ping

Geod24:
@bogdan-szabo-sociomantic : I'll leave the rest of the review to maintainers. From my PoV, an HTTP server class without routing ability is of little help over what is already existing. And it does not seem to be Ocean's goal (at least wasn't in my time) to provide those kinds of ability.

So I'd rather have someone familiar with the internal needs review this :)

ben-palmer-sociomantic:
If you could provide a good use case here it would help.

ben-palmer-sociomantic:
Is this still active? I will close if we are no longer working on this.

bogdan-szabo-sociomantic:
we can close all of my branches :) we already have the code in production in our private repos

PR #724, opened 2019-08-09
Replace getIeeeFlags with version in phobos

johannes-riecken-dunnhumby:
ocean almost compiles with gdc, if it weren't for this function and
    a few other small things. Sadly I have no idea, what this function does,
    but a function with the same name exists in phobos, which in contrast to
    the ocean one compiles with both dmd and gdc, so it seems cool to me.

### DIFF
--- a/src/ocean/math/IEEE.d
+++ b/src/ocean/math/IEEE.d
@@ -63,6 +63,16 @@ version (Naked_D_InlineAsm_X86) {
     // Needed for cos(), sin(), tan() on GNU.
     static import core.stdc.math;
 }
+
+version (D_InlineAsm_X86)
+{
+    version = InlineAsm_X86_Any;
+}
+else version (D_InlineAsm_X86_64)
+{
+    version = InlineAsm_X86_Any;
+}
+
 static import tsm = core.stdc.math;
 
 // Standard Tango NaN payloads.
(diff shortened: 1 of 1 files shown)


johannes-riecken-dunnhumby:
Withdrawn as it's too much effort to get it working under d1

johannes-riecken-dunnhumby:
Withdrawn as it's too much effort to get it working under d1

Geod24:
> Withdrawn as it's too much effort to get it working under d1

You can just target v5.x.x or later which is D2 only

joseph-wakeling-frequenz:
The compilation issue should have been fixed by https://github.com/sociomantic-tsunami/ocean/pull/714.  Perhaps you could backport that to v4.x.x if it is convenient?

johannes-riecken-dunnhumby:
> The compilation issue should have been fixed by #714. Perhaps you could backport that to v4.x.x if it is convenient?

Cool, very nice! I sadly don't have time for that this week, but I'll try to look at it next month if nobody else does it.

joseph-wakeling-frequenz:
BTW, I'm with @Geod24 on this, it's probably better to just push forward into D2-only land rather than try to backport to D1.  But to do that it would be nice to get a new 5.0.x point release with the https://github.com/sociomantic-tsunami/ocean/pull/714 fix in it (or at least to merge it into 5.x.x so it will be in the next minor release).

PR #732, opened 2019-09-03
[v5.x.x][Blocked] Deprecate most common D2 transition utilities

Geod24:
As usual, better reviewed commit-by-commit.

EDIT: As it triggers a compiler bug, this is blocked. I will split the PR further.


Geod24:
Well this is a bummer: https://issues.dlang.org/show_bug.cgi?id=20190

joseph-wakeling-frequenz:
Bit of a belated comment, but why not just replace all usage in ocean code and then deprecate the whole _module_ ... ?

Also, should this PR be closed in light of https://github.com/sociomantic-tsunami/ocean/pull/740 ... ?

Geod24:
> Bit of a belated comment, but why not just replace all usage in ocean code and then deprecate the whole module ... ?

Because there are some utilities that are not to be deprecated. Also user experience would be terrible.

> Also, should this PR be closed in light of #740 ... ?

No #740 only addressed symbols which could be deprecated, and removed usage of some symbols that weren't so frequent (`Inout`, `Immut`, basically anything but `Const`, although I did miss some instances). I'll get to fixing the compiler bug at some point in the future and then we can properly deprecate things. It's not really urgent and doesn't block anything.

joseph-wakeling-frequenz:
> Because there are some utilities that are not to be deprecated.

If you're referring to `SliceIfD1StaticArray`, do I understand right that the problem is not that it can't be deprecated, but that the deprecation message doesn't show up?  (Hence the motivation for deprecating the whole module instead.)

Other than that, what is there?  I don't follow why (after a full D2 transition) one would not be able to (and want to) migrate away from all the transitional utilities.

> Also user experience would be terrible.

Do you mean lack of per-symbol deprecation messages and recommendations, or something else?  If it's about what to do for each individual symbol, that could be addressed in `ocean.transition` module docs (for example).

> It's not really urgent and doesn't block anything.

Agreed, no pressure intended.

Geod24:
> If you're referring to SliceIfD1StaticArray, do I understand right that the problem is not that it can't be deprecated, but that the deprecation message doesn't show up? (Hence the motivation for deprecating the whole module instead.)

`SliceIfD1StaticArray` or just about any template

> Other than that, what is there? I don't follow why (after a full D2 transition) one would not be able to (and want to) migrate away from all the transitional utilities.

`ocean.core.meta.types.Qualifiers` has `Unqual`. Also I don't see why the aliases `cstring` & `mstring` would be deprecated. `ocean.transition` has a bunch of other stuff which are not so transitional, e.g. `gc_usage`, `isMutable`, `Octal`... Maybe when Phobos is acceptable.

> Do you mean lack of per-symbol deprecation messages and recommendations, or something else? If it's about what to do for each individual symbol, that could be addressed in ocean.transition module docs (for example).

Yes, lack of per-symbol solution. Having the fix in the doc doesn't reach my threshold for usability.

On a side note, this is not the first time I've came across bugs with `deprecated`. Those are usually fairly easy to solve and provide a lot of value, while working around them is usually shifting the burden to someone else.

joseph-wakeling-frequenz:
> `SliceIfD1StaticArray` or just about any template

Well, that in itself isn't a blocker.  If it was only those cases it would be an argument _for_ just deprecating the whole module (just as a workaround for the fact that the individual-symbol deprecations don't get displayed by the compiler).  But as for the other cases ...

> Also I don't see why the aliases `cstring` & `mstring` would be deprecated.

Those are also available in `ocean.meta.types.Qualifiers`.  There's no reason to keep using the `ocean.transition` import of them.

> `ocean.transition` has a bunch of other stuff which are not so transitional, e.g. `gc_usage`, `isMutable`, `Octal`... Maybe when Phobos is acceptable.

If stuff needs to outlive the D2 transition itself, it should be moved outside `ocean.transition` (e.g. `isMutable` should probably go to `ocean.meta`).

`gc_usage` can probably be dropped once everyone is using a D2 compiler that supports GC stats.

Point is we shouldn't really be having any blockers at this stage to being able to drop the `ocean.transition` module for projects that are using the D2-only libs.

jens-mueller-sociomantic:
@Geod24 Would you like to finish this PR with the workaround suggested in https://github.com/sociomantic-tsunami/ocean/pull/774? This should work for `SliceIfD1StaticArray`. It seems for `TypeofThis` we don't even need it because it is mixed in.
I would also suggest to deprecate `enableStomping`, `genOpCmp` and `genOpEquals` and `gc_usage` and move `isMutable` and `Octal` which allows us to deprecate `ocean.transition`.

Geod24:
@jens-mueller-sociomantic came up with a much more comprehensive diff.

PR #734, opened 2019-09-04
[v5.x.x] Remove static if conditions that can never trigger

Geod24:
```
Since Ocean v5.x.x does not support DMD < 2.072.x,
those can be safely removed.
As a consequence, an empty module has been deprecated.
```


Geod24:
Finally green, but looks like Codecov gave up before I did.

matthias-wende-sociomantic (approved):

PR #735, opened 2019-09-04
[v5.x.x] Test, format and simplify ocean.util.log.LayoutDate

Geod24:
Not sure if this would be better to target v4.x.x, but all my development is done in D2 so that's why I base everything off this branch.


matthias-wende-sociomantic:
> Not sure if this would be better to target v4.x.x, but all my development is done in D2 so that's why I base everything off this branch.

I'd say that's fine as the major objective is to get rid of D1 :).

matthias-wende-sociomantic (approved):

PR #741, opened 2019-09-11
Logger: Remove static variable beginTime

Geod24:
Users care more about time since program start than time since Hierarchy instantiation.
And since Logger is almost exclusively used combined with a static module constructor,
those two are essentially the same.

### DIFF
--- a/relnotes/logger.feature.md
+++ b/relnotes/logger.feature.md
@@ -1,0 +1,9 @@
+### `Logger.runtime()` now uses `Clock.startTime()`
+
+* `ocean.util.log.Logger`
+
+The `Logger` now uses `Clock.startTime()` instead of the hierarchy instantiation time.
+This means that all hierarchy will have the same `runtime`.
+In practice, programs only ever had one hierarchy,
+and since the common usage is to instantiate `Logger`s from the module constructor,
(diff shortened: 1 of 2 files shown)


matthias-wende-sociomantic:
> Users care more about time since program start than time since Hierarchy instantiation.
And since Logger is almost exclusively used combined with a static module constructor,
those two are essentially the same.

I agree that most likely Hierarchy instantiation and program start time fall together. But since you're planing to remove the distinction in this module I'd like to know the motivation behind it.
And what is the advantage?

Geod24:
The advantage is just simplifying code. Since the module already depends on `Clock`, there's no reason not to use the static variable.

ben-palmer-sociomantic (approved):
LGTM, I guess we can't add a test to make code coverage happy can we?

ben-palmer-sociomantic:
Oh and we should have a release note seeing as the behaviour is changing a bit.

Geod24:
> LGTM, I guess we can't add a test to make code coverage happy can we?

Everything is possible if you ask nicely :)

ben-palmer-sociomantic:
> Everything is possible if you ask nicely :)

Thankyou!

PR #745, opened 2019-09-17
Make BufferedDirectReadFile.DirectReadFile protected, not private

joseph-wakeling-frequenz:
At least one downstream overrides this nested static class, and this appears to be intended usage.  As of DMD 2.087.0 `private` protection will be properly enforced, and builds of such downstreams will fail.

This patch therefore switches `private` to `protected`, matching what is already in place for `BufferedDirectWriteFile.DirectWriteFile`.

### DIFF
--- a/src/ocean/io/device/DirectIO.d
+++ b/src/ocean/io/device/DirectIO.d
@@ -574,7 +574,7 @@ public class BufferedDirectReadFile: InputStream
 
     ***************************************************************************/
 
-    static private class DirectReadFile : File
+    static protected class DirectReadFile : File
     {
         override public void open (cstring path, Style style = this.ReadExisting)
         {


joseph-wakeling-frequenz:
For context, this is needed to allow this class definition to work:
https://github.com/sociomantic-tsunami/dhtnode/blob/c3ff4abe7d003b994792dbdf91ff8d23b94ce365/src/dhtnode/storage/DirectIO.d#L289

I'm not aware of other use-cases but this seems an important one, and it does seem intended that the nested class be `protected` and not `private`.

joseph-wakeling-frequenz:
Those build failures look like they aren't to do with the code.  Can someone re-run, please? :-)

ben-palmer-sociomantic:
Restarted the test but code change LGTM.

joseph-wakeling-frequenz:
Thanks!

Can I get a restart on just the one remaining "failing" job, please?  Looks like some networking or other internal Travis issues were responsible. :-P

Geod24:
@shiv-tiwari-dunnhumby : Looks like this is missing form the release.

joseph-wakeling-frequenz:
Yea, we need a new v5.1.x release, otherwise this won't get merged into v5.x.x.

ben-palmer-sociomantic:
Ah I see, I didn't see that this was targeting v5.1.x

PR #746, opened 2019-09-17
Add usage warning to LRUCache

andrew-stevenson-sociomantic:
`get()` doesn't count as usage from the point of view of LRU - you need to use
the special methods with `refresh` in their names.


ben-palmer-sociomantic:
LGTM

Issue #747, opened 2019-09-17
Transition from D1 operator overload to D2 operator overloads


Geod24:
This needs to be done in v6.0.0, as D1 operators are virtual, while D2 are template instantiation (final), but can be made virtual if needed.
Some part of Ocean make use of it, e.g. [this](https://github.com/sociomantic-tsunami/ocean/blob/f253c5374d85948937fc8861804071b23752dacb/src/ocean/core/Enum.d#L153-L218).

joseph-wakeling-frequenz:
Note that there are a few finnicky isues that come out of how various ocean functionality implements the older D1 operator overloads.

A quite common pattern is to have the implementation inside a local method, then use an alias of that to define the D1 operator overload.  The local implementations are in at least some cases overridden in subclasses (the only explicit example I found was in unittests, but we can't exclude that there are some other downstream examples).

See https://github.com/sociomantic-tsunami/ocean/pull/748 for an example of what I thought was the best way to deal with this -- I'm interested in yours (and others') feedback on this.

There are also some cases, like https://github.com/sociomantic-tsunami/ocean/pull/749, where the D1 operator definitions just seem _weird_, and make it look like they were added without really thinking through the use-cases.

Geod24:
As far as I can tell, this has been completed, thanks to @don-clugston-sociomantic & others.

PR #748, opened 2019-09-17
Replace D1 operator overloads in SuspendableThrottlerCount

joseph-wakeling-frequenz:

`SuspendableThrottlerCount` has a slightly finnicky implementation of these, so it seems worth updating it in a separate patch.  Its original design defined local methods (`inc`/`dec`, `add`/`remove`) that provide the actual implementation, and used aliases to these to implement the actual operator overloads.  This is a problem because these methods may be (and in at least some test cases are) overridden by subclasses.  We therefore cannot just replace them with standard D2 operator overloads, as templated methods cannot be overridden by subclasses.

The simplest solution therefore seems to be to continue to support these custom local methods, and replace each D1 operator overload alias with a D2 operator overload implemented as a thin wrapper of the local method.

This should allow overloads of the implementations to continue to work, while having no impact on subclasses or users of the operators.

Part of https://github.com/sociomantic-tsunami/ocean/issues/747.

### DIFF
--- a/relnotes/d1-ops.migration.md
+++ b/relnotes/d1-ops.migration.md
@@ -1,0 +1,13 @@
+### Replace D1 operator overloads with D2 operator overloads
+
+`ocean.io.model.SuspendableThrottlerCount`
+
+DMD 2.088.0 has deprecated D1 operator overloads, which are expected to
+be replaced with the newer D2 `op{Unary,Binary,BinaryRight,OpAssign}`
+templated methods: <https://dlang.org/changelog/2.088.0.html#dep_d1_ops>
+
+Since templated methods cannot be overridden this may in some rare cases
+cause breaking change for classes that implemented the old operators.
+In general however this will likely make no practical difference, other
+than the absence of scores of deprecation messages with more recent DMD
+versions.
--- a/src/ocean/io/model/SuspendableThrottlerCount.d
+++ b/src/ocean/io/model/SuspendableThrottlerCount.d
@@ -78,8 +78,6 @@ public class SuspendableThrottlerCount : ISuspendableThrottlerCount
 
         Increases the count of pending items and throttles the suspendables.
 
-        Aliased as opPostInc.
-
     ***************************************************************************/
 
     public void inc ( )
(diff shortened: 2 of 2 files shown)


joseph-wakeling-frequenz:
Updated with a cross-ref to the issue, which I hadn't seen before submitting.

ben-palmer-sociomantic:
Should we take the same approach here as we did in https://github.com/sociomantic-tsunami/ocean/pull/749 ?

joseph-wakeling-frequenz:
I don't think so, because these seem like reasonable use-cases for the operators -- we're literally incrementing or decrementing a count -- and unlike that PR, these are "in-place" updates.

Unless anyone has a different opinion?

ben-palmer-sociomantic:
LGTM then, maybe we should have some release notes though. @don-clugston-sociomantic could you maybe have a look at this as well?

joseph-wakeling-frequenz:
> LGTM then, maybe we should have some release notes though.

I was planning on doing one big release-notes section for replacement of operator overloads, after everything was done.  Is that OK or would you prefer me to go module by module?

ben-palmer-sociomantic:
> Is that OK or would you prefer me to go module by module?

Hmm, up to you. Might be easier to manage having every PR independent.

joseph-wakeling-frequenz:
> Hmm, up to you. Might be easier to manage having every PR independent.

OK.  I'll write up release notes for this patch covering the general principle of replacing D1 operator overloads, and update them module by module.

ben-palmer-sociomantic:
Just waiting on the release notes for this before merging.

joseph-wakeling-frequenz:
Oh, sorry!  I'll try to get that done later today.

joseph-wakeling-frequenz:
Release notes added :-)  I will continue to extend these in future PRs.

ben-palmer-sociomantic:
Cool, looks good. Just one thing that occurred to me. I think it would be worth adding a short unit test for these changes.

joseph-wakeling-frequenz:
It's good that you suggested that, because it turns out there is a bug in the original `opSubAssign`, which aliases to `add` instead of the correct `remove` (!).

I guess that means that probably no one is actually using that operator :-P

joseph-wakeling-frequenz:
I've pushed a bugfix patch with the operator overload migration rebased on top.  How would you like to handle this?  Would you prefer that I submit the bugfix patch against v5.1.x first ... ?

joseph-wakeling-frequenz:
Updated with a rewritten commit message and release notes for the bugfix patch: I realized it's probably best to describe the fix in terms of the operator rather than `opSubAssign`. since we're getting rid of the latter.

ben-palmer-sociomantic:
> Would you prefer that I submit the bugfix patch against v5.1.x first ... ?

Yeah I think that would be the best option. Then we can also merge the original changes in to v5.x.x

joseph-wakeling-frequenz:
Here you go: https://github.com/sociomantic-tsunami/ocean/pull/771 :-)

joseph-wakeling-frequenz:
Rebased on latest v5.x.x branch.  This should be good to go, I think ... ?

ben-palmer-sociomantic (approved):
LGTM

PR #752, opened 2019-09-24
Remove assert (verify) which wrongly checks user-input

stefan-koch-sociomantic:


stefan-koch-sociomantic:
please merge this @sociomantic-tsunami/ocean-team (please only those who know how to use neptune)

don-clugston-sociomantic (changes requested):
I believe the code change is correct. However, this needs:
* A commit message which explains why this is being removed.
* An entry for the release log.
* The PR should contain a link to the issue where this code was added.

Why this code should be removed:
* What's being checked here is the command-line arguments. Thus, a `verify` is completely inappropriate. It is normal for there to be typos on the command line, it is not exceptional in any way.
* The only reason why the check was added, was that somebody noticed that the command-line option for setting a config file value is named "override". It is a really annoying loss of functionality to disallow setting arbitrary values.

ben-palmer-sociomantic:
Agree with the above from Don. Also you might want to target v5.0.x depending in when this change was introduced.

ben-palmer-sociomantic:
Comments still need to be addressed.

ben-palmer-sociomantic:
Closing without merging as it doesn't seem to be active.

PR #754, opened 2019-09-25
Add enableStomping() calls

ghost:
Issue found by @jens-mueller-sociomantic

### DIFF
--- a/relnotes/enableStomping.bug.md
+++ b/relnotes/enableStomping.bug.md
@@ -1,0 +1,8 @@
+### Add missing enableStomping calls.
+
+`ocean.util.container.map.model.BucketInfo`,
+`ocean.util.container.map.model.BucketSet`
+
+The enableStomping calls preserve D1 semantics of
+overwriting array memory.
+
--- a/src/ocean/util/container/map/model/BucketInfo.d
+++ b/src/ocean/util/container/map/model/BucketInfo.d
@@ -512,7 +512,9 @@ class BucketInfo
 
     package void clearResize ( size_t n )
     {
+        enableStomping(this.buckets);
         this.buckets.length             = n;
+        enableStomping(this.bucket_list_indices);
         this.bucket_list_indices.length = n;
 
         /*
--- a/src/ocean/util/container/map/model/BucketSet.d
+++ b/src/ocean/util/container/map/model/BucketSet.d
@@ -36,6 +36,7 @@
 
 module ocean.util.container.map.model.BucketSet;
 
+import ocean.transition;
 
 import ocean.util.container.map.model.Bucket,
        ocean.util.container.map.model.BucketInfo,
@@ -613,6 +614,7 @@ public abstract class BucketSet ( size_t V, K = hash_t ) : IBucketSet
                 // Resize the array of buckets and the bucket_info and calculate
                 // the new bucket_mask.
 
+                enableStomping(this.buckets);
                 this.buckets.length = n_new;
 
                 .clear(this.buckets[0 .. (n_prev < $)? n_prev : $]);


Geod24:
Calling `enableStomping` is not needed before the change to `length`, but it is needed after.

ghost:
> Calling `enableStomping` is not needed before the change to `length`, but it is needed after.

This is the case when the size is being decreased. In case of extending, our guidance mentions that it can be called before. Since we cannot be sure in that case we added it before. Maybe a more correct approach would be to call it twice (before & after)? :-)

shiv-tiwari-dunnhumby (approved):
LGTM

shiv-tiwari-dunnhumby:
> Calling `enableStomping` is not needed before the change to `length`, but it is needed after.

Within enableStomping(), it calls assumeSafeAppend(). The example of the correct use of assumeSafeAppend() is calling before the change is made in the array.

https://dlang.org/phobos/core_internal_array_capacity.html#.assumeSafeAppend

Geod24:
>  The example of the correct use of assumeSafeAppend() is calling before the change is made in the array.

"The change is made" is not quite accurate. It is called after the resize (in the case of this example, the slicing), but before the append, which is what I was recommending...

> This is the case when the size is being decreased. In case of extending, our guidance mentions that it can be called before. Since we cannot be sure in that case we added it before. Maybe a more correct approach would be to call it twice (before & after)? :-)

Hum, yes indeed they mentioned that. The only reason that we would do it *before* setting the length is to be forgiving to callers that didn't call `enableStomping()` after reducing the length, and that's the only situation where it can be useful.
Note that, before I left, it was measured that `enableStomping()` has a noticeable performance impact in D2, because of all the GC metadata lookup.

andrew-stevenson-sociomantic:
A summary for the less educated among us.

We have a slice (which I will call `slice`).

A slice is backed by a block of memory (which I will call `memory`).

A slice tracks its length (i.e. `slice.length`). A block of memory tracks the highest slot that ever has been written to (`memory.used` I think - that's what I will call it here anyway).

If you have multiple slices backed by the same memory, or if you shorten a slice, then `memory.used` may not be the same as the last element in your slice (assuming here, for simplicity's sake, the slice begins at the beginning of the memory block then that's `slice.length`).

When you extend a slice (e.g. set `slice.length` to something bigger, or `~` an extra element on) then the run-time checks if `memory.used == slice.length`. If it is then the next slots in the memory block must be free and can be used safely for the new data (running out of room in the block is a separate matter which we won't address here). If it is not, that is `memory.used > slice.length`, then it is _possible_ someone else is using the data in the slots after our slice. They might not be but we don't know and the D run-time doesn't want to take the risk. Therefore a new memory block (`memory2`) is allocated and the data the slice points to is copied. Now we know that `memory2.used` == `slice.length` so appending is safe.

In our normal use case - buffers (slices) that we lengthen and then set to length 0 before appending more stuff the next time around (e.g. upon the next request) - we know that no one is using the memory slots beyond 0 once the slice's length is set to 0. Normally there are no other slices referencing this memory, or if there are we know they are invalid upon a new request. Therefore we don't need or want the new memory block to be allocated.

Calling `allowStomping` sets `memory.used` to be equal to `slice.length`. You are telling the run-time that you know no one else is using those later slots in `memory`.

It comes from this that there are two strategies that can be employed. Call `allowStomping` before every extension of a slice, or call it after every shortening. Either works (though the former is more efficient if you ever shorten multiple times between lengthening). The important thing is to be consistent for a single buffer. For reasons relating to code complexity we usually use the latter method. I conject that we almost never shorten multiple times between lengthening so there is no real difference in efficiency for us.

As `allowStomping` may be expensive (in the case of a miss for the GC metadata IIUC) you don't want to call it more often than you have to.

On this specific PR - I don't know the code but from a quick glance it seems that both of these slices are internal to the classes. That means we are free to use either convention without forcing a change on the users - though there might be an argument to stick with the one convention throughout all the code. This change seems to use the "before extension" convention though. That's fine as long as it is before every extension (and for efficiency, not after a shortening). I think there might be places where that's not the case. e.g. https://github.com/konstantinos-stamatiadis-sociomantic/ocean/blob/2a72f934c2f199c1871d81d76d3a49d1561831ed/src/ocean/util/container/map/model/BucketInfo.d#L350 . That just means the change is incomplete but not actually incorrect. I stress this is from a quick visual inspection so I might be wrong.

PR #756, opened 2019-09-25
Add rankOfArray to meta.traits.Arrays

don-clugston-sociomantic:
This code is directly copied from core.Traits, which will be deprecated
in ocean 5.x.x

When the functions were copied from core.Traits into meta,*, this function was accidentally omitted.


don-clugston-sociomantic:
Releasing this, is too complicated to be worthwhile. I'll just a workaround instead.

PR #761, opened 2019-09-30
Replace useless `(&this)` with `this`

Geod24:
```
Those `(&this)` were automatically added by d1to2conv.
However, most of them are unnecessary, and can be trivially replaced by `this`.
This is due to the fact that, within a struct, this was a pointer to the type,
hence typeof(this) would be `Struct*`, while in D2, it is a reference to the type,
hence typeof(this) is `Struct`."

The following rules were following for this change:
1) If the `(&this)` did not matter, it was simply removed.
   This is the most frequent case, and applies to *any* member access.
   E.g. `(&this).foo` can always be converted to `this.foo` in Ocean
   (this is not necessarily true in other code bases, because UFCS).
2) The second most common use case was returning a pointer to the struct,
   e.g. `return (&this);`; this was converted to `return &this;`.
   It was mostly present in containers.
3) If within a typeof, and the pointer type is wanted, e.g. `typeof((&this))`,
   the second level of parenthesis was removed;
4) Likewise, some places used `auto x = (&this)` and `funcCall((&this))`,
   and the second level of parenthsesis was removed.
```

I can split it up by module / package if it's any easier to review.

### DIFF
--- a/src/ocean/application/components/PidLock.d
+++ b/src/ocean/application/components/PidLock.d
@@ -70,7 +70,7 @@ public struct PidLock
 
     public void parseConfig ( ConfigParser config )
     {
-        (&this).pidlock_path = config.get("PidLock", "path", "").dup;
+        this.pidlock_path = config.get("PidLock", "path", "").dup;
     }
 
     /***************************************************************************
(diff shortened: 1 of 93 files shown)


matthias-wende-sociomantic (approved):
I'd have thought this already happened. At least the trivial member access case.

PR #764, opened 2019-10-01
Replace output in unit test with check on function return value

ben-palmer-sociomantic:
Using output in unit tests can cause problems with the closure
allocation check.


Geod24:
Can you also remove the import ?

ben-palmer-sociomantic:
> Can you also remove the import ?

Ah yes. Will do.

Geod24 (approved):

stefan-koch-sociomantic (approved):
LGTM

PR #765, opened 2019-10-02
Deprecate trivial function ocean.text.Util : layout

Geod24:
And a bunch of other fixes that arose from this.
As I was fixing the Logger's `Layout` classes, I realized this function was useless, and mostly a relic of the time when the Tango `Layout` class was extremely expensive.


Geod24:
Rebased on v5.x.x

matthias-wende-sociomantic (approved):

PR #766, opened 2019-10-02
Update gitignore to be more DUB friendly

Geod24:


daniel-zullo (approved):
I am not sure if we need a release note for this case but it might be worth adding it.
Other than that LGTM

Geod24:
I would not think so ? It's not code

ben-palmer-sociomantic (approved):
LGTM

PR #767, opened 2019-10-02
Use non-deprecated import in LayoutSimple test

Geod24:
This new import was introduced at the same time
the import was deprecated, leading to v5.x.x being broken
on the commit that precede this one.


daniel-zullo (approved):
LGTM, waiting for the CI green light

PR #769, opened 2019-10-02
Remove build file .D2-ready

daniel-zullo:
It no longer has any effect. The file was used by jenkins
which is no longer used for CI.


daniel-zullo:
It seems beaver or travis still use `.D2-ready` file. I'll check/assess what other changes are needed and re-open this PR if I found out a solution.

PR #771, opened 2019-10-15
Fix -= operator implementation in SuspendableThrottlerCount

joseph-wakeling-frequenz:
The `-=` operator overload (D1 `opSubAssign`) was inexplicably aliased to the `add` method rather than the correct `remove`.  This probably means that no one was ever actually using the operator overload, but it seems a good idea to fix it in any case.


ben-palmer-sociomantic:
LGTM, I have no idea how you managed to reduce code coverage though.

joseph-wakeling-frequenz:
> LGTM, I have no idea how you managed to reduce code coverage though.

I think CodeCov's stats are suspect ;-)

PR #772, opened 2019-10-15
Replace D1 unary operator overloads in ocean.math

joseph-wakeling-frequenz:

`ocean.math` contains surprisingly few such operators, and they are easy to replace.  For `WideUInt` there should be no resulting problems, as it is a struct, but `Distribution` and `SlidingAverageTime` are classes, so there is a risk this will be a breaking change for downstream subclasses if any attempt to override the older D1 methods.

### DIFF
--- a/src/ocean/math/Distribution.d
+++ b/src/ocean/math/Distribution.d
@@ -121,7 +121,7 @@ public class Distribution ( T )
 {
     /***************************************************************************
 
-        List of values, appended to by the opAddAssign method.
+        List of values, appended to by the opOpAssign!"~" method.
 
     ***************************************************************************/
 
@@ -132,7 +132,7 @@ public class Distribution ( T )
 
         Indicates whether the list of values has been sorted (which is required
         by the methods: lessThanCount, greaterThanCount, percentValue).
-        opAddAssign() and clear() reset the sorted flag to false.
+        opOpAssign() and clear() reset the sorted flag to false.
 
         TODO: it might be better to always maintain the list in sorted order?
 
@@ -161,11 +161,12 @@ public class Distribution ( T )
         Adds a value to the list.
 
         Params:
+            op = operation to perform
             value = value to add
 
     ***************************************************************************/
 
-    public void opCatAssign ( T value )
+    public void opOpAssign ( string op ) ( T value ) if (op == "~")
     {
         this.values.append(value);
         this.sorted = false;
--- a/src/ocean/math/SlidingAverage.d
+++ b/src/ocean/math/SlidingAverage.d
@@ -491,4 +491,13 @@ public class SlidingAverageTime ( T ) : SlidingAverage!(T)
 unittest
 {
     auto avg_stats = new SlidingAverageTime!(size_t)(100, 50, 1000);
+
+    test!("==")(avg_stats.current, 0, "Failed SlidingAverageTime initial value");
+
+    avg_stats++;
+    test!("==")(avg_stats.current, 1, "Failed SlidingAverageTime increment value");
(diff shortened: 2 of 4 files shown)


src/ocean/math/SlidingAverage.d
```diff
@@ -478,7 +478,7 @@ public class SlidingAverageTime ( T ) : SlidingAverage!(T)
 
     ***************************************************************************/
 
-    public T opAddAssign ( T val )
+    public T opOpAssign ( string op ) ( T val ) if (op == "+")
```
ben-palmer-sociomantic, line 485: It might be worth to have more in depth unit tests for this change. At the moment it seems to just instantiate it.

ben-palmer-sociomantic (commented):
Changes look fine. As a general comment I would say it would be good to make sure they are all covered by unit tests.

joseph-wakeling-frequenz:
@ben-palmer-sociomantic I'm sorry but I won't have time to focus on writing new unittests any time soon. :-\

src/ocean/math/Distribution.d
```diff
@@ -165,7 +165,7 @@ public class Distribution ( T )
 
     ***************************************************************************/
 
-    public void opCatAssign ( T value )
+    public void opOpAssign ( string op ) ( T value ) if (op == "~")
```
ben-palmer-sociomantic, line 169: This actually covered by the test cases using https://github.com/sociomantic-tsunami/ocean/blob/v5.x.x/src/ocean/math/Distribution.d#L453

src/ocean/math/WideUInt.d
```diff
@@ -336,7 +336,7 @@ public struct WideUInt ( size_t N )
 
     ***************************************************************************/
 
-    public void opAddAssign ( uint rhs )
+    public void opOpAssign ( string op ) ( uint rhs ) if (op == "+")
```
ben-palmer-sociomantic, line 340: This is covered here: https://github.com/sociomantic-tsunami/ocean/blob/v5.x.x/src/ocean/math/WideUInt.d#L349

ben-palmer-sociomantic (commented):
Adding one unit test for appending to SlidingAverageTime

ben-palmer-sociomantic:
@daniel-zullo-sociomantic maybe you want to have another look at this.

src/ocean/math/Distribution.d
daniel-zullo: This change will need to update documentation parameters

src/ocean/math/SlidingAverage.d
```diff
@@ -460,7 +460,7 @@ public class SlidingAverageTime ( T ) : SlidingAverage!(T)
 
     ***************************************************************************/
 
-    public T opPostInc ( )
+    public T opUnary ( string op ) ( ) if (op == "++")
```
daniel-zullo, line 466: Needs documentation parameter

src/ocean/math/SlidingAverage.d
daniel-zullo: Needs to update documentation params too

src/ocean/math/WideUInt.d
daniel-zullo: ditto

daniel-zullo (commented):
Code wise LGTM, only documentation for template parameters is missing or need to be updated.
And it seems is well covered by unit tests.

ben-palmer-sociomantic:
Added missing documentation.

daniel-zullo (approved):
LGTM

PR #773, opened 2019-10-15
Replace D1 operator overloads in ocean.time

joseph-wakeling-frequenz:

`ocean.time.Time` contains quite a few of these operators, but replacing them is fairly straightforward.  Since all the affected data structures are structs, there should be no concern about downstream overrides, so all should be drop-in replacements.

No other `ocean.time` module implements any of these operators, so this patch in effect migrates the entire package (the remaining deprecation message reported for `ocean.time.timeout.TimeoutManager` is for the use of a D1 operator overload in `ocean.util.container.AppendBuffer`, which should clear up when that module is migrated).

### DIFF
--- a/src/ocean/time/Time.d
+++ b/src/ocean/time/Time.d
@@ -149,52 +149,32 @@ struct TimeSpan
         `));
 
         /**
-         * Add the TimeSpan given to this TimeSpan returning a new TimeSpan.
+         * Add or subtract the TimeSpan given to this TimeSpan returning a
+         * new TimeSpan.
          *
-         * Params: t = A TimeSpan value to add
-         * Returns: A TimeSpan value that is the sum of this instance and t.
+         * Params:  op = operation to perform
+         *          t = A TimeSpan value to add or subtract
+         * Returns: A TimeSpan value that is the result of this instance, the
+         *          operation, and t.
          */
-        TimeSpan opAdd(TimeSpan t)
+        TimeSpan opBinary (string op) (TimeSpan t) if (op == "+" || op == "-")
         {
-                return TimeSpan(ticks_ + t.ticks_);
+            mixin("return TimeSpan(ticks_ " ~ op ~ " t.ticks_);");
         }
 
         /**
-         * Add the specified TimeSpan to this TimeSpan, assigning the result
-         * to this instance.
+         * Add or subtract the specified TimeSpan to this TimeSpan, assigning
+         * the result to this instance.
          *
-         * Params: t = A TimeSpan value to add
-         * Returns: a copy of this instance after adding t.
+         * Params:  op = operation to perform
+         *          t = A TimeSpan value to add or subtract
+         * Returns: a copy of this instance after adding or subtracting t.
          */
-        TimeSpan opAddAssign(TimeSpan t)
-        {
-                ticks_ += t.ticks_;
-                return this;
-        }
 
-        /**
-         * Subtract the specified TimeSpan from this TimeSpan.
-         *
-         * Params: t = A TimeSpan to subtract
-         * Returns: A new timespan which is the difference between this
-         * instance and t
-         */
-        TimeSpan opSub(TimeSpan t)
+        TimeSpan opOpAssign (string op) (TimeSpan t) if (op == "+" || op == "-")
         {
-                return TimeSpan(ticks_ - t.ticks_);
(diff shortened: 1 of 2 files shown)


ben-palmer-sociomantic (commented):
Again might be worth checking we have unit tests.

Geod24 (approved):
LGTM

joseph-wakeling-frequenz:
> Again might be worth checking we have unit tests.

... do we not already?  I'm really sorry but I won't have time to write them any time soon.

src/ocean/time/Time.d
```diff
@@ -166,7 +166,7 @@ struct TimeSpan
          * Params: t = A TimeSpan value to add
          * Returns: a copy of this instance after adding t.
          */
-        TimeSpan opAddAssign(TimeSpan t)
+        TimeSpan opOpAssign (string op) (TimeSpan t) if (op == "+")
```
alaksiej-stankievicx-dunnhumby: The `+` and `-` operators look very similar, why no use common string mixin trick?
```d
    TimeSpan opBinary (string op) (TimeSpan t)
        if (op == "+" || op == "-")
    {
        mixin("return TimeSpan(ticks_ " ~ op ~ " t.ticks_);");
    }
```

Geod24:
Anything holding this back ?

ben-palmer-sociomantic:
> Anything holding this back ?

I would prefer to have unit tests added but we could also merge this and add unit tests in a new PR.

src/ocean/time/Time.d
don-clugston-sociomantic: Worth noting that the main reason why the D1 overloads were replaced with the D2 syntax, is so that this trick can be used. It was noticed that practically every class which overloaded operators had a huge amount of code duplication.

joseph-wakeling-frequenz: Fair enough.  If I get time I'll do a rewrite, but if anyone else wants to take this over in the meanwhile, feel free (things are busy:-).

daniel-zullo:
> Anything holding this back ?

Documentation needs to be updated for pretty much all the changes in the function signatures.

ben-palmer-sociomantic:
Addressed comments from @alaksiej-stankievicx-sociomantic and added unit tests.

ben-palmer-sociomantic:
@daniel-zullo-sociomantic if you have time maybe you could give this one a check as well.

src/ocean/time/Time.d
```diff
@@ -149,131 +149,166 @@ struct TimeSpan
...
+            res = time_span / 5;
+            test!("==")(res.ticks(), 1);
         }
 
         /**
          * Scales this TimeSpan and assigns the result to this instance.
          *
-         * Params: v = A multipler to use for scaling
+         * Params: op = operation to perform
+         *         v = A multipler to use for scaling
```
daniel-zullo: maybe A multipler *or divisor*?

don-clugston-sociomantic (approved):
LGTM.

src/ocean/time/Time.d
```diff
@@ -149,131 +149,166 @@ struct TimeSpan
...
          *
-         * Params: t = A TimeSpan to subtract
-         * Returns: A new timespan which is the difference between this
-         * instance and t
+         * Params:  op = operation to perform
+         *          t = A TimeSpan value to add or subtract
+         * Returns: a copy of this instance after adding or subtracting t.
          */
-        TimeSpan opSub(TimeSpan t)
+
```
daniel-zullo, line 190: cosmetic: you might want to remove this extra empty line

daniel-zullo (approved):
Only a couple of optional cosmetic comments.
LGTM

src/ocean/time/Time.d
ben-palmer-sociomantic: hmm, I couldn't seem to find this one.

src/ocean/time/Time.d
ben-palmer-sociomantic: will change

src/ocean/time/Time.d
daniel-zullo: I've just mentioned because there were not empty line before and now there is one...anyway let's keep it as it is

ben-palmer-sociomantic:
Rebased and addressed comments.

PR #726, opened 2019-08-15
Prepare for dlang/dmd#10124 (field of init symbol no lvalue anymore)

kinke:
According to DMD's buildkite CI service ([log](https://buildkite.com/dlang/dmd/builds/8910#23dad51f-55b0-46da-ae79-97d66a792ea1)). I haven't set up a dev environment for ocean, so I have no idea whether there are more occurrences like this. No idea which branch to target either.

PR #774, opened 2019-10-17
Deprecate templates Const, Immut and Inout

jens-mueller-sociomantic:
The templates `Const`, `Immut` and `Inout` have been used for
transitioning from D1 to D2. Now that ocean has successfully migrated to
D2 only there is no point in having these templates. This change
deprecates their usage which allows downstream user to update their code
base.

Note, that this also removes the respective unittests because
deprecating them does not work properly, i.e., it still warns about the
deprecations.

Unfortunately, just deprecating the alias won't work. There is a 6-year
old bug report at https://issues.dlang.org/show_bug.cgi?id=10181 which
also mentions a workaround that is used in this change.


src/ocean/meta/types/ReduceType.d
```diff
@@ -262,7 +262,7 @@ unittest
 unittest
 {
     assert(!ReduceType!(CheckPrimitiveReducer, CheckPrimitiveReducer));
-    assert(ReduceType!(Const!TestAggregate, CheckPrimitiveReducer));
+    assert(ReduceType!(const TestAggregate, CheckPrimitiveReducer));
```
jens-mueller-sociomantic, line 265: @tiyash-basu-sociomantic This case was a bit strange to me. Because I thought that we always have to write parentheses in D1. But apparently your original code doesn't need the parentheses. But I didn't check whether this was only ever compiled using D2. Do you remember by any chance?

tiyash-basu-sociomantic: This change was targeted onto ocean v5.x.x branch, so it was only D2.

Geod24:
What's the deprecation message though ? Looking at the code, it points within Ocean, right ?

src/ocean/meta/types/Qualifiers.d
```diff
@@ -61,12 +61,16 @@ alias char[]         mstring;
...
+    // https://issues.dlang.org/show_bug.cgi?id=10181
+    deprecated("Use const")
+    alias const(T) Temp;
+    alias Temp Const;
 }
 
 unittest
 {
-    alias Const!(int[]) Int;
+    alias const(int[]) Int;
```
Geod24: That unittest is kinda useless now. And I suppose you can't mark the unittest as `deprecated` ?

jens-mueller-sociomantic: I can also remove the unittest.

johannes-riecken-dunnhumby:
LGTM

jens-mueller-sociomantic:
> What's the deprecation message though ? Looking at the code, it points within Ocean, right ?

You mean the deprecation message the compiler will emit, it's `Use const`, etc.

johannes-riecken-dunnhumby (approved):

Geod24:
> You mean the deprecation message the compiler will emit, it's Use const, etc.

I mean that the location will be inside Ocean, it won't point where `Const` & co are actually instantiated. Moreover this will not play along well with deprecated code, e.g. if a `deprecated` symbol is used within `deprecated` code, it should not issue a message. This is why I was thinking fixing the bug might be less trouble.

jens-mueller-sociomantic:
Can somebody check whether I'm targeting the proper branch? I didn't choose `v4.x.x` because it should not be supported anymore.

alaksiej-stankievicx-dunnhumby (approved):

jens-mueller-sociomantic:
I see you point @Geod24. Let me check.

alaksiej-stankievicx-dunnhumby:
> I mean that the location will be inside Ocean, it won't point where Const & co are actually instantiated. Moreover this will not play along well with deprecated code, e.g. if a deprecated symbol is used within deprecated code, it should not issue a message. This is why I was thinking fixing the bug might be less trouble.

If do ugly hack something with string mixin inside?

jens-mueller-sociomantic:
To me the reporting by the compiler is fine. It could be more direct though. When using `Inout` inside the unittest I'll get

```
./src/ocean/meta/types/Qualifiers.d(139): Deprecation: alias ocean.meta.types.Qualifiers.Inout!(char[]).Temp is deprecated - Use inout
./src/ocean/meta/types/Qualifiers.d(144): Error: template instance ocean.meta.types.Qualifiers.Inout!(char[]) error instantiating
```

jens-mueller-sociomantic:
I also checked

```d
deprecated                                                                      
struct Foo                                                                      
{                                                                               
    Const!(int) a;                                                              
}
```

which gives me

```
./src/ocean/meta/types/Qualifiers.d(68): Deprecation: alias ocean.meta.types.Qualifiers.Const!int.Temp is deprecated - Use const
./src/ocean/meta/types/Qualifiers.d(156): Error: template instance ocean.meta.types.Qualifiers.Const!int error instantiating
```

The last line is directly pointing to were `Const` is used.

src/ocean/meta/types/ReduceType.d
jens-mueller-sociomantic: I see. Makes sense then. Thank you.

jens-mueller-sociomantic:
Let me know if you find the messages not good enough.

jens-mueller-sociomantic:
In the fixup I deleted the corresponding unittests. I think we have to because ocean treats deprecations as errors.

jens-mueller-sociomantic:
@Geod24 I just saw https://github.com/sociomantic-tsunami/ocean/pull/732. I didn't know you've been already working on this. I only addressed the things I currently need unaware of the bigger picture. I think we should gradually deprecate everything that is only needed for transitioning in v5.x.x and remove it in v6.x.x. I hope that you find the above compiler outputs sufficient. With your already merged changes for `ocean.transition` it's only left to move the useful stuff out of it and deprecate the whole module so that we can delete the file later.
What do you think?

jens-mueller-sociomantic:
To summarize the found issues:

Deprecating aliases in templates requires a workaround with current dmd which makes the compiler message less direct. Also there will be deprecation messages when `Const`, etc. is used inside already deprecated code. The unittests for `Const`, etc. have to be removed because `deprecated` does not work for them. It seems we are observing compiler shortcomings related to deprecating aliases for eponymous templates.

From my side all I care about is that user code somewhat breaks when using any transitional feature. This is to ensure that you don't silently use any transitional feature in a code base that was already transitioned.

While I agree that having the compiler work properly, I mostly care about the outcome for code using transitional features. For example I would also be fine, if there was an ocean v6.x
.x that removed all transitional features and breaks compiling this way.

Geod24:
Sorry for the delay. Indeed the solution you came up with looks good enough, IMHO.

src/ocean/meta/types/Qualifiers.d
```diff
@@ -130,6 +112,7 @@ unittest
 
 *******************************************************************************/
 
+deprecated("Use inout")
```
alaksiej-stankievicx-dunnhumby: it seems this line here by accident

jens-mueller-sociomantic: Yeah, but thanks for spotting.

jens-mueller-sociomantic:
This ready for review. I cleaned the commits a bit.

alaksiej-stankievicx-dunnhumby (approved):

ben-palmer-sociomantic (commented):
LGTM. Its a pretty major change I would like a couple more reviews.

don-clugston-sociomantic:
We should remove all use of those templates from our other libraries, before we officially release this. A quick check: swarm v6.0.1 has 124 uses of `Const!`, dhtproto has 133. So there's some work to do. That will be a good test case for this.

It shouldn't be released until our own libraries compile cleanly. Then there's the question of what branch it should target.

jens-mueller-sociomantic:
I thought more the other way around. We release this and then when users update ocean they see the deprecation messages and fix them.

don-clugston-sociomantic:
I know we've behaved that way before, but it should stop.
Failing to even update our own libraries is incredibly rude and unprofessional, and wastes a huge amount of time for every single user that uses the libraries. It is spam, essentially. 

I have created issues:
https://github.com/sociomantic-tsunami/swarm/issues/406
https://github.com/sociomantic-tsunami/dhtproto/issues/218
https://github.com/sociomantic-tsunami/dlsproto/issues/113
https://github.com/sociomantic-tsunami/dmqproto/issues/92

Probably we need one for turtle as well, but it only has a couple of modules that are affected.
It shouldn't really be very difficult to do this.

johannes-riecken-dunnhumby:
> I know we've behaved that way before, but it should stop.
> Failing to even update our own libraries is incredibly rude and unprofessional, and wastes a huge amount of time for every single user that uses the libraries. It is spam, essentially.
> 
> I have created issues:
> [sociomantic-tsunami/swarm#406](https://github.com/sociomantic-tsunami/swarm/issues/406)
> [sociomantic-tsunami/dhtproto#218](https://github.com/sociomantic-tsunami/dhtproto/issues/218)
> [sociomantic-tsunami/dlsproto#113](https://github.com/sociomantic-tsunami/dlsproto/issues/113)
> [sociomantic-tsunami/dmqproto#92](https://github.com/sociomantic-tsunami/dmqproto/issues/92)
> 
> Probably we need one for turtle as well, but it only has a couple of modules that are affected.
> It shouldn't really be very difficult to do this.

I disagree with leaving this PR hanging until PRs for the other libraries are written, approved and merged for the following reasons:
- It would create additional effort for Jens to rebase the PR when something else in ocean gets changed, since it touches so many files, talking about "wasting a huge amount of time"
- I don't see how additional deprecation warnings waste time for users of the libraries, especially since they can be turned off by adding `-d` to DFLAGS
- Adding the deprecation warnings now allows to notify developers of the library who create new PRs to understand that they shouldn't use Const, etc. anymore. For example from the logs of a discussion about mstring parameters on Friday it is clear that some employees are not aware that Const is deprecated

Geod24:
We only use Ocean, so I don't have a horse in this race, but I tend to agree with the approach @don-clugston-sociomantic recommend, for a few reasons:
- Scale: In a sane environment, there are more application developers than library developers;
- Ownership: It's easier and makes more sense for the person doing the deprecation to tackle downstream library, than to push it on whoever is more annoyed;
- Understanding: Related to the previous point, it is frequent that someone doing a deprecation misses some use case, which can be uncovered by fixing downstream libraries / one or two apps;

This is how things are usually done with language deprecations: we have a project tester, and whenever a change would affect one of the projects, it's usually tackled (the most recent example being https://github.com/sociomantic-tsunami/ocean/pull/726 ). From my experience, this is the most scalable and sensible approach.

don-clugston-sociomantic:
Note that there are two things going on in this PR! 
(a) replace `Const!` by `const`, etc inside ocean  --- this is huge!
(b) Deprecate `Const!`

The problem I describe is only caused by (b), which is only the last commit in this PR.

So I suggest: make a new PR which is the first four commits of this PR. We can merge it immediately.

Don't merge the final commit until `Const!` and friends are removed from swarm. It shouldn't create any conflicts, so there should be no maintenance burden from keeping it around until that's done.

jens-mueller-sociomantic:
I do see the benefit of converting other libraries just to test out the changes more. But eventually you will annoy people with deprecations assuming you don't control all code that is using ocean which is generally the case. That's why I thought the easiest approach for migration is to have a new ocean release issuing deprecations that the user should fix when updating. But it would be annoying when other libraries have not removed the usage of `Const` because it will flood the screen deprecation messages that the user cannot do anything about. It's a bit of a chicken-and-egg problem.

I'll create another PR without the last commit.

jens-mueller-sociomantic:
Closed in favor of sociomantic-tsunami/ocean#777.

PR #777, opened 2019-10-29
Replace Const, Immut and Inout with language builtins

jens-mueller-sociomantic:
This PR replaces all usages of the templates `Const`, `Immut` and `Inout` with
their appropriate D2 equivalent, i.e., `const`, `immutable` and `inout`,
respectively. This should cover all usages (except in the respective unittests)
of these transitional templates at this point in time. It may be that usages of
these will be reintroduced to the code base but nevertheless it should be little
and by accident which can be addressed later when deprecating these templates.


don-clugston-sociomantic (approved):
LGTM

ben-palmer-sociomantic (approved):
LGTM

PR #779, opened 2019-10-29
Fix function to check for UTF-8 string type

### LINKED ISSUE #778: Check and fix isStringType() deprecation
`isStringType()` was deprecated and `isUTF8StringType()` should be used instead but the latter fails if a static array is checked.

```D
unittest
{
    static assert (isStringType!(char[6])); // true -> passes
    static assert (isUTF8StringType!(char[6])); // false -> fails
}
```

daniel-zullo:
The function `ocean.meta.traits.Arrays.isUTF8StringType()` is suggested to
replace the deprecated `ocean.core.Traits.isStringType()` but it failed to
check for static arrays and now it is fixed to support both basic kind of
arrays.

Fixes #778

### DIFF
--- a/relnotes/fix-is-string-type.bug.md
+++ b/relnotes/fix-is-string-type.bug.md
@@ -1,0 +1,10 @@
+### Fix function to check for UTF-8 string type
+
+`ocean.meta.traits.Arrays`
+
+The function `ocean.meta.traits.Arrays.isUTF8StringType()` is suggested to
+replace the deprecated `ocean.core.Traits.isStringType()` but it failed to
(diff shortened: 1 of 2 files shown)


alaksiej-stankievicx-dunnhumby (approved):
LGTM.

ben-palmer-sociomantic (approved):
LGTM

Geod24:
FYII this is consistent with what upstream is doing: static arrays are not ranges, only dynamic arrays.
Although in this case it's only used in one place where it makes sense to accept static arrays.

PR #780, opened 2019-11-05
Speed up compilation of AutoSmartEnum by 50X

don-clugston-sociomantic:
By replacing recursive template instantiation with a single CTFE function, the compilation time is reduced by a factor of ~50 for enums with a large number of members, and the compiler's memory usage is reduced. There is no change whatsoever to the code which is generated.

The examples which I tested, went from 0.5 seconds compile time, to 0.01 seconds, per smart enum. The impact on compiler memory usage may be more significant.

The templates which this replaced, were doing nothing except to create a string. This PR just speeds up the way the string is made, there is no change to the string, so it doesn't break anything. 

This fix doesn't improve `SmartEnum`, only `AutoSmartEnum`. The interface for `SmartEnum` relies on templates so it can't be sped up as easily.
It's likely that these classes aren't needed at all in D2 (they are only for backwards compatibility), so it would make no sense to change the interface.

### DIFF
--- a/src/ocean/core/SmartEnum.d
+++ b/src/ocean/core/SmartEnum.d
@@ -720,30 +720,6 @@ unittest
     test!("==")(count, 2);
 }
 
-/*******************************************************************************
-
-    Template to create a tuple of enum codes from 0 to the length of the passed
-    tuple of strings.
-
-    Params:
-        BaseType = base type of enum
-        i = counter (used recursively)
-        Strings = variadic list of descriptions of the enum values
-
-*******************************************************************************/
-
-private template CreateCodes ( BaseType, uint i, Strings ... )
-{
-    static if ( Strings.length == 1 )
-    {
-        alias Tuple!(SmartEnumValue!(BaseType)(Strings[0], i)) CreateCodes;
-    }
-    else
-    {
-        alias Tuple!(SmartEnumValue!(BaseType)(Strings[0], i), CreateCodes!(BaseType, i + 1, Strings[1 .. $])) CreateCodes;
-    }
-}
-
 
 /*******************************************************************************
 
@@ -762,7 +738,8 @@ public template AutoSmartEnum ( istring Name, BaseType, Strings ... )
 {
     static assert ( is(typeof(Strings[0]) : istring), "AutoSmartEnum - please only give immutable strings as template parameters");
 
-    static immutable AutoSmartEnum = SmartEnum!(Name, CreateCodes!(BaseType, 0, Strings));
+    static immutable AutoSmartEnum = autoSmartEnumImpl(Name, BaseType.stringof,
+        Strings);
 }
 
 unittest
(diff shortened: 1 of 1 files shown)


src/ocean/core/SmartEnum.d
```diff
@@ -786,6 +763,65 @@ unittest
...
+        if (s.length < minlen )
+        {
+            minlen = s.length;
+        }
+
+        // declstr will be of the form "A=0,B=1,C=2"
+
+        if ( i > 0 ) declstr ~= ",";
+
+        declstr ~= s ~"=" ~ CTFE.toString(i);
```
alaksiej-stankievicx-dunnhumby: You have missed a space here `~"="`

src/ocean/core/SmartEnum.d
```diff
@@ -786,6 +763,65 @@ unittest
...
+
+        // mapstr will be of the form  `map["A"]=A;map["B"]=B;`
+
+        mapstr ~= `map["` ~ s ~`"]=` ~ s ~ `;`;
+    }
+
+    return  "class " ~ name ~ " : ISmartEnum {" ~
+        " alias " ~ basetype ~ " BaseType; enum : BaseType {"
+        ~ declstr ~ "} enum length = " ~ CTFE.toString(members.length) ~
```
alaksiej-stankievicx-dunnhumby: you could be more consistent with the position of `~` operator (either on the very begin of newline or in the very end). I personally prefer the operator in the very begin because it is a focused area. I even have something like this in my code:
```d
auto result = first_long_name_variable
    + second_long_name variable;
```

However, whatever you prefer, but consistent.

alaksiej-stankievicx-dunnhumby (commented):
Codewise LGTM. But you can make the style more consistent.

Also, you could add unittest, that don't only make happy me but also code coverage checker.

don-clugston-sociomantic:
Repushed with an extra test, and some formatting fixes. I have no idea how to make Travis retest it.

jens-mueller-sociomantic:
You can follow the link and login at Travis to restart the build. But something seems messed up. Can you try force pushing?

alaksiej-stankievicx-dunnhumby:
BTW, do we need also stuff in release notes?

jens-mueller-sociomantic:
> BTW, do we need also stuff in release notes?

To me they are not needed because only the implementation is improved. The semantics are the same. One could advertise it as a feature though.

src/ocean/core/SmartEnum.d
```diff
@@ -786,6 +763,77 @@ unittest
...
+    // the map.
+    // Also determine the minimum and maximum length of each member name
+
+    istring declstr = null;
+    istring mapstr = null;
+
+    long maxlen = members[0].length;
+    long minlen = members[0].length;
+
+    foreach (i, s ; members )
```
ben-palmer-sociomantic: Inconsistent white space after `foreach (i`

src/ocean/core/SmartEnum.d
```diff
@@ -786,6 +763,77 @@ unittest
...
+    {
+        if (s.length > maxlen)
+        {
+            maxlen = s.length;
+        }
+        if (s.length < minlen )
```
ben-palmer-sociomantic: Inconsistent white space above in the `if` statements compared to below.

ben-palmer-sociomantic (commented):
Made two very minor formatting comments and restarted the travis build but it looks like it is still having issues.

don-clugston-sociomantic:
Repushed with formatting fixes and now it works..... :confused:

alaksiej-stankievicx-dunnhumby (approved):
LGTM

ben-palmer-sociomantic (approved):
LGTM

PR #783, opened 2019-11-10
Refactor unit test to avoid an optimisation bug

daniel-zullo:
A couple of applications that import `ocean.math.IncrementalAverage`
as a dependency for unit tests hit a compiler optimisation bug (using
flags -O -inline).
It was only observed to happen with dmd-transitional so far and it seems
the order in which the dependencies for unit tests are added matters.

Anyway the functionality in the unit test remains the same as it
only refactors the post-condition of the for-loop and this change
improves the code readability.


ben-palmer-sociomantic (approved):
LGTM

Issue #781, opened 2019-11-07
BitArray.init name is problematic


AndrejMitrovic:
https://github.com/sociomantic-tsunami/ocean/blob/ad2c368b338d74cdd6991e5c32b1c042250ebef0/src/ocean/core/BitArray.d#L206

This is causing issues with generic code. For example, `BitArray.init` cannot be used because it's "overriden" here, leading to a compiler error.

I suggest renaming it to `initialize`. But I don't know if this is used anywhere within Sociomantic, so it's up to you guys to decide if it's worth renaming or if it has to go through a deprecation phase first. I think it's probably not used anywhere because there are no annotations in this struct.

don-clugston-sociomantic:
This appears to be fixed by #784.

PR #784, opened 2019-11-12
Deprecate BitArray.init and introduce initialize()

AndrejMitrovic:
The `BitArray` struct has an `init` method which was used for initializing
with a set of parameters. However, this makes `BitArray.init` unusable in
some contexts in generic code, as it would trigger compiler errors.

The new `initialize` method should be used instead.

Part of #781


alaksiej-stankievicx-dunnhumby (approved):
LGTM

ben-palmer-sociomantic (approved):
LGTM

PR #786, opened 2019-11-28
Replace opCatAssign with opOpAssign in Buffer

Geod24:
Since Buffer is a struct, there is no concern with making this function a template.
Additionally, since Buffer is widely used in Ocean,
it is one of the main source of deprecations.

### DIFF
--- a/src/ocean/core/buffer/NoIndirections.d
+++ b/src/ocean/core/buffer/NoIndirections.d
@@ -117,11 +117,12 @@ template NoIndirectionsBufferImpl ( )
         Appends to current buffer
 
         Params:
+            op = operation to perform
             rhs = array or element to append
 
     ***************************************************************************/
 
-    void opCatAssign ( in T rhs )
+    void opOpAssign (string op) ( in T rhs ) if (op == "~")
     {
         this.length = this.data.length + 1;
         this.data[$-1] = rhs;
@@ -133,7 +134,7 @@ template NoIndirectionsBufferImpl ( )
 
     ***************************************************************************/
 
-    void opCatAssign ( in T[] rhs )
+    void opOpAssign (string op) ( in T[] rhs ) if (op == "~")
     {
         this.length = this.data.length + rhs.length;
         this.data[$-rhs.length .. $] = rhs[];
--- a/src/ocean/core/buffer/Void.d
+++ b/src/ocean/core/buffer/Void.d
@@ -96,11 +96,12 @@ template VoidBufferImpl ( )
         Appends to current buffer
 
         Params:
+            op = operation to perform
             rhs = array or element to append
 
     ***************************************************************************/
 
-    void opCatAssign ( in ubyte rhs )
+    void opOpAssign (string op) ( in ubyte rhs ) if (op == "~")
     {
         this.length = this.data.length + 1;
         this.data[$-1] = rhs;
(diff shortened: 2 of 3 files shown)


ben-palmer-sociomantic:
I would recommend we take the opportunity to add unit tests for these methods as we have done for the other changes if possible (unless I missed them).

src/ocean/core/buffer/NoIndirections.d
```diff
@@ -121,7 +121,7 @@ template NoIndirectionsBufferImpl ( )
 
     ***************************************************************************/
 
-    void opCatAssign ( in T rhs )
+    void opOpAssign (string op) ( in T rhs ) if (op == "~")
```
daniel-zullo, line 125: Needs to update documentation

src/ocean/core/buffer/Void.d
```diff
@@ -100,7 +100,7 @@ template VoidBufferImpl ( )
 
     ***************************************************************************/
 
-    void opCatAssign ( in ubyte rhs )
+    void opOpAssign (string op) ( in ubyte rhs ) if (op == "~")
```
daniel-zullo, line 104: Needs to update documentation

src/ocean/core/buffer/WithIndirections.d
```diff
@@ -102,7 +102,7 @@ template WithIndirectionsBufferImpl ( )
 
     ***************************************************************************/
 
-    void opCatAssign ( T rhs )
+    void opOpAssign (string op) ( T rhs ) if (op == "~")
```
daniel-zullo, line 106: Needs to update documentation

daniel-zullo (commented):
Documentation and unit tests are welcome. And other than that LGTM

src/ocean/core/buffer/NoIndirections.d
Geod24: How so ? `op` can only add one value. Also `T` isn't documented as well. Should I add the trivial documentation for both of them ?

daniel-zullo: I think only for `op` as `rhs` is documented and it isn't needed for `T`.
In previous PRs `op` was documented as `op = operation to perform`

Geod24:
> I would recommend we take the opportunity to add unit tests for these methods as we have done for the other changes if possible (unless I missed them).

`NoIndirection.d`'s `opOpAssign`, both overloads:
https://github.com/sociomantic-tsunami/ocean/blob/aaa18cd89f4f2d106b659edc27150fe0ef68e607/src/ocean/core/Buffer_test.d#L40-L43

`WithIndirection.d`'s `opOpAssign`, both overloads:
https://github.com/sociomantic-tsunami/ocean/blob/aaa18cd89f4f2d106b659edc27150fe0ef68e607/src/ocean/core/Buffer_test.d#L126-L129

`Void.d`'s `opOpAssign`, both overloads:
https://github.com/sociomantic-tsunami/ocean/blob/aaa18cd89f4f2d106b659edc27150fe0ef68e607/src/ocean/core/Buffer_test.d#L182-L185

Added documentation.

daniel-zullo (approved):

Geod24:
ping ?

alaksiej-stankievicx-dunnhumby (approved):
LGTM

PR #787, opened 2019-12-03
Replace D1 operator overloads with D2's in BitArray

Geod24:
Since BitArray is a struct, virtuality of method is not an issue.

Documentation added, and all those methods have unittests (the deprecation message only triggers on usage, not on declaration).

### DIFF
--- a/src/ocean/core/BitArray.d
+++ b/src/ocean/core/BitArray.d
@@ -582,10 +582,13 @@ struct BitArray
      * Generates a copy of this array with the unary complement operation
      * applied.
      *
+     * Params:
+     *  op = operation to perform
+     *
      * Returns:
      *  A new array which is the complement of this array.
      */
-    BitArray opCom()
+    BitArray opUnary ( string op )( ) if (op == "~")
     {
         auto dim = this.dim();
 
@@ -613,20 +616,24 @@ struct BitArray
     }
 
     /**
-     * Generates a new array which is the result of a bitwise and operation
+     * Generates a new array which is the result of bitwise operation `op`
      * between this array and the supplied array.
      *
+     * Supported are bitwise binary operation "&amp;", "|", "^".
+     *
      * Params:
-     *  rhs = The array with which to perform the bitwise and operation.
+     *  op = operation to perform
+     *  rhs = The array with which to perform the bitwise operation.
      *
      * In:
      *  rhs.length must equal the length of this array.
      *
      * Returns:
-     *  A new array which is the result of a bitwise and with this array and
-     *  the supplied array.
+     *  A new array which is the result of a bitwise operation between
+     *  this array and the supplied array.
      */
-    BitArray opAnd( BitArray rhs )
+    BitArray opBinary ( string op )( BitArray rhs )
+        if (op == "&" || op == "|" || op == "^")
     {
         verify(len == rhs.length);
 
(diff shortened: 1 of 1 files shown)


alaksiej-stankievicx-dunnhumby (approved):
LGTM

@daniel-zullo-sociomantic, could you provide second review?

src/ocean/core/BitArray.d
```diff
@@ -626,7 +632,8 @@ struct BitArray
      *  A new array which is the result of a bitwise and with this array and
      *  the supplied array.
```
daniel-zullo: I think documentation here needs to be updated too

src/ocean/core/BitArray.d
```diff
@@ -880,26 +832,30 @@ struct BitArray
 
 
     /**
-     * Updates the contents of this array with the result of a bitwise and
+     * Updates the contents of this array with the result of a bitwise `op`
      * operation between this array and the supplied array.
      *
+     * Supported are bitwise assignment "&amp;=", "|", "^"
```
daniel-zullo: shoudn't it be "|=",  "^=" instead?

daniel-zullo (commented):
I only have a couple of comments about documentation and code-wise LGTM

src/ocean/core/BitArray.d
alaksiej-stankievicx-dunnhumby: In any case consistent

Geod24: Well the `op` parameter takes the value `|`, not `|`. But to show the difference with a non-assign version, I wrote "bitwise *assignment*"

daniel-zullo: then you could remove the assignment from `&=` to keep it consistent as @alaksiej-stankievicx-sociomantic suggested

Geod24: Ah right, I didn't see that. Will fix.

Geod24:
Fixed documentation

src/ocean/core/BitArray.d
```diff
@@ -1063,7 +973,7 @@ struct BitArray
     }
 
     /** ditto */
-    BitArray opCatAssign( BitArray rhs )
+    BitArray opOpAssign(string op)( BitArray rhs ) if (op == "~")
```
alaksiej-stankievicx-dunnhumby: There is a missed space between the name of the function and the first parenthesis.

Now I see that was missed in initial code, but think, we can fix the style with this PR

alaksiej-stankievicx-dunnhumby: Also, personally I don't like the style of spaces between parentheses and parameter declaration, but it code style of the `ocean`. So it needs to be respected for template parameters as well.

Geod24:
Updated style

src/ocean/core/BitArray.d
```diff
@@ -1005,7 +914,7 @@ struct BitArray
      * Returns:
      *  A shallow copy of this array.
      */
-    BitArray opSubAssign( BitArray rhs )
+    BitArray opOpAssign (string op )( BitArray rhs ) if (op == "-")
```
alaksiej-stankievicx-dunnhumby: and the very last space here `( string`

Geod24: Erg sorry, fixed

alaksiej-stankievicx-dunnhumby (approved):
LGTM

@daniel-zullo-sociomantic, do you have any other comments?

daniel-zullo (approved):
LGTM too

ben-palmer-sociomantic (approved):
LGTM

PR #789, opened 2019-12-10
Add support for Ubuntu bionic

daniel-zullo:
- Remove D1 flags which are no longer needed as the code base is only D2.
- Set libssl and libcrypto to version 1.0.0 supported in Ubuntu xenial and bionic.
- The dmd-transitional compiler is available for Ubuntu bionic only in vesion 2.078.3.s*.
- Update beaver build image to support Ubuntu xenial and bionic distributions.
- Remove package d1to2fix that is no longer required as the code base is only D2.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,20 +33,20 @@ jobs:
     include:
         # Test matrix
         - <<: *test-matrix-d2
-          env: DMD=2.071.2.s* F=production
+          env: DIST=xenial DMD=2.071.2.s* F=production
         - <<: *test-matrix-d2
-          env: DMD=2.071.2.s* F=devel
+          env: DIST=xenial DMD=2.071.2.s* F=devel
         - <<: *test-matrix-d2
-          env: DMD=2.078.3.s* F=production
+          env: DIST=xenial DMD=2.078.3.s* F=production
         - <<: *test-matrix-d2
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -47,6 +47,17 @@ jobs:
           env: DMD=2.085.* F=production
         - <<: *test-matrix-d2
           env: DMD=2.085.* F=devel
+        - <<: *test-matrix-d2
+          env: DIST=bionic DMD=2.078.3.s* F=production
```
mihails-strasuns, line 51: Adding explicit `DIST=xenial` to old rows would make it slightly clearer without any further context.

mihails-strasuns (approved):

.travis.yml
daniel-zullo: right, I will add it to the old rows soon

daniel-zullo:
This is now blocked due to the issue that Don reported in 
https://issues.dlang.org/show_bug.cgi?id=20441

Geod24:
That bug has been fixed

daniel-zullo:
Updated to test dmd 2.091

daniel-zullo:
@Geod24 you might want to review it.

Geod24 (approved):
LGTM

PR #790, opened 2019-12-13
Zero pad milliseconds to 3 places

andrew-stevenson-sociomantic:
Reverts behavioural change from dcf428aa

### DIFF
--- a/src/ocean/util/log/layout/LayoutStatsLog.d
+++ b/src/ocean/util/log/layout/LayoutStatsLog.d
@@ -68,7 +68,7 @@ public class LayoutStatsLog : Appender.Layout
         const dt = (localTime) ? WallClock.toDate(tm) : Clock.toDate(tm);
 
         // format date according to ISO-8601 (lightweight formatter)
-        sformat(dg, "{u4}-{u2}-{u2} {u2}:{u2}:{u2},{u2} {}",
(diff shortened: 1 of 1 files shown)


andrew-stevenson-sociomantic:
@Geod24 FYI

alaksiej-stankievicx-dunnhumby (approved):
LGTM

You probably may comment in commit message more on the reason why this reverting is important.
And also if this behaviour is required it is good to add unittest on it to prevent problems in the future.

andrew-stevenson-sociomantic:
Well its not really that important apart from stopping it looking odd. If other things happen to have depended on the format always using 0 padding then that is really more a bug with them.

The real problem was the un-noticed behavioural change - but that was presumably accidental. A unittest might have caught that but given this method now largely just calls sformat (presumably already tested) we are really only testing no one changes the format string - and that seems unlikely for anyone to do without due consideration. If someone does decide to re-write it again then comparing old a new behaviour can be a task for them.

christoph-wasicki-sociomantic (approved):

Geod24:
Oops, good catch!

> A unittest might have caught that

There was no unittest in place for those classes at the time. So the first thing I did was to introduce an unittest, but the value I used did not exhibit the issue.

Geod24 (approved):

ben-palmer-sociomantic (approved):
LGTM

PR #791, opened 2019-12-17
Replace deprecated operator overloads in Map and container modules

don-clugston-sociomantic:
In dmd 2.088, the D1 operator overloads are deprecated. This causes almost 2000 deprecation warnings when compiling most apps using our libraries.

This PR deals with the straightforward cases, where the operator is a member of a struct. In this case it is a simple substitution, that won't break any existing code.

(The cases where the operator is a member of a class are far more difficult, because the D1 operator overloads are virtual, whereas the D2 overloads are static. Those cases are not dealt with in this PR).

Additionally in this PR I have fixed one 'shadowing' warning in a unit test, and cleaned up a broken doc comment.

### DIFF
--- a/src/ocean/util/container/more/Vector.d
+++ b/src/ocean/util/container/more/Vector.d
@@ -32,7 +32,11 @@ struct Vector (V, int Size = 0)
 {
         alias add       push;
         alias slice     opSlice;
-        alias push      opCatAssign;
+
+        public template opOpAssign ( string op : "~" )
+        {
+            alias opOpAssign = push;
+        }
 
         static if (Size == 0)
                   {
--- a/src/ocean/core/BitArray.d
+++ b/src/ocean/core/BitArray.d
@@ -588,7 +588,7 @@ struct BitArray
      * Returns:
      *  A new array which is the complement of this array.
      */
-    BitArray opUnary ( string op )( ) if (op == "~")
+    BitArray opUnary ( string op : "~" )( )
     {
         auto dim = this.dim();
 
@@ -705,7 +705,7 @@ struct BitArray
      * Returns:
      *  A new array which is the result of this array minus the supplied array.
      */
-    BitArray opBinary ( string op )( BitArray rhs ) if (op == "-")
+    BitArray opBinary ( string op : "-" )( BitArray rhs )
     {
         verify(len == rhs.length);
 
@@ -745,7 +745,7 @@ struct BitArray
      *  A new array which is the result of this array concatenated with the
      *  supplied array.
      */
-    BitArray opBinary ( string op )( bool rhs ) if (op == "~")
+    BitArray opBinary ( string op : "~" )( bool rhs )
     {
         BitArray result;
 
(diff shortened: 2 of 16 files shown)


src/ocean/util/container/more/Vector.d
```diff
@@ -32,7 +32,7 @@ struct Vector (V, int Size = 0)
 {
         alias add       push;
         alias slice     opSlice;
-        alias push      opCatAssign;
+        public template opOpAssign(string op : "~") { alias opOpAssign = push; }
```
alaksiej-stankievicx-dunnhumby: Don't we prefer a different way:
```d
public template opOpAssign (string op) if(op == "~") {alias opOpAssign = push;}
```

alaksiej-stankievicx-dunnhumby (commented):
So, IIRC,  we prefer template constraints not template specification in D2. Do you have any reason to do this in the current way?

src/ocean/util/container/more/Vector.d
don-clugston-sociomantic: What we want here is not a constrained template but a specifically defined one (all parameters specified).
The compilation time is substantially better with specialisation, the code is shorter and more readable, and the behaviour is the same.

stefan-koch-sociomantic:
template constraints get evaluated by the CTFE whereas specialization gets resolved in the template dedicated code.
It leads to compile time savings not to use constraints.

alaksiej-stankievicx-dunnhumby:
I'm mostly about that in other PR for `ocean` we have constraints (e.g see this https://github.com/Geod24/ocean/blob/6f2181843bbd033f534489be78f3d055391206c7/src/ocean/core/BitArray.d#L591). And if there is no heavy drawbacks I prefer consistent approach.

stefan-koch-sociomantic:
it also looks nicer.

don-clugston-sociomantic:
Good point. I don't think there's any reason those ones were done in that way.

Geod24:
> Good point. I don't think there's any reason those ones were done in that way.

AFAIR I tried specialization first (it looks nicer) but AFAIR it didn't work. I can't quite remember the specific though.

don-clugston-sociomantic:
Repushed with the other concatenation operators changed to use specialization. If it doesn't work, hopefully we'll see why.

don-clugston-sociomantic:
I found two more modules with overloaded operators that can be changed to D2-style without problems.

ben-palmer-sociomantic:
Changes LGTM. I haven't checked what the unit test coverage is for the changes though. Could be argued to be out of scope of the PR but would be nice to check.

don-clugston-sociomantic:
The code coverage actually increases as a result of this PR (since a new unittest is added). But the *reported* code coverage goes down, because we're replacing ordinary functions with templated functions.
The code coverage stats ignore templates. So you get no credit for testing them. 
It's a structural bug in code coverage.

ben-palmer-sociomantic (approved):

ben-palmer-sociomantic:
Any other comments before I merge?

Issue #730, opened 2019-08-27
Many ocean.math unittests fail when built with LDC


joseph-wakeling-frequenz:
I've given a go to building ocean v5.0.x tests with `ldmd2` instead of `dmd` (v5.0.x because it includes the `getIeeeFlags` fix that was a blocker to using compilers other than DMD).  The command used was:

```
DC=ldmd2 make V=1 test ALLOW_DEPRECATIONS=1
```

... with LDC 1.16.0 being used to test (matching DMD 2.086.1).

My suspicion (not yet confirmed) is that all of these reflect LDC using different core math routines to DMD (e.g. implementation of cos, sin, sqrt, etc.).  I suspect that similar results might be obtained for GDC, for similar reasons.

No worries if this is a WONTFIX (or at least very low priority) from the maintainers' side, but since I went through this all systematically today I thought I'd share the results.

The following unittests all fail when LDC is used:

`ocean.math.Elliptic`:
https://github.com/sociomantic-tsunami/ocean/blob/63ef0e886955986a99c89aeb4d0f939149c5975a/src/ocean/math/Elliptic.d#L349

`ocean.math.Probability`:
https://github.com/sociomantic-tsunami/ocean/blob/63ef0e886955986a99c89aeb4d0f939149c5975a/src/ocean/math/Probability.d#L261-L271
https://github.com/sociomantic-tsunami/ocean/blob/63ef0e886955986a99c89aeb4d0f939149c5975a/src/ocean/math/Probability.d#L368

`ocean.math.Math`:
https://github.com/sociomantic-tsunami/ocean/blob/63ef0e886955986a99c89aeb4d0f939149c5975a/src/ocean/math/Math.d#L422
https://github.com/sociomantic-tsunami/ocean/blob/63ef0e886955986a99c89aeb4d0f939149c5975a/src/ocean/math/Math.d#L428-L429
https://github.com/sociomantic-tsunami/ocean/blob/63ef0e886955986a99c89aeb4d0f939149c5975a/src/ocean/math/Math.d#L435
https://github.com/sociomantic-tsunami/ocean/blob/63ef0e886955986a99c89aeb4d0f939149c5975a/src/ocean/math/Math.d#L490
https://github.com/sociomantic-tsunami/ocean/blob/63ef0e886955986a99c89aeb4d0f939149c5975a/src/ocean/math/Math.d#L513-L514
[...]

PR #792, opened 2019-12-19
Add DUB unittest configuration to specify library dependencies

joseph-wakeling-frequenz:
This should help avoid linker errors when running `dub test`.  Note that `libssl` (and hence `libcrypto`) have to be set to v1.0.0 instead of the default v1.1.0, as ocean requires symbols that are not available via the latter, specifically:

  - `SSL_library_init`
  - `SSL_load_error_strings`
  - `SSLv23_method`

Placing these settings inside the unittest configuration should ensure that downstreams can avoid linking against library dependencies they do not need: however, it comes at the cost that downstreams have to specify their library dependencies manually.

Some libraries specified in `Build.mak` appear to be unnecessary and are therefore not included in the DUB config:

  - bz2
  - history
  - rt
  - xml2
  - zlib (`-lz`) [probably because `zlib` is already in phobos]

The `ocean.core.UnitTestRunner` module is excluded from unittest builds, as its `main` would clash with the DUB auto-created `main`.  A different option would be to explicitly use the ocean unittest runner, but this is left for a later decision.

Finally, note that this setup will only run unittests, not integration tests: addressing this is left for future work.

As written, it is possible to run `dub test` successfully with upstream DMD.  `Using --compiler=dmd-transitional` fails with an error message:

```
Error: module typetuple is in file 'std/typetuple.d' which cannot be read
```

while using `--compiler=ldc2` or `--compiler=ldmd2` fails because of https://github.com/sociomantic-tsunami/ocean/issues/730.

### DIFF
--- a/dub.sdl
+++ b/dub.sdl
@@ -7,3 +7,9 @@ copyright "Copyright (c) 2004 dunnhumby Germany GmbH and Tango contributors. All
 systemDependencies "Depending on the component used you might need one of the following libraries (check the used modules for details): -lglib-2.0 -lpcre -lxml2 -lxslt -lreadline -lhistory -llzo2 -lbz2 -lz -ldl -lgcrypt -lgpg-error -lrt -lebtree (version from https://github.com/sociomantic-tsunami/ebtree)"
 importPaths "src"
 targetType "sourceLibrary"
+
+configuration "unittest" {
+    excludedSourceFiles "src/ocean/core/UnitTestRunner.d"
+    libs "glib-2.0" "pcre" "ebtree" "readline" "lzo2" "gcrypt" "gpg-error"
+    lflags "-l:libcrypto.so.1.0.0" "-l:libssl.so.1.0.0"
+}


Geod24:
> Note that libssl (and hence libcrypto) have to be set to v1.0.0 instead of the default v1.1.0, as ocean requires symbols that are not available via the latter, [..]

I am wondering if those usages could be replaced. I would be highly surprised if there was not a common subset that works for both libraries. Perhaps look into how Vibe.d did it ?

joseph-wakeling-frequenz:
> I am wondering if those usages could be replaced. I would be highly surprised if there was not a common subset that works for both libraries. Perhaps look into how Vibe.d did it ?

I'm sure that's possible, but this was the simplest and smallest change required to get `dub test` working -- I'm not sure how much time I have to go chasing into the depths of how ocean uses libssl ;-)

The risk is that fixing it would be a breaking change.

ben-palmer-sociomantic:
I don't see anything too wrong with this approach however my dub knowledge is pretty much 0. Maybe @don-clugston-sociomantic or @stefan-koch-sociomantic could take a look?

don-clugston-sociomantic:
> I am wondering if those usages could be replaced. I would be highly surprised if there was not a common subset that works for both libraries. Perhaps look into how Vibe.d did it ?

You may be surprised. libssl is far worse than you would imagine. From v1.0 to v1.1 they completely changed the way it is initialised.  A rational person would have changed the old functions to be no-ops, but instead they made a breaking change. Of course it would be better to just abandon v1.0 but at the time there was a reason we needed to support it. That might not be true any more.

ben-palmer-sociomantic:
> Of course it would be better to just abandon v1.0 but at the time there was a reason we needed to support it. That might not be true any more.

I think for this PR we can just merge these changes and look at upgrading SSL in a later PR.

joseph-wakeling-frequenz:
Does that mean that we can take this in? :-P

don-clugston-sociomantic:
Yeah, I'll take that as an LGTM. I also approve.

PR #796, opened 2020-01-13
Introduce ZipStreamDecompressor, to support streaming of PKZIP files

don-clugston-sociomantic:
The existing `ZlibStreamDecompressor` supports streaming of gzip files.
This module extends that class to also support streamable pkzip files. User code does not need to know if it is processing a gzip or a pkzip archive; they are treated identically.

Streaming of PKZIP files isn't actually possible, but this module does it anyway :smile: 

In practice it works for pretty much any zip file that has only a single file inside it.
The weakness of the pkzip format is that it is not possible to know if you have a pathological case until you have received the entire archive. So if faced with a pathological case, it will just throw an exception.

### DIFF
--- a/src/ocean/io/compress/ZipStream.d
+++ b/src/ocean/io/compress/ZipStream.d
@@ -1,0 +1,492 @@
+/*******************************************************************************
+
+    Zlib decoder which supports pkzip and gzip archives, and can be stored in a
+    pool.
+
+    In general it is not possible to stream pkzip archives. This is because the
+    format supports some quirky features which were important in the days of
+    floppy diskettes.
+
+    This class supports an important case where streaming is
+    possible: an archive which consists of a single file stored at
+    the start of the archive, using DEFLATE compression.
+
+    Needs linking with -lz.
+
+    Usage example:
+
+    ---
+
+        import ocean.io.compress.ZipStream;
+
+        auto unzipper = new ZipStreamDecompressor;
+
+        unzipper.reset();
+
+        try
+        {
+            unzipper.start();
+        }
+        catch (Exception e)
+        {
+            // Error!
+        }
+
+        // `downloader` is a hypothetical source which provides chunks of
+        // compressed data, eg downloaded from a socket
+        // Before processing, it may be wise to check that the first bytes in
+        // the file are equal to GZipFileSignature (for gzip files) or
+        // ocean.util.compress.c.Zip.ZipLocalFileHeaderSignature (for pkzip); an
+        // exception will be thrown if this is not true.
+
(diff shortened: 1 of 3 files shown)


ben-palmer-sociomantic:
Should this target v5.x.x?

src/ocean/io/compress/ZipStream.d
```diff
@@ -0,0 +1,492 @@
...
+    possible: an archive which consists of a single file stored at
+    the start of the archive, using DEFLATE compression.
+
+    Needs linking with -lz.
+
+    Usage example:
+
+    ---
+
+        import ocean.io.compress.ZipStream;
```
ben-palmer-sociomantic, line 20: Normally I would be a fan of having example code as unit tests. Not sure if this is possible for this module though.

ben-palmer-sociomantic (commented):
Only one comment from me about the example code.

src/ocean/io/compress/ZipStream.d
Geod24: Even if you don't run the code, wrapping the code inside a function that is never executed is already a step up because it ensures it'll always compile.

don-clugston-sociomantic: It's not real code though, it is pseudocode, It uses a hypothetical function, which would presumably be called from a separate fiber. So it will not compile.
It's a modified version of the example in `ZlibStreamDecompressor`.

don-clugston-sociomantic:
> Should this target v5.x.x?

Dunno. The problem is, there is an important commit in v5.3.x which isn't in v5.x.x. I don't know which of these branches is actually the head.

ben-palmer-sociomantic:
> Dunno. The problem is, there is an important commit in v5.3.x which isn't in v5.x.x. I don't know which of these branches is actually the head.

Oops, that was a mistake that should now be fixed.

don-clugston-sociomantic:
Changed to v5.x.x, now that the branches have been harmonised.

src/ocean/io/compress/ZipStream.d
ben-palmer-sociomantic: So there is no way we could just replace it with a dummy function?

don-clugston-sociomantic: I think that would clutter the example so much it would be largely useless as documentation. The point of it is to tell the higher-level reasoning behind using the code. Ie it is the WHY, not the WHAT.
The unittest which I have added, covers what's actually required to use the class.

stefan-koch-sociomantic:
This looks good to me!

PR #797, opened 2020-01-17
Prepare for dlang/dmd#10124 (field of init symbol no lvalue anymore) #2

kinke:
A follow-up to #726; should be the last one. ;) - I hope the branch is correct.

### DIFF
--- a/src/ocean/meta/traits/Basic.d
+++ b/src/ocean/meta/traits/Basic.d
@@ -674,7 +674,8 @@ unittest
 
 bool isManifestConstant ( T, string name ) ( )
 {
-    mixin(`return is(typeof(T.init.`~name~`)) && !is(typeof(&T.init.`~name~`));`);
+    T instance;
+    mixin(`return is(typeof(instance.`~name~`)) && !is(typeof(&instance.`~name~`));`);
 }
 
 unittest


ben-palmer-sociomantic:
LGTM, @don-clugston-sociomantic  or @UplinkCoder could you take a quick look as well?

stefan-koch-sociomantic (approved):
That's correct.

Geod24:
That template looks buggy BTW. 

`struct` are the only place where `d1to2conv` replaced `const X = Y;` with `enum`. The other places used `static immutable`. And given the test, the following won't pass:
```D
class Foo
{
    static immutable Manifest = "Hello World";
}
static assert(isManifestConstant!(Foo, "Manifest"));
```

A better test would be to check if a `static immutable` can be initialized from the field, because `enum` and `static immutable` with initializer are just two kinds of manifest constant, one being a rvalue and the other being a lvalue, but `isManifestConstant` explicitly excludes the later (Which contradicts the doc).

UplinkCoder:
static immutable is not a manifest constant.

at least not necessarily.

you can do
```
static immutable sim =  int(4);
@unsafe void f()
{
    (*cast(int*)&sim) = 12;
}
```

Am Mo., 20. Jan. 2020 um 12:45 Uhr schrieb Mathias LANG <
notifications@github.com>:

Geod24:
> static immutable is not a manifest constant.

It is.

> you can do

That's undefined behavior.

ben-palmer-sociomantic:
So this needs a fix or a revert?

Geod24:
No @kinke 's fix is right, but IMO the original definition was wrong and should be fixed (but I doubt you use that template much, otherwise the issue would have shown up given how prevalent `static immutable` is in your code).
@don-clugston-sociomantic might have an opinion on this.

PR #798, opened 2020-01-23
Remove use of hacky genOpCmp and genOpEquals mixins

don-clugston-sociomantic:
These mixins were required for the D2 transition. They can now be replaced with the code which they generated.
In the case of `opEquals` they were only used inside unit tests.


stefan-koch-sociomantic (approved):
This looks good.
A nice addition and removal of legacy code.

PR #799, opened 2020-01-24
Replace 'import ocean.transition' inside ocean.core modules

don-clugston-sociomantic:
The `ocean.transition` module was used during D2 transition. Some of the aliases which it defined (`istring`, `cstring`, `mstring`) turned out to be useful, and were moved to another module.  This PR updates the imports to point to the new module. A couple of modules used `gc_stats` which doesn't have a replacement, but in those cases I made the transition import into a local import in a single function.

This PR cleans up all the modules in the `ocean.core.*` package. The rest of ocean will have to be updated at some time in the future.

### DIFF
--- a/src/ocean/io/console/AppStatus.d
+++ b/src/ocean/io/console/AppStatus.d
@@ -66,9 +66,9 @@ import ocean.io.Console;
 import ocean.io.Stdout;
 import ocean.io.Terminal;
 import ocean.io.model.IConduit;
+import ocean.meta.types.Qualifiers;
 import ocean.text.convert.Formatter;
 import ocean.time.MicrosecondsClock;
-import ocean.transition;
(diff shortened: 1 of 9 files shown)


stefan-koch-sociomantic (approved):
Small inconsistencies in the imports but it does not matter either way.

src/ocean/io/console/AppStatus.d
```diff
@@ -907,6 +907,8 @@ public class AppStatus
 
     public bool getMemoryUsage ( out float mem_allocated, out float mem_free )
     {
+        import ocean.transition;
```
stefan-koch-sociomantic, line 910: this could and should be replaced with core.memory.

don-clugston-sociomantic:
I have split this into two different PRs.

PR #800, opened 2020-01-27
Remove all remaining uses of "import ocean.transition"

don-clugston-sociomantic:
This PR minimizes the remaining usage of `ocean.transition` inside ocean. 
The modules shown here use `gc_usage` or `Octal` from `ocean.transition` which are not present anywhere else in ocean, which prevents the import from being completely removed. However this PR removes all other usage of `ocean.transition` from those modules.

I do not know if it's actually worth merging this, since it doesn't actually get rid of any of the transition imports. But it's maybe useful as a reference of what still remains to be done.

### DIFF
--- a/integrationtest/unixsockext/main.d
+++ b/integrationtest/unixsockext/main.d
@@ -15,7 +15,7 @@
 
 module integrationtest.unixsockext.main;
 
-import ocean.transition;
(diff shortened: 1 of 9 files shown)


Geod24:
Closing as this has been abandoned.

PR #801, opened 2020-01-30
Remove workarounds used during D2 transition

don-clugston-sociomantic:
This PR cleans up the main workarounds used during D2 transition.
These are the code changes required before `import ocean.transition` can be removed (the remaining required changes only involve changes to import statements).
None of the changes in this PR actually affect any generated code.

### DIFF
--- a/src/ocean/math/Range.d
+++ b/src/ocean/math/Range.d
@@ -50,8 +50,8 @@ public struct Range ( T )
 
     ***************************************************************************/
 
-    import ocean.transition: TypeofThis, assumeUnique;
-    mixin TypeofThis!();
+    alias typeof(this) This;
+
 
     /***************************************************************************
 
@@ -215,6 +215,8 @@ public struct Range ( T )
     {
         public istring toString()
         {
+            import ocean.core.TypeConvert: assumeUnique;
+
             auto msg = format("{}({}, {}", This.stringof, this.min_, this.max_);
 
             if (this.is_empty)
--- a/src/ocean/util/Convert.d
+++ b/src/ocean/util/Convert.d
@@ -916,7 +916,8 @@ D toStringFromString(D,S)(S value)
     static if (is(S : D))
         return value;
 
-    else static if (!isMutable!(DElem) || !isMutable!(SElem))
+    else static if ( is(DElem == const) || is(DElem == immutable)
+        || is(SElem == const) || is(SElem == immutable) )
         // both cases require creating new string which makes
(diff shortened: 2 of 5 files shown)


src/ocean/util/Convert.d
```diff
@@ -916,7 +916,8 @@ D toStringFromString(D,S)(S value)
     static if (is(S : D))
         return value;
 
-    else static if (!isMutable!(DElem) || !isMutable!(SElem))
```
Geod24, line 919: `isMutable` still looks better TBH

src/ocean/math/Range.d
```diff
@@ -215,6 +215,8 @@ public struct Range ( T )
     {
         public istring toString()
         {
+            import ocean.core.TypeConvert: assumeUnique;
```
jens-mueller-sociomantic, line 218: This seems unrelated to the commit. Looks like you removed an unnecessary import here. But I don't see by which import it is actually covered.

jens-mueller-sociomantic (commented):
Looks good to me. The only thing that confuse me is the removal of this one import statement.

src/ocean/util/Convert.d
jens-mueller-sociomantic: True, what were the reasons to drop it?

don-clugston-sociomantic: You reckon? Personally I don't like the double negative (if it is not not const and not not immutable ...) But the point is, the use of `isMutable`  is a blocker to removing ocean.transition, it doesn't exist elsewhere in ocean.

Geod24: Maybe `isMutable` should be made a type traits then ?
Or use a different condition, such as `is(Unqual!T == T)` to show that only unqualified types are accepted ?

src/ocean/math/Range.d
don-clugston-sociomantic: This was the only remaining use of `ocean.transition` in this module, which is the reason for making these changes. It will be removed from everywhere else in ocean in the next PR.

src/ocean/util/Convert.d
don-clugston-sociomantic: I do not think anything like that is justified for a single expression which is only used in one place in the internals of a function. I am trying to make the absolutely minimal changes required to finish the ocean D2 transition. I can't justify anything more complicated than that.

Geod24 (approved):
LGTM

PR #802, opened 2020-01-30
Replace ocean.transition imports where trivially possible

don-clugston-sociomantic:
The ocean.transition module was created to ease D2 conversion. It can now be replaced with more appropriate imports.

This PR changes the import statements and makes no other changes.

The replacement was done by executing the command
    `find . -type f -name "*.d" | xargs -o sed -i "s/import ocean.transition;/import ocean.meta.types.Qualifiers;/g"`

Virtually all modules use `cstring`, so the automatic replacement will not produce very many unnecessary imports.

I then added imports for `Typedef` and `assumeUnique` when required.
In many cases these were only required inside unit tests, so in those cases the imports were added inside `version (unittest)` blocks.

For the few modules in `ocean.core.array` I did the replacement carefully, ensuring that the imports are actually minimal, and I also cleaned up the import order. I didn't bother doing this with the rest.

Once this PR has been merged, the only remaining modules which import `ocean.transition` are:
* four modules which use `gc_usage`.
* six modules which use `Octal`.

Neither of those functions have direct replacements in ocean at present.

### DIFF
--- a/example/collectd/main.d
+++ b/example/collectd/main.d
@@ -20,7 +20,7 @@
 
 module example.collectd.main;
 
-import ocean.transition;
+import ocean.meta.types.Qualifiers;
 import ocean.core.Test;
 import ocean.net.collectd.Collectd;
 
--- a/integrationtest/asyncio/main.d
+++ b/integrationtest/asyncio/main.d
@@ -15,7 +15,7 @@
 
 module integrationtest.asyncio.main;
 
-import ocean.transition;
+import ocean.meta.types.Qualifiers;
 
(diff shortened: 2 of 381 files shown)


Geod24 (approved):
Most import are not ordered anymore, but I guess we don't care that much :)

don-clugston-sociomantic:
Yeah, though I noticed that many of the modules didn't have consistent ordering anyway. I know that years ago I saw a tool to fix the ordering, maybe it can be automated. If anyone actually cares enough to do it.

PR #805, opened 2020-02-04
Replace SortedMap.opIn_r with opBinaryRight!"in"

don-clugston-sociomantic:
Even though SortedMap is a class, the opIn_r function was marked as `final` so the fact that the D2 overload is non-virtual is not a problem.

This fixes a large number of deprecation warnings in dmd 2.090


don-clugston-sociomantic:
The codecov check is incorrectly marked as 'fail'. 0 lines are added, of which 0 are covered. That should be 'pass'.

alaksiej-stankievicx-dunnhumby (approved):

PR #806, opened 2020-02-07
Create D2 operator overloads to avoid deprecation messages

don-clugston-sociomantic:
D1-style operator overloads are deprecated in dmd 2.090.
This PR silences most of the remaining deprecation warnings in the upcoming dmd 2.091.

There are a few remaining deprecation warnings.
(1) from `SmartEnum`, which requires a little more care;
(2) some really dreadful operator overloads, which deserve to generate warnings:

* There is an absolutely disgusting opSubAssign in IEBTree. (`--this` is used to update the reference count)
* There is an opSub in SignalMask which is a bad idea.
* This is an opCatAssign in FilePath which appends a "/" before doing the append. I don't think ~= should be used for this.

### DIFF
--- a/src/ocean/core/Enum.d
+++ b/src/ocean/core/Enum.d
@@ -184,6 +184,17 @@ public interface IEnum
 
     /***************************************************************************
 
+        Support for the 'in' operator
+
+        Aliased to opIn_r, for backwards compatibility
+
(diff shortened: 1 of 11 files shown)


Geod24 (approved):
LGTM. Can we get this merged ?

PR #808, opened 2020-03-16
 Make ocean.core.array.Search cross platform

Geod24:
It's the same, but portable.
CC @don-clugston-sociomantic

### DIFF
--- a/src/ocean/core/array/Search.d
+++ b/src/ocean/core/array/Search.d
@@ -28,7 +28,13 @@ import ocean.core.Verify;
 import ocean.meta.traits.Basic;
 import ocean.meta.types.Qualifiers;
 
-import core.sys.posix.sys.types; // ssize_t;
+// ssize_t is only available on this header, but should really be in `object`
+version (Posix)
+    import core.sys.posix.sys.types;
+else static if (size_t.sizeof == 8)
+    alias ssize_t = long;
+else
+    alias ssize_t = int;
 
 version (unittest)
 {


alaksiej-stankievicx-dunnhumby (approved):
LGTM

don-clugston-sociomantic:
I don't like this idea. It perpetrates a bit of C naming stupidity. Semantically, it's wrong: it is most definitely not a pointer difference. It is a signed size. And I think that in general, the difference of two pointers is going to be unsigned, not signed. I cannot understand why C chose such a stupid name, the types of size_t and ptrdiff_t are backwards.
Is there any concrete reason why we need to use the stupid name instead of the slightly more sensible one?

Geod24:
Why would pointer difference be signed? I did see a few pointers computation resulting in negative values, e.g. in C where slices are not available.
Why `ptrdiff_t` ? Simply because it's the standard name and it's portable. We use a subset of Ocean (mostly the logger and formatter) and are adding Windows to our list of CIs. This is one of the things that stood out as non portable.

don-clugston-sociomantic:
If you mean "why would pointer difference be unsigned", it's because a pointer can point to anywhere in memory. 
And in fact it really is a modulo arithmetic. Incrementing the memory address 0xFFFFFFFFF gives you 0. Subtracting two arbitrary pointers gives you an unsigned value of the number of bytes between them. It's not a negative value, if you do that number of increments you do get to the other value. So it is one of the few things that is genuinely unsigned. 

Now, in C it technically isn't actually legal to subtract two pointers unless they both point to the same array. So it isn't actually allowed to wrap. BUT that it would only fit into a signed value if you are guaranteed that arrays never exceed half the memory space. And you do not have that guarantee.

Pretending that the difference between two pointers is signed, is silly. I don't understand how that happened.

Geod24:
> Pretending that the difference between two pointers is signed, is silly. I don't understand how that happened.

But it did happen, and it's the way most languages that support raw pointers, including D, work nowadays. The argument I can see for it is one of symmetry: If you subtract `b` from `a` and it gives you `o`, then adding `o` to `b` should give you `a`, bar any overflow. But whether or not that's sensible, it is currently how most language work and we have to deal with it.

> And in fact it really is a modulo arithmetic.

That's not quite how it works currently though. If you want modulo arithmetic, you have to do it yourself. And the model we use for computation assumes infinite memory.

> Now, in C it technically isn't actually legal to subtract two pointers unless they both point to the same array.

And getting a negative value is something one can do, e.g.when using `strstr` and the likes.

But I personally don't care either way. If you don't want `ptrdiff_t` that's fine, just let me know what way you'd prefer to see this done. An `ssize_t` alias in a `core` module perhaps ? I simply used `ptrdiff_t` because it fits the purpose by the current standard, even if that standard can seem a bit weird.

joseph-wakeling-frequenz:
> It's the same, but portable.

What's the portability issue?

Geod24:
https://github.com/sociomantic-tsunami/ocean/pull/808#issuecomment-599607324

joseph-wakeling-frequenz:
So this is about ensuring Windows compatibility?  Are there any other platforms you care about that are problematic?

FWIW I'm not unsympathetic to the thought "This is the standard name/type that the world has chosen for this purpose, and if it's not ideal, it's still the reality."  But given @don-clugston-sociomantic's principled objections, I think we should at least ask what the alternative options are.

Geod24:
> So this is about ensuring Windows compatibility? Are there any other platforms you care about that are problematic?

Yes. Not at the moment.

> I think we should at least ask what the alternative options are.

I mean that's what my last message is saying. Waiting for @don-clugston-sociomantic 's reply.

Geod24:
@don-clugston-sociomantic : So what do you think would be more appropriate than `ptrdiff_t` here ?

Geod24:
@don-clugston-sociomantic : So, what would be your prefered course of action here ?

don-clugston-sociomantic:
Yeah, this is a problem, and really should be fixed at the druntime level. I really object to `ptrdiff_t` because there are no pointers in this code at all. In fact `ssize_t` was created because `ptrdiff_t` is so wrong. Unfortunately some bad actors haven't support that.

Here is why I say it uses unsigned arithmetic. 
```
void main()
{
        ubyte[3] x;
        auto p = &x[2];
        auto arr = p[0..size_t.max];

        arr[0]=7;
        arr[$-1]=3;
        assert(x[2] == 7);
        assert(x[0] == 3);
}
```
Now maybe that's not legal, but you could in theory create ` auto arr = new ubyte[size_t.max/2 +1];
which has  `&arr[0] > null && &arr[$-1] < null` which tells me the whole thing is a big mess.
Pointer comparison is signed, which is just insane.

What we should have, I think, is a  static if that if `ssize_t` doesn't exist already, do `alias ssize_t = ptrdiff_t;` But that should really be in object.d or something.

Geod24:
What about this ?

don-clugston-sociomantic:
That's great! LGTM

andrew-stevenson-sociomantic:
> And I think that in general, the difference of two pointers is going to be unsigned, not signed. I cannot understand why C chose such a stupid name, the types of size_t and ptrdiff_t are backwards.

I think Don already knows this but for clarity for everyone else, in the C world you get a ptrdiff_t by subtracting two pointers - hence the name.

You are only allowed (IIRC) to subtract two pointers if they point to elements in the same array (with the usual caveat of the magic "extra" element off the end of the array). Everything else is undefined.

The confusing part, which I think is what Don was alluding to, is that the result of the diff is the difference in the indexes of those two elements and not some measure of bytes. That's consistent with other C things though - e.g. adding 1 to an `int` pointer moves you on one `int`, not 1 byte. If you want to do things with bytes then you are going to be un-portable and you can use an integral type that corresponds to bytes on your machines (`uint8_t` working on every modern architecture I imagine).

This result of the index subtraction can, of course, be signed because either pointer may be the larger one.

If they didn't create a new type for this then what should the result be? Clearly it must be a signed integer of some sort. Is there anything else in the C standard already that could work? Don already pointed out it could be a very large number and will need to be different sizes on different platforms.

I see `ssize_t` mentioned but it isn't in the C spec but comes from SUS. Even if it were available its range is only guaranteed to be from -1 to 32767 - its to allow functions to return the number of bytes read/written or -1 to indicate an error. Using it here would change its semantics.

don-clugston-sociomantic:
> This result of the index subtraction can, of course, be signed because either pointer may be the larger one.

Right, but this is essentially an unsigned operation. If the value is negative it means that either second pointer was further from the start of the array than the first was, OR that the array is in bytes and is greater than half of the memory space.

don-clugston-sociomantic:
I guess the only reason `ptrdiff_t` exists at all (as distinct from `size_t`) is because of segmented architectures like 8086 where you can have weird stuff like pointers that contain 16 bit segment + 16 bit offset, but the memory space is at most 24 bits?

I don't really care what we use, as long as it the same size as `size_t` but signed. It shouldn't be called ptrdiff_t when it doesn't involved pointers, though it can be an alias to that if implemented that way.

andrew-stevenson-sociomantic:
> If the value is negative it means that either second pointer was further from the start of the array than the first was, OR that the array is in bytes and is greater than half of the memory space.

If the result can't fit in a `ptrdiff_t` then the result is undefined so you shouldn't have done it ;-)
I think therefore the sign can only indicate the order of the two indexes.

`size_t` IIRC just has to be big enough to store the size of the largest object in bytes.There can be a limit other than the memory space on that - though I don't imagine there often is.

> I guess the only reason `ptrdiff_t` exists at all (as distinct from `size_t`) is because of segmented architectures like 8086 where you can have weird stuff like pointers that contain 16 bit segment + 16 bit offset, but the memory space is at most 24 bits?

I think the difference of two indexes is inherently signed so a signed type was needed and that rules out `size_t` without even thinking about weird architectures. Weird architectures will be why C doesn't really expose byte based operations though I would guess.

> I don't really care what we use, as long as it the same size as `size_t` but signed. It shouldn't be called ptrdiff_t when it doesn't involved pointers, though it can be an alias to that if implemented that way.

I'm in no way commenting on what should be done in D. D wrote off all small and older systems by design so probably can assume flat memory spaces and 8 bit bytes etc. If you at @Geod24 think its a good idea in D I'm sure I agree :-)

don-clugston-sociomantic:
That's interesting, that seems to imply that in C, the appropriate type for the number of elements in an array is `ptrdiff_t`, not `size_t`. 

>I think the difference of two indexes is inherently signed so a signed type was needed and that rules out size_t without even thinking about weird architectures. 
Yes, but you could rule out ptrdiff_t as well, because it isn't big enough! 

Fundamentally the problem is that there is no perfect solution. Take the 16 bit case where this was important. If you have two 16 bit indexes into an array, and you subtract them, you need 16 bits for the magnitude, and then if you want it to be signed you need 17 bits. In asm you have the carry flag for this. In C you don't have the option, so I guess that instead C has given you the two imperfection options of signed and unsigned 16 bits, and you choose which of the two matches your problem better. In practice this works well enough.

The problem is that D has stupidly made array.length an unsigned type, which means that trivial, innocent-looking things like:

``` if (x.length - 2 > y.length) ```  
are bugs in D because it's an unsigned operation, whereas
`` if (xend - xstart - 2 > yend - ystart)`` is fine in C because it is ptrdiff_t and thus a signed comparison.

So what I think *should* have happened is something like, create a type ``length_t`` which is a signed `size_t` (same as ``ptrdiff_t``) and which is the type returned by array.length. Unfortunately that's not what happened.  :(

andrew-stevenson-sociomantic:
> the appropriate type for the number of elements in an array is `ptrdiff_t`, not `size_t`

The spec just says an integral type IIRC. I guess that means you could use `ptrdiff_t` though I think both it and `size_t` are defined only in the context of operations that produce them (e.g. `sizeof`, `offsetof`, `alignof` et al) so I think it would be abuse.

> Fundamentally the problem is that there is no perfect solution. Take the 16 bit case where this was important. If you have two 16 bit indexes into an array, and you subtract them, you need 16 bits for the magnitude, and then if you want it to be signed you need 17 bits

If 16 bits is the absolute maximum integer size then its awkward but many systems will allow some form of long larger than the native word size. If 16 bits is the largest you've got then you probably don't wan to allocate an array that big cause you have no memory for anything else ;-)

> The problem is that D has stupidly made array.length an unsigned type

I don't think that is a problem per se - I think its that (assuming it doesn't warn you) you can write a dangerous expression without knowledge. You should have to indicate such unsafe practices are deliberate, or prove they aren't unsafe by e.g. testing values before hand, or get a compiler warning.

don-clugston-sociomantic:
> If 16 bits is the absolute maximum integer size then its awkward but many systems will allow some form of long larger than the native word size. If 16 bits is the largest you've got then you probably don't want to allocate an array that big cause you have no memory for anything else ;-)

Obviously you wouldn't be using the entire memory space, just more than half of it. It was pretty common in the 16 bit days, you might have a text editor where the buffer was 40Kb in size.

> I don't think that is a problem per se - I think its that (assuming it doesn't warn you) you can write a dangerous expression without knowledge. You should have to indicate such unsafe practices are deliberate, or prove they aren't unsafe by e.g. testing values before hand, or get a compiler warning.

The problem is that the dangerous operations include things like
`if (arr.length - 3 > 1)`
and if you do something like:
`if (a.length  - 1 > b.length)`
there isn't even a signed-unsigned comparison you could warn about.

And the reason for this is just so can support the extremely obscure case of gigantic arrays of byte-sized objects which occupy more than half on the memory space, which was nearly impossible to any system that was ever supported by D and is impossible on those which are most commercially relevant today. (Eg in standard 32 bit windows it was impossible but there was a special version of Windows which allowed you to have a single process with >2GB of RAM; even so you could still only have one such array).

It may well be a lost cause. Regardless, I'm not going to be able to come up with a solution today :)

PR #809, opened 2020-03-16
TimeSpan: Replace implementation with core.time : Duration

Geod24:
This reimplements TimeSpan as a small wrapper around `core.time : Duration`.
The intent is to ultimately replace `TimeSpan` with `Duration`.

This will have multiple advantages:
- Less code in Ocean;
- More attention to this code (this hasn't been touched in over a decade);
- Better D2 compatibility (attributes & co);
- Safer and easier API;
- Better compatibility with other type (e.g. `Thread.sleep`);
- Less reliance on platform-specific API;

Like the old utilities in Phobos, the `time` package of Ocean is built around ticks,
with 1 tick equivalent to 1 hnsecs, or 100 nano seconds.

However, Phobos moved away from using ticks directly, as they do not accurately
represent time: for example, some API acception time objects might expect monotonic times,
while other might expect "user time" (leap second and other non monotonic behavior).

In order to adapt Ocean to this distinction, the first step is to reduce the amount
of duplication and special casing, of which removing `TimeSpan` contribute.
This does not actually deprecate `TimeSpan` yet (as many places still rely on it),
but allow to start replacing it.

### DIFF
--- a/src/ocean/time/Time.d
+++ b/src/ocean/time/Time.d
@@ -20,9 +20,9 @@
 
 module ocean.time.Time;
 
-import ocean.meta.types.Qualifiers;
-
+import core.time;
 import core.stdc.time: time_t;
+import ocean.meta.types.Qualifiers;
 
 version (unittest)
 {
@@ -62,7 +62,12 @@ version (unittest)
 struct TimeSpan
 {
         // this is the only member of the struct.
-        package long ticks_;
+        package Duration duration_;
+
+        /// Convenience function (older compilers don't allow public alias to package)
+        public Duration duration () const { return this.duration_; }
+        /// Allow implicit conversion from `TimeSpan` to `Duration`
+        alias duration this;
 
         // useful constants.  Shouldn't be used in normal code, use the
         // static TimeSpan members below instead.  i.e. instead of
@@ -99,30 +104,49 @@ struct TimeSpan
         /**
          * Minimum TimeSpan
          */
-        enum TimeSpan min = {long.min};
+        enum TimeSpan min = TimeSpan(Duration.min);
 
         /**
          * Maximum TimeSpan
          */
-        enum TimeSpan max = {long.max};
+        enum TimeSpan max = TimeSpan(Duration.max);
 
         /**
          * Zero TimeSpan.  Useful for comparisons.
          */
-        enum TimeSpan zero = {0};
+        enum TimeSpan zero = TimeSpan(Duration.zero);
+
+        /// Compatibility constructors
+        public this (long ticks)
+        {
+            this.duration_ = ticks.hnsecs;
+        }
+
+        ///
+        public this (Duration dur)
+        {
+            this.duration_ = dur;
+        }
 
         /**
-         * Get the number of ticks that this timespan represents.  This can be
-         * used to construct another TimeSpan:
+         * Get the number of ticks that this timespan represents.
          *
+         * A tick correspond to an hecto-nanosecond, or 100ns.
+         * This is the way this struct stores data internally,
+         * which is the representation `core.time : Duration` uses.
+         *
+         * This method can be used to construct another `TimeSpan`:
          * --------
          * long ticks = myTimeSpan.ticks;
          * TimeSpan copyOfMyTimeSpan = TimeSpan(ticks);
          * --------
+         *
+         * See_Also:
+         * https://dlang.org/phobos/core_time.html
          */
         long ticks()
         {
-                return ticks_;
+            return this.duration_.total!"hnsecs";
         }
 
         /**
@@ -130,7 +154,7 @@ struct TimeSpan
          */
         equals_t opEquals(TimeSpan t)
         {
-                return ticks_ is t.ticks_;
+                return this.duration_ == t.duration_;
         }
 
         /**
@@ -138,13 +162,7 @@ struct TimeSpan
          */
         public int opCmp ( const typeof(this) rhs ) const
         {
-            if (ticks_ < rhs.ticks_)
-                return -1;
-
-            if (ticks_ > rhs.ticks_)
-                return 1;
-
-            return 0;
+            return this.duration_.opCmp(rhs.duration_);
         }
 
         /**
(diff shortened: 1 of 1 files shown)


Geod24:
Fixed the test failure and reduced the diff

src/ocean/time/Time.d
```diff
@@ -122,29 +139,23 @@ struct TimeSpan
          */
         long ticks()
         {
-                return ticks_;
+            return this.impl.total!"hnsecs";
```
joseph-wakeling-frequenz: What's the basis for assuming that `ticks == hnsecs` is valid?  It's not at a glance an obvious equivalence, so it would be good to have a comment (and ideally, reference) explaining the validity.

src/ocean/time/Time.d
```diff
@@ -62,7 +62,12 @@ version (unittest)
 struct TimeSpan
 {
         // this is the only member of the struct.
-        package long ticks_;
+        package Duration impl;
```
joseph-wakeling-frequenz: Minor/bikesheddy: something like `duration_` would be a more informative name.

I guess you want to stress the fact that you're reworking the struct as just a thin wrapper around the true implementation, but you know that anyway, so you might as well use a field name that conveys what the field actually is.

src/ocean/time/Time.d
```diff
@@ -62,7 +62,12 @@ version (unittest)
...
+
+        /// Convenience function (older compiler don't allow public alias to package)
+        public Duration duration () const { return this.impl; }
+        /// Allow implicit convertion from `TimeSpan` to `Duration`
```
joseph-wakeling-frequenz: Spelling: `conversion` ;-)

src/ocean/time/Time.d
Geod24: An hecto-nanoseconds is 100ns (https://dlang.org/phobos/core_time.html).
And we have this constant: https://github.com/sociomantic-tsunami/ocean/pull/809/files#diff-77aa1635ec08d59b7a587a49057b8b0fR79

joseph-wakeling-frequenz: Gotcha.  Would you be happy to add a small cross-ref to those definitions as a code comment before the `return` statement?  It would make the relationship more obvious at first glance.

(You could probably just make it a cross-ref to the `NanosecondsPerTick` value above, as you only really have to show consistency with ocean.)

src/ocean/time/Time.d
```diff
@@ -865,14 +876,18 @@ struct TimeOfDay
...
-                t.millis %= 1000;
-                return t;
+            return TimeOfDay(modulo24(ticks));
+        }
+
+        /// Ditto
+        static TimeOfDay opCall (Duration dur)
+        {
+            // Note: Since this represent a time of day, hours needs to be
+            // 0 <= hours < 24. Usinf `days` ensures that, even if we don't
```
joseph-wakeling-frequenz: Spelling: `Using` ;-)

src/ocean/time/Time.d
```diff
@@ -865,14 +876,18 @@ struct TimeOfDay
...
-                t.hours   = (t.millis / 3_600_000) % 24;
-                t.millis %= 1000;
-                return t;
+            return TimeOfDay(modulo24(ticks));
+        }
+
+        /// Ditto
+        static TimeOfDay opCall (Duration dur)
+        {
+            // Note: Since this represent a time of day, hours needs to be
```
joseph-wakeling-frequenz: Grammar: `represents`

src/ocean/time/Time.d
```diff
@@ -62,7 +62,12 @@ version (unittest)
 struct TimeSpan
 {
         // this is the only member of the struct.
-        package long ticks_;
+        package Duration impl;
+
+        /// Convenience function (older compiler don't allow public alias to package)
```
joseph-wakeling-frequenz: Grammar: `compilers` rather than `compiler`

joseph-wakeling-frequenz:
Just to get a sense where this is going: is the idea here to preserve the public `TimeSpan` APIs, but to allow all private ocean functionality to be rewritten using `Duration`?  Then add public `Duration` APIs and deprecate `TimeSpan` and public APIs that use it?

Geod24:
Done:
- Improved documentation of `ticks` to mention hnsecs;
- Renamed `impl` to `duration_`;
- Addressed reviews about grammar;

> Just to get a sense where this is going [...]

Yes, that's it. I was looking at this package and the amount of duplication / unused modules. This was a low hanging fruit. There are other places that I was looking at, but given that Ocean won't be used by its creators anymore, I am not sure how worth it is.

FYI this is our current diff: https://github.com/bpfkorea/ocean/commit/b50157ea3025b9d354b85ac33d8c2975c7e6cb17

src/ocean/time/Time.d
```diff
@@ -20,9 +20,9 @@
 
 module ocean.time.Time;
 
-import ocean.meta.types.Qualifiers;
-
+import core.time;
 import core.stdc.time: time_t;
+import ocean.meta.types.Qualifiers;
```
joseph-wakeling-frequenz, line 25: Minor: probably best not to mix up style changes like the moving of `import ocean.meta.types.Qualifiers` with the functional changes this patch intends.

A question of taste: maybe use a selective import to make clear what `core.time` symbols you are using (I assume just `Duration`)?  And is there any reason not to use a nested import inside `TimeSpan` ... ?

joseph-wakeling-frequenz:
Needs release notes (this change should be highlighted, since even though it should theoretically be non-breaking, it's possible there will be some unanticipated behavioural change).

Have you benchmarked just in case there are some performance impacts?  I wouldn't see that as a blocker but it would be good to know.  (No worries if not.)

Other than that, and the minor remarks above, LGTM.

Geod24:
> Needs release notes 

We've never done release notes for refactoring. I plan to deprecate it in the future, that's where the release notes should be.

> Have you benchmarked [...]

No

joseph-wakeling-frequenz:
> We've never done release notes for refactoring.

It's not mere refactoring (this patch adds a new constructor), and in any case, if an implementation has changed significantly enough that it could have downstream impact, it's only polite to alert users to that possibility.

joseph-wakeling-frequenz:
The `alias this` and `duration` method are also new.  So, this genuinely does change what the data structure can do, and what users can do with it.  It's not just under the hood changes.

Geod24:

don-clugston-sociomantic:
Honestly I don't see any need to be pedantic about release notes, since very few people are affected. I'll give the two of you one day more to battle it out, and then merge it. FYI: I have 8 more days with commit rights to this repo, I doubt there will be any more activity on it after that. So I really don't want perfection to get in the way of good enough.
I'd love to see the end of hnsecs, which seems to not exist anywhere in the world outside of Phobos, and which is clearly somewhat platform-specific.

joseph-wakeling-frequenz:
Well, I've given my suggestions.  No big deal if there is disagreement ;-)

Geod24:

don-clugston-sociomantic:
LGTM

PR #810, opened 2020-03-17
dub.sdl: Exclude deprecated modules from source file

Geod24:
Avoid deprecation warnings when compiling.

### DIFF
--- a/dub.sdl
+++ b/dub.sdl
@@ -8,6 +8,10 @@ systemDependencies "Depending on the component used you might need one of the fo
 importPaths "src"
 targetType "sourceLibrary"
 
+// Deprecated files are excluded
+excludedSourceFiles "src/ocean/stdc/posix/arpa/inet.d" "src/ocean/stdc/posix/net/if_.d" "src/ocean/stdc/posix/netinet/in_.d" "src/ocean/stdc/posix/stdlib.d" \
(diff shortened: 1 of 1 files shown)


joseph-wakeling-frequenz:
I'm not familiar with the modules, but ... is this wise?  Where do these deprecation warnings show up, if one doesn't either import the modules directly, or import modules that import them?

I'm concerned that excluding these source files in this way may either make them unusable by DUB projects or stop them being unittested via DUB.

joseph-wakeling-frequenz:
I suppose the attitude might be "No new project should use these modules and by definition any DUB project using them is a new project", but generally it seems better to be comprehensive with testing in all circumstances, and bear the price of deprecations if necessary.

Geod24:
Deprecated modules are not tested anyway:
https://github.com/sociomantic-tsunami/ocean/blob/af8c85cb6b92817946ae1d79d73c2ff4f2c5f50a/Build.mak#L44-L51

Geod24:
Plus, none of those modules have tests anyway, they are merely thin wrappers for other modules

joseph-wakeling-frequenz:
Fair enough, objections withdrawn :-)

Geod24:

don-clugston-sociomantic:
LGTM

PR #811, opened 2020-03-18
Add `return ref` annotations when required

don-clugston-sociomantic:
From dmd2.092 onwards, when a function returns a ref parameter, the parameter
must be marked with 'return', so that memory corruption can be prevented.
Functions returning references to their data members must be similarly marked.
See DIP25.

### DIFF
--- a/src/ocean/text/Search.d
+++ b/src/ocean/text/Search.d
@@ -456,7 +456,7 @@ public struct SearchFruct
 
      ***********************************************************************/
 
-    Substitute tokens (cstring content, cstring sub = null)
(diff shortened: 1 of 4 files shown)


joseph-wakeling-frequenz:
Is it worth adding `-dip25` to the CI setup to make sure this gets validated with earlier compiler versions?  (I'm actually a little surprised that the code passes CI without it, I forget how DMD treats these `return` attributes/qualifiers when `-dip25` is not used.)

Geod24 (approved):

don-clugston-sociomantic:
I will make a separate PR for -dip25.

PR #813, opened 2020-03-23
Add -dip25 to build, to check compatibility with dmd 2.092

don-clugston-sociomantic:
DMD 2.092 has not actually been released yet, but the main breaking change it has it that -dip25 is enabled. So -dip25 should be set for all supported compilers.


joseph-wakeling-frequenz:
There's a spelling mistake (`annotation`, not `annotaion`) in the one commit title.  Other than that LGTM.

don-clugston-sociomantic:
I fixed the typo. It passes, so I will merge it.

PR #814, opened 2020-03-23
Remove ++this from IEBTree, replace with increaseNodeCount

don-clugston-sociomantic:
Using ++this, --this to maintain the counter was operator abuse and totally unnecessary, as these were `protected` implementation functions used only once per class.

The operators generate warnings in dmd2.090 onwards.

I have renamed them rather than added a new alias because `--this` doesn't deserve to live. In any case they are wrappers around C libraries so the three classes existing in ocean are the only ones that make sense.

It's weird that it supports increasing and decreasing by `n`, when it's only called with `n==1`. But the whole design is a bit broken, the modules are very tightly coupled.

After this change, ocean finally compiles cleanly with dmd 2.092 beta.

### DIFF
--- a/relnotes/ebtree-ops.migration.md
+++ b/relnotes/ebtree-ops.migration.md
@@ -1,0 +1,8 @@
+### Replacement of IEBtree protected operators
+
+* `ocean.util.container.ebtree.IEBTree`
+
+The `opAddAssign` and `opSubAssign` operators have been replaced with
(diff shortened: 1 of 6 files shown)


Geod24:
Perhaps a release notes to notify the breaking change ? Otherwise LGTM

joseph-wakeling-frequenz:
Oh, bravo.  I remember these showing up during the operator migration and thinking they need to die.  LGTM with the migration notes included.

PR #815, opened 2020-03-24
More changes required by 2.092

don-clugston-sociomantic:
A couple more deprecation warnings which are generated when compiling ocean unit tests.

* Two cases of D1-style operator syntax, where it is just an alias rather than a virtual function;
* Some variable shadowing in foreach statements
* An inconsequential `printf` inside a unittest.


Geod24 (approved):
LGTM. Last commit message gas a typo ("foreeach") but that's about it :)

alaksiej-stankievicx-dunnhumby (commented):
Except mentioned typo, LGTM

don-clugston-sociomantic:
Typo fixed.

PR #816, opened 2020-03-26
Reduce Ocean's dependency on `ocean.stdc`

Geod24:
`ocean.stdc` is a relic from the Tango days.
Unfortunately, some C bindings are still lacking from druntime (mostly POSIX 2008) so it can't be eliminated right away, but a few modules can be nuked.
In order to deprecated it, we first need to get rid of the imports in ocean. This PR does just that, it converts any import to `ocean.stdc.sys.posix.MOD` to it's `core.sys.posix.sys.MOD` equivalent, save for `socket` and `un`, which still have things that druntime is lacking.

### DIFF
--- a/src/ocean/io/device/FileMap.d
+++ b/src/ocean/io/device/FileMap.d
@@ -17,12 +17,11 @@
 
 module ocean.io.device.FileMap;
 
+import ocean.io.device.Array;
+import ocean.io.device.File;
 import ocean.meta.types.Qualifiers;
 import ocean.stdc.posix.sys.mman;
 import ocean.sys.Common;
-
-import ocean.io.device.File,
-               ocean.io.device.Array;
 
 
 /*******************************************************************************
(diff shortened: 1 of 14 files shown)


Geod24:
👋 @ben-palmer-sociomantic

src/ocean/io/device/FileMap.d
```diff
@@ -18,7 +18,7 @@
 module ocean.io.device.FileMap;
 
 import ocean.meta.types.Qualifiers;
-import ocean.stdc.posix.sys.mman;
+import core.sys.posix.sys.mman;
```
ben-palmer-sociomantic: Should we re-order these to keep them alphabetical?

ben-palmer-sociomantic (commented):
Just one minor cosmetic comment.

Geod24:
Addressed review, rebased on v5.x.x

Geod24:
@don-clugston-sociomantic ?

don-clugston-sociomantic:
LGTM

PR #817, opened 2020-03-26
Deprecate `ocean.stdc.posix.netinet.tcp` & improve segmentation

Geod24:
As can be seen from the diff, this deprecate a module that was missed in a previous pass, and improve segmentation of C binding so that future contributors know when a specific binding can be dropped from Ocean.

See https://github.com/dlang/druntime/pull/2789


Geod24:
Ping @ben-palmer-sociomantic :)

ben-palmer-sociomantic:
LGTM @don-clugston-sociomantic any comments?

don-clugston-sociomantic:
LGTM

PR #821, opened 2020-03-31
Clear up some more extern(C) declarations

Geod24:
Better reviewed commit-by-commit.

### DIFF
--- a/src/ocean/sys/SignalFD.d
+++ b/src/ocean/sys/SignalFD.d
@@ -83,34 +83,21 @@
 
 module ocean.sys.SignalFD;
 
-
-import ocean.core.Exception;
-import ocean.sys.ErrnoException;
-
-import ocean.sys.SignalMask;
-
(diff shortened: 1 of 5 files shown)


Geod24:
Travis, why are you like this ? Can someone retrigger the CI ? It's a network error...
CC @ben-palmer-sociomantic @don-clugston-sociomantic

don-clugston-sociomantic:
This basically looks good, but could you please move the deprecation commits (timerfd, pthread) into a PR which is different from the cleanup stuff.  I notice that ThreadWorker is modified in both, so the cleanup one would have to be merged first,
If you split it off into a separate PR I will just merge them, we don't need to wait for Travis since we know it passes.

Geod24:
Removed the two deprecation commits, will submit a new PR once this is merged.

don-clugston-sociomantic:
LGTM

PR #823, opened 2020-04-01
Deprecate module ocean.core.Traits

Geod24:


don-clugston-sociomantic:
It turns out that this template doesn't actually have a replacement.

```
deprecated("Use ocean.meta.types.Function.ReturnTypeOf")
template ReturnTypeOf( alias fn )
{
    alias ReturnTypeOf!(typeof(fn)) ReturnTypeOf;
}
```
The code in ocean.core.Traits works for member functions. The one in meta does not, because it ignores the 'this' pointer.

This is used in dhtproto.

Geod24:
Do you have a link ? Github doesn't match anything.

don-clugston-sociomantic:
It is in dhtproto.client.internal.NodeHashRanges

The relevant code is:

```
private class NodeHashRangesBase
{
    import ocean.meta.types.Function : ReturnTypeOf;
    import swarm.neo.AddrPort;

    /// Convenience alias for the comparison type of AddrPort.
    private alias ReturnTypeOf!(AddrPort.cmp_id) Addr;
```

And swarm.neo.AddrPort is:
```
public struct AddrPort
{
    public long cmp_id ( )  { .. }
}
```
I get:
./submodules/ocean/src/ocean/meta/types/Function.d(112): Error: need this for cmp_id of type long()

Geod24:
Can you try:
```D
    private alias ReturnTypeOf!(&AddrPort.cmp_id) Addr;
```

From your error message, it looks like the method is implicitly called...

don-clugston-sociomantic:
That doesn't work either.

./src/dhtproto/client/internal/NodeHashRanges.d(278): Error: expression & cmp_id is not a valid template value argument

Geod24:
Submitted a PR to fix the code, so merging this.

PR #826, opened 2020-07-08
Support `scope cstring` in the FormatterSink

AndrejMitrovic:
I couldn't run the test-suite locally but hopefully this works.

CC @Geod24

### DIFF
--- a/relnotes/conduit-read-write.migration.md
+++ b/relnotes/conduit-read-write.migration.md
@@ -1,0 +1,18 @@
+`InputStream.read`, `OutputStream.write` et al now take a `scope` argument
+
+* ``
+
+The following method have been modified, so that either their argument is `scope`,
+or the delegate they accept takes a `scope` argument (when multiple `scope`s are present,
+the change will be highlighted as `*scope*`).
+
+In `ocean.io.model.IConduit`:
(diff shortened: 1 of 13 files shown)


AndrejMitrovic:
Argh that didn't work.

Geod24:
There's a few places where `FormatterSink` should be used instead of ad-hoc delegates, e.g. in the `format` overloads of the `Appender` or in `Stdout`, so you'd have to fix that first.

AndrejMitrovic:
I forgot to mention, there were too many breaking changes when I tried to change this alias. I'm not sure this is an easy change. It's only easy in our fork, since we exclude most files in the build.

ben-palmer-sociomantic:
Seems to be failing the tests at the moment

AndrejMitrovic:
@Geod24 do you wanna take over this one? You're the formatting expert.

Geod24:
Just took a look, this will take a lot of fixes to make it work without breaking change, because the implementation of contravariance for delegate is botched :/
Might as well go into v6.x.x, but is there any interest to make such a release ?

Geod24:
Done in: https://github.com/sociomantic-tsunami/ocean/pull/855

PR #829, opened 2020-08-13
Remove v2.071.2.s* from the test matrix

Geod24:
```
This compiler is long gone and there is no interest in supporting it in the future.
This is the lowest hanging fruit in the matrix and will allow upstream to move
forward immediately.
```

CC @joseph-wakeling-frequenz

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,10 +33,6 @@ jobs:
     include:
         # Test matrix
         - <<: *test-matrix-d2
-          env: DIST=xenial DMD=2.071.2.s* F=production
-        - <<: *test-matrix-d2
-          env: DIST=xenial DMD=2.071.2.s* F=devel
-        - <<: *test-matrix-d2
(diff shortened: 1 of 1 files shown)


ben-palmer-sociomantic:
LGTM

ben-palmer-sociomantic:
LGTM

joseph-wakeling-frequenz:
Just to check, are there any downstream apps/libs in tsunami still using (relying on?) this compiler?

Geod24:
Yes, the nodes, but they will be updated in due time.

joseph-wakeling-frequenz:
Yea, fair enough.  We can tweak their requirements once they update their dependencies.

joseph-wakeling-frequenz:
(As a general rule I prefer to operate the other way round -- drop the downstream requirement first, then drop support upstream -- but in this case I think we can be practical:-)

PR #830, opened 2020-08-13
Use FormatterSink more consistently instead of delegate

Geod24:
This will make sure documentation is readable,
and any future update we make will be slightly easier.


ibuclaw (approved):
Reviewer note, checked it is an alias for `delegate(cstring)`.

PR #831, opened 2020-08-16
Fix 827 - SmartUnion generates conflicting opCalls

MoonlightSentinel:
Only generate `opCall` for members with unique types. This bug wasn't
reported due to [1] but the generated `opCall`s were unusable
anyway (calling them caused an ambiguity error).

[1] https://issues.dlang.org/show_bug.cgi?id=2789

Note: I hope this is the correct target branch(?)

### DIFF
--- a/relnotes/smartunion.bug.md
+++ b/relnotes/smartunion.bug.md
@@ -1,0 +1,10 @@
+### `SmartUnion` will no longer produce conflicting `opCall`
+
+* `ocean.core.SmartUnion`
+
+Before this change, `SmartUnion` would generate duplicated `static opCall`
+for initialization. However those `opCall` were not callable,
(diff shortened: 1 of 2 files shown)


Geod24:
So... It will just generate an `opCall` for whatever first method is encountered ?

MoonlightSentinel:
Not anymore. The generated `opCall`s weren't callable anyway due to the ambiguity so this shouldn't break other code.

Geod24:
SGTM. Just needs a [release note](https://github.com/sociomantic-tsunami/neptune/blob/v0.x.x/doc/library-contributor.rst)

Geod24:
Added release notes, fixed the typos.

MoonlightSentinel:
Thank you @Geod24

PR #835, opened 2020-08-19
Remove support for compilers older than 2.077

ibuclaw:
The minimum supported compiler version is now 2.078


Geod24 (approved):
This could also go to v5.x.x if you want.

Geod24:
Although we didn't really have a support policy in v5.x.x, so v6.x.x is probably better.

ibuclaw:
> Although we didn't really have a support policy in v5.x.x, so v6.x.x is probably better.

So I'll have to write up a changelog say that there is now a support policy. ;-)

PR #836, opened 2020-08-21
Remove templates Const, Immut, Inout

Geod24:
As mentioned in the release notes, those are an heritage of the D2 transition.
Since all downstream projects are D2-only, it is trivial to just replace them.


ibuclaw (approved):
OK.

PR #837, opened 2020-08-25
[v6.0.0] Drop compatibility for anything less than v2.092

Geod24:
```
Since we want to drop support for dmd-transitional,
we need to settle on the minimum supported version of DMD.
Due to the various deprecations and some codeggen bugs in -O -inline,
only v2.092.0 and higher is able to compile our code correctly.

Additionally we now only test on bionic instead of xenial,
and reduced coverage to only the latest version of the compiler.
```

As was already done in the nodes, and will be done in the protos / swarm once the deprecation is fixed.


ibuclaw (approved):
OK

PR #840, opened 2021-02-25
HttpStack: Use assert(0) instead of abort()

Geod24:
assert(0) will be lowered to an HLT in release mode.
This blocks upstream DMD, as DIP1034 will introduce `noreturn`,
which correctly detects that `abort` does not return,
and subsequently triggers a "Statement is not reachable" on the `return`.


Geod24:
Of course 0% of diff hit, it's an `assert(0)`.

PR #841, opened 2021-02-25
Replace Travis CI with a Github workflow

Geod24:
This works rather well: https://github.com/Geod24/ocean/runs/1977736677?check_suite_focus=true


.github/workflows/ci.yml
```diff
@@ -0,0 +1,54 @@
...
+name: CI
+
+on: [push, pull_request]
+
+jobs:
+  main:
+    name: Run
+    strategy:
+      # Disable `fail-fast` because we want the whole test suite to run even if one
+      # of the nigthly is broken
```
joseph-wakeling-frequenz: typo: should be `nightly` :-)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,54 @@
...
+    # Checkout this repository and its submodules
+    - uses: actions/checkout@v2
+      with:
+        submodules: true
+
+    - name: Setup environment
+      run: |
+        echo "$(git config -f .gitmodules submodule.beaver.path)/bin" >> $GITHUB_PATH
+        echo "COV=${{ matrix.coverage }}" >> $GITHUB_ENV
+        echo "DIST=${{ matrix.dist }}" >> $GITHUB_ENV
```
joseph-wakeling-frequenz, line 35: It should be possible to get this with `lsb_release -cs` without needing to include a separate `dist` parameter in the matrix config.  That might help avoid future errors where the `os` setting gets updated but the `dist` does not.

joseph-wakeling-frequenz (dismissed):
One typo and one suggestion for a possible improvement (up to you whether to follow it:-).

The rest looks theoretically good, except for the fact that I don't see the GitHub tests actually running for this PR :-\

If I was going to be really fussy I'd suggest adding the GitHub workflow in one PR (so we see the GitHub and Travis tests running side by side) and dropping the Travis config in a follow-up one.

Finally: is pushing straight to the latest 6.x.x branch sufficient?  Do we want to apply this to any earlier versions and merge into later versions and the major-version dev branch(es)?  Up to you as the driver of active development right now.

.github/workflows/ci.yml
joseph-wakeling-frequenz: BTW I assume here you are talking about nightly compiler builds (which will presumably be added later as they are not in the current setup)?  It might be worth clarifying that in the comment.

Also, it's not just about nightlies!  I've experienced failures with more recent DMD releases that passed with earlier ones (usually because of a deprecation-related issue), or failures with prod that did not show up with devel.  So it's useful to see all the matrix builds even if no nightlies are involved.

Maybe rephase:

```
Disable `fail-fast` because even if things fail for one compiler release
or build flavour we still want to see how things perform with the others
```

Geod24:
> The rest looks theoretically good, except for the fact that I don't see the GitHub tests actually running for this PR :-\

It will only start to show up once merged, hence why I linked to the run in my fork.

.github/workflows/ci.yml
Geod24: Good point, will do.

Geod24:
> Finally: is pushing straight to the latest 6.x.x branch sufficient? Do we want to apply this to any earlier versions and merge into later versions and the major-version dev branch(es)? Up to you as the driver of active development right now.

I really don't want to spend more time on Ocean that is necessary, unless we see more users. I'm currently the sole maintainer and developer, and while I will follow the deprecation process and SemVer guarantee, I have no intention to provide support for older version for non-existing users.

.github/workflows/ci.yml
Geod24: I actually want to get rid of this and move the work from beaver to the Github action directly.
Currently the Github action runs in a container, which itself spawns a container. That's quite superfluous.
Plus, if we want to test on other platforms (e.g. OSX) we can't rely on beaver / `lsb_release`.

Geod24:
I'm moving ahead with this because upstream DMD needs a new release to make progress on DIP1034. See last comment in https://github.com/dlang/dmd/pull/12222 .

PR #843, opened 2021-03-01
Replace beaver with a pure Github Action solution

Geod24:
```
This will allow us to get rid of the extra Docker image used for testing,
which was previously necessary because of dmd-transitional.
In addition of speeding up the test, it will also allow us to make the
test-suite cross platform (beaver currently only supports Ubuntu).
```

I wanted to add support for `dmd-latest` (2.095), but unfortunately: https://issues.dlang.org/show_bug.cgi?id=21670
There also seems to be a bug with devel + 2.094, so I decided to stick with the supported versions for now.


Geod24:
CC @llucax @ibuclaw

ibuclaw (approved):
OK.

.github/workflows/ci.yml
```diff
@@ -13,42 +13,67 @@ jobs:
...
+    - name: 'Download ebtree from bintray'
+      if: steps.cache-ebtree.outputs.cache-hit != 'true'
       run: |
-        echo "$(git config -f .gitmodules submodule.beaver.path)/bin" >> $GITHUB_PATH
-        echo "COV=${{ matrix.coverage }}" >> $GITHUB_ENV
-        echo "DIST=${{ matrix.dist }}" >> $GITHUB_ENV
-        echo "DMD=${{ matrix.dmd }}" >> $GITHUB_ENV
-        echo "F=${{ matrix.flavor }}" >> $GITHUB_ENV
+        wget -O ${{ github.workspace }}/tools/libebtree.deb https://bintray.com/sociomantic-tsunami/dlang/download_file?file_path=libebtree6_6.0.s7-rc5-xenial_amd64.deb
+        wget -O ${{ github.workspace }}/tools/libebtree-dev.deb https://bintray.com/sociomantic-tsunami/dlang/download_file?file_path=libebtree6-dev_6.0.s7-rc5-xenial_amd64.deb
```
llucax, line 38: What happens if you change these lines because the version is upgraded or something? Will this step be run even if there is a cache hit or do you have to somehow clear the cache manually?

llucax (approved):

.github/workflows/ci.yml
Geod24: If there's a version change, I'll probably update the `key` to include the distribution, so there will be no cache hit.
We can also clear the cache manually if we wish, but the `key` approach works best IMO.
Example usage (although it's a bit useless since it only runs on `windows`): https://github.com/dlang/druntime/blob/306bd965ea9d83bad7e5444ff9d5e1af1a6d934a/.github/workflows/cxx.yml#L220-L227

Geod24:
Moving ahead with this then, thanks for the review!

PR #847, opened 2021-04-20
Replace 'body' with 'do'

Geod24:
This should make Ocean usable with v2.096 which deprecates `body` in favor of `do`.


Geod24:
Not usable with v2.096 yet because of https://issues.dlang.org/show_bug.cgi?id=21670 but much better usability nonetheless.

PR #850, opened 2021-06-01
CI: Download and build ebtree ourselves

Geod24:
Previously we used bintray, however the service has now shut down.
Checkout the repository and build instead (but cache the result).


Geod24:
That was fairly easy.

ibuclaw (approved):

PR #851, opened 2021-06-10
Deprecate functions using creal or ireal

Geod24:
At least I know who'll review this :)

### DIFF
--- a/relnotes/creal_ireal.deprecation.md
+++ b/relnotes/creal_ireal.deprecation.md
@@ -1,0 +1,7 @@
+### `creal` / `ireal` functions in `ocean.math` are deprecated
+
+`ocean.math.IEEE`, `ocean.math.Math`, `ocean.meta.traits.Basic`
+
+Upstream deprecated support for `creal` and `ireal` types in v2.097.0.
+Functions in `ocean.math.{IEEE,Math}` that used them are now deprecated.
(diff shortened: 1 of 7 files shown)


alaksiej-stankievicx-dunnhumby (approved):
LGTM
Do you expect me to review it?

alaksiej-stankievicx-dunnhumby:
Hmm, it seems these changes somewhat affect template instantiation somewhere else.

ibuclaw:
Yeah, the testsuite results say that there's a little more that needs to be done in order to get it working.

Geod24:
Yep, needs to deal with `isFloatingPoint`.

Geod24:
Worked around by not deprecating the `isComplexType`. I don't think there's any users, and if there is, well the next major is going to break their code.

alaksiej-stankievicx-dunnhumby:
> Worked around by not deprecating the `isComplexType`. I don't think there's any users, and if there is, well the next major is going to break their code.

Maybe add something via pragma(msg) to notify people to stop using this function but without actual depreciation?

Geod24:
> Maybe add something via pragma(msg) to notify people to stop using this function but without actual depreciation?

Well it's used in `isFloatingPoint`, so it would trigger in unrelated code.

Geod24:
Rebasing now, and will test in our application to see if other deprecations are being triggered.

Geod24:
I can confirm, all gone :)

PR #852, opened 2021-11-14
Explicitly discard `SmartUnion` in unittest

dkorpel:
Buildkite fails on this dmd PR:
https://github.com/dlang/dmd/pull/12890

```
./submodules/ocean/src/ocean/core/SmartUnion.d(549): Warning: calling `ocean.core.SmartUnion.SmartUnion!(HasDuplicates).SmartUnion.opCall` without side effects discards return value of type `SmartUnion!(HasDuplicates)`; prepend a `cast(void)` if intentional
```

### DIFF
--- a/src/ocean/core/SmartUnion.d
+++ b/src/ocean/core/SmartUnion.d
@@ -545,9 +545,9 @@ unittest
     alias U = SmartUnion!HasDuplicates;
 
     // These calls are unambiguous.
-    U(Object.init);
-    U("c");
-    U([true]);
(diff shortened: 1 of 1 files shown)


ibuclaw (approved):
Seems reasonable.

dkorpel:
Can someone please merge this?

dkorpel:
@Geod24 Can you release a new version? buildkite is checking out 'tag: v6.1.0' which doesn't have this fix, so it's still red.

PR #854, opened 2021-12-08
Logger: Add two extra log levels (Debug, Verbose)

Geod24:
```
Experience has shown that the log level provided by Ocean's logger
are not always suitable. Notably, they only offer one level for debugging,
with the rest supposed to be reserved for informational messages,
warnings, or more severe events.
However, while debugging an application, it is often necessary to filter
events by different levels of severity or frequency, something not possible
with a single 'Trace' level.
```

CC @hewison-chris

### DIFF
--- a/src/ocean/util/log/Logger.d
+++ b/src/ocean/util/log/Logger.d
@@ -72,9 +72,6 @@ version (unittest)
 
     These represent the standard LOG4J event levels.
 
-    Note that `Debug` is called `Trace` here, because debug is a reserved word
-    in D.
-
 *******************************************************************************/
 
 public alias ILogger.Level Level;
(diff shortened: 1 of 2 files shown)


hewison-chris (approved):
Nice

src/ocean/util/log/Logger.d
```diff
@@ -463,6 +474,24 @@ public final class Logger : ILogger
...
+        Params:
+            Args = Auto-deduced format string arguments
+            fmt = Format string to use.
+                  See `ocean.text.convert.Formatter` documentation for
+                  more informations.
+            args = Arguments to format according to `fmt`.
+
+    ***************************************************************************/
+
+    public void dbg (Args...) (cstring fmt, Args args)
```
HaraldZealot, line 493: you could mention in documentation for the function that it is named `dbg` instead `debug` (which would be preferable) as the latter is reserved word in D.

Geod24: Fixed

PR #858, opened 2021-12-31
Fix Formatter to handle inout(char) properly

Geod24:
```
We know this construct doesn't allocate, and avoids the need
for Unqual entirely.
```


Geod24:
As a first step, I only committed the test, and it failed as expected:
```
./src/ocean/text/convert/Formatter.d(549): Error: cannot modify `inout` expression `b[1]`
./src/ocean/text/convert/Formatter.d(279): Error: template instance `ocean.text.convert.Formatter.handle!(inout(char))` error instantiating
./src/ocean/text/convert/Formatter.d(137):        instantiated from here: `sformat!(inout(char))`
./src/ocean/text/convert/Formatter_test.d(575):        instantiated from here: `format!(inout(char))`
make: *** [/home/runner/work/ocean/ocean/build/prod/tmp/allunittests] Error 1
```

PR #860, opened 2022-02-18
Replace all usages of 'istring' with 'string'

Geod24:
```
Now that Ocean is D2-only, this trivial alias doesn't make sense,
and make code harder to use outside Ocean.
```


Geod24:
Heh there's some `typeid` shenanigans

ibuclaw (approved):

Geod24:
Found the bug, it was a name conflict with a `string` member. Will submit as a separate PR to make sure the commit is atomic.

PR #865, opened 2022-02-21
Log: Make Logger and Appender name property string

Geod24:
In practice, they are always string, and not referring to some internal buffer.


Geod24:
No other tsunami project implements `Appender`.

ibuclaw (approved):

PR #866, opened 2022-03-14
Formatter: Support Phobos like OutputRange interface

Geod24:
Before this patch, printing a SysTime would fail because we were passing
a literal, which would not implicitly convert to a 'ref'.

### DIFF
--- a/src/ocean/text/convert/Formatter.d
+++ b/src/ocean/text/convert/Formatter.d
@@ -464,9 +464,13 @@ private void handle (T) (T v, FormatInfo f, scope FormatterSink sf, scope ElemSi
     // toString hook: Give priority to the non-allocating one
     // Note: sink `toString` overload should take a `scope` delegate
     else static if (is(typeof(v.toString(sf))))
+    {
+        // Some `toString` signatures expect a `ref` sink
+        scope sinkRef = (in cstring e) { se(e, f); };
         nullWrapper(&v,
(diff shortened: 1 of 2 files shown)


Geod24:
Using the CI to test as I don't have a Linux at hand. Added the test, should go red, will add the fix after, which is one line.

Geod24:
```
./src/ocean/text/convert/Formatter.d(468): Error: template `ocean.text.convert.Formatter_test.__unittest_L580_C1.SysTime.toString` cannot deduce function from argument types `!()(void delegate(const(char[]) e) @system)`, candidates are:
./src/ocean/text/convert/Formatter_test.d(585):        `toString(W)(ref W writer)`
./src/ocean/text/convert/Formatter.d(279): Error: template instance `ocean.text.convert.Formatter.handle!(SysTime)` error instantiating
./src/ocean/text/convert/Formatter.d(137):        instantiated from here: `sformat!(SysTime)`
./src/ocean/text/convert/Formatter_test.d(587):        instantiated from here: `format!(SysTime)`
make: *** [/home/runner/work/ocean/ocean/build/prod/tmp/allunittests] Error 1
submodules/makd/Makd.mak:574: recipe for target '/home/runner/work/ocean/ocean/build/prod/tmp/allunittests' failed
```

Bingo

ibuclaw (approved):

PR #879, opened 2023-06-07
Build for OPENSSLv3 if variable is defined

alexandrumc:
This will help to fix the Buildkite build after this https://github.com/dlang/ci/pull/483 gets merged.

### DIFF
--- a/Build.mak
+++ b/Build.mak
@@ -16,6 +16,11 @@ ifneq ($(ALLOW_DEPRECATIONS),1)
 	override DFLAGS += -de
 endif
 
+# Enables https://github.com/sociomantic-tsunami/ocean/blob/v6.x.x/src/ocean/net/ssl/openssl/OpenSsl.d#L529
+ ifneq ($(OPENSSL_V3),1)
+ »   override DFLAGS += -version=Ocean_OpenSSL3
+ endif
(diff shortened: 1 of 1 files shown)


Build.mak
```diff
@@ -16,6 +16,11 @@ ifneq ($(ALLOW_DEPRECATIONS),1)
 	override DFLAGS += -de
 endif
 
+# Enables https://github.com/sociomantic-tsunami/ocean/blob/v6.x.x/src/ocean/net/ssl/openssl/OpenSsl.d#L529
+ ifneq ($(OPENSSL_V3),1)
+ »   override DFLAGS += -version=Ocean_OpenSSL3
+ endif
```
Geod24:
```suggestion
 ifeq ($(OPENSSL_V3),1)
     override DFLAGS += -version=Ocean_OpenSSL3
 endif
```

You might also want to add support in `ci.yml`, currently we only run on Ubuntu 20.04

alexandrumc: > You might also want to add support in ci.yml, currently we only run on Ubuntu 20.04

That would probably require updating all the other dependencies as well, so I'm not quite interested TBH :grin:. I will try to find some time in the near future.

Issue #890, opened 2025-07-26
ExpiringLRUCache has a null dereference in getExpiringOrCreate


rikkimax:
The following statement can dereference null, even though its within a null check.

https://github.com/sociomantic-tsunami/ocean/blob/b06be4d9c6099bb756a801e13e8b945ea05c2abe/src/ocean/util/container/cache/ExpiringLRUCache.d#L342

Null was set: https://github.com/sociomantic-tsunami/ocean/blob/b06be4d9c6099bb756a801e13e8b945ea05c2abe/src/ocean/util/container/cache/ExpiringLRUCache.d#L335

PR #891, opened 2025-10-01
Fix issue #890 - ExpiringLRUCache has a null dereference in getExpiringOrCreate

rikkimax:


Geod24:
Any chance you can add an unittest that exercise this path ?

